PR #222, opened 2015-02-11
Add support for OpenPathSegment

ificator:
This is a small change to allow the handling of requests that access an open property. It does not contain and changes to make sure $select etc works (nor did I see if it just magically works), since my current scenario does not require that.

OneDrive was unable to upgrade to WebApi OData 5.4 due to a change in how open properties were parsed (probably in one of the ODatalib changes - we started getting back OpenPropertySegment instead of UnresolvedSegment, which resulted in WebApi OData throwing a not implemented error), and so we'll need this change (or something similar) before we can consume any newer versions.

Email bcleaver [Microsoft] for more details.

### DIFF
--- a/OData/src/System.Web.OData/OData/Routing/OpenPropertyPathSegment.cs
+++ b/OData/src/System.Web.OData/OData/Routing/OpenPropertyPathSegment.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Web.Http;
+using Microsoft.OData.Edm;
+
+namespace System.Web.OData.Routing
+{
+    /// <summary>
+    /// An <see cref="ODataPathSegment"/> implementation representing an open property.
+    /// </summary>
+    public class OpenPropertyPathSegment : ODataPathSegment
+    {
+        /// <summary>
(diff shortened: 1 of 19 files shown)


OData/test/System.Web.OData.Test/OData/Routing/ODataPathRouteConstraintTest.cs
```diff
@@ -149,35 +149,6 @@ public void Match_ReturnsTrue_IfODataPathCanBeParsed()
             Assert.Same(_pathHandler, request.ODataProperties().PathHandler);
         }
 
-        [Fact]
```
ificator: This isn't ideal, but there weren't any other segments that weren't implemented, and it seems impossible to override ODataPathSegmentTranslator to force a not implemented failure

OData/src/System.Web.OData/OData/Routing/OpenPropertyPathSegment.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// Returns a <see cref="System.String" /> that represents this instance.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="System.String" /> that represents this instance.
+        /// </returns>
+        public override string ToString()
+        {
+            return this.PropertyName;
```
congysu: Nit: Remove this., and others.

LianwMS:
submit with 16d5d4c20df66e66be9cb479c246bb97fda8d9d1

PR #228, opened 2015-02-20
Allow multiple ODataRoutePrefix attributes

ificator:
OneDrive has scenarios where it exposes objects of the same type at different points in the model. Since the actions that can be performed on the objects are the same irrespective of location we're currently decorating the action methods with a number of almost identical ODataRoute attributes, where the only real difference is prefix. Instead we'd like to define all the applicable prefixes at the controller level, and just decorate the actions with a single ODataRoute that identifies when the action should trigger for the object represented by the controller.

### DIFF
--- a/OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
+++ b/OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
@@ -207,7 +207,7 @@ namespace System.Web.OData.Routing.Conventions
                     ILookup<string, HttpActionDescriptor> actionMapping = actionSelector.GetActionMapping(controller);
                     HttpActionDescriptor[] actions = actionMapping.SelectMany(a => a).ToArray();
  
-                    foreach (string prefix in AttributeRoutingConvention.GetODataRoutePrefixes(controller))
+                    foreach (string prefix in GetODataRoutePrefixes(controller))
                     {
                         foreach (HttpActionDescriptor action in actions)
                         {
--- a/OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
+++ b/OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
@@ -205,15 +205,17 @@ namespace System.Web.OData.Routing.Conventions
                 {
                     IHttpActionSelector actionSelector = controller.Configuration.Services.GetActionSelector();
                     ILookup<string, HttpActionDescriptor> actionMapping = actionSelector.GetActionMapping(controller);
-                    IEnumerable<HttpActionDescriptor> actions = actionMapping.SelectMany(a => a);
-                    string prefix = GetODataRoutePrefix(controller);
-
-                    foreach (HttpActionDescriptor action in actions)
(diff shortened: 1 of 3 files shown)


OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
```diff
@@ -205,15 +205,17 @@ public string SelectController(ODataPath odataPath, HttpRequestMessage request)
...
                     {
-                        IEnumerable<ODataPathTemplate> pathTemplates = GetODataPathTemplates(prefix, action);
-                        foreach (ODataPathTemplate pathTemplate in pathTemplates)
+                        foreach (HttpActionDescriptor action in actions)
                         {
-                            attributeMappings.Add(pathTemplate, action);
+                            IEnumerable<ODataPathTemplate> pathTemplates = GetODataPathTemplates(prefix, action);
+                            foreach (ODataPathTemplate pathTemplate in pathTemplates)
+                            {
+                                attributeMappings.Add(pathTemplate, action);
```
congysu: Add logic to handle duplicated route prefixes:
[ODataRoutePrefix("VipCustomer")]
[ODataRoutePrefix("VipCustomer")]

OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
```diff
@@ -205,15 +205,17 @@ public string SelectController(ODataPath odataPath, HttpRequestMessage request)
                 {
                     IHttpActionSelector actionSelector = controller.Configuration.Services.GetActionSelector();
                     ILookup<string, HttpActionDescriptor> actionMapping = actionSelector.GetActionMapping(controller);
-                    IEnumerable<HttpActionDescriptor> actions = actionMapping.SelectMany(a => a);
-                    string prefix = GetODataRoutePrefix(controller);
-
-                    foreach (HttpActionDescriptor action in actions)
+                    HttpActionDescriptor[] actions = actionMapping.SelectMany(a => a).ToArray();
+ 
+                    foreach (string prefix in AttributeRoutingConvention.GetODataRoutePrefixes(controller))
```
congysu: nit: Can remove AttributeRoutingConvention.

OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs
ificator: The existing code already seemed to handle the scenario by virtue of how the templates are stored in the dictionary. So:
[ODataRoute("entityset({key})")]
[ODataRoute("entityset({key})")]

Ended up collapsing to a single entry since they are duplicates.

Because of that I didn't worry about doing anything special here. We *could* potentially detect duplicates and throw an error but semantically it didn't seem worth while - there's no ambiguity introduced by having it declare twice, if anything it just pollutes the source.

MrTomWhite:
What's the status of this PR?

ificator:
I believe it may have already landed in Odata:master, if that's the case we probably just need to close this PR. I'll let the OData guys handle that though just in case their process requires this open for some reason.

VikingsFan:
checked in 34b5125da2052cb076b499ce7dca07304e8f9a47

MrTomWhite:
Awesome PR! :tada:

Issue #231, opened 2015-02-26
DeleteRef fails if navigation property is derived type and there is no corresponding entity set


pschmiedge:
I am reporting this as a variant of issue #173, as the fix for that issue does not resolve this case.  Consider the following scenario, based on the existing scenario in ODataRoutingModel.cs and ODataRoutingTest.cs:

Model:

    public class RoutingCustomer
    {
        public int ID { get; set; }
        public string Name { get; set; }
        public virtual List<SpecialProduct> SpecialProducts { get; set; }
    }

    public class Product
    {
        public int ID { get; set; }
        public string Name { get; set; }
    }

    public class SpecialProduct : Product
    {
        public int Value { get; set; }
        public virtual List<RoutingCustomer> RoutingCustomers { get; set; }
    }

Controller setup:

    ODataConventionModelBuilder builder = new ODataConventionModelBuilder(configuration);
    builder.EntitySet<RoutingCustomer>("RoutingCustomers");
    builder.EntitySet<Product>("Products");

Notably, a `SpecialProducts` route is *not* mapped.  The only way to access a `SpecialProduct` is through `Products`.

The following two requests will fail with a 404:

    DELETE RoutingCustomers(1)/SpecialProducts/$ref?$id=http://localhost/Products(5)
    DELETE RoutingCustomers(1)/SpecialProducts/$ref?$id=http://localhost/Products(5)/System.Web.OData.Routing.SpecialProducts

I would expect at least one of these to succeed.

Like #173, the issue is still in DefaultODataPathHandler:

    try
    {
        entityIdSegment = uriParser.ParseEntityId();

        if (entityIdSegment != null)
        {
            // Create another ODataUriParser to parse $id, which is absolute or relative.
            ODataUriParser parser = new ODataUriParser(model, serviceRootUri, entityIdSegment.Id);
            id = parser.ParsePath().LastSegment as Semantic.KeySegment;
        }
    }
    catch (ODataException)
    {
        // Exception was thrown while parsing the $id.
        // We will throw another exception about the invalid $id.
        exceptionThrown = true;
    }

    if (exceptionThrown ||
        (entityIdSegment != null &&
            (id == null || !id.EdmType.IsOrInheritsFrom(lastSegmentEdmType.ElementType.Definition))))
    {
        throw new ODataException(Error.Format(SRResources.InvalidDollarId, queryString.Get("$id")));
    }

[...]

PR #237, opened 2015-03-05
Issue #231: DeleteRef fails if navigation property is derived type and there is no corresponding entity set 

VikingsFan:
DELETE RoutingCustomers(1)/SpecialProducts/$ref?$id=http://localhost/Products(5) will work after this fix
Modify the original test case to cover those two sittuations


pschmiedge:
Thanks for looking into this so quickly!  The fix looks good to me.

VikingsFan:
Pushed as commit ff354b22aaaaf4a4d793b9339b8d2f9b6043fc2d

Issue #238, opened 2015-03-09
SelectExpandQueryOption xml comment typo


congysu:
v3: OData/src/System.Web.Http.OData/OData/Query/SelectExpandQueryOption.cs
and
v4: OData/src/System.Web.OData/OData/Query/SelectExpandQueryOption.cs

The related pull request is #213

PR #213, opened 2015-02-06
Update SelectExpandQueryOption.cs

JScharloo:
Fix unjustified copy of a wrong value on function summary.

PR #239, opened 2015-03-09
SelectExpandQueryOption xml comment #238

congysu:
This is a follow up for pull request #213.
This change is for OData v3:
OData/src/System.Web.Http.OData/OData/Query/SelectExpandQueryOption.cs


congysu:
Checked in 42da57b6bc094650377885d781ce6c3c83bc8627

Issue #251, opened 2015-03-18
$filter in $expand throws if Linq-to-entities


congysu:
The issue seems related to null propagation requirement of different Linq providers.

Request:
http://localhost:53372/odata/Products?$expand=Order_Details($filter=UnitPrice gt 18.0)

Error:
```json
{
  "error":{
    "code":"","message":"An error has occurred.","innererror":{
      "message":"The 'ObjectContent`1' type failed to serialize the response body for content type 'application/json; odata.metadata=minimal'.","type":"System.InvalidOperationException","stacktrace":"","internalexception":{
[...]

PR #253, opened 2015-03-18
$filter in $expand throws if Linq-to-entities #251

congysu:
No null propagation for Linq-to-entities.


OData/test/System.Web.OData.Test/OData/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.OData.Core;
 using Microsoft.OData.Core.UriParser;
 using Microsoft.OData.Core.UriParser.Semantic;
+using Microsoft.OData.Core.UriParser.TreeNodeKinds;
```
congysu: Will remove this.

Issue #300, opened 2015-04-09
[Bug] Function Date doesn't work with Linq To Entity


katan4ig:
Filter for example:
```
~/odata/Customers?$filter=date(Birthday) eq 2015-03-23
```
Field Birthday has type DateTimeOffset
After converting this function to Linq query, it looks as follows:
```
.Where($it => new Date($it.Birthday.Year, $it.Birthday.Month, $it.Birthday.Day) == 2015-03-23)
```
But Linq To Entity cannot convert this query to SQL, and throws next exception:
```
System.NotSupportedException: Only parameterless constructors and initializers are supported in LINQ to Entities
```
I use latest version Entity Framework(6.1.3) and Web Api Odata Libraries (6.11 and 5.5.1)

VikingsFan:
time(DateTime) also doesn't work with same error

xuzhg:
@katan4ig, would you please check my pull request works for your scenario?

katan4ig:
@xuzhg, Thanks, your changes work fine.

katan4ig:
@xuzhg I found that filter by "equal" doesn't work. 
For example:
~/odata/Customers?$filter=date(Birthday) eq 2015-03-23
If field Birthday contains Time in DB (2015-03-23 + Time), this filter returns 0 rows

xuzhg:
@katan4ig 

date() function is to return the "Date" part of the DateTimeOffset, That's the year, month & day. While the "Date" part maybe different for the same DateTimeOffset in different TimeZone.
For example, for current implementation:

if: Birthday = 2015-03-23T01:02:03.004+8:00
then: date(Birthday ) =  2015-03-23

if: Birthday = Birhthday.ToUniversalTime;
then: date(Birthday ) =  2015-03-22

I don't know the type of "Birthday", Can you share me your sample codes? Thanks.

xuzhg:
Checked in: 424764a

xuzhg:
@katan4ig  please re-open or file a new issue if you have further concern.

simoco:
Hello! Why this is not working
`~/odata/Customers?$filter=date(Birthday) eq 2015-03-23`

when Birthday is `DateTime` opposed to `DateTimeOffset`?
I got `No function signature for the function with name 'date' matches the specified arguments. The function signatures considered are: date(Edm.DateTimeOffset); date(Edm.DateTimeOffset Nullable=true).`. 
[...]

PR #303, opened 2015-04-14
Function Date() doesn't work with Linq To Entities

xuzhg:
It's for Function Date doesn't work with Linq To Entity: https://github.com/OData/WebApi/issues/300


congysu:
Change title to: Function Date() doesn't work with Linq To Entities

xuzhg:
Merged as 424764a 
And see the E2E test cases: b2d9358

Issue #312, opened 2015-04-20
Null-valued open properties not getting serialized


MaxWilson:
Description:

Open properties with null values are omitted from serialized output. Makes it impossible to distinguish unset open properties from properties set to null.

Minimal repro steps:
Create an open type. Make a service operation which returns an instance of that open type, with its dynamic property collection set to new Dictionary<string, object> { { "MyNullProperty", null } }. Call the service operation. Property will not be there.

Example (only NonNullProperty shows up in output):
        // GET: odata/Stores
        [EnableQuery]
        public IQueryable<Store> GetStores()
        {
            var retval = new Store
            {
                StoreID = 8,
                StoreName = "My store",
                DynamicProperties = new Dictionary<string, object> { { "MyNullProperty", null }, { "MyNonNullProperty", 42 } }
            };
            return new[] { retval }.AsQueryable();
        }

Expected result: 
Null properties become ODataNull type.

Actual result:
ODataEdmTypeSerializer.AppendDynamicProperties() skips over null-valued properties as if they don't exist. Output in screenshot:

![image](https://cloud.githubusercontent.com/assets/6829565/7235463/333b9022-e743-11e4-9f04-f7b6298242ee.png)

Further technical details:
As best I can tell, the OData spec does not restrict the values of open properties, so it should be possible to have null-valued open properties.

MaxWilson:
Sorry, paste error excluded type parameters. Corrected:

        // GET: odata/Stores
        [EnableQuery]
        public IQueryable GetStores()
        {
            var retval = new Store
            {
                StoreID = 8,
                StoreName = "My store",
                DynamicProperties = new Dictionary<string, object> { { "MyNullProperty", null }, { "MyNonNullProperty", 42 } }
            };
            return new[] { retval }.AsQueryable();
        }

congysu:
Thanks @MaxWilson for reporting the issue.  Please feel free to submit a pull request on it.

xuzhg:
@MaxWilson Skip the null dynamic property to serialize is by design. What's your scenario to have a null-dynamic property in the payload?

I do think you can set the value as null if you can't find the dynamic property in the payload. Please let us know your thought. thanks.

MaxWilsonMS:
Scenario is letting end-users have a property bag for business data that actually shows everything that's in the bag (so that adding new properties doesn't take a dev cycle). Also, the fact that the OData spec says that open properties can be nullable, so I should think the scenario is also "implement behavior from the OData spec."

[...]

PR #313, opened 2015-04-21
Allow null dynamic property values in serialized open types

MaxWilsonMS:
Description: there is currently no way to serialize an open propery with a null value ("MyNullProperty" : null). There can be semantic differences between absent properties and null-valued properties, and it should be possible to represent both. This changeset serializes null values as ODataNullValue instances.

Issue: https://github.com/OData/WebApi/issues/312


congysu:
Syncing with Max offline.

LianwMS:
Have synced and closed

forsajt:
When this change is planned to be released? It was not a part of 5.7 release as far I am concerned.

xuzhg:
@forsajt Would you please help review the change in https://github.com/OData/WebApi/pull/573? Any feedback is welcome. Thanks.

forsajt:
@xuzhg  As far as I can say, the change looks good. Thanks

xuzhg:
Thanks.

Issue #166, opened 2015-01-12
v5.6: Incorrect OData-Version header will lead to no matching route


LianwMS:
Send a request containing an OData-Version header value set to any other than
4.0.  

Then the service code would got false in ODataVersionConstraint's Match
method, thus would not find a proper matching route for this. And resulting a
5xx error without detail information.  

Actually according to spec, service should return 4xx here:  
If present on a request, the service MUST interpret the request according to
the rules defined in the specified version of the protocol, or fail the
request with a 4xx response code.  
## Work Item Details

**Original CodePlex Issue:**    [Issue 2180](http://aspnetwebstack.codeplex.com/workitem/2180)
**Status:** Proposed
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Nov 6, 2014 at 1:46 AM
**Reported by:**    karata
**Updated on:** Tue at 7:12 AM
**Updated by:** cysu

LianwMS:
On *2015-01-06 15:12:28 UTC*, **cysu** commented:

Suggest fix as a bug.

PR #322, opened 2015-05-01
Write nextpagelink/count if they can be retrieved from the current reque...

chinese007:
This change is the second part to fix the issue (https://github.com/OData/odata.net/issues/140). In this change, the ODataCollectionSerializer will construct the ODataCollectionStart object by assigning the nextPageLink and count retrieved from the Request.ODataProperties().

After that, write.WriteStart() method will trigger the underlying writer to have those properties properly written to response. 

This pull request has a dependency on the other request (https://github.com/OData/odata.net/pull/166). The checkin can only happen after pull request #166 gets submitted and new OData core library becomes available.

### DIFF
--- a/OData/src/System.Web.Http.OData/OData/Formatter/Serialization/ODataCollectionSerializer.cs
+++ b/OData/src/System.Web.Http.OData/OData/Formatter/Serialization/ODataCollectionSerializer.cs
@@ -10,6 +10,8 @@ using Microsoft.Data.OData;
 
 namespace System.Web.Http.OData.Formatter.Serialization
 {
+    using Extensions;
+
     /// <summary>
     /// ODataSerializer for serializing collection of Entities or Complex types or primitives.
     /// </summary>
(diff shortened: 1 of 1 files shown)


OData/src/System.Web.Http.OData/OData/Formatter/Serialization/ODataCollectionSerializer.cs
```diff
@@ -10,6 +10,8 @@
 
 namespace System.Web.Http.OData.Formatter.Serialization
```
congysu: Do you want to change `System.Web.OData`, which is for OData v4?  `System.Web.Http.OData.*` is for OData v3.

OData/src/System.Web.Http.OData/OData/Formatter/Serialization/ODataCollectionSerializer.cs
```diff
@@ -10,6 +10,8 @@
...
 {
+    using Extensions;
```
congysu: nit: Normally, `using` is before `namespace`.

VikingsFan:
@chinese007 please also add test cases since the change of ODataCollectionSerializer.cs. You can see a page of add unit test. http://odata.github.io/WebApi/#09-01-unittest-e2etest

xuzhg:
@chinese007 
ODL V3 doesn't modify to support, however V4 does. Here's https://github.com/OData/odata.net/pull/166

So, based on your pull request, I changed the V4. Here's the PR: #407.

xuzhg:
@chinese007 I have ported your change to 88376cf. Thanks your contribution.

chinese007:
Thanks!

Sent from my Windows Phone
________________________________

PR #333, opened 2015-05-08
build command case insensitive, rename functional to unit

VikingsFan:


xuzhg:
Merged at: 

 c73b081

Issue #334, opened 2015-05-08
Time zone conversions are not DST aware


mj1856:
In [`OData.Formatter.EdmPrimitiveHelpers.ConvertPrimitiveValue`](https://github.com/OData/WebApi/blob/bad3b8cbec3af117903dcdd2748b1626f9f88c5c/OData/src/System.Web.OData/OData/Formatter/EdmPrimitiveHelpers.cs#L87) and [`OData.Formatter.Serialization.ODataPrimitiveSerializer.ConvertUnsupportedPrimitives`](https://github.com/OData/WebApi/blob/bad3b8cbec3af117903dcdd2748b1626f9f88c5c/OData/src/System.Web.OData/OData/Formatter/Serialization/ODataPrimitiveSerializer.cs#L148-L154), values are being incorrectly converted to a time zone using the time zone's `BaseUtcOffset`.

This essentially ignores all daylight saving time rules contained within the `TimeZoneInfo` object, and assumes that the time zone is in a single fixed offset for all values - which is not correct.  (See "Time Zone != Offset in [the timezone tag wiki on StackOverflow](http://stackoverflow.com/tags/timezone/info).)

The same mistake is repeated throughout many of the unit tests.  The tests also have a lot of noisy conversions back and forth between UTC and local time, many of which are unnecessary.

xuzhg:
Merged the PR. Thanks.

PR #335, opened 2015-05-08
Fix time zone issues

mj1856:
This fixes the issues described in #334.

### DIFF
--- a/OData/src/System.Web.OData/OData/Formatter/EdmPrimitiveHelpers.cs
+++ b/OData/src/System.Web.OData/OData/Formatter/EdmPrimitiveHelpers.cs
@@ -84,7 +84,7 @@ namespace System.Web.OData.Formatter
                     {
                         DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                         TimeZoneInfo timeZone = TimeZoneInfoHelper.TimeZone;
-                        dateTimeOffsetValue = dateTimeOffsetValue.ToUniversalTime().ToOffset(timeZone.BaseUtcOffset);
+                        dateTimeOffsetValue = TimeZoneInfo.ConvertTime(dateTimeOffsetValue, timeZone);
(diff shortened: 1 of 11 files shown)


OData/src/System.Web.OData/OData/Formatter/EdmPrimitiveHelpers.cs
```diff
@@ -84,7 +84,7 @@ public static object ConvertPrimitiveValue(object value, Type type)
                     {
                         DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                         TimeZoneInfo timeZone = TimeZoneInfoHelper.TimeZone;
-                        dateTimeOffsetValue = dateTimeOffsetValue.ToUniversalTime().ToOffset(timeZone.BaseUtcOffset);
+                        dateTimeOffsetValue = TimeZoneInfo.ConvertTime(dateTimeOffsetValue, timeZone);
```
xuzhg: what I do here is try to resolve the daylight saving problem.
This can make sure that I set a -8 timezone, and the payload result has -8:00, not -7:00.

mj1856: That's invalid, sorry.  The result is *required* to be -7 for values that are during daylight time.  By trying to "resolve the daylight saving problem", you're actually ignoring the reality that daylight saving time exists, and are thus *creating* a problem.

Remember, "time zone != offset".  Setting "Pacific Time" does not set "a -8 time zone". It sets a time zone whose *standard* offset is -8, and whose *daylight* offset is -7, with specific dates and times for the transitions between the two offsets, and specific history of the changes in those transitions.  (Time zones are much more complex than just a single number can offer.)

xuzhg:
@mj1856, I merged your PR. Thanks your contribution .

Issue #317, opened 2015-04-29
Support nullable referential constraint with conventional model builder


congysu:
For now, it is only supported with model builder, but not conventional model builder.
Consider to support the [following scenario](http://www.asp.net/web-api/overview/odata-support-in-aspnet-web-api/odata-v4/entity-relations-in-odata-v4):
```c# 
    public class Product
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public decimal Price { get; set; }
        public string Category { get; set; }

        // New code:    
        [ForeignKey("Supplier")]
        public int? SupplierId { get; set; }
        public virtual Supplier Supplier { get; set; }
    }
```

xuzhg:
Fixed at:  ef8c917

Issue #216, opened 2015-02-09
Nullable foreign key in 5.4.0


pascalberger:
After Update to Version 5.4.0 I receive the following error in my application:

> Error: The 'ObjectContent`1' type failed to serialize the response body for content type 
> 'application/json; odata=fullmetadata; charset=utf-8'.; The property 'MyProperty[Nullable=False]' of 
> type 'Edm.String' has a null value, which is not allowed.;

The correspondig model looks like this:

    public class MyModel
    {
        [Key]
        public int Id { get; set; }

        [ForeignKey("MyProperty")]
        public string MyPropertyId { get; set; }

        public virtual MyReferenceType MyProperty { get; set; }
    }

In the case where the error is thrown MyProperty has no value assigned. This worked fine in 5.3.1. I assume this has to do with the new [referential constraints](http://odata.github.io/WebApi/referential-constraint/), but I'm not sure what I need to change to make my code working again (meaning MyProperty should allow null)

xuzhg:
@pascalberger Did you make sure the MyProperty is an optional navigation property?

pascalberger:
How do you mean making it an optional navigation property? Since it is a string it can be null. I don't explicitely set it somewhere optional, but also never did in 5.3.1.

xuzhg:
@pascalberger 

[OData V4 spec](http://docs.oasis-open.org/odata/odata/v4.0/errata02/os/complete/part3-csdl/odata-v4.0-errata02-os-part3-csdl-complete.html#_Toc406397969) says:
The type of the dependent property MUST match the type of the principal property. If the navigation property on which the referential constraint is defined or the principal property is nullable, then the dependent property MUST be nullable. If both the navigation property and the principal property are not nullable, then the dependent property MUST be marked with the Nullable="false" attribute value.

==================================
So, the nullable value of "MyPropertyId" depends on the nullable of principal property and the nullable of navigation property.

Supposed the principal property is "PropertyId", So, the nullable mapping can be shown as:

|PropertyId|MyProperty |MyPropertyId 
--------------|----------------|------------------
true|true|true
true|false|true
false|true|true
false|false|false

note: true means nullable, false means non-nullable. 
So, can you show us your model builder codes?

pascalberger:
It's with OData v3 (not v4). 

Here's my (simplified) builder:

    var builder = new ODataConventionModelBuilder();
    builder.EntitySet<MyModel>("MyModels");
[...]

PR #340, opened 2015-05-12
Support nullable referential constraint with conventional model builder.

xuzhg:
The corresponding issues are: #216 and #317.


xuzhg:
Merged at: 

 ef8c917

PR #363, opened 2015-05-29
vNext alpha-1

lewischeng-ms:


lewischeng-ms:
Merged as https://github.com/OData/WebApi/commit/909465b1e1b3e4016be3cc49ecaa5c6d119a20a3

PR #374, opened 2015-06-03
NullReferenceException during serialization the complex value if it e…

xuzhg:
NullReferenceException during serialize the complex value if it equals to null.


VikingsFan:
Merged

brjohnstmsft:
Looks good to me.

Issue #129, opened 2015-01-12
Make ODataPathSegmentTranslator public so that developers can write their own odata path handler.


LianwMS:
Now if a developer write his own odata path handler, he has to duplicate the
ODataPathSegmentTranslator to translate the ODL odata path to webapi odata
path as it is an internal class.  

Work Item Details
--------------------
**Original CodePlex Issue:**	[Issue 2052](http://aspnetwebstack.codeplex.com/workitem/2052)
**Status:**	Proposed
**Reason Closed:**	Unassigned
**Assigned to:**	xuzhg
**Reported on:**	Jul 8, 2014 at 3:18 AM
**Reported by:**	jinfutan
**Updated on:**	Sep 30, 2014 at 2:55 AM
**Updated by:**	joeloughton

congysu:
Key as segment is a scenario for this issue.

xuzhg:
Changed. Thanks.

PR #378, opened 2015-06-05
Make the ODataPathSegmentTranslator class public.  

dmcallis:
This enables developers to write custom odata path handlers. Fix for bug: https://github.com/OData/WebApi/issues/129

### DIFF
--- a/OData/src/System.Web.OData/OData/Routing/ODataPathSegmentTranslator.cs
+++ b/OData/src/System.Web.OData/OData/Routing/ODataPathSegmentTranslator.cs
@@ -18,7 +18,7 @@ namespace System.Web.OData.Routing
     /// <summary>
     /// Translator to convert an OData path segment to WebAPI path segment.
     /// </summary>
-    internal class ODataPathSegmentTranslator : PathSegmentTranslator<IEnumerable<ODataPathSegment>>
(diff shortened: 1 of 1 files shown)


VikingsFan:
@dmcallis Thanks for your pull request! It's ok to set ODataPathSegmentTranslator into public, but you need to add corresponding unit test for it's public method.

please build before pull request, I found a CA1709 error after it turns to public, please help to suppress it.

OData/src/System.Web.OData/OData/Routing/ODataPathSegmentTranslator.cs
```diff
@@ -18,7 +18,7 @@ namespace System.Web.OData.Routing
     /// <summary>
     /// Translator to convert an OData path segment to WebAPI path segment.
     /// </summary>
-    internal class ODataPathSegmentTranslator : PathSegmentTranslator<IEnumerable<ODataPathSegment>>
+    public class ODataPathSegmentTranslator : PathSegmentTranslator<IEnumerable<ODataPathSegment>>
```
congysu: Change TranslateODLPathToWebAPIPath to such as: TranslateODataLibPathToWebApiPath, or add to a dictionary https://github.com/OData/WebApi/blob/master/OData/src/CodeAnalysisDictionary.xml

xuzhg:
Merged. Thanks.

Issue #377, opened 2015-06-03
Please make it easier to implement server-driven paging with custom queries


brjohnstmsft:
Currently server-driven paging is easy to support if you're using Web API OData with LINQ. `ODataQueryOptions.ApplyTo()` does the work for you. However, for services that don't use LINQ and instead process the OData query options ($filter, $orderby, etc.) and execute them directly, there is a lot of logic to implement. In particular, generating `@odata.nextLink` is non-trivial.

As a workaround, we are calling one of the overloads of `ODataQueryOptions.GetNextLink()` using reflection. It would be preferable if this method were public so that the logic can be re-used by anyone wanting to implement server-driven paging without using `ApplyTo()`.

This is a small change, and if you agree with it I would be happy to make the change myself and submit a PR.

congysu:
Thanks @brjohnstmsft very much for the feedback.  Please feel free to submit a PR.

brjohnstmsft:
Created: https://github.com/OData/WebApi/pull/384

VikingsFan:
Checkin 604b16c

julealgon:
@brjohnstmsft I just ran into exactly the scenario you describe and created an issue here:
https://github.com/OData/WebApi/issues/1633

Could you please advise on how to properly support NextLink and TotalCount using a purely custom query there? Even if I rely on the `Request.GetNextPageLink(int)` call, it still generates the link everytime so I wonder if I'd have to handle when to show and not to show it manually?

In my particular case, I cannot apply the query options since I'm using Dapper and stored procs to do the heavy lifting.

brjohnstmsft:
@julealgon You will need to implement your own logic to determine when to set `NextLink` and `TotalCount`. Actually setting them is pretty easy:

```
request.ODataProperties().TotalCount = totalCount;
request.ODataProperties().NextLink = request.GetNextPageLink(pageSize);
```
where `totalCount` and `pageSize` are values that you have to calculate yourself. Tip: to detect whether results fit entirely within a page or are truncated (i.e. -- to help you decide whether you need a `NextLink` or not), request one more row/record/document from your underlying data store than the page size. If the extra row comes back, you need a `NextLink`, otherwise you don't.

PR #384, opened 2015-06-08
Making ODataQueryOptions.GetNextPageLink public.

brjohnstmsft:
This will be useful for anyone processing OData query options directly (i.e. -- not using LINQ and ApplyTo()).

### DIFF
--- a/OData/src/System.Web.OData/OData/Query/ODataQueryOptions.cs
+++ b/OData/src/System.Web.OData/OData/Query/ODataQueryOptions.cs
@@ -516,11 +516,19 @@ namespace System.Web.OData.Query
         /// <returns>A next page link.</returns>
         public static Uri GetNextPageLink(HttpRequestMessage request, int pageSize)
         {
-            Contract.Assert(request != null);
-            Contract.Assert(request.RequestUri != null);
-            Contract.Assert(request.RequestUri.IsAbsoluteUri);
+            if (request == null || request.RequestUri == null)
+            {
+                throw Error.ArgumentNull("request");
+            }
 
-            return GetNextPageLink(request.RequestUri, request.GetQueryNameValuePairs(), pageSize);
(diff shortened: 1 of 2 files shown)


brjohnstmsft:
This is for this issue: https://github.com/OData/WebApi/issues/377

congysu:
Thanks @brjohnstmsft. Looks good to me.

OData/src/System.Web.OData/OData/Query/ODataQueryOptions.cs
```diff
@@ -508,7 +508,13 @@ public static IQueryable<T> LimitResults<T>(IQueryable<T> queryable, int limit,
...
-        internal static Uri GetNextPageLink(HttpRequestMessage request, int pageSize)
+        /// <summary>
+        /// Creates a link for the next page of results; To be used as the value of @odata.nextLink.
+        /// </summary>
+        /// <param name="request">The request on which to base the next page link.</param>
+        /// <param name="pageSize">The number of results allowed per page.</param>
+        /// <returns>A next page link.</returns>
+        public static Uri GetNextPageLink(HttpRequestMessage request, int pageSize)
         {
             Contract.Assert(request != null);
```
congysu: Thanks @brjohnstmsft. We need to check if the `request` is not null, and also the `request.RequestUri` is not null etc. for a public API; and add a few test cases.

brjohnstmsft: Ok; working on it now

brjohnstmsft: Fixed

VikingsFan:
Thanks for your pull request! Squash those two commit as one will be better :) ,merged 604b16ccbfee805f5fcaea2997680c3152119fcd

Issue #355, opened 2015-05-21
Convention model builder: duplicate properties


Casimodo72:
Using ASP identity v2 (Microsoft.AspNet.Identity.EntityFramework v2), the ODataConventionModelBuilder creates duplicate properties.
I'm using "Microsoft.AspNet.OData" version="5.6.0-rc-150519-Nightly"

Scenario: 

I'm deriving all identity classes.
E.g. `public class Role : IdentityRole<Guid, UserRole>`

I'm using the ODataConventionModelBuilder to build the identity EDM types. Note that I also have to explicitely define the key property, otherwise OData will bark about the missing key on that type.

EntityTypeConfiguration<Role> role = builder.EntitySet<Role>("Roles").EntityType;
role.HasKey(p => p.Id); 

Calling builder.GetEdmModel(), results in two "Id" properties being created on that type.
```xml
<EntityType Name="Role" OpenType="true">
    <Key>
        <PropertyRef Name="Id" />
    </Key>
    <Property Name="Id" Type="Edm.Guid" Nullable="false" />
    <Property Name="Index" Type="Edm.Int32" />
    <Property Name="DisplayName" Type="Edm.String" Nullable="false" />
    <Property Name="Id" Type="Edm.Guid" Nullable="false" />
</EntityType>
```
Apparently building the other identity relationship types also produces duplicate properties as reported here: http://stackoverflow.com/questions/30316344/duplicate-columns-in-odata-metadata

Trying to query for my Roles results in the following error:

"error": {
     "code": "",
     "message": "An error has occurred.",
     "innererror": {
         "message": "The 'ObjectContent1 type failed to serialize the response body for content type 'application/json; odata.metadata=minimal'.",
         "type": "System.InvalidOperationException",
         "stacktrace": "",
         "internalexception": {
             "message": "Ein primitiver Wert wurde angegeben, es wurde jedoch ein Wert des nicht primitiven Typs '' erwartet.",
             "type": "Microsoft.OData.Core.ODataException",
             "stacktrace": "   
bei Microsoft.OData.Core.ValidationUtils.ValidateIsExpectedPrimitiveType(Object value,
		 IEdmPrimitiveTypeReference valuePrimitiveTypeReference, IEdmTypeReference expectedTypeReference, Boolean bypassValidation)
bei Microsoft.OData.Core.JsonLight.ODataJsonLightValueSerializer.WritePrimitiveValue(Object value, IEdmTypeReference expectedTypeReference)
[...]

Issue #217, opened 2015-02-10
Same property exposed twice in complex type.


xuzhg:
For example:

```C#

    [DataContract]
    public class People
    {
        [IgnoreDataMember]
        public decimal Height { get; set; }
    }

    [DataContract]
    public class Employee : People
    {
        [DataMember]
        public long Id { get; set; }

        [DataMember]
        public int Age { get; set; }
    }

 var builder = new ODataConventionModelBuilder();
 var employee = builder.ComplexType<Employee>();
 employee.Property(p => p.Age).IsOptional();
 employee.Property(p => p.Height).IsOptional();
```

Edm XML schema generated:

```xml
<ComplexType Name="Employee">
<Property Name="Age" Type="Edm.Int32"/>
<Property Name="Height" Type="Edm.Decimal"/>
<Property Name="Id" Type="Edm.Int64"/>
<Property Name="Height" Type="Edm.Decimal" Nullable="false"/>      2nd Height Property?
</ComplexType>
```
Two questions here:
1.Although Height is marked as an [IgnoredDataMember], why is it still exposed in the Edm schema in OData V4? (note that in OData V3 the Height property in this case won’t be exposed though marked as optional)

2. Both Age and Height are set as IsOptional() properties, but why is Height property being exposed twice? 

3. The same problem maybe exist in entity type.

xuzhg:
It's the reason that the

```C# 
PropertyInfo
 ```
has the different value when access the same property from different class.

**Comment by Casimodo72** [comment]
@xuzhg,
this issue is not related to the PropertyInfo type, because in ODataConventionModelBuilder.MapStructuralProperty the property is looked up by its name.
Here the explicit property is set on the reflected property:
``` addedEdmProperty.AddedExplicitly = addedExplicitly; ```
This results in duplicate properties, plus conventions will *not* be applied to any of both properties. 
For me this behavior does not seem to make sense.

On the other hand, explicit navigation properties are overwriting reflected properties. In MapEntityType we have:
```c#
// don't add this property if the user has already added it.
if (!entity.NavigationProperties.Any(p => p.Name == property.Name))
```

In (https://github.com/OData/WebApi/issues/355) I have asked what ODataConventionModelBuilder is intended for. Does anyone actually know? Is there a design description?

Regards,
Kasimier Buchcik

Casimodo72:
@VikingsFan,
in the other thread you write that fixing this bug could break other assumptions. Can you point me to those assumptions or tests?

[...]

PR #385, opened 2015-06-08
Fix Issue #355 #217

VikingsFan:
Use name instead of propertyInfo to lookup

### DIFF
--- a/OData/src/System.Web.OData/OData/Builder/StructuralTypeConfiguration.cs
+++ b/OData/src/System.Web.OData/OData/Builder/StructuralTypeConfiguration.cs
@@ -268,21 +268,13 @@ namespace System.Web.OData.Builder
             ValidatePropertyNotAlreadyDefinedInDerivedTypes(propertyInfo);
 
             // Remove from the ignored properties
-            if (RemovedProperties.Contains(propertyInfo))
+            if (RemovedProperties.Any(prop => prop.Name.Equals(propertyInfo.Name)))
             {
-                RemovedProperties.Remove(propertyInfo);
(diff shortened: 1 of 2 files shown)


OData/test/UnitTest/System.Web.OData.Test/OData/Builder/Conventions/ODataConventionModelBuilderTests.cs
```diff
@@ -2671,6 +2671,58 @@ public void GetEdmModel_ThrowsExpcetion_ForRecursiveLoopOfComplexType()
                 "The complex type 'System.Web.OData.Builder.Conventions.RecursiveEmployee' has a reference to itself " +
                 "through the property 'Manager'. A recursive loop of complex types is not allowed.");
         }
+
```
xuzhg: remove

xuzhg:
So far, it looks good to me. However, it will be better to cover more scenarios with test cases.

xuzhg:
Merged. Thanks.

OData/src/System.Web.OData/OData/Builder/StructuralTypeConfiguration.cs
```diff
@@ -268,21 +268,13 @@ public virtual PrimitivePropertyConfiguration AddProperty(PropertyInfo propertyI
             ValidatePropertyNotAlreadyDefinedInDerivedTypes(propertyInfo);
 
             // Remove from the ignored properties
-            if (RemovedProperties.Contains(propertyInfo))
+            if (RemovedProperties.Any(prop => prop.Name.Equals(propertyInfo.Name)))
```
congysu: Does this work, if with property name aliasing, such as `[DataMember(Name="abc")]`?

Issue #394, opened 2015-06-10
Web API OData $expand doesn't return complex types


Sedatk:
this issue is mentioned in the following stackoverflow question

http://stackoverflow.com/questions/20137681/web-api-odata-expand-doesnt-return-complex-types

xuzhg:
@Sedatk  I saw that you had been fixing. What's about it?

xuzhg:
Close it. See more context in https://github.com/OData/WebApi/pull/396.

PR #396, opened 2015-06-10
Issue #394 Web API OData $expand doesn't return complex types

Sedatk:


OData/src/System.Web.OData/OData/Query/Expressions/SelectExpandWrapper.cs
```diff
@@ -128,8 +128,13 @@ public bool TryGetPropertyValue(string propertyName, out object value)
             // The user asked for all the structural properties on this instance.
             if (Instance != null)
             {
-                foreach (IEdmStructuralProperty property in type.StructuralProperties())
+                foreach (IEdmProperty property in type.DeclaredProperties)
                 {
+                    if(!property.Type.IsStructured())
```
congysu: Thanks @Sedatk.  Can you add a few test cases?

VikingsFan:
@Sedatk Thanks for you pull request, this is a bug!.
Actually I found the problem is in `Extensions\HttpActionDescriptorExtensions.cs`, line 37, the `isQueryCompositionMode` set to true, so when you use api controller, odata webapi can't discover the complex type, why not use odata controller as a work around?
or you can fix this issue, but it try not to break the case like `EnableQueryAttributeTest\CountValueReturnsAsContent_CountRequest` , which need to build as entity to enable query.

LianwMS:
@Sedatk Could you try ODataController instead of APIController?

Sedatk:
No i' ve written all my codes using api controller,
Thanks

VikingsFan:
@Sedatk your pull request break some test cases, please make sure all unit test is passed.
FYI http://odata.github.io/WebApi/#09-01-unittest-e2etest

Sedatk:
Thanks for your reply, but when i run unit tests under visual studio, it gives the following errors, so i ve not seen results of all cases, i think vs 2013 is incompatible with xunit ver 1.9.2, vs2012 gives the same errors too, i ve prepared my environment as described in 'http://odata.github.io/WebApi/#09-01-unittest-e2etest', waiting for your helps

[xUnit.net 00:00:00.0570404] Skipping: Microsoft.TestCommon (could not find dependent assembly 'Microsoft.TestCommon, Version=0.0.0')
[xUnit.net 00:00:00.1321294] Skipping: System.Web.Http.OData.Test (could not find dependent assembly 'System.Web.Http.OData.Test, Version=0.0.0')
[xUnit.net 00:00:00.2234573] Skipping: System.Web.OData.Test (could not find dependent assembly 'System.Web.OData.Test, Version=0.0.0')

VikingsFan:
In case xUnit Visual Studio Runner does not detect any test cases.  Try to install 0.9.5.4 instead the latest version.
  [xunit.runner.visualstudio-0.9.5.4.vsix](https://xunit.codeplex.com/downloads/get/755091)

We will consider upgrading to xUnit 2.0.

Sedatk:
Thanks for you reply, i ve updated my codes

congysu:
With query composition, types are mapped as entity types.  With auto expand, will this work? http://odata.github.io/WebApi/#04-15-autoexpand-attribute

Sedatk:
If i understand you correctly, 
no adding AutoExpand attribute on complex types or on container of complex types doesnt change the results in case "without my changes with this pull request", but it works as expected with my changes.

I d seen an easy fix as changing the value of parameter isQueryCompositionMode to false at the line 37 in HttpActionDescriptorExtensions.cs but it breaks many other test cases.

it seems a complex type should also be a structural type, without having any key, but the StructuralProperties extension method doesnt return complex types, so i have added IsStructuralOrComplex method, 

if something is wrong with my pull request please inform me

thanks

VikingsFan:
@Sedatk I have tried autoexpand, and it's working on your scenario, 

    public class Course
    {
       public int CourseId { get; set; }
       [AutoExpand]
       public virtual ICollection<TeeSet> TeeSets { get; set; }
    }
    public class TeeSet
    {
        public int TeeSetId { get; set; }
        public int CourseId { get; set; }
        [AutoExpand]
        public CourseRating MensRating { get; set; }
    }

Sedatk:
i had tested with my production code without virtual keyword(No load on demand scenario), because using odata with entity framework, expand query option can expand lists or collections without virtual keyword, in such scenario with virtual keyword there is no problem with complex types and no need AutoExpand attribute because when serializer needs such property then value is loaded already,

Thanks

VikingsFan:
@Sedatk your code is quiet suitable and working for your scenario! But this is a breaking change for our user, the code has changed the original behavior, our major work is support ODataController, so it is a long time for this behavior, I think most ApiController users are comfortable with this, if you want a private build, we are glad to help, or you can add something like a option to allow customer to choose turn on or off.

Sedatk:
Thanks, 
Will my codes acceptible if i add an attribute like AutoExpand to choose the behavior?

PR #416, opened 2015-06-24
Update to ODL 6.13-beta.

xuzhg:
Update to ODL 6.13-beta.
To use the capabilities vocabulary


congysu:
Looks good to me.  Do we also need to change the tooling in TFS?

xuzhg:
@congysu: the version in tooling is already changed to 5.7 beta.  And the tooling is in VSO.

xuzhg:
I also updated the reference in the E2E test project. See the commit. 820d37fe4e0677f9f27858e4a694158a96eebdbc

PR #424, opened 2015-07-06
Update HttpConfigurationExtensions.cs

victorantos:
fix variable misspelling


xuzhg:
Merged. Thanks for contribution.

congysu:
Thanks @victorantos.  We shall also fix:
https://github.com/OData/WebApi/blob/41b457a981b95545970b9ebf5a99bc517e66cff1/OData/src/System.Web.OData/OData/Routing/Conventions/AttributeRoutingConvention.cs#L56-L59

Issue #447, opened 2015-07-28
Default behaviour change for batch error handling in 5.7.0-beta


karataliu:
In previous version, the default behavior for handling batch is continue-on-error, but on 5.7.0 beta the default behavior becomes stop-on-error. This could be introduced by continue-on-error feature, please consider adding a flag to prevent this behavior change.

congysu:
This dues to fix of: https://github.com/OData/odata.net/issues/9.  We shall mention this in release notes, and allow service configured to use `continue-on-error`, and set batch preference-applied header.

VikingsFan:
checked in 9cbe805c0d75721d8ea5530fa7c12546c41aa3b9

PR #457, opened 2015-07-31
Issue #447 Default behaviour change for batch error handling in 5.7.0-beta

VikingsFan:


VikingsFan:
@chgarg we found last change for 'continue-on-error' is a breaking change (before 5.7 bata, this header is true by default, now is false), so we decide to add an flag in server side to decide turn on that or not, if you have time, would you please try it out and let us know any issues. Thanks.

OData/src/System.Web.OData/OData/Batch/DefaultODataBatchHandler.cs
```diff
@@ -42,7 +44,12 @@ public DefaultODataBatchHandler(HttpServer httpServer)
             IList<ODataBatchRequestItem> subRequests = await ParseBatchRequestsAsync(request, cancellationToken);
 
             string preferHeader = RequestPreferenceHelpers.GetRequestPreferHeader(request);
-            if (preferHeader != null && preferHeader.Contains(PreferenceContinueOnError))
+            var configuration = request.GetConfiguration();
+            if (configuration == null)
+            {
+                throw Error.InvalidOperation(SRResources.RequestMustContainConfiguration);
```
xuzhg: Can we not throw this exception and let it do the following if configuration == null

OData/test/UnitTest/System.Web.OData.Test/OData/Batch/UnbufferedODataBatchHandlerTest.cs
```diff
@@ -192,7 +206,7 @@ public void ProcessBatchAsync_ContinueOnError()
             });
             UnbufferedODataBatchHandler batchHandler = new UnbufferedODataBatchHandler(server);
             HttpRequestMessage batchRequest = new HttpRequestMessage(HttpMethod.Post, "http://example.com/$batch")
-            {
+            {
```
xuzhg: why add this?

xuzhg:
Others look good to me。

VikingsFan:
merged 9cbe805c0d75721d8ea5530fa7c12546c41aa3b9

PR #464, opened 2015-08-10
added UrlConventions support in DefaultODataPathHandler

BarclayII:


OData/src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
```diff
@@ -31,6 +32,15 @@ internal ODataUriResolverSetttings ResolverSetttings
         }
 
         /// <summary>
+        /// Gets or sets the <see cref="ODataUrlConventions"/> used by the parser.
+        /// </summary>
+        public ODataUrlConventions UrlConventions
```
congysu: Enable key as segment, while the OData () key syntax still works.

congysu:
Pending the new ODataSimplified key-as-segment convention to be released in OData v4 Lib 6.14.

congysu:
OData v4 Lib 6.14 released.  For new key-as-segment convention, please use [ODataUrlConventions ODataSimplified](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/ODataUrlConventions.cs#L58).

VikingsFan:
Merged 33c370b79fe055a57fbd6e3767a007437f2f8c41

Issue #459, opened 2015-08-03
GetConcurrencyProperties should look at OptimisticConcurrency annoataion


karataliu:
https://github.com/OData/WebApi/blob/d21a410fb1b36b9455796dd18d5bbdcf10c2f5ad/OData/src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs#L421-L425

Here it refers to EdmConcurrencyMode, which is obsoleted in OData V4. This leads incorrect ETag computation in service.

Issue #190, opened 2015-01-15
Add Concurrency Support In WebAPI V3


LianwMS:
Support ConcurrencyCheck attribute in reflector model 
Support set ConcurrencyCheck on a primitive property 
Support Etag

LianwMS:
Check in with commit 03af4da00d97af7eb465688e74279d813782cf54

PR #474, opened 2015-08-12
ETag annotation support

xuzhg:
Related issues:

https://github.com/OData/WebApi/issues/459

#190


xuzhg:
@Teams, I can't see the value to save the concurrencyProperties in EntityInstanceContext.  EntityInstanceContext is used temporary in entity serializer and will be destroyed after serialization. It's same for creating ETag, we only create once for each entity instance. Besides, we can't save it globally in static field because EntityInstanceContext is used for any navigation source.  
However, I also don't think it will be a performance problem.

Issue #479, opened 2015-08-15
xUnit runner extension to nuget


congysu:
The runner extension won't install, and it not supported.  Shall change to nuget runner.

```
8/15/2015 6:52:31 PM - Microsoft VSIX Installer
8/15/2015 6:52:31 PM - -------------------------------------------
8/15/2015 6:52:31 PM - Initializing Install...
8/15/2015 6:52:31 PM - Extension Details...
8/15/2015 6:52:31 PM - 	Identifier      : xunit.runner.visualstudio.Microsoft.623f78dd-6d4b-48f3-a2c7-4bbb381bb146
8/15/2015 6:52:31 PM - 	Name            : xUnit.net runner for Visual Studio
8/15/2015 6:52:31 PM - 	Author          : Outercurve Foundation
8/15/2015 6:52:31 PM - 	Version         : 0.99.9
8/15/2015 6:52:31 PM - 	Description     : NO LONGER NEEDED. Please un-install this extension. See http://xunit.github.io/docs/vsix-deprecated.html
8/15/2015 6:52:31 PM - 	Locale          : en-US
8/15/2015 6:52:31 PM - 	MoreInfoURL     : http://xunit.github.io/docs/vsix-deprecated.html
8/15/2015 6:52:31 PM - 	InstalledByMSI  : False
8/15/2015 6:52:31 PM - 	SupportedFrameworkVersionRange : [3.5,)
8/15/2015 6:52:31 PM - 
8/15/2015 6:52:31 PM - 	Supported Products : 
8/15/2015 6:52:31 PM - 		Microsoft.VisualStudio.Ultimate
8/15/2015 6:52:31 PM - 			Version : [11.0]
8/15/2015 6:52:31 PM - 		Microsoft.VisualStudio.Ultimate
8/15/2015 6:52:31 PM - 			Version : [12.0]
8/15/2015 6:52:31 PM - 		Microsoft.VisualStudio.Ultimate
8/15/2015 6:52:31 PM - 			Version : [14.0]
8/15/2015 6:52:31 PM - 		Microsoft.VisualStudio.Premium
8/15/2015 6:52:31 PM - 			Version : [11.0]
8/15/2015 6:52:31 PM - 		Microsoft.VisualStudio.Premium
8/15/2015 6:52:31 PM - 			Version : [12.0]
8/15/2015 6:52:31 PM - 		Microsoft.VisualStudio.Premium
8/15/2015 6:52:31 PM - 			Version : [14.0]
[...]

PR #480, opened 2015-08-15
xUnit runner extension to nuget #479

congysu:
Use nuget version of xUnit Visual Studio runner.


LianwMS:
Submit in cc639c352a22a3cfeacfa36444a625a0e182ae77

Issue #501, opened 2015-09-03
Migrate vNext to Beta7


gfrappier:
Now that the beta7 framework is officially available, we should move vNext to this version.

lewischeng-ms:
Merged pull request #502. Thanks for the contribution.

PR #502, opened 2015-09-03
Beta7

gfrappier:
Issue: https://github.com/OData/WebApi/issues/501

- Migrated the code to use the beta7 runtime
- Added support for primitive types in the DefaultODataModelProvider since strings and doubles where mapped to System.String or System.Double instead of Edm.String and Edm.Double
- Added support for route prefix in ODataProperties which is used in the UrlHelperExtensions to create OData link from the request
- Simplified the customer controller action calls


lewischeng-ms:
@gfrappier Thanks for your contribution. Merged as https://github.com/OData/WebApi/commit/aa48fa9d273e2f42537d4a324c5fdb22835e3e30 to https://github.com/OData/WebApi/commit/bc25e08957324911a576f57e007849ccc7abde65.

PR #527, opened 2015-10-13
Support for beta 8

EvanMachusak:
Updated project references to beta 8 and sometimes RC 1.
Fixed resultant compilations errors.
Dropped Helios references in the sample project.
Tested the sample project lightly (i.e., confirmed the Products controller returns a data set as expected).


vNext/Microsoft.AspNet.OData.6.0.0-beta8.nuspec
```diff
@@ -0,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
+    <metadata>
+        <id>Microsoft.AspNet.OData</id>
+        <version>6.0.0-beta8-16000</version>
+        <title>Microsoft.AspNet.OData</title>
+        <authors>Evan</authors>
```
lewischeng-ms: Please remove personal information. Also in line 9.

vNext/src/Microsoft.AspNet.OData/Query/ODataQueryOptions.cs
```diff
@@ -74,6 +76,22 @@ public ODataQueryOptions(ODataQueryContext context, HttpRequest request)
         public SelectExpandQueryOption SelectExpand { get; private set; }
 
         /// <summary>
+        /// Specifies a non-negative integer n that excludes the first n items of the queried collection from the result.         
+        /// </summary>
+        /// <remarks>
+        /// Corresponds to the $skip query option.  The service returns items starting at position n+1.
+        /// </remarks>
+        public int? Skip { get; private set; }
```
lewischeng-ms: We can consider use `long?` for both `Skip` and `Top` instead of `int?`.

vNext/src/Microsoft.AspNet.OData/Query/ODataQueryOptions.cs
```diff
@@ -97,6 +115,25 @@ public virtual IQueryable ApplyTo(IQueryable query, ODataQuerySettings querySett
...
+            {
+                take = Math.Min(querySettings.PageSize.Value, int.MaxValue);
+            }
+            if (Top.HasValue)
+            {
+                take = Math.Min(Top.Value, take ?? int.MaxValue);
+            }
+            if (take.HasValue)
+            {
+                query = ExpressionHelpers.Take(query, take.Value, Context.ElementClrType, false);
```
EvanMachusak: The ExpressionHelpers class uses reflection to invoke the LINQ extension methods Skip and Take respectively, and those methods both accept int arguments, not longs.  If we want to promote our $skip and $top parameters to support long integers, we'd need to implement our own edition of the Skip and Take methods.  Thoughts?

vNext/src/Microsoft.AspNet.OData/Query/ODataQueryOptions.cs
ilmax: Can we use unsigned here? does it make sense to take -3 rows?

EvanMachusak: The code which parses the value of $skip and $top throws an ODataException for negative values.  I prefer to avoid uints because they're not CLS compliant.  uint doubles the range of the parameter - but do we think a $skip or $top query with a value above Int32.MaxValue is particularly useful?
Date: Thu, 15 Oct 2015 11:20:38 -0700

VikingsFan:
Merged with https://github.com/OData/WebApi/commit/ec2047a0894e3cf352e0c61021a5b2060e312cd2

Thanks for your contribution:)

PR #530, opened 2015-10-19
Update DeltaOfTEntityType.cs

jhgoodwin:
Add better handling for Enum types which do not seem to come over right from json during patch.

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -1,10 +1,13 @@
 OData-WebAPI
 
-Copyright (c) Microsoft Corporation
+Copyright (c) Microsoft. All rights reserved.
 
-All rights reserved. 
+Material in this repository is made available under the following terms:
+  1. Code is licensed under the MIT license, reproduced below.
+  2. Documentation is licensed under the Creative Commons Attribution 3.0 United States (Unported) License.
+     The text of the license can be found here: http://creativecommons.org/licenses/by/3.0/legalcode
 
-MIT License
+The MIT License (MIT)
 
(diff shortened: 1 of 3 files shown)


VikingsFan:
Thanks @jhgoodwin for your contribution! Is this problem also occur in V4? under /OData/src/System.Web.OData/

Also please add test case to verify your change, instruction: https://github.com/OData/WebApi/blob/master/CONTRIBUTION.md

xuzhg:
@jhgoodwin Would you please give us more context by filing a bug? Because, as I investigate, it seems the enum type works in DeltaOfTEntityType. See my PR #535. Would you please take a look and let us know your concern? Thanks.

jhgoodwin:
I'll write a unit test and file a bug.  Just had a deadline we had to meet (delivery is tomorrow).  In the PR #535, the proof of concept only seems to show string -> enum conversions, but in my case, int -> enum conversions were failing.

xuzhg:
@jhgoodwin I updated my PR #535 to add the case about int -> enum conversions. It works at my side. So, I am appreciate that you can write an unit test and file a bug. Thanks.

jhgoodwin:
I put a comment on PR #535 .  This sort of workaround will only be necessary if you want to build in testing for how your library will interact with a project using the Newtonsoft.Json library instead of your serializers.

xuzhg:
@jhgoodwin Thanks your feedback and warmly welcome more feedbacks from you. If you have further concern, please reopen it. Thanks.

PR #535, opened 2015-10-23
Add e2e test cases for enum type in DeltaOfTEntity

xuzhg:
For #530 PR, I add some e2e test cases to verify.

### DIFF
--- a/OData/packages/repositories.config
+++ b/OData/packages/repositories.config
@@ -8,6 +8,7 @@
   <repository path="..\test\E2ETest\WebStack.QA.Instancing\packages.config" />
   <repository path="..\test\E2ETest\WebStack.QA.Share\packages.config" />
   <repository path="..\test\E2ETest\WebStack.QA.Test.OData\packages.config" />
+  <repository path="..\test\E2ETestV3\WebStack.QA.Test.OData\packages.config" />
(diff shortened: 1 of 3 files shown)


jhgoodwin:
Hi xuzhg, okay, I see the problem.

Our app is using Newtonsoft.Json for serialization while these tests are using the OData ones. 

Minor suggestion: DeltaOfEntityType.cs line 54 </param> should be moved up one line.
Medium suggestion your unit tests do not test both input *AND* output to verify the saved values were applied as expected.  I tried to do this by adding:
        private const string PreferHeaderName = "Prefer";
        private const string ReturnContentHeaderValue = "return-content";
Then in the Patch method:
        request.Headers.Add(PreferHeaderName, ReturnContentHeaderValue)
Then reading the content:
        var content = response.Content.ReadAsStringAsync();
But, the problem is that the OData Serializer doesn't seem to like your nullable collection of colors that have null items.  I think you may have to work around it by invoking a get after the patch to verify the object is correctly set.

This lead me to the answer of what is happening.  Basically, the OData serializer does a Convert to get the value from the json stream right onto the destination property, where the Newtonsoft.Json library uses their own dynamic intermediary object.  This seems to be causing the issue I am encountering.

Issue #482, opened 2015-08-18
Web API does not support Edm.Date literal in $filter when the property is Edm.Date [Nullable=True] and the backend is EF


lewischeng-ms:
I am using EF as the backend LINQ provider. If I have a property called `Birthday` of type `Edm.Date` with `Nullable=True`, the following query:

    $filter=Birthday eq 1986-01-01

would cause EF to throw exception: 

    Unable to create a constant value of type 'System.Nullable`1[[Microsoft.OData.Edm.Library.Date, Microsoft.OData.Edm, Version=6.13.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35]]'. Only primitive types or enumeration types are supported in this context.

If I set the `Nullable` to `False`, everything is fine.

xuzhg:
Fixed. Thanks.

PR #541, opened 2015-10-28
Issue #482: Web API does not support Edm.Date literal in $filter when…

xuzhg:
Issue #482: Web API does not support Edm.Date literal in $filter when the property is Edm.Date [Nullable=True] and the backend is EF. Thanks.

### DIFF
--- a/OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
+++ b/OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
@@ -1522,7 +1522,7 @@ namespace System.Web.OData.Query.Expressions
                 left = ConvertToEnumUnderlyingType(left, enumType, enumUnderlyingType);
                 right = ConvertToEnumUnderlyingType(right, enumType, enumUnderlyingType);
             }
-
+           
             if (leftUnderlyingType == typeof(DateTime) && rightUnderlyingType == typeof(DateTimeOffset))
             {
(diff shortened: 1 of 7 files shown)


OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
```diff
@@ -1692,6 +1696,39 @@ private Expression CreateTimeBinaryExpression(Expression source)
...
 
+        private Expression ConvertToDateTimeRelatedConstExpression(Expression source)
+        {
+            var parameterizedConstantValue = ExtractParameterizedConstant(source);
+            if (parameterizedConstantValue != null)
+            {
+                var dateTimeOffset = parameterizedConstantValue as DateTimeOffset?;
+                if (dateTimeOffset != null)
+                {
+                    source = Expression.Constant(dateTimeOffset.Value, typeof(DateTimeOffset));
```
lewischeng-ms: better directly **return** in each if-body to avoid redundant conditions.

lewischeng-ms:
Looks good to me. @xuzhg Thanks for fixing the issue.

OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
xuzhg: Good.

xuzhg:
Merged. Thanks.

Issue #427, opened 2015-07-10
CultureInfo property can't be seraialized


marinkobabic:
The entity type 'System.Globalization.Calendar' cannot be configured as a complex type because the derived type 'System.Globalization.GregorianCalendar' is already configured as an entity type.

To reproduce the exception:

            ODataModelBuilder modelBuilder = new ODataConventionModelBuilder();
            modelBuilder.EntitySet<CultureInfo>("cultureInfo");
            IEdmModel model = modelBuilder.GetEdmModel();
            config.MapODataServiceRoute("ODataRoute", "odata", model);

or use a Entity which has CultureInfo as property. This worked in previous versions.

congysu:
@marinkobabic Which versions does this work in?  Thanks.

xuzhg:
Fixed. Thanks.

PR #542, opened 2015-10-28
Issue #427: CultureInfo property can't be serialized.

xuzhg:
Issue #427: CultureInfo property can't be serialized.


xuzhg:
Merged.

Issue #302, opened 2015-04-10
Swagger json generation for OData - compatible with Azure App Service


joshbooker:
Azure App Service requires swagger metadata.  Swashbuckle is used in VS APIApp tooling to create the metadata using APIExplorer.  There are two problems preventing this from working in WebAPI OData projects:
1)  Odata Controller explicitly prevents APIExplorer
2)  Swashbuckle cannot handle routes and parameters for OData endpoints

Related issues:
https://github.com/OData/WebApi/issues/278
https://github.com/OData/ODataSamples/issues/13
https://github.com/OData/RESTier/issues/69

What is needed:
1) ability to shut off ignore APIExplorerSettings
2) swagger metadata builder tool, preferably customized swashbuckle swagger.config, to generate swagger metadata compatible with Azure APIApp.

Thanks for listening.

congysu:
Need to support e.g.: `/odata/swagger.json`

congysu:
Related issue: https://github.com/OData/odata.net/issues/129

abkmr:
JSON metadata is also needed for Dynamics CRM

joshbooker:
@TomDu,

I see from the latest PR #520 that swagger.json generation was added to WebAPIOdata...many thanks to OData team for that.

However, there is still a problem with the resulting swagger output in interoperability with Azure.  For example, AutoRest is a tool which generates client code from swagger.json and it requires OperationIDs on all operations.

https://github.com/Azure/autorest

The swagger being output from WebAPIOData is missing the required OperationIDs.

Not sure if this can be fixed before 5.8 release but it would be really swell if the resulting swagger was Azure compliant.

xuzhg:
Added. Thanks.

congysu:
operationId support is added in commit https://github.com/OData/WebApi/commit/d63e4e3e927754ecad45949d766a60cab6362704.
A pending work item is to ensure that the generated swagger.json passes swagger validation https://github.com/OData/lab/issues/25.

PR #552, opened 2015-11-11
Issue #302 Add operationId for Swagger json generation

TomDu:
Generate OperationIDs  in Swagger json payload.

### DIFF
--- a/OData/src/System.Web.OData/OData/ODataSwaggerUtilities.cs
+++ b/OData/src/System.Web.OData/OData/ODataSwaggerUtilities.cs
@@ -33,6 +33,7 @@ namespace System.Web.OData
                 {
                     "get", new JObject()
                         .Summary("Get EntitySet " + entitySet.Name)
+                        .OperationId(entitySet.Name + "_Get")
                         .Description("Returns the EntitySet " + entitySet.Name)
                         .Tags(entitySet.Name)
                         .Parameters(new JArray()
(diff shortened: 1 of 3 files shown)


joshbooker:
@TomDu,

Wow...Thanks! ...Ask and you a shall receive! I haven't tried this yet, but have one more suggestion:

OperationIds are used in AutoRest to generate Classes using a naming convention. For example, consider the following swagger:

"paths": {
 "/Customers": {
 "get": {
 "operationId": "Customers_Get",
 ..........
 "/Customers({Id})": {
 "get": {
 "operationId": "Customers_GetById",

If there is an underscore in the OperationId, AutoRest will use the convention [ClassName_MethodName] to generate a Customers class having .Get and .GetById as methods.

More info:
https://github.com/Azure/autorest/blob/master/Documentation/defining-clients-swagger.md#generating-operation-classes

It would be really swell if OData swagger used this convention for OperationIds.

Also see this early example of how WebAPI team handles multi parameter operations:

https://azure.microsoft.com/en-us/documentation/articles/app-service-api-dotnet-swashbuckle-customize/

I believe WebAPI team recently settled on a swashbuckle IOperationFilter that handles this, but cannot find a blog about that. Maybe @bradygaster has more info.

Thanks for listening!

xuzhg:
Merged. Thanks.

Issue #543, opened 2015-10-28
EditLink is wrong in 5.7


wkit23:
In order to obtain the URL of a referred entity, referred to by a navigation property we use the full metadata like e.g.

```
http://localhost/MyResource(1)/MyNavigationProperty?$format=application/json;odata.metadata=full
```

which will return the following result in 5.6

```
{
    "@odata.editLink":"http://localhost/MyNavigationProperty(1)"
}
```

However when we upgraded to 5.7, the returned result became

```
{
    "@odata.editLink":"http://localhost/MyResource(1)/MyNavigationProperty(1)"
}
```

which apparently is wrong. 

I suspect this is caused by [https://github.com/OData/WebApi/pull/372]. Any idea? We mainly use this link to patch the odata entity.

VikingsFan:
@ificator I've found that your PR #372 is for navigation prop, not only for contained navigation prop, so it's a breaking change for `id` and `editlink`, I'm going to fix this bug but I'm not sure it will break your service or not because you just add one test case. Can your tell me your scenario or provide a fix?

VikingsFan:
@wkit23 Would you like to review my PR #554 ? Thanks!

wkit23:
@VikingsFan I tried using your PR #554. It seems to work perfectly. Thanks!

VikingsFan:
Merged 86e222607410a7bb61698b8f2ce69a0b1f24d901

PR #554, opened 2015-11-12
Fix Issue #543 EditLink is wrong in 5.7

VikingsFan:
Use default behavior when target is not contained navigation property
When target is contained navigation property, just keep the last non-contained navigation srouce.

### DIFF
--- a/OData/test/E2ETest/WebStack.QA.Test.OData/ODataPathHandler/LinkGenerationTests.cs
+++ b/OData/test/E2ETest/WebStack.QA.Test.OData/ODataPathHandler/LinkGenerationTests.cs
@@ -1,4 +1,7 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Web.Http;
 using System.Web.OData;
 using System.Web.OData.Builder;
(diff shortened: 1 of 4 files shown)


OData/test/E2ETest/WebStack.QA.Test.OData/ODataPathHandler/LinkGenerationTests.cs
```diff
@@ -89,5 +105,19 @@ public void GeneratedLinkShouldMatchRequestRouting()
             content = this.Client.GetStringAsync(this.BaseAddress + "/v2/LinkGeneration_Model2").Result;
             Assert.DoesNotContain(@"/v1/LinkGeneration_Model2", content);
         }
+
+        [Theory]
+        [InlineData("/v1/LinkGeneration_Model1(1)/NonContainedNavigationProperty", "/v1/LinkGeneration_Model2(2)")]
```
xuzhg: Only one test case, it's better to use [Fact] and put the cast condition into test body.

OData/test/E2ETest/WebStack.QA.Test.OData/ODataPathHandler/LinkGenerationTests.cs
```diff
@@ -89,5 +105,19 @@ public void GeneratedLinkShouldMatchRequestRouting()
...
+        public void GeneratedLinkTestForNavigationProperty(string requestUrl, string expectLinkUrl)
+        {
+            string AcceptJsonFullMetadata = "application/json;odata.metadata=full";
```
xuzhg: can remove the random variable if it's only used once.

xuzhg:
looks good.

VikingsFan:
Merged 86e222607410a7bb61698b8f2ce69a0b1f24d901

PR #553, opened 2015-11-11
Issue #349: $expand contain $count don't work

xuzhg:
Issue #349: $expand contain $count don't work

PR #556, opened 2015-11-13
Issue #560: Unnecessary casts in expression when querying properties of the base class for the inheritor

postoronnimw:
Remove unnecessary casts from expression when querying properties of the base class for the inheritor.

### DIFF
--- a/OData/src/System.Web.Http.OData/OData/Query/Expressions/SelectExpandBinder.cs
+++ b/OData/src/System.Web.Http.OData/OData/Query/Expressions/SelectExpandBinder.cs
@@ -115,32 +115,30 @@ namespace System.Web.Http.OData.Query.Expressions
             IEdmEntityType declaringType = property.DeclaringType as IEdmEntityType;
             Contract.Assert(declaringType != null, "only entity types are projected.");
 
-            Expression propertyName;
-
             // derived navigation property using cast
             if (elementType != declaringType)
             {
+                Type originalType = EdmLibHelpers.GetClrType(elementType, _model);
                 Type castType = EdmLibHelpers.GetClrType(declaringType, _model);
                 if (castType == null)
                 {
                     throw new ODataException(Error.Format(SRResources.MappingDoesNotContainEntityType, declaringType.FullName()));
                 }
 
-                // Expression
-                //          source is navigationPropertyDeclaringType ? propertyName : null
-                propertyName = Expression.Condition(
-                    test: Expression.TypeIs(source, castType),
-                    ifTrue: Expression.Constant(property.Name),
-                    ifFalse: Expression.Constant(null, typeof(string)));
-            }
-            else
-            {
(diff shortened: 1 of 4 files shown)


VikingsFan:
Hi @postoronnimw , thanks for contribute to our lib! Would you like to add some test to verity your code and later code change won't break your change. You can refer to other PR like #553 or this [page](https://github.com/OData/WebApi/blob/master/CONTRIBUTION.md)

OData/src/System.Web.Http.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -115,32 +115,30 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
-                    test: Expression.TypeIs(source, castType),
-                    ifTrue: Expression.Constant(property.Name),
-                    ifFalse: Expression.Constant(null, typeof(string)));
-            }
-            else
-            {
-                // Expression
-                //          "propertyName"
-                propertyName = Expression.Constant(property.Name);
+                if (!castType.IsAssignableFrom(originalType))
```
xuzhg: Is this check necessary?

postoronnimw:
I created Issue #560 .

OData/src/System.Web.Http.OData/OData/Query/Expressions/SelectExpandBinder.cs
postoronnimw: This check allows to remove cast from expression.

OData/src/System.Web.Http.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -115,32 +115,30 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
+                {
+                    // Expression
+                    //          source is navigationPropertyDeclaringType ? propertyName : null
+                    return Expression.Condition(
+                        test: Expression.TypeIs(source, castType),
+                        ifTrue: Expression.Constant(property.Name),
+                        ifFalse: Expression.Constant(null, typeof(string)));
+                }
```
VikingsFan: Is it possible that castType.IsAssignableFrom(originalType) is false, and Expression.TypeIs(source, castType) is true?

postoronnimw: @VikingsFan yes, it is possible. You may review test SelectExpandBinderTest.CreatePropertyNameExpression_DerivedProperty_ReturnsConditionalExpression

VikingsFan:
merged 621f256847ca2923a8c5000bff84333236c9f7bc revert the V3 part as we didn't have plan to release it.

PR #573, opened 2015-12-01
Enable to serialize the 'null' dynamic propery by config

xuzhg:
See related issue: https://github.com/OData/WebApi/pull/313


xuzhg:
Merged. Thanks.

PR #580, opened 2015-12-05
ProcedureConfiguration API to support types known at runtime

YogiBear52:
- ProcedureConfiguration - Add a parameter by the given clrType (not only with the generic API)
- FunctionConfiguration/ActionConfiguration - Set a return type with the given clrType(not only with the generic API)

### DIFF
--- a/OData/src/System.Web.OData/OData/Builder/ActionConfiguration.cs
+++ b/OData/src/System.Web.OData/OData/Builder/ActionConfiguration.cs
@@ -133,19 +133,33 @@ namespace System.Web.OData.Builder
         /// Established the return type of the Action.
         /// <remarks>Used when the return type is a single Primitive or ComplexType.</remarks>
         /// </summary>
-        [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "In keeping with rest of API")]
-        public ActionConfiguration Returns<TReturnType>()
+        public ActionConfiguration Returns(Type clrReturnType)
         {
-            Type returnType = typeof(TReturnType);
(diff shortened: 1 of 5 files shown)


xuzhg:
@YogiBear52  Would you please give us a scenario description why these functions (APIs) are necessary for you? Thanks.

YogiBear52:
Sure,
I am trying to build and EdmModel out of a DataModel which is known only at runtime.
Entities,entity sets, functions and complex types.
These days I'm writing code in OData.WebApi and ODataLibs to help customize and be build all odata endpoint functonality for data models known at runtime.

The current API of function configuration is the generic method AddParameter<T>. Since I don't know the type T at compile time, I can't use this method (unless I use reflection, but it is not a good coding..)

Thanks.

OData/src/System.Web.OData/OData/Builder/ActionConfiguration.cs
```diff
@@ -133,22 +133,36 @@ public ActionConfiguration HasActionLink(Func<EntityInstanceContext, Uri> action
         /// Established the return type of the Action.
         /// <remarks>Used when the return type is a single Primitive or ComplexType.</remarks>
         /// </summary>
-        [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "In keeping with rest of API")]
-        public ActionConfiguration Returns<TReturnType>()
+        public ActionConfiguration Returns(Type clrReturnType)
```
VikingsFan: can you add test case for this new API, same for the others

xuzhg:
Merged. Thanks.

PR #584, opened 2015-12-09
Issue #559 : Using object key for null check in expression

### LINKED ISSUE #559: Query from DB only properties that declared in the projection
I have domain model something like this:

```C#
public class TestEntity
{
    public virtual long Id { get; protected set; }
    public virtual TestDerivedSubEntity SubEntity { get; set; }
}
public class TestSubEntity
{
    public virtual long Id { get; protected set; }
    public virtual string Code { get; set; }
}
public class TestDerivedSubEntity : TestSubEntity
{
    public virtual string Name { get; set; }
	...
}
```

I execute following query from client:

```C#
var query = _dataContext.TestEntities.Select(c => new {c.Id, c.SubEntity.Code});
```

wich builds the following request:

```
http://localhost:51000/ResearchService/TestEntities?$expand=SubEntity($select=Code)&$select=Id
```

In built expression i see property

```C#
IsNull = $var1.SubEntity == null
```

Consequently in sql query (with NHibernate mapper) i see all properties of the TestDerivedSubEntity (not only Code):

```SQL
select testentity0_.id as col_0_0_,
	 case when testsubent1_.e_type='Derived' then 1 else 0 end as col_1_0_,
	 testentity0_.id_subentity as col_2_0_,
	 testsubent1_.code as col_3_0_,
	 testsubent1_.id as col_4_0_,
	 testsubent1_.id as id1_,
	 testsubent1_.code as code1_,
	 testsubent1_.name as name1_,
	 testsubent1_.e_type as e2_1_ 
from t_entity testentity0_ 
left outer join t_subentity testsubent1_ 
	on testentity0_.id_subentity=testsubent1_.id
```

I have deleted check `IsNull = $var1.SubEntity == null` from expression and now sql query contains only needed properties.
Also i noticed that comment on `NamedPropertyExpression.NullCheck` contains text about EntityFramework specifics:

```C#
// Checks whether this property is null or not. This is required for expanded navigation properties that are null as entityframework cannot
// create null's of type SelectExpandWrapper<ExpandedProperty> i.e. an expression like 
//       => new NamedProperty<Customer> { Value = order.Customer == null : null : new SelectExpandWrapper<Customer> { .... } } 
// cannot be translated by EF. So, we generate the following expression instead,
//       => new ExpandProperty<Customer> { Value = new SelectExpandWrapper<Customer> { .... }, IsNull = nullCheck }
// and use Value only if IsNull is false.
public Expression NullCheck { get; set; }
```

Which is there a correct way to select only need properties? Is it specifically for NHibernate only?

postoronnimw:
This PR resolves #559
Before: IsNull = (Param_1.Customer == null)
Now: IsNull = (Convert(Param_1.Customer.ID) == null)

It is necessary for Query from DB only properties that declared in the projection (with NHibernate mapper).

### DIFF
--- a/OData/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
+++ b/OData/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
@@ -391,7 +391,7 @@ namespace System.Web.OData.Query.Expressions
                 Expression propertyName = CreatePropertyNameExpression(elementType, propertyToExpand, source);
                 Expression propertyValue = CreatePropertyValueExpressionWithFilter(elementType, propertyToExpand, source,
                     expandItem.FilterOption);
-                Expression nullCheck = Expression.Equal(propertyValue, Expression.Constant(null));
+                Expression nullCheck = GetNullCheckExpression(propertyToExpand, propertyValue, projection);
 
                 Expression countExpression = CreateTotalCountExpression(propertyValue, expandItem);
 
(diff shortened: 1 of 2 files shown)


OData/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -391,7 +391,7 @@ private Expression CreateTotalCountExpression(Expression source, ExpandedNavigat
                 Expression propertyName = CreatePropertyNameExpression(elementType, propertyToExpand, source);
                 Expression propertyValue = CreatePropertyValueExpressionWithFilter(elementType, propertyToExpand, source,
                     expandItem.FilterOption);
-                Expression nullCheck = Expression.Equal(propertyValue, Expression.Constant(null));
+                Expression nullCheck = GetNullCheckExpression(propertyToExpand, propertyValue, projection);
```
VikingsFan: https://github.com/OData/WebApi/blob/f4573605bd88dabe882c4dc45e4a4ed0e55b32e0/OData/src/System.Web.OData/OData/Query/HandleNullPropagationOptionHelper.cs#L10

Can we check the LINQ provider and change the behavior?

postoronnimw: @VikingsFan `IsNull` property presence does not depend on the value of HandleNullPropagation.
When `HandleNullPropagation == HandleNullPropagationOption.False` then 
```C#
IsNull = (Convert(Param_1.Customer.ID) == null)
```
When `HandleNullPropagation == HandleNullPropagationOption.True` then 
```C#
IsNull = (IIF((IIF((Param_1 == null), null, Param_1.Customer) == null), null, Convert(IIF((Param_1 == null), null, Param_1.Customer).ID)) == null)
```

VikingsFan: @postoronnimw Yes, I mean you have changed the original behavior to support NHibernate, can  we change the behavior depends on the LINQ provider, for example, is the provider is EF, keep the old behavior, if it is NHibernate, use the new expression

VikingsFan:
This PR resolves #559

VikingsFan:
Merged 14db4e498b46f810cde37f64bdf3451608fe575b

PR #585, opened 2015-12-10
[vNext] Returning proper EDM type for procedures returning primitive type

gfrappier:
Modified the default model provider to support the primitive return type
functionality. Added a simple test method to the sample.


xuzhg:
Merged at: 7843c9323d1f019c91e04ec896ea11175f677844

Thanks your contribution.

Issue #601, opened 2015-12-25
ODataRoute should accept an interface instead of ODataPathRouteConstraint


karunasagark:
The constructor of ODataRoute() accepts an object of type ODataPathRouteConstraint. However, this causes tight coupling.
Consider a multi-model typeless routing system. In this case, we would want to customize the routeConstraint to select the appropriate model. This is hard because ODataRoute() accepts ODataPathRouteConstriant, which forces us to create a subclass and duplicate the private logic internal to ODataPathRouteConstraint.
If ODataRoute() accepted an interface, say IHttpRouteConstraint, it would cause a cleaner design using composition.

xuzhg:
Hi, @karunasagark 
Would you please take a look the PR: #604? And Let me know any concern. 

Thanks.
Sam

xuzhg:
Fixed at 5ebfb93f7b8e0a524426104a15f96f10024bb754

PR #604, opened 2016-01-08
Issue #601: ODataRoute should accept an interface instead of ODataPathRouteConstraint.

xuzhg:
Issue #601: ODataRoute should accept an interface instead of ODataPathRouteConstraint.

### DIFF
--- a/OData/src/System.Web.OData/OData/Routing/ODataRoute.cs
+++ b/OData/src/System.Web.OData/OData/Routing/ODataRoute.cs
@@ -26,9 +26,17 @@ namespace System.Web.OData.Routing
         /// <param name="routePrefix">The route prefix.</param>
         /// <param name="pathConstraint">The OData path constraint.</param>
         public ODataRoute(string routePrefix, ODataPathRouteConstraint pathConstraint)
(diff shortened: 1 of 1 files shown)


karunasagark:
Hi Sam, Thanks for the PR. The changes look good. 
A new public property 'RouteConstraint' is introduced. I'm wondering why public? Can this be private? Wouldn't it cause confusion with 'PathRouteConstraint' property?

xuzhg:
@karunasagark

Yes, it's a little bit confusion. However, here's the reasons:

1. Can't delete/change the `'PathRouteConstraint` because it's public.

2. If 'RouteConstraint' is private, the developer can't get the real constraint if it is not derived from `ODataPathRouteConstraint `.

For example:

```C#
class MyRouteConstraint : IHttpRouteContraint
{}


ODataRoute route = new ODataRoute("myRoute", new MyRouteConstraint());
MyRouteConstraint constraint = route.PathRouteConstraint as MyRouteConstraint;  // constraint will be null
```

karunasagark:
@xuzhg 

Makes sense. Thank you!

Also, would you know in what version of the assembly would this fix be available?

xuzhg:
@karunasagark  It will be in 5.9

xuzhg:
Merged. Thanks.

PR #613, opened 2016-01-17
Bind Uri Functions to CLR methods

### LINKED ISSUE #612: Bound custom UriFunctions to CLR methods.
After issue OData/odata.net#430 is solved at OData.Lib level, we must support it at OData.WebApi level.

Now, after adding the custom function signature in OData.Libs and run the WebApi we get the following exception as expected: 
"message": "Unknown function '[MyCustomFunctionName]'.",
"type": "System.NotImplementedException",
"stacktrace": "   at System.Web.OData.Query.Expressions.FilterBinder.BindSingleValueFunctionCallNode(SingleValueFunctionCallNode node)

It means that there is no CLR bounded method to this specific signature.

Now we have to make it possible for clients to register their EdmFunctionSignature to a CLR MethodInfo so we could create an Expression out of it.

YogiBear52:
This PR resolves #612

Bind custom functions name to CLR MethodInfo.
The class 'UriFunctionsToClrBinder' is responsible for register ant function name to it's MethodInfo.

Now we can add a custom uri function to ODataLibs and bind it to MethodInfo in OData.WebApi.


VikingsFan:
@YogiBear52 is your change pending ODL's 6.15 release?

YogiBear52:
Yes, it is. I forgot to mention it,
Do you want me to check in OData.Libs 6.15?

YogiBear52:
OData.Libs 6.15 has just released. Now it is possible to merge this pull request.
Thank you!

OData/src/System.Web.OData/OData/ODataUriFunctions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using Microsoft.OData.Core.UriParser;
```
VikingsFan: can we put those settings into HttpConfigurationExtensions to keep consistency?

OData/src/System.Web.OData/OData/ODataUriFunctions.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        /// <param name="customFunctionName">The uri function name that appears in the OData request uri.</param>
+        /// <param name="functionSignature">The new custom function signature.</param>
+        /// <param name="methoInfo">The MethodInfo to bind the given function name.</param>
+        /// <exception cref="Exception">Any exception thrown by 'CustomUriFunctions.RemoveCustomUriFunction' and 'UriFunctionsToClrBinder.UnbindUriFunctionName' methods.</exception>
+        /// <returns>'True' if the fucntion signature has successfully removed and unbinded. 'False' otherwise.</returns>
+        public static bool RemoveCustomUriFunction(string customFunctionName, FunctionSignatureWithReturnType functionSignature, MethodInfo methoInfo)
+        {
+            return
+                CustomUriFunctions.RemoveCustomUriFunction(customFunctionName, functionSignature) &&
```
VikingsFan: what can customer do if this method return false?

OData/src/System.Web.OData/OData/Query/Expressions/UriFunctionsToClrBinder.cs
```diff
@@ -0,0 +1,198 @@
...
+
+namespace System.Web.OData.Query.Expressions
+{
+    /// <summary>
+    /// This class helps to bind uri functions to CLR.
+    /// For creating an Expression and apply it on a Queryable collection, we must get the CLR information,
+    /// i.e MethodInfo of each EdmFunction which is mentioned in the EdmModel.
+    /// If you add a custom uri function in OData.Core via 'CustomUriFunctions' class, you must bind it to it's MethodInfo.
+    /// </summary>
+    public static class UriFunctionsToClrBinder
```
VikingsFan: UriFunctionsBinder?

OData/src/System.Web.OData/OData/ODataUriFunctions.cs
YogiBear52: Nothing special. The customer will know that he has done something wrong, like trying to remove a custom function which doesn't exist.
I see it just like a 'Remove' method of any data structures.

OData/src/System.Web.OData/OData/Query/Expressions/UriFunctionsToClrBinder.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// </summary>
+        /// <param name="methodName"></param>
+        /// <param name="methodArgumentsType"></param>
+        /// <returns></returns>
+        private static string GetMethodLiteralSignature(string methodName, IEnumerable<Type> methodArgumentsType)
+        {
+            StringBuilder builder = new StringBuilder();
+            string descriptionSeparator = "";
+            builder.Append(descriptionSeparator);
+            descriptionSeparator = "; ";
```
VikingsFan: never used?

OData/src/System.Web.OData/OData/Query/Expressions/UriFunctionsToClrBinder.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            string parameterSeparator = "";
+            builder.Append(methodName);
+            builder.Append('(');
+            foreach (Type type in methodArgumentsType)
+            {
+                builder.Append(parameterSeparator);
+                parameterSeparator = ", ";
```
VikingsFan: try not to assign in the for loop?

OData/src/System.Web.OData/OData/ODataUriFunctions.cs
YogiBear52: I have considered that, but as you can see, these methods are not extension methods, they are just static.
But I don't rellay mind if you prefer to put them in HttpConfigurationExtensions.

OData/src/System.Web.OData/OData/Query/Expressions/UriFunctionsToClrBinder.cs
YogiBear52: Yea why not. I am not good with naming lol

OData/src/System.Web.OData/OData/ODataUriFunctions.cs
VikingsFan: agree ;)

OData/src/System.Web.OData/OData/ODataUriFunctions.cs
VikingsFan: then the customer don't need to learn a new public class :)

VikingsFan:
Thanks for you PR! it will be very helpful if you can resolve the comments and rebase on master.

OData/src/System.Web.OData/OData/Query/Expressions/UriFunctionsToClrBinder.cs
YogiBear52: Do you mean the two 'GetMethodLiteralSignature'?
I think both methods are used.
Line 73 and line 163.
Tell me if I am wrong because I am not using my visual studio right now.

OData/src/System.Web.OData/OData/Query/Expressions/UriFunctionsToClrBinder.cs
YogiBear52: In this line I change the parameterSeperator value from "" to ", ".  Because after the first loop we need to assign a comma between each type. Another way to do this is to hold an index variable and assin comma  if the index is above 0 otherwise assign nothing.
Another way is to do the first loop outside the for loop, and do the rest in the for loop.
Choose your ideal solution.

YogiBear52:
Thank you very much for your fast code review.
I won't be near my computer for the next two weeks(I chat with you through
my phone), so I would be happy if you could Commit the changes.

Yogev.
בתאריך 21 במרץ 2016 12:33 אחה״צ,‏ "Fan Ouyang" <notifications@github.com>
כתב:

VikingsFan:
Thanks, Merged 56c61a1b957054e02ada5c35aed9735a83bd11a2

PR #635, opened 2016-02-16
[vNext] simple fixes for EdmBuilder

SkyKnight:
Hi,

I found missing attributions filtering for conventions applyings. Also propInfo.DeclaringType.GetTypeInfo().BaseType for most cases returns just System.Object instead of base class of entity. I can't locate any unit tests for vNext stuff, so I just pulling this fixes with label "it works for me" ;)


xuzhg:
Merged at: 4edf8987bb3821919a537b2fb419ea392386de6f

Thanks your contribution.

PR #642, opened 2016-02-26
advertise action/function in feed payload

### LINKED ISSUE #637: Functions/Actions bound to entity should be advertised
See #84.

Currently, action bound to entity is advertised in payload. But not the function. please investigate to add it.

See [OData spec](http://docs.oasis-open.org/odata/odata-json-format/v4.0/errata02/os/odata-json-format-v4.0-errata02-os-complete.html#_Toc403940649):

<h3>15 Bound Function</h3>

A function that is bound to the current entity is advertised via a name/value pair where the name is a hash (#) character followed by the namespace- or alias-qualified name of the function.

Functions that are bound to a collection of entities are advertised in representations of that collection.

A function may have multiple overloads with different parameters. If function overloads exist that cannot be bound to the current entity type, the name SHOULD address a specific function overload by appending the parentheses-enclosed, comma-separated list of non-binding parameter names, see rule qualifiedFunctionName in [OData-ABNF].

If odata.metadata=full is requested, each value object MUST have at least the two name/value pairs title and target. It MAY contain annotations. The order of the name/value pairs MUST be considered insignificant.

The target name/value pair contains a bound function or action URL. If the URL in the target name/value pair cannot be used to invoke all overloads for the function, then the function name MUST be distinguished by appending the parentheses-enclosed, comma-separated list of non-binding parameter names.

The title name/value pair contains the function or action title as a string.

If odata.metadata=minimal is requested, the target name/value pair MUST be included if its value differs from the canonical function or action URL.

Example 31: minimal representation of a function where all overloads are applicable
```json

{

  "@odata.context": "http://host/service/$metadata#Employees/$entity",

  "#Model.RemainingVacation": {},

  ...

}
```

Example 32: full representation of a specific overload

```json
{

  "@odata.context": "http://host/service/$metadata#Employees/$entity",

  "#Model.RemainingVacation(Year)": {

    "title": "Remaining vacation from year...",

    "target": "Employees(2)/RemainingVacation(Year=@Year)"

  },

  ...

}
```

Example 33: full representation in a collection

```json
{

  "@odata.context": "http://host/service/$metadata#Employees",

  "#Model.RemainingVacation": {

    "title": "Remaining Vacation",

    "target": "Managers(22)/Employees/RemainingVacation"

  },

  "value": [ ... ]

```

xuzhg:
This PR resolves #637.

### DIFF
--- a/OData/src/System.Web.OData/OData/Builder/ActionConfiguration.cs
+++ b/OData/src/System.Web.OData/OData/Builder/ActionConfiguration.cs
@@ -48,11 +48,13 @@ namespace System.Web.OData.Builder
             {
                 throw new ArgumentNullException("actionLinkFactory");
             }
+
             if (!IsBindable || BindingParameter.TypeConfiguration.Kind != EdmTypeKind.Entity)
             {
                 throw Error.InvalidOperation(SRResources.HasActionLinkRequiresBindToEntity, Name);
             }
-            LinkFactory = actionLinkFactory;
+
+            ProcedureLinkBuilder = new ActionLinkBuilder(actionLinkFactory, followsConventions);
             FollowsConventions = followsConventions;
             return this;
         }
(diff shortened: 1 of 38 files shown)


VikingsFan:
Merged b8fe7618fc735015453a97619fb12debf7f70dc4

Issue #644, opened 2016-03-03
ODataMediaTypeFormatter does not expose an extensibility hook for changing uri base address


jackfree:
For servers that sit behind a load balancer, the uri that the end server sees is often times not the uri that you want to use when generating OData uris in the response.  For example, a server behind a load balancer may see requests as 'http' instead of 'https', it may see requests as using an explicit host name or different port.   

While there is a method in ODataMediaTypeFormatter for getting this -- GetBaseAddress -- it is private and static.  We need an extensibility hook that allows consumers to provide their own base address (the service root part of an OData uri) and this hook should not require consumers to re-create all the formatters.  Furthermore it should be opt-in and not change the default behavior of the class.

I propose adding a delegate, exposed via a public property on ODataMediaTypeFormatter, that allows consumers to provide their own method for generating the base address.  Furthermore, the existing GetBaseAddress method should be exposed as public so that custom implementations can call it, allowing for minor surgical alterations (such as converting the uri scheme to https whilst preserving the rest of the uri).

xuzhg:
@jackfree Would you please help verify the commit? I do a little bit change for you PR. Give me any feedback. Thanks.

jackfree:
Thanks Sam - I will verify that this change will fix the issue we have with base URIs.

lencharest:
I'd like to make a couple of suggestions regarding the API exposed by these changes.

First, the term used in the OData specs for the base URI of an OData service is _service root_. The API should use terminology consistent with the spec, so I suggest that the new public method `GetDefaultBaseAddress` be renamed to `GetDefaultServiceRoot`.

Second, class properties are typically named as nouns or noun phrases. I suggest the new property `GetBaseAddress` be renamed to `ServiceRootResolver` or `ServiceRootFactory`.

jackfree:
Those are good points.  The code has already been pushed, but I will rename the public static method to GetDefaultServiceRoot, and change the property to ServiceRootResolver

lencharest:
:+1:

jackfree:
Actually it looks like the OData folks made the name change and exposed the property as a Func called BaseAddressFactory, so we're half-way there with the naming.  I agree that ServiceRoot would be better since that's the official OData terminology though

jackfree:
Updated PR

PR #645, opened 2016-03-03
Provide an extensibility hook for consumers of ODataMediaTypeFormatter to customize base address of service root in OData uris

jackfree:
Issue #644 

- Added optional delegate for consumers of ODataMediaTypeFormatter
- Exposed existing GetBaseAddress method as public and renamed to GetDefaultBaseAddress
- Added test case to verify functionality

See https://github.com/OData/WebApi/issues/644


xuzhg:
Merged.Thanks.

PR #680, opened 2016-03-20
Fix multiple navigations in groupby

kosinsky:
Dynamic type generation for $apply has an issues for cases like: Products?$apply=groupby((Category/CategoryName, Supplier/CompanyName)) caused by the same type name for top level and nested classes


VikingsFan:
Merged f3b5bfd92087b671625ab73ea4f5249f20c1078f Thanks!

Issue #733, opened 2016-05-18
V3 does not provide BaseAddressFactory like V4 does (load balancing scenarios, #644)


wachulski:
System.Web.Http.OData.Formatter.ODataMediaTypeFormatter does not expose BaseAddressFactory hook as V4 does.

Since we expose OData V3 service for Excel (OData feed) and the communication passes load balancers, we end up in the same scenario which led to introduction BaseAddressFactory hook for V4 (see: https://github.com/OData/WebApi/issues/644).

Since V3 is still maintained I see it should also contain such a hook.

xuzhg:
Fixed at 3c2095e06f090e3608eee716d04d5e27caf43f5c

PR #734, opened 2016-05-18
Fixing #733 - [V3 only] Provide an extensibility hook for consumers of ODataMed…

wachulski:
### Issues
#733 

### Description
same as #645 for V4


xuzhg:
Thanks your input.

Issue #418, opened 2015-06-25
ODataQueryOptions should allow duplicate non-OData query options


brjohnstmsft:
Our Web API exposes both OData and custom query options. Some of our query options allow duplicate values, but  the `ODataQueryOptions` constructor throws an exception when it detects any duplicate query options (not just standard OData query options). This requires us to maintain a lot of code to work around this issue that we would rather not have to maintain.

WebAPI should have no opinions on whether it is valid to have duplicate custom query options. A Dictionary should only be used to hold OData query options, not custom query options. See this line in the `ODataQueryOptions` constructor:

```
        IDictionary<string, string> queryParameters =
            request.GetQueryNameValuePairs().ToDictionary(p => p.Key, p => p.Value);
```

congysu:
Pre-process the name value pairs, before creating the dictionary.

VikingsFan:
relay on ODL https://github.com/OData/odata.net/issues/375

TomDu:
Related issue: https://github.com/OData/odata.net/issues/245.

VikingsFan:
Merged for 6.x, need merge in 5.x when OData 5.x include this fix. a10881f462d24bca53a5373d9d3e64a3df9f7e2b

jedd-ahyoung:
@VikingsFan Hey - I just ran into this problem today. We're on .NET 4.6.1 and so I can't upgrade to OData 6.0.0 (due to the dependency injection framework). Is there any chance that this will be backported to the 5.9.x branch?

Thanks!

brjohnstmsft:
@xuzhg This issue appears to have been fixed, at least in 7.x. Can this issue be closed?

habbes:
Closing this issue. If this still affects the latest versions of the library, kindly open a new issue.

PR #748, opened 2016-06-07
issue #418 ignore non-odata query parameters when building parameter …

mpetito:
### Issues
This pull request fixes issue #418.

### Description
For `System.Web.OData`, when constructing `ODataQueryOptions` only query parameters that start with $ are included in the queryParameters dictionary.  This allows for duplicate non-OData related query parameters in the request url.

- [ x ] Test cases added

### Additional work necessary
None. Initially I wanted to use `IsSupportedQueryOption` for the filtering.  However, the parser / resolver options can't be constructed until after the parameter dictionary is built.

### DIFF
--- a/OData/src/System.Web.OData/OData/Query/ODataQueryOptions.cs
+++ b/OData/src/System.Web.OData/OData/Query/ODataQueryOptions.cs
@@ -78,10 +78,10 @@ namespace System.Web.OData.Query
             Context = context;
             Request = request;
 
-            // Parse the query from request Uri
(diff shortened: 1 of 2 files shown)


VikingsFan:
@mpetito Thanks for you contribution! But this change should based on OData lib's change: 670abe58ddfaa64202ce770fe529cc2e92f8ab19 or the request will failed because uri parser will throw exception, so I'll merge your pr into WebAPI/OData breaking change branch: OData60.

VikingsFan:
Merged a10881f462d24bca53a5373d9d3e64a3df9f7e2b

PR #839, opened 2016-10-28
reduced full MVC dependency to Mvc.Core

damienpontifex:
### Issues

Removed dependency on entire MVC package and moving to just Mvc.Core. This should be a more logical base as it doesn't require other view based features that are included in the full MVC package.
### Description

Changing dependency from `Microsoft.AspNetCore.Mvc` to `Microsoft.AspNetCore.Mvc` and updated packages for minor release
- [  ] Test cases added
- [ x ] Build and test with one-click build and test script passed ➝ built with `dotnet build` and tested with `dotnet test`on macOS

### DIFF
--- a/vNext/samples/ODataSample.Web/project.json
+++ b/vNext/samples/ODataSample.Web/project.json
@@ -1,6 +1,6 @@
 ﻿{
   "dependencies": {
-    "Microsoft.AspNetCore.Mvc": "1.0.0",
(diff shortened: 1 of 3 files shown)


xuzhg:
Merged at 99bda47aa27052b46e024e43b2d417a348a029e6.  Thanks your contribution.

PR #870, opened 2016-12-13
Create a copy constructor for SelectExpandNode

ificator:
### Description
Prior to `OData 6.0` it was possible to prevent the serialization of literal `null` using a method described in [this StackOverflow question](http://stackoverflow.com/questions/33192231/prevent-null-values-from-being-emitted-in-a-webapi-odata-v4-service). Unfortunately this stopped working as the serialization code now enumerates of the various properties of `SelectExpandNode` instead of `ODataResource.Properties`.

A new method to accomplish the same goal is to therefore override `CreateSelectExpandNode` and remove the applicable properties from the `SelectExpandNode` created in the base implementation. Unfortunately this implementation utilizes a cache, and so directly modifying those instances leads to tricky bugs. Ultimately the easiest solution here was to create a copy constructor for `SelectExpandNode` so the properties could be modified without updating the original. This still **does not** allow the overriding of the `SelectAllDynamicProperties` Boolean, but that can be worked around if needed.

### DIFF
--- a/OData/src/System.Web.OData/OData/Formatter/Serialization/SelectExpandNode.cs
+++ b/OData/src/System.Web.OData/OData/Formatter/Serialization/SelectExpandNode.cs
@@ -36,6 +36,23 @@ namespace System.Web.OData.Formatter.Serialization
         }
 
         /// <summary>
+        /// Creates a new instance of the <see cref="SelectExpandNode"/> class by copying the state of another instance. This is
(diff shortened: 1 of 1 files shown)


xuzhg:
Merged at : https://github.com/OData/WebApi/commit/ebb6aa49411e08f84ec148b02cb4f32db7cbbfc5

Thanks your contribution.

Issue #514, opened 2015-09-16
Support now()


MaxWilson:
Per oData v4 spec,

5.1.1.4.21 now
The now function has the following signature:
Edm.DateTimeOffset now()
The now function returns the current point in time (date and time with time zone) as a DateTimeOffset value. 

This doesn't seem to be supported in version 5.6 of oData.

_Example:_ http://Example.com/ProductPrices?$count=true&$filter=SKU+eq+%274F3-00015%27+and+FutureEndDate+gt+now()

_Result:_ 

``` javascript
    {
      error: {
        code: "",
        message: "The query specified in the URI is not valid. Unknown function 'now'."
      }
    }
```

congysu:
Now() is not supported for now.

robward-ms:
See #872

PR #872, opened 2016-12-15
Add support to now() function on filter clauses

athoscouto:
### Issues
*This pull request fixes issue #514.*  

### Description
*Added support to the now() function on filter queries for the WebAPI project. The odata.net project already parses and handles the now query correctly. Unit tests and E2E tests were added too.*

### Checklist
- [x] *Test cases added*

### Additional work necessary
*None*

### DIFF
--- a/OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
+++ b/OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
@@ -630,6 +630,9 @@ namespace System.Web.OData.Query.Expressions
                 case ClrCanonicalFunctions.TimeFunctionName:
                     return BindTime(node);
 
+                case ClrCanonicalFunctions.NowFunctionName:
+                    return BindNow(node);
+
                 default:
                     // Get Expression of custom binded method.
                     Expression expression = BindCustomMethodExpressionOrNull(node);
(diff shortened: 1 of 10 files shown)


OData/src/System.Web.OData/OData/Query/Expressions/FilterBinder.cs
```diff
@@ -887,6 +890,17 @@ private Expression BindDate(SingleValueFunctionCallNode node)
...
 
+        private Expression BindNow(SingleValueFunctionCallNode node)
+        {
+            Contract.Assert("now" == node.Name);
+
+            // Function Now() does not take any arguemnts.
+            Expression[] arguments = BindArguments(node.Parameters);
+            Contract.Assert(arguments.Length == 0);
+
+            return Expression.Property(null, typeof(DateTimeOffset), "Now");
```
robward-ms: Why not UtcNow instead of Now?  The spec does allow us to use any time zone (but it's example is Utc) but Utc seems to isolate us form the time zone of the server. 

From the spec:
5.1.1.6.9 now
The now function has the following signature:
Edm.DateTimeOffset now()
The now function returns the current point in time (date and time with time zone) as a DateTimeOffset value.
Services are free to choose the time zone for the current point, e.g. UTC. Services that are unable to preserve the offset of Edm.DateTimeOffset values and instead normalize the values to some common time zone SHOULD return a value in the normalized time zone (i.e., UTC).

robward-ms: @mikepizzo , any thoughts?

robward-ms:
This is a lot of commit history to add to OData, can you clean it up a bit? I'd expect 1 or 2 commits, not 10.

robward-ms:
Please ensure E2E tests pass, see Readme.md for command line instructions for E2E tests.

robward-ms:
Merged.

Issue #874, opened 2016-12-15
$apply used with Entity Framework causes memory leaks


kosinsky:
Dynamic CLR types generated to represent shape in groupby/aggregate clauses are aggressively cached by Entity Framework 6 in many places and as result never garbage collected. Entity Framework caches whole Query plan and parts of lambda expression used to represent a query. As a result, each next call make system a bit slower.

markdstafford:
Hi Konstantin, thanks for creating the issue. Can you please clarify whether this is an issue with WAO or whether it's an issue with EF? If it is our issue, are you planning to send a PR or is this something you would like us to look at?

kosinsky:
Issues is on the edge. EF caches types used in queries for performance reasons and current implementation generates at least 2 dynamic types (in two dynamic assemblies) per query and EF caches them. Original implementation with emitting dynamic types was based on assumption that generated types will be thrown away in all cases.

xuzhg:
@kosinsky Please verify the fix and close it.

kosinsky:
@xuzhg Didn't notice right away. Looks good now. Closing.

PR #875, opened 2016-12-15
Don't use dynamic types for $apply implementation

kosinsky:
### Issues
*This pull request fixes issue #874.* 

### Description
Dynamic CLR types generated to represent shape in groupby/aggregate clauses are aggressively cached by Entity Framework 6 in many places and as result never garbage collected. As a result, each next call make system a bit slower

Re-implemented AggregationBinder using the same approach as SelectExpandBinder does (linked list of Containers). A bit of complexity was added that expression generated from each groupby could be use in further groupby, filter, orderby etc. expressions. It’s why we used dynamic types in first place, but recent refactoring in ODL v7/ WebApi v6 make linked list approach possible.

- [ x ] Test cases added


OData/src/System.Web.OData/OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -235,31 +235,27 @@ private void WriteDeltaResource(object graph, ODataDeltaWriter writer, ODataSeri
         {
             var properties = new List<ODataProperty>();
             var dynamicObject = graph as DynamicTypeWrapper;
-            foreach (var prop in graph.GetType().GetProperties())
+            foreach (var prop in dynamicObject.Values)
             {
-                object value;
                 ODataProperty property = null;
```
xuzhg: move it into the inner loop

OData/src/System.Web.OData/OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -235,31 +235,27 @@ private void WriteDeltaResource(object graph, ODataDeltaWriter writer, ODataSeri
...
+                        .FirstOrDefault(p => p.Name.Equals(prop.Key));
+                    if (edmProperty != null)
                     {
-                        IEdmProperty edmProperty = entityType.Properties()
-                            .FirstOrDefault(p => p.Name.Equals(prop.Name));
-                        if (edmProperty != null)
-                        {
-                            dynamicTypeProperties.Add(edmProperty, value);
-                        }
+                        dynamicTypeProperties.Add(edmProperty, prop.Value);
```
xuzhg: add assert for the dynamic type. [Resolved].

xuzhg:
Merged with some changes at https://github.com/OData/WebApi/commit/7bcaedfebb949cdbefcb2cb9aa934a0d3a70d3f5.

@kosinsky Please verify the fix and let us know any problems.  

Thanks your contribution.

OData/src/System.Web.OData/OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -511,6 +554,154 @@ internal Expression BindHas(Expression left, Expression flag)
...
+        /// <returns>Returns null if no aggregations were used so far</returns>
+        protected Expression GetFlattenedPropertyExpression(string propertyPath)
+        {
+            if (flattenedPropertyContainer == null)
+            {
+                return null;
+            }
+
+            Expression expression;
+            if (flattenedPropertyContainer.TryGetValue(propertyPath, out expression))
```
PietheinHeemskerk: The lines 696-700 should be replaced by the following lines to allow ODataConventionModelBuilderExtensions.EnableLowerCamelCase() to work. See issue #1136

```
#            var entry = FlattenedPropertyContainer.FirstOrDefault(c => string.Equals(c.Key, propertyPath, StringComparison.OrdinalIgnoreCase));
            if (entry.Key != null)
            {
                return entry.Value;
            }
```

PR #882, opened 2016-12-27
#881: CreateRequestContainer rise system exception instead of ODataEx…

### LINKED ISSUE #881: CreateRequestContainer rise system exception instead of ODataException.
CreateRequestContainer rise system exception instead of ODataException 

### Assemblies affected
System.Web.OData.dll

### Reproduce steps
First of all it's strange that the following code can rise an exception in OData.      

> var context = new HttpContextWrapper(HttpContext.Current);
>       var routeData = RouteTable.Routes.GetRouteData(context);
>       var routeData2 = RouteTable.Routes.GetRouteData(context);

In new version of OData (6.0.0), inside of `HttpRequestMessageExtensions.CreateRequestContainer` throws InvalidOperationException (why we need the check?) if `request.Properties` already has requestContainer property. So, just call  `RouteTable.Routes.GetRouteData(context)` two times in a row, we have got such exception which not always possible to catch, because the call of CreateRequestContainer might be chained with a lot of different external libs. The `CreateRequestContainer` calls from `Match` method of `ODataPathRouteConstraint` class. The Match class contains try/catch logic, but it catch only ODataException. To make the Odata code more predictable, need to change type of exception, from InvalidException to OdataException (and still write Odata's message).
### Expected result
The 'Match' method from ODataPathRouteConstraint will catch OdataExceptions.

### Actual result
The 'Match' method doesn't catch exception throwed by `System.Web.OData.Extensions.HttpRequestMessageExtensions` class.

### Proposed fix:
Change `throw Error.InvalidOperation(SRResources.RequestContainerAlreadyExists);` 
to: `throw new ODataException(SRResources.RequestContainerAlreadyExists);`.

smitbmx:
…ception.

Done.

### Issues
#881

### Description
Fixed #881.

### DIFF
--- a/OData/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs
+++ b/OData/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs
@@ -261,7 +261,7 @@ namespace System.Web.OData.Extensions
         {
             if (request.Properties.ContainsKey(RequestContainerKey))
             {
-                throw Error.InvalidOperation(SRResources.RequestContainerAlreadyExists);
+                throw new ODataException(SRResources.RequestContainerAlreadyExists);
             }
 
             IServiceScope requestScope = request.CreateRequestScope(routeName);
--- a/OData/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs
+++ b/OData/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs
@@ -255,6 +255,9 @@ namespace System.Web.OData.Extensions
         /// <param name="request">The request.</param>
         /// <param name="routeName">The name of the route.</param>
         /// <returns>The request container created.</returns>
+        /// <exception cref="ODataException">
+        /// <paramref name="request"/> already has requestContainer property.
+        /// </exception>
         [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope",
             Justification = "requestScope will be disposed when the request ends.")]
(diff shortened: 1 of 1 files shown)


OData/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -261,7 +261,7 @@ public static IServiceProvider CreateRequestContainer(this HttpRequestMessage re
         {
             if (request.Properties.ContainsKey(RequestContainerKey))
             {
-                throw Error.InvalidOperation(SRResources.RequestContainerAlreadyExists);
+                throw new ODataException(SRResources.RequestContainerAlreadyExists);
```
AlanWong-MS: Please add an exception section in the function comments. For example,
```
        /// <exception cref="ODataException">
        /// <paramref name="request"/> already has requestContainer property.
        /// </exception>
```

AlanWong-MS:
Thanks for your contribution! Before we can complete this request, can you please add tests to your PR?

OData/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs
smitbmx: Done.

biaol-odata:
@AlanWong-MS : It seems like the PR is still not merged yet? can you please tell us what is still required to be done? Thanks!

AlanWong-MS:
@biaol-odata @smitbmx : we just need to update/add appropriate unit tests to accommodate this change.

xuzhg:
Close it. We'd love to see your further PRs. Thanks your contribution.

Issue #773, opened 2016-07-13
aggregate($count as total) doesn't work


olsh:
I'm trying to use the aggregation function with the virtual property $count as described [here](https://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/csprd02/odata-data-aggregation-ext-v4.0-csprd02.html#_Keyword_with_1)
Here is my query:
`?$apply=groupby((name, version), aggregate($count as total))`
### Assemblies affected

Microsoft.AspNet.OData 5.9.1
Microsoft.OData.Core 6.15.0
Microsoft.OData.Edm 6.15.0
### Reproduce steps

Execute aggregate function with $count
`$apply=aggregate($count as total)`
### Expected result

Total amount of rows in the set.
### Actual result

`The query specified in the URI is not valid. 'with' expected at position 17 in 'aggregate($count as total)'.`

VikingsFan:
@olsh yes, this is not supported for now, need do it in ODL and WebAPI/OData

olsh:
@VikingsFan thanks. 
Is there any other way to count entities in a set?

VikingsFan:
can you add &$count=true to get the count as a workaround ?

olsh:
I need something like this.
```
  "value": [
    { "name" : "bob", "version" 2: , "total": 5 },
    { "name" : "alice", "version" 1: , "total": 8 },
    { "name" : "alice", "version" 2: , "total": 2 },
  ]
```
I thought `&$count=true` returns count for all records for the entire query.

VikingsFan:
@olsh For now we only support filter and aggregate in groupby, or there need a property and use  `aggregate PropertyName with sum as total` to get total count.

olsh:
@VikingsFan thank you for the quick reply.
> need do it in ODL and WebAPI/OData

What is ODL?
I'll try to add the feature if I have some spare time.

VikingsFan:
You are very welcome to contribute to us, ODL odata lib, is https://github.com/OData/odata.net , our master is reference on their ODataV4-6.x

xuzhg:
We have a PR request merged at https://github.com/OData/WebApi/commit/6c18a49921f257e20ca157da6b755886f0da6c57. Please test at your side and let us know any further problem. Thanks.

PR #908, opened 2017-02-07
Virtual Property Count

athoscouto:
### Issues
*This pull request fixes issue [#773](https://github.com/OData/WebApi/issues/773) of WebAPI project.*  

- [x] Test cases added
- [x] Build and test with one-click build and test script passed

### Additional comments
This is a pretty simple (maybe hackery) way of implementing the $count virtual property for aggregate clauses. If there are better options, we should consider changing the design.

Every time we find the `$count as alias` keyword inside a aggregate clause, we just map it to a method call of the form `attr with InternalCount as alias`, where attr is just a hardcoded string in the code that won't be used (but must be defined because is checked several times for nulls) and InternalCount is a method that counts objects (is a subroutine of `countdistinct`).

### DIFF
--- a/OData/src/System.Web.OData/OData/Query/Expressions/AggregationBinder.cs
+++ b/OData/src/System.Web.OData/OData/Query/Expressions/AggregationBinder.cs
@@ -173,11 +173,18 @@ namespace System.Web.OData.Query.Expressions
 
         private Expression CreateAggregationExpression(ParameterExpression accum, AggregateExpression expression)
         {
-            LambdaExpression propertyLambda = Expression.Lambda(BindAccessor(expression.Expression),
-                this._lambdaParameter);
             // I substitute the element type for all generic arguments.                                                
             var asQuerableMethod = ExpressionHelperMethods.QueryableAsQueryable.MakeGenericMethod(this._elementType);
             Expression asQuerableExpression = Expression.Call(null, asQuerableMethod, accum);
+
+            // $count is a virtual property, so there's not a propertyLambda to create.
+            if (expression.Method == AggregationMethod.VirtualPropertyCount)
+            {
+                var countMethod = ExpressionHelperMethods.QueryableCountGeneric.MakeGenericMethod(this._elementType);
+                return Expression.Call(null, countMethod, asQuerableExpression);
+            }
+
+            LambdaExpression propertyLambda = Expression.Lambda(BindAccessor(expression.Expression), this._lambdaParameter);
 
             Expression aggregationExpression;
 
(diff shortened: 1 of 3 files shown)


OData/src/System.Web.OData/OData/Query/Expressions/AggregationBinder.cs
```diff
@@ -267,6 +279,7 @@ private Expression BindAccessor(SingleValueNode node)
                 case QueryNodeKind.SingleValueOpenPropertyAccess:
                     var openNode = node as SingleValueOpenPropertyAccessNode;
                     return Expression.Property(BindAccessor(openNode.Source), openNode.Name);
+                case QueryNodeKind.None:
```
robward-ms: Is this intended to catch the CountVirtualPropertyNode since its Kind is None?  Won't navNode be null since its not a SingleNavigationNode and crash when accessing a property, i.e. navNode.Source?

OData/src/System.Web.OData/OData/Query/Expressions/AggregationBinder.cs
```diff
@@ -173,8 +173,13 @@ private IQueryable BindSelect(IQueryable grouping)
 
         private Expression CreateAggregationExpression(ParameterExpression accum, AggregateExpression expression)
         {
-            LambdaExpression propertyLambda = Expression.Lambda(BindAccessor(expression.Expression),
-                this._lambdaParameter);
+            LambdaExpression propertyLambda = null;
+            // $count is a virtual property, so there's not a propertyLambda to create.
+            if (expression.Method != AggregationMethod.VirtualPropertyCount)
```
robward-ms: Could you reorder this code to have only one reference to VirtualPropertyCount since it is handled differently?  Something like:

var asQuerableMethod ...
Expression asQuerableExpression ...

if (expression.Method == AggregationMethod.VirtualPropertyCount)
{
    var countMethod = ExpressionHelperMethods.QueryableCountGeneric.MakeGenericMethod(this._elementType); 
    aggregationExpression = Expression.Call(null, countMethod, asQuerableExpression); 
}
else
{
  LambdaExpression propertyLambda = Expression.Lambda(BindAccessor(expression.Expression), this._lambdaParameter); 

  switch (expression.Method)
...


}

OData/src/System.Web.OData/OData/Query/Expressions/AggregationBinder.cs
athoscouto: This case should be removed. Was a piece of code that I added there when I was trying another approach and forgot to remove.

OData/src/System.Web.OData/OData/Query/Expressions/AggregationBinder.cs
athoscouto: Done.

robward-ms:
Merged.

robward-ms:
Closed by accident; not yet merged.

xuzhg:
Merged at : https://github.com/OData/WebApi/commit/6c18a49921f257e20ca157da6b755886f0da6c57

Thanks your contribution.

PR #942, opened 2017-03-17
[vNext] Fix corruption of serviceRoot value

genusP:


robward-ms:
Additionally, we've been reviewing the code in vNext and while it has tons of great changes, we're thinking that a new version of WebApi for ASP.Net Core that is a fork of the 6.x code (and technically, vNext is a fork of the 5.x code) will create some maintenance headaches. We're considering layering the vNext branch on top of the OData 6.x branch to create and ASP.Net WebApi and ASP.Net Core Mvc versions from a single, mostly common (~80%) code base. This work would borrow from the vNext branch heavily but would leave the vNext branch as non-shipping branch. Since you have contributed significantly to this branch, I wanted to get your thoughts on this approach.

genusP:
I  cleanup commit history an create new PR #944

genusP:
Hi!
Thanks for appreciating my modest contribution.
With such an amount of common code, I think it's the right approach.

Issue #930, opened 2017-03-04
Regression - ODataConventionModelBuilder.GetEdmModel Stack Overflow + Recursive Complex Types


timward60:
I am seeing an stack overflow exception when my service tries starts up and returns the EdmModel using the ODataConventionModelBuilder. I believe this is because I have a model with entity that has a recursive complex type property.

```C#
public static IEdmModel GetConventionModel()    
{  
    var builder = new ODataConventionModelBuilder();  
    builder.EntitySet<Rule>("Rules");  
    builder.EntitySet<RuleVariable>("RuleVariables");  
    return builder.GetEdmModel(); // Stack overlfow thrown here.
}
```

I don't see this issue when using older version of System.Web.OData nad Microsoft.OData.*. Sample project with recursive complex types working can be found [here].(https://github.com/timward60/RuleService/tree/microsoft_odata_client_issue)

### Assemblies affected
System.Web.OData 6.0.0.0
Microsoft.OData.Core 7.0.0
Microsoft.OData.Edm 7.0.0

### Reproduce steps
1. Create model with an entity that contains a recursive complex type.

I also have a sample project that reproduces the bug:
> 1. git clone https://github.com/timward60/RuleService.git
> 2. git checkout tags/system_web_data_issue -b hotfix/system_web_data_issue
> 3. start RuleService.sln
> 4. Launch RuleService

### Expected result
1. GetEdmModel succeeds.

### Actual result
Stack overflow

### Additional details
[Example Metadata](https://gist.github.com/timward60/32ef54d706df8fed0324a041bed7569c#file-ruleservice_metadata-edmx)
[Full Exception](https://gist.github.com/timward60/32ef54d706df8fed0324a041bed7569c#file-ruleservice_microsoft_odata_core_exception-txt)
[Sample Project](https://github.com/timward60/RuleService/tree/system_web_data_issue)

brjohnstmsft:
@robward-ms This issue is a blocker for Complex Types support in Azure Search. Seems to be a duplicate of https://github.com/OData/WebApi/issues/437  I'm also curious why the referenced PR was closed. If it solves the issue, it would be great to see it merged.

robward-ms:
Looks like this was fixed in #977 and is available in WebApi 6.1.

brjohnstmsft:
This was not completely fixed. See this issue for broader context: https://github.com/OData/WebApi/issues/1248

Issue #892, opened 2017-01-22
Bug in AssociationSetDiscoveryConvention for 6.0


lensgrinder:
I discovered an issue while upgrading from 5.9.1 to 6.0.
Please see issue logged here for details:
https://github.com/OData/odata.net/issues/753

Likely that issue was opened in the wrong place as I hadn't yet debugged to discover the root cause in the System.Web.OData binary. The fix I provide in my post works for my case but it is quick and dirty.

mtroth-microsoft:
This issue has been fixed.

PR #968, opened 2017-04-28
ODataConventionModelBuilder.GetEdmModel Stack Overflow Fix For Recursive Complex Types

timward60:
### Issues
This pull request fixes issues #930 #892.

### Description
When finding navigation properties add check to ensure we don't revisit any property types that have already been traversed.

- [ x ] Test cases added

### Additional work necessary
None.


nathansmetelak:
@timward60 

Hi.  I'm having this same issue and I was curious as to what, if any, the timeline would be for getting this into an official release.  I was hoping to avoid having to build a rogue dll in our codebase.

Thx in advance!

robward-ms:
Looks like this was fixed in #977 and is available in WebApi 6.1.

PR #977, opened 2017-05-16
fix stack overflow when loop is present in model

mtroth-microsoft:
### Issues
This pull request fixes issue #892 

### Description
Added check of type during recursion.

- [ x ] Test cases added

### Additional work necessary
No docs needed. This is a 6.0 only issue.

### DIFF
--- a/OData/src/System.Web.OData/OData/Builder/NavigationPropertyExtensions.cs
+++ b/OData/src/System.Web.OData/OData/Builder/NavigationPropertyExtensions.cs
@@ -46,6 +46,11 @@ namespace System.Web.OData.Builder
             IEnumerable<StructuralTypeConfiguration> derivedTypes = builder.DerivedTypes(configuration);
             foreach (var config in derivedTypes)
             {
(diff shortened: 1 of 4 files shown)


OData/src/System.Web.OData/OData/Builder/NavigationPropertyExtensions.cs
```diff
@@ -46,6 +46,11 @@ internal static class NavigationPropertyExtensions
             IEnumerable<StructuralTypeConfiguration> derivedTypes = builder.DerivedTypes(configuration);
             foreach (var config in derivedTypes)
             {
+                if (path.OfType<Type>().Any(p => p == config.ClrType) == true)
```
mtroth-microsoft: I think of this as a readability and consistency issue. The sugar short hand of omitting the Boolean compare is harder to read and inconsistent with the way that the false case is handled.

xuzhg:
Merged with little changes at https://github.com/OData/WebApi/commit/0d714bd96cb4c4d19fda50478d93540231a6e6f0

Thanks.

Issue #987, opened 2017-06-06
ResourceContext.ResourceInstance should be original instance regardless of select/expand


ificator:
*Short summary (3-5 sentences) describing the issue.*  

### Assemblies affected
`System.Web.OData`

### Reproduce steps
1. Override the `CreateResource` from `ODataResourceSerializer` and access `resourceContext.ResourceInstance`
2. Invoke the code using a request that includes a `$select` query string parameter

### Expected result
The instance returned from the action implementation would be returned from `ResourceInstance`

### Actual result
A new instance is returned, which is missing data that exists on the original

### Additional details
The code currently hydrates a new instance because the `EdmObject` is not a `TypedEdmStructuredObject` and is instead a descendent of `SelectExpandWrapper<>`. The `SelectExpandWrapper<>` itself has an `Instance` property, but that's not populated all of the time.

ificator:
The PR is done, this issue should be resolved.

ificator:
The PR referred to above was reverted - the issue persists.

PR #988, opened 2017-06-06
Always remember the instance provided to the SelectExpandBinder so it can be accessed during serialization

ificator:
### Issues
This pull request is a prototype fix for #987

### Description
This change exposes an `object Instance` property on `ISelectExpandWrapper` and ensures it's always populated. This value can then be used in `ResourceContext` instead of unnecessarily rehydrating a lossy instance of the original.

The semantics of the `TryGetPropertyValue` and `ToDictionary` are retained by way of a flag that indicates whether the `Instance` value should be used as a source of property values.

- [ x ] Test cases added

### Additional work necessary
This adds a new property to `ISelectExpandWrapper`, which is public. Documentation update may be required.


ificator:
I'm going to abdon this PR due to the branching change. I'll resubmit after resolving the above questions.

Issue #990, opened 2017-06-07
ResourceContext.BuildResourceInstance can null ref when model and CLR names for a property do not match


ificator:
`ResourceContext.ResourceInstance` will null ref when the instance needs to be hydrated, and the target class as a property name that is different to the name defined in the model.

### Assemblies affected
`System.Web.OData` 

### Reproduce steps
1. Create a class that has a property call "Foo", and add a `DataMember` that gives it a different name in the model.
2. Create an `ODataResourceSerializer` that attempts to access `resourceContext.ResourceInstance`
3. Expose the object via an API, and query it with a $select=PropertyName 

### Expected result
The request succeeds

### Actual result
The request fails due to a null ref in `BuildResourceInstance`

### Additional details
N/A

xuzhg:
Thanks.

PR #991, opened 2017-06-07
Fix ResourceContext.ResourceInstance null ref by using the right property name

ificator:
### Issues
This pull request fixes issue #990 

### Description
Convert the model property name to the CLR property name before attempting to set it on the resource.

- [ x ] Test cases added

### Additional work necessary
N/A

### DIFF
--- a/OData/src/System.Web.OData/OData/ResourceContext.cs
+++ b/OData/src/System.Web.OData/OData/ResourceContext.cs
@@ -243,13 +243,15 @@ namespace System.Web.OData
                 object value;
                 if (EdmObject.TryGetPropertyValue(property.Name, out value) && value != null)
                 {
+                    string propertyName = EdmLibHelpers.GetClrPropertyName(property, EdmModel);
+
                     if (value.GetType().IsCollection())
                     {
-                        DeserializationHelpers.SetCollectionProperty(resource, property, value, property.Name);
+                        DeserializationHelpers.SetCollectionProperty(resource, property, value, propertyName);
                     }
                     else
                     {
-                        DeserializationHelpers.SetProperty(resource, property.Name, value);
+                        DeserializationHelpers.SetProperty(resource, propertyName, value);
                     }
                 }
             }
--- a/OData/test/UnitTest/System.Web.OData.Test/OData/ResourceContextTest.cs
+++ b/OData/test/UnitTest/System.Web.OData.Test/OData/ResourceContextTest.cs
@@ -224,6 +224,34 @@ namespace System.Web.OData
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


OData/src/System.Web.OData/OData/ResourceContext.cs
```diff
@@ -243,13 +243,15 @@ private object BuildResourceInstance()
                 object value;
                 if (EdmObject.TryGetPropertyValue(property.Name, out value) && value != null)
                 {
+                    string propertyName = EdmLibHelpers.GetClrPropertyName(property, EdmModel);
+
                     if (value.GetType().IsCollection())
                     {
-                        DeserializationHelpers.SetCollectionProperty(resource, property, value, property.Name);
+                        DeserializationHelpers.SetCollectionProperty(resource, property, value, propertyName);
```
xuzhg: Make sense. Thanks.

OData/test/UnitTest/System.Web.OData.Test/OData/ResourceContextTest.cs
```diff
@@ -224,6 +224,34 @@ public void Property_ResourceInstance_EdmObjectHasCollectionProperty_ForComplexC
         }
 
         [Fact]
+        public void Property_ResourceInstance_HandlesModelClrNameDifferences()
+        {
+            const string ClrPropertyName = "Property";
+            const string ModelPropertyName = "DifferentProperty";
+
+            // Arrange
```
xuzhg: Move it at top. Thanks.

ificator: Does it make sense to have test specific constants defined for the entire class? Maybe for the "ClrPropertyName" but probably not for the "ModelPropertyName". I could always just leave them inline as well :)

xuzhg: I mean move // Arrange to the top of this method.

xuzhg: And I modified it here. So, don't worry about it. Thanks.

xuzhg:
Merged at: https://github.com/OData/WebApi/commit/28401a37e4f2b3d5d4dda06250b6205db72cf9a0

Thanks your contribution.

OData/test/UnitTest/System.Web.OData.Test/OData/ResourceContextTest.cs
ificator: Ah gotcha - thanks!

PR #996, opened 2017-06-14
By default, make build.cmd run all tests as to ensure nothing is missed.

robward-ms:
### Description
By default, make build.cmd run all tests as to ensure nothing is missed.
Optionally, run build.cmd quick to run only the unit tests.

- [   ] Test cases added

### Additional work necessary
None

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -12,21 +12,41 @@ REM is not supported. Also handle x86 operating systems, where %ProgramFiles(x86
 REM %MSBuild% value when setting the variable and never quote %MSBuild% references.
 set MSBuild="%ProgramFiles(x86)%\MSBuild\12.0\Bin\MSBuild.exe"
 if not exist %MSBuild% @set MSBuild="%ProgramFiles%\MSBuild\12.0\Bin\MSBuild.exe"
+set FullBuild=0
 
-if /I "%1" == "" goto BuildDefaults
+REM with no switches, run the full set of tests.
+if /I "%1" == "" (
+  set FullBuild=1
+  goto BuildDefaults
+)
+
+REM support quick build, unit tests only.
+if /I "%1" == "QUICK" (
+  set FullBuild=0
+  goto BuildDefaults
+)
+
+REM Continue to suport original switches for those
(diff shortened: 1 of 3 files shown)


build.cmd
```diff
@@ -1,4 +1,4 @@
-@echo off
```
robward-ms: The @echo off should still ne part of the file. Not ready to merge.

robward-ms: Fixed.

build.cmd
```diff
@@ -12,21 +12,41 @@ REM is not supported. Also handle x86 operating systems, where %ProgramFiles(x86
 REM %MSBuild% value when setting the variable and never quote %MSBuild% references.
 set MSBuild="%ProgramFiles(x86)%\MSBuild\12.0\Bin\MSBuild.exe"
 if not exist %MSBuild% @set MSBuild="%ProgramFiles%\MSBuild\12.0\Bin\MSBuild.exe"
+set FullBuild=0
 
-if /I "%1" == "" goto BuildDefaults
+REM with no switches, run the full set of tests.
+if /I "%1" == "" (
```
biaol-odata: Can we combine this non-switches case with the "FULL" case below since they are equivalent?

robward-ms: Sadly, no. Cmd IF syntax only supports the NOT operator: https://ss64.com/nt/if.html

build.cmd
```diff
@@ -12,21 +12,41 @@ REM is not supported. Also handle x86 operating systems, where %ProgramFiles(x86
...
+  set FullBuild=1
+  goto BuildDefaults
+)
+
+REM support quick build, unit tests only.
+if /I "%1" == "QUICK" (
```
robward-ms: Are you suggesting "quick" and "q" or just "q"? What about local customization, i.e. alias?

xuzhg: Yes. I think to add `-q` as an alternative to `QUICK` maybe easy to use. For someone who has local customization, he can use `QUICK`, else can use `-q` quickly. Any thoughts?

robward-ms: Fixed.

PR #1009, opened 2017-06-23
handle single navigation expand filter

taeinkimseattle:
### Description
Today nested $filter inside $expand of non-collection navigation property is ignored. This change introduces a ODataQuerySettings that honor the $filter. If the condition is not met, the navigation property returns null.

- [ x ] Test cases added

### DIFF
--- a/src/System.Web.OData/OData/EnableQueryAttribute.cs
+++ b/src/System.Web.OData/OData/EnableQueryAttribute.cs
@@ -161,6 +161,23 @@ namespace System.Web.OData
         }
 
         /// <summary>
+        /// Honor $filter inside $expand of non-collection navigation property.
+        /// The expanded property is only populated when the filter evaluates to true.
+        /// This setting is false by default.
+        /// </summary>
+        public bool HandleReferenceNavigationPropertyExpandFilter
+        {
+            get
+            {
+                return _querySettings.HandleReferenceNavigationPropertyExpandFilter;
+            }
+            set
+            {
+                _querySettings.HandleReferenceNavigationPropertyExpandFilter = value;
+            }
+        }
+
+        /// <summary>
         /// Gets or sets the query parameters that are allowed in queries.
         /// </summary>
         /// <value>The default includes all query options: $filter, $skip, $top, $orderby, $expand, $select, $count,
--- a/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
+++ b/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
@@ -187,9 +187,11 @@ namespace System.Web.OData.Query.Expressions
             Type nullablePropertyType = propertyValue.Type.ToNullable();
             Expression nullablePropertyValue = ExpressionHelpers.ToNullable(propertyValue);
 
-            if (filterClause != null && property.Type.IsCollection())
+            if (filterClause != null)
             {
-                IEdmTypeReference edmElementType = property.Type.AsCollection().ElementType();
+                bool isCollection = property.Type.IsCollection();
+
+                IEdmTypeReference edmElementType = (isCollection ? property.Type.AsCollection().ElementType() : property.Type);
                 Type clrElementType = EdmLibHelpers.GetClrType(edmElementType, _model);
                 if (clrElementType == null)
                 {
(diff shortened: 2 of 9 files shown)


src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -187,30 +187,64 @@ internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IE
             Type nullablePropertyType = propertyValue.Type.ToNullable();
             Expression nullablePropertyValue = ExpressionHelpers.ToNullable(propertyValue);
 
-            if (filterClause != null && property.Type.IsCollection())
+            if (filterClause != null)
             {
-                IEdmTypeReference edmElementType = property.Type.AsCollection().ElementType();
+                var collection = property.Type.IsCollection();
```
robward-ms: nit: we'd like to move away from var, can you make this bool?

src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -187,30 +187,64 @@ internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IE
...
+                    if(filterParameter == null)
+                    {   
+                        //Never seen this, but just to be safe...
+                        throw new ODataException(Error.Format(SRResources.ExpandFilterExpressionLambdaExpressionNoParameter,
+                            property.Name));                            
+                    }
+
+                    var predicateExpression = new ReferenceNavigationPropertyExpandFilterVisitor(filterParameter, nullablePropertyValue).Visit(filterLambdaExpression.Body);
+
+                    // predicateExpression == true ? nullablePropertyValue : null
```
robward-ms: what's this line for?

src/System.Web.OData/OData/EnableQueryAttribute.cs
```diff
@@ -162,6 +162,23 @@ public int PageSize
         }
 
         /// <summary>
+        /// Honor $filter inside $expand of non-collection navigation property.
+        /// The expanded property is only populated when the filter evaluates to true.
+        /// This setting is false by default.
+        /// </summary>
+        public bool HandleReferenceNavigationPropertyExpandFilter
```
robward-ms: Is the idea of the setting just to maintain back-compat with the existing behavior, as opposed to wanting to turn it on or off?

src/System.Web.OData/OData/Query/ODataQuerySettings.cs
```diff
@@ -110,13 +110,24 @@ public HandleNullPropagationOption HandleNullPropagation
...
         }
 
+        /// <summary>
+        /// Honor $filter inside $expand of non-collection navigation property.
+        /// The expanded property is only populated when the filter evaluates to true.
+        /// This setting is false by default.
+        /// </summary>
+        public bool HandleReferenceNavigationPropertyExpandFilter
+        {
+            get; set;
```
mtroth-microsoft: >get; set; [](start = 12, length = 9)

nit: fix formatting.

src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -187,30 +187,64 @@ internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IE
...
+                    filterResult = Expression.Call(
+                        ExpressionHelperMethods.QueryableWhereGeneric.MakeGenericMethod(clrElementType),
+                        filterSource,
+                        filterPredicate);
+
+                    nullablePropertyType = filterResult.Type;
+                }
+                else if(_settings.HandleReferenceNavigationPropertyExpandFilter)
+                {                    
+                    var filterLambdaExpression = FilterBinder.Bind(null, filterClause, clrElementType, _context.RequestContainer) as LambdaExpression;
```
mtroth-microsoft: >var [](start = 20, length = 3)

can fix the vars in this method?

src/System.Web.OData/OData/EnableQueryAttribute.cs
taeinkimseattle: back-compat. By making this opt-in, this change does not change existing consumers and easier to accept :)

src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
taeinkimseattle: describes what kind of expression is created. will update comment

taeinkimseattle:
i made a mistake during merge, i will fix and re-push. please don't review this change until then

PR #1013, opened 2017-06-28
Allow WebApi tests to run in Debug configuration.

robward-ms:
The Top and Skip tests that check for value greater than the maximum integer
causes assertions the require the dismal of a popup. As such, these tests
fail in an automated environment.

This change removes the assert to allow the debug versions of the tests to pass.


xuzhg:
Merged at: https://github.com/OData/WebApi/commit/e80a23c734993bb34d3562bc5068c1b97ffedc6d

PR #1022, opened 2017-07-14
Add memory cache for storing compiled select/expand delegates

jamescourtney:
Adding opt-in capability to store precompiled delegates in a memory cache using a sliding window expiration policy. Since this is in the attribute, it should be possible to specify distinct caching policies on a per controller or per method basis, if that's desirable.

Thanks!

### DIFF
--- a/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
+++ b/src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
@@ -8,6 +8,7 @@ using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Runtime.Caching;
 using System.Web.Http;
 using System.Web.OData.Formatter;
 using System.Web.OData.Formatter.Serialization;
@@ -23,6 +24,11 @@ namespace System.Web.OData.Query.Expressions
     /// </summary>
     internal class SelectExpandBinder
     {
+        static SelectExpandBinder()
+        {
+            ResetCache();
+        }
+
         private SelectExpandQueryOption _selectExpandQuery;
         private ODataQueryContext _context;
         private IEdmModel _model;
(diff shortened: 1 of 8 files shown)


src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -23,6 +24,11 @@ namespace System.Web.OData.Query.Expressions
     /// </summary>
     internal class SelectExpandBinder
     {
+        static SelectExpandBinder()
+        {
+            ResetCache();
+        }
```
mtroth-microsoft: move this down with the other constructors.

src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -82,6 +95,32 @@ private IQueryable Bind(IQueryable queryable)
             return selectMethod.Invoke(null, new object[] { queryable, projectionLambda }) as IQueryable;
         }
 
+        private Delegate GetCompiledProjectionLambda()
+        {
+            string cacheKey = String.Format("{0}?select={1}&$expand={2}", _selectExpandQuery.Context.ElementClrType.FullName, _selectExpandQuery.RawSelect, _selectExpandQuery.RawExpand);
```
mtroth-microsoft: >select [](start = 49, length = 6)

missing $?

src/System.Web.OData/OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -82,6 +95,32 @@ private IQueryable Bind(IQueryable queryable)
...
+                // Use sliding window expiration to prevent unused things from hanging around indefinitely.
+                CacheItemPolicy evictionPolicy = new CacheItemPolicy
+                {
+                    SlidingExpiration = TimeSpan.FromSeconds(_settings.SelectExpandCacheExpirationTimeSeconds)
+                };
+
+                DelegateCache.Add(cacheKey, compiledExpression, evictionPolicy);
+            }
+
+            return compiledExpression;
```
mtroth-microsoft: why is this break out of get/add more efficient than the MemoryCache's own AddOrGet method?

AlanWong-MS:
@jamescourtney, again, we highly appreciate the time and energy you contributed to the project. It seems that now may not be the best time to merge the change due to your schedule, so we will park the PR in a "Closed" state. Should you find a good opportunity to move forward with the contribution, please feel free to comment on this PR or open a separate one. Thank you!

PR #1026, opened 2017-07-18
Always persist the original instance when selecting or expanding

ificator:
During serialization there are often scenarios where custom code requires access to the item being serialized. The `ResourceContext` provides a mechanism to access this instance, but for select/expand scenarios it ends up returning a different instance to that returned by the invoked action which can result in "internal" data being missing (i.e. data that isn't part of the model but is used to track state).

This PR attempts to resolve this by modifying the SelectExpandWrapper so that it ALWAYS has the original instance.

Validated by building and running unit tests.

PS: Did the PR description template disappear?

### DIFF
--- a/test/UnitTest/System.Web.OData.Test/OData/Query/Expressions/SelectExpandBinderTest.cs
+++ b/test/UnitTest/System.Web.OData.Test/OData/Query/Expressions/SelectExpandBinderTest.cs
@@ -74,7 +74,7 @@ namespace System.Web.OData.Query.Expressions
             Assert.True(enumerator.MoveNext());
             var partialCustomer = Assert.IsAssignableFrom<SelectExpandWrapper<Customer>>(enumerator.Current);
             Assert.False(enumerator.MoveNext());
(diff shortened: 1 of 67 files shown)


test/UnitTest/System.Web.OData.Test/OData/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -74,7 +74,7 @@ public void Bind_GeneratedExpression_ContainsExpandedObject()
             Assert.True(enumerator.MoveNext());
             var partialCustomer = Assert.IsAssignableFrom<SelectExpandWrapper<Customer>>(enumerator.Current);
             Assert.False(enumerator.MoveNext());
-            Assert.Null(partialCustomer.Instance);
+            Assert.NotNull(partialCustomer.Instance);
```
robward-ms: Can you add some additional validation to ensure Instance is set to the correct object?

PR #1030, opened 2017-07-20
Update NuSpec files for generating Nuget packages.

robward-ms:
### Description
Update NuSpec files for generating Nuget packages.
 - Add NuSpec files to build packages
 - Add GetNugetPackageMetadata.proj to set build variables to pass to Nuget.exe pack command.
 - Add WebStack.versions.settings.targets to control versioning
 - Remove company, copyright, version from CommonAssemblyInfo.cs in favor of generated file
   from WebStack.versions.settings.targets
A

- [   ] Test cases added

### DIFF
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -6,10 +6,7 @@ using System.Reflection;
 using System.Resources;
 using System.Runtime.InteropServices;
 
-#if !BUILD_GENERATED_VERSION
-[assembly: AssemblyCompany("Microsoft Corporation.")]
-[assembly: AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
-#endif
+[assembly: AssemblyProduct("Microsoft OData Web API")]
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyTrademark("")]
 [assembly: ComVisible(false)]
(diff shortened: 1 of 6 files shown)


src/CommonAssemblyInfo.cs
```diff
@@ -18,16 +15,3 @@
...
 [assembly: AssemblyMetadata("Serviceable", "True")]
-
-// ===========================================================================
-//  DO NOT EDIT OR REMOVE ANYTHING BELOW THIS COMMENT.
-//  Version numbers are automatically generated based on regular expressions.
-// ===========================================================================
-
-#if ASPNETODATA
-#if !BUILD_GENERATED_VERSION
-[assembly: AssemblyVersion("5.9.0.0")] // ASPNETODATA
```
mtroth-microsoft: >sembly: AssemblyVersion("5.9.0.0")] [](start = 3, length = 35)

if i link directly to the project what version will it have?

robward-ms: WebStack.versions.settings.targets generates a cs assembly info file from the data at the top of that file. The cs file is dropped into the obj dir during compilation. When compiling in VS, you get the same behavior as the build system so the version is whatever is in WebStack.versions.settings.targets.

mtroth-microsoft:
:shipit:

PR #1031, opened 2017-07-20
Update NuSpec files for generating Nuget packages. for maintenance-v4

robward-ms:
### Description
Update NuSpec files for generating Nuget packages.
 - Add NuSpec files to build packages
 - Add GetNugetPackageMetadata.proj to set build variables to pass to Nuget.exe pack command.
 - Add WebStack.versions.settings.targets to control versioning
 - Remove company, copyright, version from CommonAssemblyInfo.cs in favor of generated file
   from WebStack.versions.settings.targets
A

- [   ] Test cases added


mtroth-microsoft:
:shipit:

Issue #784, opened 2016-08-01
OData query with many fields produces StackOverflowException 


Casimodo72:
Hi,

a javascript client odata query produces a StackOverflowException on the server in EntityFramework when the query has lots of fields.
In my scenario the query succeeds with 94 fields (main and 13 $expanded tables), but not with 96 fields (i.e. 2 fields added to query over main table). Note that I do not select all fields of any of those tables.

The exception is thrown for me in EntityFramework's Coordinator<T>.ReadNextElement in the line:
element = TypedCoordinatorFactory.Element(shaper);
I'm not able (as in dunno how) to debug further down into the TypedCoordinatorFactory.Element method.

This issue is probably related to the bug described here: https://github.com/dotnet/corefx/issues/7953. In this scenario the OP seems to have a query over a single table. Also his stack trace does not include the Coordinator<T>.ReadNextElement method.
The folks over at corefx are asking for a LINQ-expression-only repro. Is someone able to produce such an expression, reflecting the shape of expression as produced by WebApi OData and EF? This should also include the $top and $count query options.

I'm not sure where to put this bug, because it might be a bug in the Lambda compilation, but I can only reproduced this using WebApi OData + EntityFramework.

Is someone already aware of this issue?

I'm using .NET 4.6, ASP.NET 4, MVC 5, OData v4 (6.15.0).

Regards,
Kasimier Buchcik

VikingsFan:
if there in no issue when you turn the data source into in-memory, then this should be EF's problem, since the exception is throwing in EF, waiting the repro on reference EF issue.

mikepizzo:
This appears to be an issue in .NET (see https://github.com/dotnet/corefx/issues/7953) fixed by https://github.com/dotnet/corefx/pull/11091. There is an open question as to whether OData may be building queries that exacerbate the issue.  This appears to affect a number of folks; @kosinsky has been involved in the discussion and may have more context.

kosinsky:
@mikepizzo @robward-ms Should issue be closed? Fix was merged

kevinayers:
Did this issue get ported over to .NET core?  I am having a very similar issue in .NET core v2.2 and I don't see the presence of the System.Linq.Expressions.StackGuard class.

Kevat:
[...]

PR #1033, opened 2017-07-24
Use "tree" instead of "linked list" for $select/$expand to decrease recursion for EF

kosinsky:
### Issues
*This pull request fixes issue #784.*  

### Description
For $select/$expand expression OData generates following expression tree
`.Select($it => new PropertyContainer() { Name = "Col1", Value = $it.Col1, Next = new PropertyContainer() { Name = "Col2", Value = $it.Col2, Next =....  } }`
when used with Entity Framework  with expression should be compiled. However Expression.Compile() fail with stack overflow, because very deep recursion.
To avoid deep recursion will generate expression in b-tree format
`.Select($it => new PropertyContainer() { Name = "Col1", Value = $it.Col1, Next1 = new PropertyContainer() { Name = "Col2", Value = $it.Col2, Next1 =....  } , Next2 = new PropertyContainer() { Name = "Col3", Value = $it.Col3, Next1 =....  }, ... }`

That means for 1K of fields we will have only Log(1k) level of recursion

- [ x ] Test cases added

### Additional work necessary
None


src/System.Web.OData/OData/Query/Expressions/PropertyContainer.cs
```diff
@@ -233,57 +273,108 @@ public override object GetValue()
...
+
+        internal class NamedPropertyWithNext<T> : NamedPropertyWithNextLeftOnly<T>
+        {
+            public PropertyContainer RightNext { get; set; }
+
+            public override void ToDictionaryCore(Dictionary<string, object> dictionary, IPropertyMapper propertyMapper,
+                bool includeAutoSelected)
+            {
+                base.ToDictionaryCore(dictionary, propertyMapper, includeAutoSelected);
+                if (RightNext != null)
```
robward-ms: Isn't NamedPropertyWithNext used in the case when LeftNext != null && RightNext != null already? Maybe leaving this out would provide more diagnostic information, i.e. crash because NamedPropertyWithNext was used with RightNext == null? 

Also wondering why this has a check for RightNext but NamedPropertyWithNextLeftOnly does not have a check for LeftNext.

kosinsky: With more columns Left/Right isn't enough to avoid stackoverflow. I'm implementing B-tree (with 32 branches for now)

src/System.Web.OData/OData/Query/Expressions/PropertyContainer.generated.tt
```diff
@@ -0,0 +1,64 @@
+﻿<#@ template debug="false" hostspecific="false" language="C#" #>
```
xuzhg: Do we really need the tt file to generate the codes?

kosinsky: It's 32*4 classes.

src/System.Web.OData/OData/Query/Expressions/PropertyContainer.generated.cs
```diff
@@ -0,0 +1,1693 @@
+﻿using System.Collections.Generic;
```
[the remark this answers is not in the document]
kosinsky: I've used option 3 from https://shishkin.wordpress.com/2008/07/08/stylecop-how-to-ignore-generated-code/

robward-ms:
@kosinsky - One last thing, can you rebase? This change currently fails to build in the CI environment but it looks like you are just missing a change made to the performance solution to fix an ODL version inconsistency.

PR #1038, opened 2017-08-04
Url too long Error

haoyang1989:
"$filter" requires a lot of "or" queries, but when the Url is too long, the page will prompt for "The Uri string is too long"

### DIFF
--- a/src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
+++ b/src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
@@ -112,12 +112,10 @@ namespace System.Web.OData.Routing
             {
                 Contract.Assert(serviceRoot != null);
 
-                serviceRootUri = new Uri(
-                    serviceRoot.EndsWith("/", StringComparison.Ordinal)
-                        ? serviceRoot
-                        : serviceRoot + "/");
+                string serviceRootStr = serviceRoot.EndsWith("/", StringComparison.Ordinal) ? serviceRoot : serviceRoot + "/";
+                serviceRootUri = new Uri(serviceRootStr);
 
-                fullUri = new Uri(serviceRootUri, odataPath);
+                fullUri = new Uri(serviceRootStr + odataPath);
                 queryString = fullUri.ParseQueryString();
                 uriParser = new ODataUriParser(model, serviceRootUri, fullUri, requestContainer);
             }


src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
```diff
@@ -112,12 +112,10 @@ private ODataPath Parse(string serviceRoot, string odataPath, IServiceProvider r
             {
                 Contract.Assert(serviceRoot != null);
 
-                serviceRootUri = new Uri(
-                    serviceRoot.EndsWith("/", StringComparison.Ordinal)
-                        ? serviceRoot
-                        : serviceRoot + "/");
+                string serviceRootStr = serviceRoot.EndsWith("/", StringComparison.Ordinal) ? serviceRoot : serviceRoot + "/";
+                serviceRootUri = new Uri(serviceRootStr);
```
robward-ms: serviceRootUri is now unused, please delete it here and the definition above.

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
```diff
@@ -112,12 +112,10 @@ private ODataPath Parse(string serviceRoot, string odataPath, IServiceProvider r
             {
                 Contract.Assert(serviceRoot != null);
 
-                serviceRootUri = new Uri(
-                    serviceRoot.EndsWith("/", StringComparison.Ordinal)
-                        ? serviceRoot
-                        : serviceRoot + "/");
+                string serviceRootStr = serviceRoot.EndsWith("/", StringComparison.Ordinal) ? serviceRoot : serviceRoot + "/";
```
robward-ms: I'm unclear how it solves the problem, do you know why?

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
```diff
@@ -112,12 +112,10 @@ private ODataPath Parse(string serviceRoot, string odataPath, IServiceProvider r
...
+                serviceRootUri = new Uri(serviceRootStr);
 
-                fullUri = new Uri(serviceRootUri, odataPath);
+                fullUri = new Uri(serviceRootStr + odataPath);
```
robward-ms: Can you add a test for long Uris?

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
haoyang1989: serviceRootUri is used below.
This is the original code.
uriParser = new ODataUriParser(model, serviceRootUri, fullUri, requestContainer);

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
haoyang1989: I'm not sure why. I found the same problem on Google, but no one knows how to solve it, so I tried to modify the source code and successfully implemented my OData query.

krzysztofwierzbicki: Problem is connected with such creation:
new Uri(new Uri, relativeUri)

I have really simillar problem, but the error i get is "Invalid URI: The Uri scheme is too long." 
It seems that microsoft library System.Uri have some bugs there. For example it takes relative uri and try to find scheme there. So if my OData request have any(a:(...)) with not encoded colon, it takes this colon as end of scheme. And there is limitation to scheme to 1024 signs. 

At least this is my case. I don't know is this the same case, but solution that you gave will probably solve my problem too.

haoyang1989: @krzysztofwierzbicki Yes, my solution now is to write a new class that inherits the IODataPathHandler interface and instantiate my new class when calling the MapODataServiceRoute method in the WebApiConfig file.

krzysztofwierzbicki: Its great idea! Thanks! But still i think about it as temporary solution, as the best would be fix it here.

haoyang1989: Yes, this is only a temporary solution.If can fix it here, that's the best.

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
robward-ms: Ah, missed it. My browser search was "serviceRootUri " instead of "serviceRootUri". Thanks.

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
robward-ms: This change will require adding a test.

xuzhg:
@haoyang1989 would you please response the @robward-ms 's comments?

src/System.Web.OData/OData/Routing/DefaultODataPathHandler.cs
haoyang1989: Sorry, I'm a rookie. I don't know how to write this test.

robward-ms: Take a look at the tests in DefaultODataPathHandlerTest.cs. Many of them call parse just like you would need to, you just need to construct a long uri similar to the issue you have been seeing.

xuzhg:
I have to close it. Please re-open it or send us new PRs. Thanks for your inputs.

PR #1052, opened 2017-08-16
[M-v4] modify the password with random guid to make CredScan happy

xuzhg:
Web API Security CredScan detects two passwords used in Web API Source code. 

They looks like:
`
var certificate = new X509Certificate2(
                certificateFilePath, "1234", X509KeyStorageFlags.MachineKeySet);
`

This change makes the CredScan happy.


xuzhg:
merge at https://github.com/OData/WebApi/commit/46585b3a87930698fc4a689e3ca53be60bfcf042

PR #1053, opened 2017-08-16
V3: modify the password with random guid to make CredScan happy 

xuzhg:
Web API Security CredScan detects two passwords used in Web API Source code. 

They looks like:
`
var certificate = new X509Certificate2(
                certificateFilePath, "1234", X509KeyStorageFlags.MachineKeySet);
`

This change makes the CredScan happy.


xuzhg:
merge at https://github.com/OData/WebApi/commit/af627749fd42a865017602e630ad735704a907c2

Issue #669, opened 2016-03-10
[vNext] Support for Entity Framework Core


pisees:
Hi,

What is the roadmap for support of Entity Framework Core?

Thanks,

xuzhg:
@pisees which version do you want to support EF core? Can you mention more?

pisees:
Hi,

I am looking for Entity Framework Core support in OData v4 with release that lines up with ASP.NET and Entity Framework Core. 

EF Core Roadmap lines up with ASP.NET Core: https://github.com/aspnet/EntityFramework/wiki/Roadmap

Overall trying to understand how the roadmaps of ASP.NET, Entity Framework and OData v4 line up.

Thanks, Kurt

robward-ms:
@pisees - We are working on an AspNetCore compatible version of OData WebApi (see #939) but that does not include the work for EFCore yet, we'll use this item to track it. But we'd likely only look at making our AspNetCore version work with the EFCore. Is this what you are looking for?

grahamehorner:
with regards to efcore support will odata also support entities that have no primary key as announced by the ef core v2.1  ?

robward-ms:
Users have had some success with OData WebApi for NetCore and EFCore. One notable exception, Aggregation fails due to lack of AsQueryable support, is discussed in #1154.

One obstacle is the lack of lazy-loading in [EFCore](https://docs.microsoft.com/en-us/ef/core/querying/related-data):

```
### Lazy loading
Lazy loading is not yet supported by EF Core. You can view the lazy loading item on our backlog to track [this feature](https://github.com/aspnet/EntityFramework/issues/3797).

```

OData WebApi expects that navigation properties can be loaded lazily. Using [EnableQuery] essential says "return to me a collection of something and I'll run queries against it for you". When those queries involve navigation properties, the lack of lazy loading causes the query to fail.

I have a branch with the EFCore version of Aggregation tests + solution here: [https://github.com/robward-ms/WebApi/tree/NetCore-E2ECore-EFCore](https://github.com/robward-ms/WebApi/tree/NetCore-E2ECore-EFCore)

keatkeat87:
when can we migrate from asp.net 4.x + ef + web api + odata 
to asp.net core, Is there an estimate time ?
or it's still a long way off.

xuzhg:
[...]

PR #1063, opened 2017-08-19
Removed support for deprecated linq to sql's Binary type >> mapping & expression building

ysmoradi:
Linq to sql itself is deprecated. I think it's not necessary to have a support for its Binary type anyway.

Additional info:
When I tried to run Microsoft.AspNet.OData 6.1 (The main one, I'm not talking about the new asp.net core one) on top of .net core, that was the only problem I faced. (I passed mono version of System.Data.Linq at runtime in AppDomain.CurrentDomain.AssemblyResolve event, and everything was working fine after that).

Thanks in advance

### DIFF
--- a/test/PerfTest/WebApiPerformance.Service/WebApiPerformance.Service.csproj
+++ b/test/PerfTest/WebApiPerformance.Service/WebApiPerformance.Service.csproj
@@ -26,16 +26,16 @@
   <Choose>
     <When Condition=" '$(TestType)'=='Private' ">
       <ItemGroup>
-        <Reference Include="Microsoft.OData.Core, Version=7.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-          <HintPath>..\..\..\sln\packages\Microsoft.OData.Core.7.2.0\lib\portable-net45+win8+wpa81\Microsoft.OData.Core.dll</HintPath>
+        <Reference Include="Microsoft.OData.Core, Version=7.3.1.10814, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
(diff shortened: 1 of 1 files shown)


ysmoradi:
That test was failing before I touch anything (:

mtroth-microsoft:
@ysmoradi  sorry for the build issues. there was a problem with our build trigger. I have corrected the problem and your latest build has run successfully.

ysmoradi:
@mtroth-microsoft YW :) Let's merge this PR!

robward-ms:
@ysmoradi - Did you mean Microsoft.AspNet.OData 6.1.0 beta? And the intention of the change is so allow 6.1.0 to run on top of .net core?

Dropping support for Binary could impact customers who rely on it. Seems like taking it out would constitute a breaking change.

I know you are active on #939 as well, do you agree that that is a better long term approach? I have prototyped the OData WebApi .NetCore support to find out that this bit of code becomes conditional in ASP.Net Core. Our current plan is to leave it in for ASP.Net and take is out for ASP.Net Core.

ysmoradi:
@robward-ms The initial work was started when there were no news about asp.net core version of OData. 

"... the intention of the change is so allow 6.1.0 to run on top of .net core?" Yes

"... do you agree that that is a better long term approach" Yes

We can ignore this PR, but I'm working on another PR for Microsoft.Extensions.DependencyInjection problem (Which prevents us from using libraries like EF Core 2.0 in asp.net based projects alongside MS.AspNet.OData 6.0). I'm not submitting that PR for the same reason as I submitted this PR. We'll talk about this later (;

robward-ms:
@ysmoradi  - Given your response, I'd like to reject this PR as it is now. We could consider taking it after evaluating how many people we might break.

For EF Core 2.0, maybe track this under #669?

AlanWong-MS:
Closing this PR per comments. Thank you for the support and contribution! Please feel free to comment for continued conversation and we may re-open as necessary.

PR #1066, opened 2017-08-29
Move System.Web.OData code to Microsoft.OData.WebApi-AspNet\ folder.

robward-ms:
Move System.Web.OData code from System.Web.OData\OData\ to Microsoft.AspNet.OData\ folder.
Move shared components to Microsoft.AspNet.OData.Common\ folder
    
The code does not compile in this change.
    
Commands:
mkdir .\src\Microsoft.AspNet.OData.Shared\Common
git mv .\src\Common\ .\src\Microsoft.AspNet.OData.Shared\Common\
git mv .\src\System.Web.OData\Properties\SRResources.Designer.cs .\src\Microsoft.AspNet.OData.Shared\Common\SRResources.Designer.cs
git mv .\src\System.Web.OData\Properties\SRResources.resx .\src\Microsoft.AspNet.OData.Shared\Common\SRResources.resx
    
mkdir .\src\Microsoft.AspNet.OData
git mv .\src\System.Web.OData\Extensions\ .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\Properties\ .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\Results\ .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\Batch\ .\src\Microsoft.AspNet.OData\
    
git mv .\src\System.Web.OData\GlobalSuppressions.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\Microsoft.AspNet.OData.Nightly.Release.nuspec .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\Microsoft.AspNet.OData.Release.nuspec .\src\Microsoft.AspNet.OData
git mv .\src\System.Web.OData\packages.config .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\Settings.StyleCop .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\System.Web.OData.csproj .\src\Microsoft.AspNet.OData\Microsoft.AspNet.OData.csproj
    
git mv .\src\System.Web.OData\OData\EnableQueryAttribute.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\ETagMessageHandler.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\FromODataUriAttribute.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\NonValidatingParameterBindingAttribute.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\ODataController.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\ODataFormattingAttribute.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\ODataNullValueMessageHandler.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\ODataQueryParameterBindingAttribute.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\ODataRoutingAttribute.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\PerRequestActionValueBinder.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\PerRequestContentNegotiator.cs .\src\Microsoft.AspNet.OData\
git mv .\src\System.Web.OData\OData\PerRequestParameterBinding.cs .\src\Microsoft.AspNet.OData\
    
mkdir .\src\Microsoft.AspNet.OData\Routing
git mv .\src\System.Web.OData\OData\Routing\ODataActionSelector.cs .\src\Microsoft.AspNet.OData\Routing
git mv .\src\System.Web.OData\OData\Routing\ODataPathParameterBindingAttribute.cs .\src\Microsoft.AspNet.OData\Routing
git mv .\src\System.Web.OData\OData\Routing\ODataPathRouteConstraint.cs .\src\Microsoft.AspNet.OData\Routing
git mv .\src\System.Web.OData\OData\Routing\ODataRoute.cs .\src\Microsoft.AspNet.OData\Routing
git mv .\src\System.Web.OData\OData\Routing\ODataValueProviderFactory.cs .\src\Microsoft.AspNet.OData\Routing
git mv .\src\System.Web.OData\OData\Routing\ODataVersionConstraint.cs .\src\Microsoft.AspNet.OData\Routing
    
mkdir .\src\Microsoft.AspNet.OData\Routing\Conventions
git mv .\src\System.Web.OData\OData\Routing\Conventions\IODataRoutingConvention.cs .\src\Microsoft.AspNet.OData\Routing\Conventions
git mv .\src\System.Web.OData\OData\Routing\Conventions\ODataRoutingConventions.cs .\src\Microsoft.AspNet.OData\Routing\Conventions
    
mkdir .\src\Microsoft.AspNet.OData\Formatter\
git mv .\src\System.Web.OData\OData\Formatter\DefaultODataETagHandler.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\IETagHandler.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataCountMediaTypeMapping.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataRawValueMediaTypeMapping.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\QueryStringMediaTypeMapping.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataBinaryValueMediaTypeMapping.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataEnumValueMediaTypeMapping.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataPrimitiveValueMediaTypeMapping.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataMediaTypeFormatters.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataMediaTypeFormatter.cs .\src\Microsoft.AspNet.OData\Formatter
git mv .\src\System.Web.OData\OData\Formatter\ODataModelBinderProvider.cs  .\src\Microsoft.AspNet.OData\Formatter
    
mkdir .\src\Microsoft.AspNet.OData\Formatter\Deserialization\
git mv .\src\System.Web.OData\OData\Formatter\Deserialization\ODataDeserializerProviderProxy.cs .\src\Microsoft.AspNet.OData\Formatter\Deserialization
    
mkdir .\src\Microsoft.AspNet.OData\Query\
git mv .\src\System.Web.OData\OData\Query\QueryFilterProvider.cs .\src\Microsoft.AspNet.OData\Query
    
mkdir .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\Builder\ .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\Formatter\ .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\Query\ .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\Routing\ .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ClrPropertyInfoAnnotation.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ClrTypeAnnotation.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\CompiledPropertyAccessor.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ContentIdHelpers.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\CustomAggregateMethodAnnotation.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\DefaultContainerBuilder.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\DeltaOfTStructuralType.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\Delta.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmChangedObjectCollection.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmComplexCollectionObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmComplexObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaCollectionType.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaComplexObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaDeletedEntityObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaDeletedLink.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaEntityKind.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaEntityObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaLink.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmDeltaType.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmEntityCollectionObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmEntityObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmEnumObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmEnumObjectCollection.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmModelExtensions.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmStructuredObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\EdmTypeExtensions.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ExpressionHelperMethods.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ExpressionHelpers.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\FastPropertyAccessor.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\FunctionImportComparer.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IDelta.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmChangedObject.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmComplexObject.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmDeltaDeletedEntityObject.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmDeltaDeletedLink.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmDeltaLink.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmDeltaLinkBase.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmEntityObject.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmEnumObject.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmObject.cs .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\IEdmStructuredObject.cs  .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\MetadataController.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\NavigationPropertyQueryableConfiguration.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\NavigationPropertyQueryableConfigurationAnnotation.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\NullEdmComplexObject.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataActionParameters.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataQueryContext.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataQueryContextExtensions.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataSwaggerConverter.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataSwaggerUtilities.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataUntypedActionParameters.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ODataUriFunctions.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\PageResult.cs .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\PageResultOfT.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\PropertyAccessor.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\QueryableRestrictions.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\QueryableRestrictionsAnnotation.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\RequestPreferenceHelpers.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ResourceContext.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ResourceContextOfTStructuredType.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\ResourceSetContext.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\TimeZoneInfoHelper.cs   .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\TypedDelta.cs .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\TypeHelper.cs .\src\Microsoft.AspNet.OData.Shared\
git mv .\src\System.Web.OData\OData\UnqualifiedCallAndEnumPrefixFreeResolver.cs   .\src\Microsoft.AspNet.OData.Shared\


mtroth-microsoft:
Why WebApi-AspNet instead of just WebApi?

xuzhg:
@robward-ms 

1.  do you think we should use "/" in branch name "OData:feature/netcore" ? 
2.  Can we name "Microsoft.OData.WebApi.Shared" as "Microsoft.AspNet.OData" or just "Microsoft.OData"?  "WebApi" looks not good be a part of a name?

robward-ms:
@mtroth-microsoft - My intention was to leave room for WebApi-AspNetCore, which comes after WebApi-AspNet is working again.

@xuzhg - feature/netcore is a standard git naming schema, one reference for it is here: https://github.com/agis/git-style-guide

@xuzhg - The AspNetCore layer's namespace needs to be considered. We could make this Microsoft.AspNet.OData and the NetCore layer Microsoft.AspNetCore.OData but will be a mix of Microsoft.AspNetCore.OData and Microsoft.AspNet.OData since the common layers expose public Api we don't want to wrap. That said, anything we choose would lead to that since we want to stay backwards compatible as much as possible, i.e. the AspNet and Shared layers needs to share a namespace. I picked Microsoft.OData.WebApi since this this is the WebApi layer of OData and we use Microsoft.OData.* for ODL.

robward-ms:
@xuzhg - After thinking about it some more, and considering what the AspNetCore version should be called, I think your suggestion makes a lot of sense. I pushed a new change with the intention of implementing the following in future PRs:

Code Structure:
•	Folder/Project: Microsoft.AspNet.OData – AspNet-specific code
•	Folder/Project: Microsoft.AspNet.OData.Shared – WebApi common code for both versions. I could also simply name the folder “Common”
•	Folder/Project: Microsoft.AspNetCore.OData – AspNetCore-specific code

AspNet Version
•	Namespace: Microsoft.AspNet.OData (platform-specific & shared code)
•	Binary: Microsoft.AspNet.OData.dll
•	Package: Microsoft.AspNet.OData (same as today)

AspNetCore Version:
•	Namespace: Microsoft.AspNetCore.OData (platform-specific) and Microsoft.AspNet.OData (shared code)
•	Binary: Microsoft.AspNetCore.OData.dll
•	Package: Microsoft.AspNetCore.OData

robertmclaws:
I like this structure a lot. But instead of using shared projects, why not just ship another binary in a separate package called Microsoft.AspNet.OData.Common that has the shared code? That way it can be revved independently as needed? That way we're not using complicated compiler tricks to get what we want, just a clean and simple structure where there is a 1-1 mapping between projects and binaries.

robward-ms:
@robertmclaws - Thanks for the feedback! While I agree that a shared binary would simplify project layout in that there is a 1-1 mapping between project and binary, I don’t agree that it simplifies release management and deployment. But first, a bit on the code.

I did prototype both a shared binary and a shared project but landed on a shared project for a couple of reasons. The first is framework compatibility. I started the shared binary as a .Net Standard project but quickly learned that to get access to all the APIs I needed, I had to target a version of .Net Standard that resulted in changing the OData WebApi dependency from Net Fx 4.5 to 4.6+. We’re trying to make the next version of OData WebApi for Asp.Net backwards compatible so this is not a dependency we wanted to introduce for this change. Another option would be using VS2017 ‘s multi-target binaries, meaning we could target Net Fx 4.5 and Net Core 1.x in a single project.  However, any platform difference that needs to be accommodated would leverage conditional compilation, which is the same as shared projects: either using symbols we define for shared project or the symbols VS defines for multi-target projects. Shared projects won’t wind up being a compiler trick any more than multi-target in that the code compiles for both platforms at the same time.

However, there were a few cases where a class that required access to a ASP .Net-specific classes was highly intertwined with the common code. For instance, ODataQueryParameterBindingAttribute is used in common code but requires access to ASP .Net-specific classes. Once solution we found for this was partial classes: the class is defined in the common code so it can be referenced there but the bulk of the implementation is defined in the ASP .Net-specific project. There are a few other examples where a traditional abstraction layer simply didn’t work without a more significant refactor, something else we wanted to avoid to maximize backward compatibility.

Additionally, we felt that adding a binary made development, release management and deployment more complex. If the common code binary is truly an independent package, and assuming a new feature would almost certainly need to modify common and platform-specific code, we’d find ourselves in an analogous situation to WebAPI and ODL development where it’s difficult to do side-by-side development. WebAPi depends on the public package of ODL, which is correct for deployment purposes, but developing, testing, validating and committing a WebApi change dependent on a unreleased ODL change requires quite a bit from the engineering system and release management. We could opt to not make them truly independent, much like ODL where we release all three packages for every version, which simplifies development and release management, but the truly independent package buys us very little. For deployment, we felt adding a new binary might only complicate deployment for existing customers who expect a single binary today. 

We do have some feature asks to make parts of WebApi consumable without requiring all the ASP.Net specific classes, such as using ModelBuilder in a stand-alone way. We looked at that as well and came to the conclusion we could refactor model build as part of ODL, specifically Edm. That said, a new package for WebAPI code  could be in our future but we think it’s more appropriate for something that is truly independent of the ASP.Net components but I think that is further down the road for WebApi.

robertmclaws:
Hey Rob, thanks so much for taking the time to explain all of this. Sorry I didn't get back to you sooner, I was on vacation. Ultimately the choice makes sense. The refactoring has been great so far. Looking forward when those efforts make their way up the chain to Restier.

Thanks!

Issue #975, opened 2017-05-11
[vNext] Even ASP.NET Core 2 Preview 1 announced, OData WebAPI vNext still no updates


alvachien:
Just found the ASP.NET Core 2 Preview 1 has been announced officially:
https://blogs.msdn.microsoft.com/webdev/2017/05/10/aspnet-2-preview-1/

But OData WebAPI vNext branch still not update to run on even for ASP.NET Core 1!

xiaoxin1983:
+1

ethnoplex:
+1

harsimranb:
The `Microsoft.AspNet.OData` assembly is available in Asp.Net Core 2.0, but there is no way to really use it as explained in the docs.  Is there any kind of documentation, or is the .Net Core 2.0 support just a fluke?

vitaliipanchenko:
+1

robward-ms:
Hi! Thanks to everyone for their patience. Yes, we will be releasing a version of OData WebApi that supports ASP.Net Core. We hope to have an alpha on NuGet in the next several weeks. After that, we’d like to get feedback before releasing an RTM version. If you are interested in help out, please let us know.

The current plan involves breaking changes. We want to release the ASP.Net Core version in a Microsoft namespace as opposed to the current System namespace. We decided to change the existing System namespace to a Microsoft namespace while leaving the existing APIs signature as-is. This does constitute a breaking change, albeit one we think is easy to accommodate. We’d love to get your feedback on this approach and its impact to you. If you have a different idea or just want to chime in with a pro or con, please let us know.

At a high level, the plan is to combine the code from the master branch (formerly OData60) and the maintenance-aspnetcore branch (formerly vNext) to create two version of OData WebApi from a common codebase: a ASP.Net WebApi 2 compatible version and an ASP.Net Core compatible version. Below, you’ll find some details on how we arrived at this plan. We plan to add a feature branch to the OData\WebApi repository and begin consolidating the changes there.

Feel free to chime in with thoughts and concerns regarding this plan on either this issue or a new GitHub issue. Again, thanks for your patience. – The OData Team.
---------------------------------------
[...]

Issue #772, opened 2016-07-12
[vNext] Roadmap for OData WebAPI run on ASP.NET Core 1.0


alvachien:
The ASP.NET Core 1.0 has officially released several weeks ago, the OData vNext (with alpha) still rely on 1.0 RC2 and cannot install from Nuget so far.

Does there any roadmap exist to support ASP.NET Core 1.0? 

Thanks,
Alva

VikingsFan:
@alvachien we will have a discuss about the roadmap after the 6.0.0 release, thanks :)

mterwoord:
@VikingsFan Would contributions for asp.net core support be accepted?

VikingsFan:
@mterwoord sure, contribution are always warmly welcome.

mterwoord:
What's the branch to pick?

VikingsFan:
vNext branch, but now we are not actively develop in this branch, we'll have a roadmap for this after 6.0.0 release

andygjp:
@VikingsFan when will v6 be released?

VikingsFan:
@alvachien 2-3 weeks after [OData Lib](https://github.com/OData/odata.net) release

aacanakin:
@alvachien Are there any updates on release date?

bpenggithub:
Any updates on supporting ASP.NET Core 1.0?

xuzhg:
@bpenggithub 

Please follow up https://github.com/OData/WebApi/tree/vNext

And You are welcome to share your contributions.

Thanks.

mowali:
@xuzhg and @VikingsFan are we as of now at the rtm phase of ODATA support to Core? I'm not interested in the alpha or beta, only stable release will convince my other coworkers in the team. 
I downloaded the sample and played with it. Most of its methods not working (I post a question about that) and I see that the sample project is using 6.0.0-alpha1-rtm, does it mean the ODATA is already rtm!? If not, when?

xuzhg:
@mowali 
The OData support to Asp.Net Core is not finished yet. The development is still going on. However such development slowed down in the past three months. I think once new resources are founded, the development will speed up. 

We do hope all of you can share us your contributions on the support OData for Core. Thanks!

candidodmv:
I'm very excited about this.

akorchev:
Hi. Is this feature still worked on?

xuzhg:
@akorchev and @all

[...]

Issue #628, opened 2016-02-04
vNext OData port


joshcomley:
Hello all,

Seeing as ASP.NET 5/ASP.NET Core/vNext/whatever-you-want-to-call-it is most certainly the future, will there be any concerted effort to continue the port? I see it was last touched 8 months ago. I have a fork that works with RC2, but it is still a very limited port (no $expand, $skip, $top, support for nested properties etc.).

Does anyone have any thoughts on this?

Many thanks

DickvdBrink:
I'm really interested in this too! I tried it a few days ago and it was a bit painfull to get it working with a later release.

@joshcomley, the one in master is indeed 8 months ago - but in the [vnext](https://github.com/OData/WebApi/tree/vNext) **branch** (not the folder) there is still some progress :)

YehudahA:
+1

grahamehorner:
the odata library namespace should be renamed to Microsoft.AspCore.Odata aligning the namespace with the standardisation that has been selected for EF  & MVC etc. which targets the Core CLR.

joshcomley:
Hey I am sorry I didn't get notified of these messages! I have a branch that I maintain that I keep working up to date with the latest ASP.NET Core version. I am unsure whether I should bother with pull requests to this, or as it looks relatively abandoned simply start a new open source OData project with the implementation I've adapted from here as a base. Does anyone know if this is legal to do?

joshcomley:
My branch can be found here:
https://github.com/joshcomley/WebApi/

It is up to date with ASP.NET Core, and with the namespace change, too.

joshcomley:
This also includes some support I've added for OData actions in the vNext implementation. Check my commit history for details of updates.

joshcomley:
FYI I have a pull request to get my work (and @grahamehorner's original work that mine was based on) merged:

https://github.com/OData/WebApi/pull/611

egbertn:
Still no clue :(
OData is not ported yet to ASPNET.Core

YehudahA:
OData is dead?

egbertn:
It's open source. Microsoft has used it for Azure (e.g.) and I guess, they have no huge team on it.

YehudahA:
No updates, no ecosystem, no contributes.

skysigal:
Sporadic equally spaced blog updates avoiding the elephant in the room. Vague mentions of reshuffling. 

This is starting to feel like Silverlight all over again... 

[...]

Issue #229, opened 2015-02-20
[vNext] Port ASP.NET Web API OData to ASP.NET 5/MVC 6


PinpointTownes:
I expect Web API OData to be really painful to port given that ASP.NET 5 and MVC 6 are radically different from the previous versions, but I think it's definitely worth it (sure, that's a totally biased opinion... I'm a huge fan of OData :dancer:).

For a complete port, we'd need a .NET core version of the ODataLib and EdmLib, but targeting `aspnet50` might be a reasonable objective for a first step.

I'll probably start working on a straight `aspnet50` port in the next couple of days, so let me know if you wanna participate :smile:

congysu:
@PinpointTownes
Thank you:)  Let's see how we can provide .NET core version of the ODataLib and EdmLib to you.  Shall we set up a phone call on this?

PinpointTownes:
@congysu thank you for your interest! :sake:

A few remarks concerning my progression:
* I almost got rid of `HttpRequestMessage` and `HttpResponseMessage` and replaced them with the new `HttpContext`. It was painful, but it's now done. That said, there is still a place I can't remove them from yet: in `MediaTypeMapping` (because `Microsoft.Net.Http.Headers` doesn't seem to have an equivalent for that). I'll ping @Tratcher when he's back.

* I removed the batching support from the first bits: I realize that it's of course a crucial feature, but MVC 6 doesn't support it yet, and it'll likely be implemented at the middleware level. This would need much more thinking, so I prefer removing it for now.

* I'm still undecided concerning OData routing. In the first bits, I removed the custom action selector and all the conventions to exclusively favor the attribute routing approach (that'll become the first-class citizen in MVC 6). This new version might also be a great opportunity to unify `ODataRoute` and `Route`/`HttpGet` attributes. The first tests using standard MVC 6 attributes are extremely encouraging, but only one EDM model per app is supported with this approach. Comments are welcome on these points :smile:

```csharp
public class MyController
{
    [HttpGet("Entities")]
    public IActionResult Get()
    {
        return new ObjectResult(new[]
        {
            new MyEntity { ID = 4 },
            new MyEntity { ID = 5 }
        });
    }

    [HttpGet("Entities({id})")]
    public object Get(int id, ODataQueryOptions<MyEntity> options)
    {
        return new MyEntity { ID = id };
    }

    [HttpGet("Entities({id})/Children")]
    public object GetChildren(int id)
    {
        return new ObjectResult(new[]
        {
[...]

PR #1075, opened 2017-09-13
Introduce WebApi interface layer in Microsoft.AspNet.OData.Shared

robward-ms:
### Issues

### Description

Introduce WebApi interface layer in Microsoft.AspNet.OData.Shared.
Adapt Microsoft.AspNet.OData.Shared to use interface layer.
Add complex return type to SelectControllerResult.
Abstract the creation of action mapping.
Move HttpMethod definitions to a common location.
Remove reference to ODataMediaTypeFormatter in resource files.

The code does not compile in this change.

- [   ] Test cases added

### Additional work necessary

 - Step 4: Implement interface layer in Microsoft.AspNet.OData.
 - Step 5: Move unit test code
 - Step 6: Adjust unit test code namespaces 
 - Step 7: Get unit tests to compile, run, and pass
 - Step 8: Get E2E tests to compile, run, and pass
 - Step 9: Update performance solutions & projects


src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.AspNet.OData
     /// <summary>
     /// The default container builder implementation based on the Microsoft dependency injection framework.
     /// </summary>
-    public class DefaultContainerBuilder : IContainerBuilder
+    public class DefaultContainerBuilder : Microsoft.OData.IContainerBuilder
```
mtroth-microsoft: >Microsoft.OData. [](start = 43, length = 16)

did you opt for this approach rather than adding a using statement because it was easier to find replace with this approach? bit of a nit, i guess, but if this were just a single file change we'd prefer to see the using statement.

src/Microsoft.AspNet.OData.Shared/HttpMethodHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// A helper class for comparing Http methods.
+    /// </summary>
+    internal class HttpMethodHelper
+    {
+        /// <summary>
+        /// "Get"
+        /// </summary>
+        public const string HttpGet = "GET";
```
mtroth-microsoft: >public const  [](start = 8, length = 13)

public const are kinda risky, want to use a static readonly instead?

src/Microsoft.AspNet.OData.Shared/ODataQueryParameterBindingAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using Microsoft.AspNet.OData.Query;
```
mtroth-microsoft: what is using this?

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -86,37 +87,29 @@ public class ODataConventionModelBuilder : ODataModelBuilder
...
-
-        /// <summary>
         /// Initializes a new <see cref="ODataConventionModelBuilder"/>.
         /// </summary>
-        /// <param name="configuration">The <see cref="HttpConfiguration"/> to use.</param>
-        public ODataConventionModelBuilder(HttpConfiguration configuration)
-            : this(configuration, isQueryCompositionMode: false)
+        /// <param name="resolver">The <see cref="IWebApiAssembliesResolver"/> to use.</param>
+        internal ODataConventionModelBuilder(IWebApiAssembliesResolver resolver)
+            : this(resolver, isQueryCompositionMode: false)
```
mtroth-microsoft: breaking? how will public call ctor?

src/Microsoft.AspNet.OData.Shared/Common/SRResources.Designer.cs
```diff
@@ -39,7 +39,20 @@ internal class SRResources {
         internal static global::System.Resources.ResourceManager ResourceManager {
             get {
                 if (object.ReferenceEquals(resourceMan, null)) {
-                    global::System.Resources.ResourceManager temp = new global::System.Resources.ResourceManager("System.Web.OData.Properties.SRResources", typeof(SRResources).Assembly);
+#if NETFX_CORE
+                    var assembly = typeof(SRResources).GetTypeInfo().Assembly;
+#else
+                    var assembly = typeof(SRResources).Assembly;
```
mtroth-microsoft: do we want to use var in this context?

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -843,6 +842,55 @@ public static bool IsNullable(Type type)
...
+            else
+            {
+                TryGetInnerTypeForDelta(ref type);
+                expectedPayloadType = model.GetEdmTypeReference(type);
+            }
+
+            return expectedPayloadType;
+        }
+
+        internal static bool TryGetInnerTypeForDelta(ref Type type)
```
mtroth-microsoft: >internal static bool TryGetInnerTypeForDelta(ref Type type) [](start = 8, length = 59)

why no comments on this function?

src/Microsoft.AspNet.OData.Shared/Formatter/ODataMediaTypes.cs
```diff
@@ -13,121 +13,39 @@ namespace Microsoft.AspNet.OData.Formatter
...
-            get { return (MediaTypeHeaderValue)((ICloneable)_applicationJsonStreamingTrue).Clone(); }
-        }
-
-        public static MediaTypeHeaderValue ApplicationXml
-        {
-            get { return (MediaTypeHeaderValue)((ICloneable)_applicationXml).Clone(); }
-        }
-
-        public static ODataMetadataLevel GetMetadataLevel(MediaTypeHeaderValue contentType)
+        public const string ApplicationJson = "application/json";
```
mtroth-microsoft: >public const [](start = 8, length = 12)

public static readonly better practice than public const

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
```diff
@@ -24,74 +24,23 @@ namespace Microsoft.AspNet.OData.Query
...
-        /// <summary>
         ///  Gets the given <see cref="ODataQueryContext"/>
         /// </summary>
         public ODataQueryContext Context { get; private set; }
 
         /// <summary>
         /// Gets the request message associated with this instance.
         /// </summary>
-        public HttpRequestMessage Request { get; private set; }
+        internal IWebApiRequestMessage InternalRequest { get; private set; }
```
mtroth-microsoft: are we enforcing sort based on public/internal/private? if so, internal comes after public.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -69,20 +69,7 @@ public abstract class ExpressionBinderBase
...
-        /// </summary>
-        /// <param name="requestContainer">The request container.</param>
-        protected ExpressionBinderBase(IServiceProvider requestContainer)
-        {
-            Contract.Assert(requestContainer != null);
-
-            QuerySettings = requestContainer.GetRequiredService<ODataQuerySettings>();
-            Model = requestContainer.GetRequiredService<IEdmModel>();
-            AssembliesResolver = requestContainer.GetRequiredService<IAssembliesResolver>();
-        }
```
mtroth-microsoft: you aren't sealing the class you are removing the protected ctor. does the new use case you are proposing work?

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
robward-ms: Step4 adds it back on the AspNet side, hence the partial tag being added here.

src/Microsoft.AspNet.OData.Shared/Common/SRResources.resx
```diff
@@ -121,7 +121,7 @@
     <value>Cannot serialize a null '{0}'.</value>
   </data>
   <data name="TypeCannotBeSerialized" xml:space="preserve">
-    <value>'{0}' cannot be serialized using the {1}.</value>
+    <value>'{0}' cannot be serialized using the ODataMediaTypeFormatter.</value>
```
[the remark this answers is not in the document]
robward-ms: Most of the places that use this string pass typeof(ODataMediaTypeFormatter) and most of those are in Shared code. ODataMediaTypeFormatter will be an AspNet-only class since it deals with AspNets-specific classes like MediaTypeFormatter and MediaTypeHeaderValue.

src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
robward-ms: I thought it was because of confusion between OData vs DependencyInjection classes but that appears to not be the case, the using works fine in this class.

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
robward-ms: It moved from a location where it had none but that should be fixed.

src/Microsoft.AspNet.OData.Shared/ODataQueryParameterBindingAttribute.cs
robward-ms: ODataQueryOptions.cs, ODataQueryOptionsOfTEntity.cs

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
robward-ms: The ctor is added back in Step4 on the AspNet side.

src/Microsoft.AspNet.OData.Shared/HttpMethodHelper.cs
robward-ms: This change is a bit harder due to these constants being used in switch/case statement. I push a change for all other comments, still working on this one.

src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
```diff
@@ -20,9 +21,9 @@ public class DefaultContainerBuilder : IContainerBuilder
         /// <param name="lifetime">The lifetime of the service to register.</param>
         /// <param name="serviceType">The type of the service to register.</param>
         /// <param name="implementationType">The implementation type of the service.</param>
-        /// <returns>The <see cref="IContainerBuilder"/> instance itself.</returns>
-        public virtual IContainerBuilder AddService(
-            ServiceLifetime lifetime,
+        /// <returns>The <see cref="Microsoft.OData.IContainerBuilder"/> instance itself.</returns>
+        public virtual Microsoft.OData.IContainerBuilder AddService(
```
mtroth-microsoft: > Microsoft.OData. [](start = 22, length = 17)

so do you still need these?

src/Microsoft.AspNet.OData.Shared/ODataQueryParameterBindingAttribute.cs
mtroth-microsoft: i meant, the using statement. doesn't seem needed.

---
In reply to: [139828269](https://github.com/OData/WebApi/pull/1075#discussion_r139828269) [](ancestors = 139828269)

robward-ms: It's because of <see cref="ODataQueryOptions"/> in the comments.

---
In reply to: [139848296](https://github.com/OData/WebApi/pull/1075#discussion_r139848296) [](ancestors = 139848296,139828269)

src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
robward-ms: Not the ones on IContainerBuilder but the others are needed to disambiguate OData vs DependencyInjection.

---
In reply to: [139848023](https://github.com/OData/WebApi/pull/1075#discussion_r139848023) [](ancestors = 139848023)

PR #1076, opened 2017-09-14
Add try logic for the host server create for KatanaSelfHost

xuzhg:
### Issues

Occasionally, we get test failures on WebApi. 

Event Log shows a lot of Error logged from E2E test running:

for example:

**Unable to bind to the underlying transport for [::]:9177. The IP Listen-Only list may contain a reference to an interface which may not exist on this machine.  The data field contains the error number.**

That's because the required port is being used for other process. 

### Description

Add a try logic to create the host server using different port.

- [ x ] Test cases added

### Additional work necessary


xuzhg:
Merged at: a973a83470a8cf104e0d5e634a8bcdab98a5b973

PR #1079, opened 2017-09-18
Update performance test project to use ODL 7.3.0

AlanWong-MS:
### Issues
No GitHub issues created for this error. The WebApiPerformance.Service.csproj project is breaking.

### Description
- Updating WebApiPerformance.Service.csproj to use ODL 7.3.0 to fix build break

- [ x ] Test cases added

### Additional work necessary
N/A


robward-ms:
You checked "test cases added" but didn't add test cases.

AlanWong-MS:
Thanks for the catch.

PR #1058, opened 2017-08-17
[maintaince-aspnetcore] Add support Asp.Net Core 2.0

genusP:

PR #1082, opened 2017-09-20
Call BuildServiceProvider using reflection to make it compatible with both MS.Extensions.DependencyInjection 1 & 2

ysmoradi:
### Issues

Short version of story:

Microsoft.Extensions.DependencyInjection has a class called **ServiceCollectionContainerBuilderExtensions** which has a method named **BuildServiceProvider**. In MS.Extensions.DependencyInjection 1.0, it was returning **IServiceCollection** and in version 2.0 it returns **ServiceCollection**

[This is a breaking change](https://github.com/aspnet/DependencyInjection/issues/550) and you've to either recompile your code or use dynamic method call using reflection. I'm not a fan of recompile against MS.Extensions.DependencyInjection 2.0, because it forces all developers who are using Microsoft.AspNet.OData 6.0+ to use .NET standard 2.0 based tools such as VS 2017.3 and the project itself should be updated from .NET 4.5 to .NET 4.6.1

Using dynamic method call, MS.AspNet.OData 6.0+ becomes compatible with both MS.Extensions.DependencyInjection 1 & 2.  **I've tested that code a lot and it works like a charm** and it makes **no breaking change** to this fantastic library at all.

This allows us to use libraries such as EntityFramework Core 2.0 and lots of other libraries on nuget.org which are dependent on MS.Extensions.DependencyInjection 2.0.

Thanks in advance

Issue link >> https://github.com/OData/WebApi/pull/1058#issuecomment-326710746

### Additional work necessary

**Change nuspec** dependencies from:

```
Microsoft.Extensions.DependencyInjection (>= 1.0.0 && < 2.0.0) 
Microsoft.Extensions.DependencyInjection.Abstractions (>= 1.0.0 && < 2.0.0) 
```

to

```
Microsoft.Extensions.DependencyInjection (>= 1.0.0) 
Microsoft.Extensions.DependencyInjection.Abstractions (>= 1.0.0) 
```

### DIFF
--- a/src/System.Web.OData/OData/DefaultContainerBuilder.cs
+++ b/src/System.Web.OData/OData/DefaultContainerBuilder.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+using System.Reflection;
 using System.Web.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
 using ServiceLifetime = Microsoft.OData.ServiceLifetime;
-using System.Reflection;
 
 namespace System.Web.OData
 {
--- a/src/System.Web.OData/OData/DefaultContainerBuilder.cs
+++ b/src/System.Web.OData/OData/DefaultContainerBuilder.cs
@@ -5,6 +5,7 @@ using System.Web.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
 using ServiceLifetime = Microsoft.OData.ServiceLifetime;
+using System.Reflection;
 
 namespace System.Web.OData
 {
@@ -78,7 +79,7 @@ namespace System.Web.OData
         /// <returns>The container built by this builder.</returns>
         public virtual IServiceProvider BuildContainer()
         {
-            return services.BuildServiceProvider();
+            return (IServiceProvider)services.GetType().GetTypeInfo().Assembly.GetType(typeof(ServiceCollectionContainerBuilderExtensions).GetTypeInfo().FullName).GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) }).Invoke(null, new object[] { services });
         }
 
         private static Microsoft.Extensions.DependencyInjection.ServiceLifetime TranslateServiceLifetime(
--- a/src/System.Web.OData/OData/DefaultContainerBuilder.cs
+++ b/src/System.Web.OData/OData/DefaultContainerBuilder.cs
@@ -79,7 +79,17 @@ namespace System.Web.OData
         /// <returns>The container built by this builder.</returns>
         public virtual IServiceProvider BuildContainer()
         {
-            return (IServiceProvider)services.GetType().GetTypeInfo().Assembly.GetType(typeof(ServiceCollectionContainerBuilderExtensions).GetTypeInfo().FullName).GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) }).Invoke(null, new object[] { services });
+            /* "services.BuildServiceProvider()" returns IServiceProvider in Microsoft.Extensions.DependencyInjection 1.0 and ServiceProvider in Microsoft.Extensions.DependencyInjection 2.0
+            * (This is a breaking change)[https://github.com/aspnet/DependencyInjection/issues/550].
+            * To support both versions with the same code base in OData/WebAPI we decided to call that extension method using reflection.
+            * More info at https://github.com/OData/WebApi/pull/1082
+            */
+
+            Assembly microsoftExtensionsDependencyInjectionAssembly = services.GetType().GetTypeInfo().Assembly;
+            TypeInfo serviceCollectionContainerBuilderExtensionsType = microsoftExtensionsDependencyInjectionAssembly.GetType(typeof(ServiceCollectionContainerBuilderExtensions).GetTypeInfo().FullName).GetTypeInfo();
+            MethodInfo buildServiceProviderMethod = serviceCollectionContainerBuilderExtensionsType.GetMethod("BuildServiceProvider", new[] { typeof(IServiceCollection) });
+
+            return (IServiceProvider)buildServiceProviderMethod.Invoke(null, new object[] { services });
         }
 
         private static Microsoft.Extensions.DependencyInjection.ServiceLifetime TranslateServiceLifetime(


robward-ms:
Build error:

OData\DefaultContainerBuilder.cs(82,174): error CS0103: The name 'nameof' does not exist in the current context [C:\BuildAgent\_work\116\s\src\System.Web.OData\System.Web.OData.csproj]

src/System.Web.OData/OData/DefaultContainerBuilder.cs
```diff
@@ -78,7 +79,7 @@ public class DefaultContainerBuilder : IContainerBuilder
         /// <returns>The container built by this builder.</returns>
         public virtual IServiceProvider BuildContainer()
         {
-            return services.BuildServiceProvider();
+            return (IServiceProvider)services.GetType().GetTypeInfo().Assembly.GetType(typeof(ServiceCollectionContainerBuilderExtensions).GetTypeInfo().FullName).GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) }).Invoke(null, new object[] { services });
```
robward-ms: I like the approach but the code is a bit unreadable. Can you break it up on multiple lines?

Also, can you add some comments? Without an explanation, someone is likely to wonder why we go through all this trouble instead of calling the method directly.

ysmoradi:
I try to improve it.

1- I'll remove **nameof**. (I was trying to make this dynamic call a little strongly typed)

2- I'll break it up on multiple lines

3- I'll add comments

ysmoradi:
I'm afraid, I wasn't able to find [CONTRIBUTION.md](https://github.com/OData/WebApi/blob/master/CONTRIBUTION.md) file, because its link in readme file is broken. May I create a separate branch from master for every PR I create? If yes, let me close this one, and after that, I'll submit another PR with change requests applied in separate branch.

AlanWong-MS:
@ysmoradi, thanks for pointing out the broken link. I'll fix that.

Yes, please create a separate branch from master for each PR you do. Is there a particular reason why you want to close this one as opposed to applying your changes here? (Or perhaps submitting this one as its own since we're almost there.)

ysmoradi:
@AlanWong-MS "Or perhaps submitting this one as its own since we're almost there" No reason other than following best practices. Ok, I'll continue to make this PR complete on its own way (master branch), and I'll create a separate branch for each PR after that.

AlanWong-MS:
@ysmoradi our build failed due to a code style check: there's a small adjustment you have to make in DefaultContainerBuilder.cs, which is to move the `using System.Reflection;` above the other namespaces:

src\System.Web.OData\OData\DefaultContainerBuilder.cs(8,1): Error : SA1208: System using directives must be placed before all other using directives.

ysmoradi:
Sorry about that. I'm gonna fix it soon

AlanWong-MS:
No worries. Please let us know if the build didn't tell you about the style issue when you built locally. We aim to let developers know about these checks as soon as possible to reduce time back-and-forth.

ysmoradi:
@AlanWong-MS Thanks. I read contribution (The link is working now), and I understood build.cmd's role, when I executed that, I received the same error ("System using directives must be placed before all other using directives.") Then I fixed namespaces sort, and now I see no error about that.

But everytime I ran build.cmd, I receive different test failures, even when I revert all my changes. May I create a separate issue for that?

AlanWong-MS:
@ysmoradi, great thanks so much for contribution! Glad to hear that the build.cmd is doing the checks. We're striving to improve the experience for devs when they run the command--please do create a separate issue for errors that you see. The build passed on our end so I'm going to merge your PR. :)

ysmoradi:
As I see in https://www.myget.org/feed/webapinightly/package/nuget/Microsoft.AspNet.OData, I think we've forgotten to apply additional works.

### Additional work necessary
**Change nuspec** dependencies from:
```
Microsoft.Extensions.DependencyInjection (>= 1.0.0 && < 2.0.0) 
Microsoft.Extensions.DependencyInjection.Abstractions (>= 1.0.0 && < 2.0.0) 
```
to
```
Microsoft.Extensions.DependencyInjection (>= 1.0.0) 
Microsoft.Extensions.DependencyInjection.Abstractions (>= 1.0.0) 
```
I had no access to those configurations in source code as I think they're managed in CI build variables.

Could you please fix this?

Thanks in advance.

robward-ms:
Actually, its on line 16 here: https://github.com/OData/WebApi/blob/master/tools/WebStack.versions.settings.targets

Since our packages refer to 1.x, I think the dependency is expressed correctly. We happen to be compatible with 2.0 due to your change but we still compile against 1.0.

I expect that for folks like yourself that want to use 2.0, a bindingRedirect in app.config would be the appropriate way to point towards the 2.0 binary.

ysmoradi:
The problem is, I can't even install MS.Extensions.DependencyInjection 2.0 library in my problems, because of this. The binding redirect is for the next step.
I really appreciate **Microsoft.Extensions.DependencyInjection (>= 1.0.0)** and after that, we need a simple assembly redirect, but **Microsoft.Extensions.DependencyInjection.Abstractions (>= 1.0.0 && < 2.0.0)** prevents me from installing MS.Extensions.DependencyInjection ):

ysmoradi:
We may change this:
```xml
<DependencyInjectionPackageDependency>[1.0.0, 2.0.0)</DependencyInjectionPackageDependency>
```
to:
```xml
<DependencyInjectionPackageDependency>1.0.0</DependencyInjectionPackageDependency>
```

robward-ms:
I understand the change we would need to make but that's the product change. 

Currently, we test with 1.1 (see https://github.com/OData/WebApi/blob/master/test/UnitTest/System.Web.OData.Test/packages.config).

If we want to change the Nuget dependency, we'd also need to make sure it works by testing it, right? So we won't just want to change the Nuget dependency but also setup a test project to make sure DI 2.0 works as expected.

Your original change was "make it work through reflection", which I think it a fine change but didn't change the support statement; you essential took on the risk of making sure it worked for you. Changing the Nuget dependency changes the support statement and we'd need to make sure it works for everybody. Without testing it, we don't know for sure whether it works.

Make sense?

ysmoradi:
It makes sense (As a matter of fact, I'm happy with this kind of project managements which results into a high quality product).

But, is there anything we can do about this?

If I can't install package V 2.0 because of nusepc, with or without reflection call, it doesn't work at all. By design, we can implement OData's IContainerBuilder interface with any container other than MS.Extensions.DependencyInjection. Using reflection based call + providing new implementation for IContainerBuilder using UseCustomerContainer method + assembly redirect + IL modification at runtime + everything I know in C# .NET,  **there's no way I can add any .net  standard 2.0 library** such as Ef core 2.0 to my project ):

ysmoradi:
Maybe we run tests twice against both versions automatically. I can submit a PR for that

ysmoradi:
We can make test project multi target against  .net 4.5 and 4.6.1 it doesn't broke anything. 4.5 is something like what we've at the moment, and a project which uses MS.Ext 2.0 is .NET 4.6.1 at least because of .net standard 2.0 for .net projects requirement. We add MS.Ext 1.1 for .net 4.5 target and 2.0 in 4.6.1 target. We may change build.cmd to reflect this changes (it should run both outputs), and in customer's perspective, there is no change.

Thanks for your attention

robward-ms:
Ya, that's what I was thinking; we'd create a test project for 2 versions each with a different DI dependency and then run both for testing. The existing XUnit test project, which is enough for checking DI I think, runs pretty quick so it's not lots of overhead for both.

ysmoradi:
:)
I'm ready to do the right thing. Shall we start with multi-target project ( I've no idea about how ready is the CI server for this ), or do you have any other method to discuss? Anyway, count on me.

robward-ms:
I'll handle the CI side of things. You should not be breaking anything by adding this so the existing CI builds should pass. Once the change is committed, I'll include it in the CI and official test runs.

As for layout, I started here: https://docs.microsoft.com/en-us/visualstudio/ide/visual-studio-multi-targeting-overview: "You could also create a solution that contains projects that target different versions of the framework."

VS2017 has some fancy multi-target features but we still use VS2013 to build WebApi and we have some changes coming to the project format for the NetCore support, at which point WebApi (as we know it today) will build using VS2015. I'd rather not churn the VS requirement for now.

In order to keep using VS2013, I think a separate project is what is needed. I'd add a new test project, just like the one that exists today, that produces a second binary and targets DI 2.x.  We would then want to run this in build.cmd (probably only under the full profile as opposed to quick). And of course, the packaging change in https://github.com/OData/WebApi/blob/master/tools/WebStack.versions.settings.targets 

Thanks for your contribution to this!

ysmoradi:
I understand the whole conceptual side of things, but for implementation with support for VS 2013, I've no practical idea other than creating a new project which adds another project's files [**as link**]
(https://grantwinney.com/visual-studio-add-file-as-link/)

But this is not possible in our scenario, because **adding a .net standard 2.0 library such as MS.DI 2.0 requires its own tooling**. VS 2017.3 have support for it built-in, vs 2015 can support this using nuget 3.6 and separate installation of support package.

[From dot net announcements:](https://github.com/dotnet/announcements/issues/32#issue-261095818)
If you use Visual Studio 2015 and use NuGet 3.6, we'll prompt you to install a support package which will handle copying the files to the output directory.

Compufreak345:
@robward-ms 
Is there any progress regarding this issue? As far as I understand this is what is preventing us from using this library with EFCore 2.1 Preview 1 and the only alternative would be using an inofficial package?

Thanks & best regards,
Christoph

ysmoradi:
As I know, the code itself has no problem. It's all about nuspec. MS.Extensions.DI's version is locked to 1, because automated tests are being executed against that version only. As more and more libraries use .net standard approach without multi targeting, and as usage of .net standard 2 becomes more and more, this issue becomes even worse. I think at least we can allow people use MS.Extensions.DI on their own risk. An exception can tell them a story behind this "if they use version 2", and they should use an method on http configuration to prevent this exception from being raised. Nuget package manager installs dependencies by lowest versions, so "existing applications will continue to work without any interrupt." If someone uses a library which is depend on version 2, they'll recieve that exception. @robward-ms I can handle this if you're ok with the concept.

robward-ms:
@Compufreak345 - This bug was for the AspNet/6.0 version, which does not work with .NetCore.  Our AspNetCore0-compatible package, Microsoft.AspNetCore.OData, is listed as compatible with Microsoft.Extensions.DependencyInjection (>= 2.0.0 && < 3.0.0) , see the [Nuget package](https://www.nuget.org/packages/Microsoft.AspNetCore.OData/7.0.0-beta2).

So there should be nothing blocking integration with EFCore 2.1 other than the fact that it is not fully supported yet, see #669.

The code in AspNet and AspNetCore is similar enough to thing that the AspNet version (both in feature/netcore and in master) will work with Microsoft.Extensions.DependencyInjection (>= 2.0.0 && < 3.0.0). I think @ysmoradi is correct, i.e. the code is fine.

@ysmoradi - I'm OK with the concept. Did you plan to update master and feature/netcore branches? Note that the DI version dependency is in [this file](https://github.com/OData/WebApi/blob/master/tools/WebStack.versions.settings.targets), which is in the same place for both branches.

ysmoradi:
@robward-ms Yeah. Thanks (-: I'm gonna handle that soon.

ysmoradi:
@robward-ms Sorry for my late response. I was too busy. Shall we remove DefaultContainerBuilder.cs from shared and develop aspnet | aspnetcore specific implementations?
It has two benefits in my point of view:
1- AspNetCore's minimal version of MS.Ext.DI is 2.0.0 and the reflection based invocation of BuildServiceProvider is not required, because of the MS.Ext.DI's breaking change relates to upgrading from 1 to 2, not 2 to 3.
2- After we changed aspnet version of nuspc for MS.Ext.DI from (>= 1.0.0 && < 2.0.0) to (>= 1.0.0 && < 3.0.0), as we talked we need to throw an exception by default for those folks which are interested in using MS.Ext.DI greater than 1 to ask their explicit permission. It can be achieved by asking them explicitly inherit from DefaultContainerBuilder class which has a virtual method in aspnet version only, and override that to explicitly allow usage of MS.Ext.DI 2.

robward-ms:
I wrote out a fairly long response to this with lots of options but I think the easiest is this: just upgrade AspNet classic to DI 2.0.

@ysmoradi - I think I was always trying to figure out how to get DI 2.0 working for the 6.x series. It seems from your response that you are looking at the 7.0 code, i.e. a 7.0 only solution could work for you?

I considered supporting both and how we might go about testing it, then just considered the upgrade and when we'd want to do it. A major release is the ideal time to make a breaking package change and we do indeed have a major upgrade in master now (AspNet classic 7.0). We could just update DefaultContainerBuilder.cs to use the 2.0 signature, tweak the nuget packaging spec and be done with it.

@xuzhg - Thoughts?

ysmoradi:
@robward-ms Totally agree 🤝 Let's introduce some useful breaking changes in MS.AspNet.OData 7.0. I'm waiting for @xuzhg thoughts. If he agrees too, I can submit another PR.

ysmoradi:
@xuzhg ?

xuzhg:
@ysmoradi Sorry, I missed your previous message. Please file PR for us. 

BTW, as .NET Core 2.1 released, do you think we should to target that version because .NET Core team said that 2.1 will be a recommendation version to develop.

PR #1084, opened 2017-09-22
Implement WebApi interface layer in Microsoft.AspNet.OData.

robward-ms:
### Issues

### Description

Implement WebApi interface layer in Microsoft.AspNet.OData. 
Adapt WebApi to use interface layer in Microsoft.AspNet.OData.

The Microsoft.AspNet.OData project now compile.

- [   ] Test cases added

### Additional work necessary

 - Step 5: Move unit test code
 - Step 6: Adjust unit test code namespaces 
 - Step 7: Get unit tests to compile, run, and pass
 - Step 8: Get E2E tests to compile, run, and pass & Update performance solutions & projects


build.cmd
```diff
@@ -7,11 +7,11 @@ mkdir bin
 
 :Build
 
-REM Find the most recent 32bit MSBuild.exe on the system. Require v12.0 (installed with VS2013) or later since .NET 4.0
+REM Find the most recent 32bit MSBuild.exe on the system. Require v14.0 (installed with VS2013) or later since .NET 4.0
```
AlanWong-MS: "Require v14.0 (installed with **VS2013**)" => "Require v14.0 (installed with **VS2015**)"

src/Microsoft.AspNet.OData/Adapters/WebApiActionDescriptor.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// The inner action wrapped by this instance.
+        /// </summary>
+        private HttpActionDescriptor innerDescriptor;
+
+        /// <summary>
+        /// Initializes a new instance of the WebApiActionDescriptor class.
+        /// </summary>
+        /// <param name="actionDescriptor">The inner descriptor.</param>
+        public WebApiActionDescriptor(HttpActionDescriptor actionDescriptor)
+            ////WebApiControllerDescriptor controllerDescriptor)
```
AlanWong-MS: Rogue comment

src/Microsoft.AspNet.OData/Adapters/WebApiActionMap.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        /// <summary>
+        /// Determines whether a specified key exists.
+        /// </summary>
+        /// <param name="name">The key.</param>
+        /// <returns>True if the key exist; false otherwise.</returns>
+        public bool Contains(string name)
+        {
+            return this.innerMap.Contains(name);
```
AlanWong-MS: How are the values (HttpActionDescriptor) of the map utilized? Should there be another method here for the HttpActionDescriptor?

src/Microsoft.AspNet.OData/Adapters/WebApiContext.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
+        public string RouteName
+        {
+            get { return this.innerContext.RouteName; }
+        }
+
+        /// <summary>
+        /// Gets the data store used by <see cref="IODataRoutingConvention"/>s to store any custom route data.
+        /// </summary>
+        /// <value>Initially an empty <c>IDictionary&lt;string, object&gt;</c>.</value>
```
AlanWong-MS: Text encoding/decoding hiccup?

src/Microsoft.AspNet.OData/Adapters/WebApiActionDescriptor.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// Gets the collection of supported HTTP methods for the descriptor.
+        /// </summary>
+        private IList<ODataRequestMethod> supportedHttpMethods;
+
+        /// <summary>
+        /// The inner action wrapped by this instance.
+        /// </summary>
+        private HttpActionDescriptor innerDescriptor;
+
+        /// <summary>
```
AlanWong-MS: Put <exception>...</exception> in the summaries? There are several constructors that require non-null parameters.

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        /// <summary>
+        /// Gets the HTTP method used by the HTTP request message.
+        /// </summary>
+        public ODataRequestMethod Method
+        {
+            get
+            {
+                string method = this.innerRequest.Method.ToString().ToUpperInvariant();
```
AlanWong-MS: nit: is `.ToUpperInvariant()` necessary if you're ignoring case later below?

src/Microsoft.AspNet.OData/Results/CreatedODataResult.cs
```diff
@@ -142,7 +144,7 @@ internal Uri EntityId
 
         internal IHttpActionResult GetInnerActionResult()
         {
-            if (RequestPreferenceHelpers.RequestPrefersReturnNoContent(Request))
+            if (RequestPreferenceHelpers.RequestPrefersReturnNoContent(new WebApiRequestHeaders(Request.Headers)))
```
AlanWong-MS: For debugging purposes, I would save `new WebApiRequestHeaders(Request.Headers)` into a `WebApiRequestHeaders` object and then pass the object to `RequestPrefersReturnNoContent`.

src/Microsoft.AspNet.OData/Adapters/WebApiActionDescriptor.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <returns>A list of attributes of type T.</returns>
+        public IEnumerable<T> GetCustomAttributes<T>(bool inherit) where T : Attribute
+        {
+            return this.innerDescriptor.GetCustomAttributes<T>(inherit);
+        }
+
+        /// <summary>
+        /// Determine if the Http method is a match.
+        /// </summary>
+        public bool IsHttpMethodMatch(ODataRequestMethod method)
```
biaol-odata: >IsHttpMethodMatch [](start = 20, length = 17)

nit: better name to be 'IsHttpMethodSupported'?

src/Microsoft.AspNet.OData/ResourceContext.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <see cref="M:NavigationSourceConfiguration.HasReadLink"/>
+    /// ) and navigation link (
+    /// <see cref="M:NavigationSourceConfiguration.HasNavigationPropertyLink"/>,
+    /// <see cref="M:NavigationSourceConfiguration.HasNavigationPropertiesLink"/>
+    /// ) builders and can be used by the link builders to generate links.
+    /// </summary>
+    public partial class ResourceContext
+    {
+        /// <summary>
+        /// Gets or sets the HTTP request that caused this instance to be generated.
```
biaol-odata: >caused [](start = 47, length = 6)

nit: "causes"?

src/Microsoft.AspNet.OData/ResourceSetContext.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// Gets or sets the <see cref="IEdmModel"/> to which this instance belongs.
+        /// </summary>
+        public IEdmModel EdmModel
+        {
+            get { return Request.GetModel(); }
+        }
+
+        /// <summary>
+        /// Gets or sets the HTTP request that caused this instance to be generated.
```
biaol-odata: >caused [](start = 47, length = 6)

nit:  "causes"?

src/Microsoft.AspNet.OData/Batch/ODataBatchRequestItem.cs
```diff
@@ -48,12 +49,27 @@ public abstract class ODataBatchRequestItem : IDisposable
...
+        private static void AddLocationHeaderToMapping(
+            HttpResponseMessage response,
+            IDictionary<string, string> contentIdToLocationMapping,
+            string contentId)
+        {
+            Contract.Assert(response != null);
+            Contract.Assert(contentIdToLocationMapping != null);
+            Contract.Assert(contentId != null);
+
+            if (response.Headers.Location != null)
```
biaol-odata: >response.Headers [](start = 16, length = 16)

Add !null check for response.Headers?

src/Microsoft.AspNet.OData/Adapters/WebApiActionMap.cs
robward-ms: They get used at a higher layer which is AspNet-version specific. At the lower layers, i.e. where the adapter is used, the Contains check is used to validate that action name and then the action name is returned. The name is then used in the higher layer to get the action descriptor. ASPNet has churned in this area quite a bit, replacing a controller and action description with a single descriptor and changing the collection used to map them.

src/Microsoft.AspNet.OData/Adapters/WebApiContext.cs
robward-ms: Note, that is intended to produce IDictionary<string, object> but you have to escape the <>'s in comments.

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
robward-ms: Not needed, it's a left-over from a time when a string was returned from this property instead of an enum.

src/Microsoft.AspNet.OData/Adapters/WebApiActionDescriptor.cs
robward-ms: I don't like "Won't Fix"ing comments but I am going to won't fix this. The null check pattern is used heavily in all OData code without the exception tags. I like the ide and it is proper but it's beyond the scope of this feature to add the tags everywhere that they are needed.

Issue #1078, opened 2017-09-16
GroupBy stop working at version 6.1.0 of Microsoft.AspNet.OData


jvitor83:
At version 6.0.0 of Microsoft.AspNet.OData it works. After upgrade to 6.1.0 GroupBy stop working!

### Assemblies affected

https://www.nuget.org/packages/Microsoft.AspNet.OData/6.1.0

### Reproduce steps

Make a GroupBy request to a OData endpoint.

- Clone this repo -> https://github.com/jvitor83/OData.Sample
- Run the solution
- Make a request to `http://localhost:60469/People?$apply=groupby((Id))`

### Expected result

The request return.

At the Sample: `[
    {
        "Id": 1
    },
    {
        "Id": 2
    }
]`

### Actual result

The following error at response:
```
{
    "Message": "Ocorreu um erro.",
    "ExceptionMessage": "O tipo 'ObjectContent`1' não pôde serializar o corpo da resposta para o tipo de conteúdo 'application/json; charset=utf-8'.",
    "ExceptionType": "System.InvalidOperationException",
    "StackTrace": null,
    "InnerException": {
        "Message": "Ocorreu um erro.",
        "ExceptionMessage": "Error getting value from 'NestedValue' on 'System.Web.OData.Query.Expressions.AggregationPropertyContainer+LastInChain'.",
        "ExceptionType": "Newtonsoft.Json.JsonSerializationException",
[...]

PR #1090, opened 2017-09-26
Support newtonsoft.json serialization for grouby results

kosinsky:
### Issues

*This pull request fixes issue #1078.*

### Description

Added converter for DynamicTypeWrapper to support cases when basic newtonsoft.json serialization is used instead of WebAPI formatters

- [ x ] Test cases added

### DIFF
--- a/src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
+++ b/src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
@@ -12,7 +12,6 @@ namespace System.Web.OData.Query.Expressions
     /// <summary>
     /// Represents a container class that contains properties that are grouped by using $apply.
     /// </summary>
-    [JsonConverter(typeof(DynamicTypeWrapperConverter))]
     public abstract class DynamicTypeWrapper
     {
         /// <summary>
@@ -33,6 +32,7 @@ namespace System.Web.OData.Query.Expressions
         }
     }
 
+    [JsonConverter(typeof(DynamicTypeWrapperConverter))]
     internal class GroupByWrapper : DynamicTypeWrapper
     {
         private Dictionary<string, object> _values;
--- a/src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
+++ b/src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using Newtonsoft.Json;
 [module: SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleClass", Justification = "Extra needed to workaorund EF issue with expression shape.")]
 
 namespace System.Web.OData.Query.Expressions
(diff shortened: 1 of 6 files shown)


src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
```diff
@@ -4,13 +4,15 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
robward-ms: Yep, it already exists. it's been there a while but sadly it's not listed in the Nuget package dependency.

test/UnitTest/System.Web.OData.Test/PublicApi/System.Web.OData.PublicApi.bsl
```diff
@@ -3216,6 +3216,9 @@ public class System.Web.OData.Formatter.Serialization.SelectExpandNode {
 	public static void GetStructuralProperties (Microsoft.OData.Edm.IEdmStructuredType structuredType, System.Collections.Generic.HashSet`1[[Microsoft.OData.Edm.IEdmStructuralProperty]] structuralProperties, System.Collections.Generic.HashSet`1[[Microsoft.OData.Edm.IEdmStructuralProperty]] nestedStructuralProperties)
 }
 
+[
+JsonConverterAttribute(),
```
robward-ms: This seems to be the only instance of this attribute in our public API.

kosinsky: In #1078 PeopleController.cs (https://github.com/jvitor83/OData.Sample/blob/master/OData.Sample/Controllers/PeopleController.cs) Is derived from ApiController instead of ODataController as  a result OData serialization will not be used. Instead default ASP.NET json serialization will happen and it uses newtonsoft.json.

Please, take a look on https://github.com/OData/WebApi/blob/master/src/System.Web.OData/OData/Query/Expressions/SelectExpandWrapper.cs it uses the same approach

src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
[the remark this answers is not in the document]
kosinsky: In https://github.com/OData/WebApi/blob/master/src/System.Web.OData/packages.config I can see: <package id="Newtonsoft.Json" version="6.0.4" targetFramework="net45" />

kosinsky:
In #1078 PeopleController.cs (https://github.com/jvitor83/OData.Sample/blob/master/OData.Sample/Controllers/PeopleController.cs) Is derived from ApiController instead of ODataController as  a result OData serialization will not be used. Instead default ASP.NET json serialization will happen and it uses newtonsoft.json.

Please, take a look on https://github.com/OData/WebApi/blob/master/src/System.Web.OData/OData/Query/Expressions/SelectExpandWrapper.cs it uses the same approach

test/UnitTest/System.Web.OData.Test/PublicApi/System.Web.OData.PublicApi.bsl
kosinsky: SelectExpandWrapper that uses converter too is internal => not visible for Public API.
I can make DynamicTypeWrapper internal too, but it will breaking change for public API

kosinsky: I moved to class that doing real job. Public API contains not changes now

src/System.Web.OData/OData/Query/Expressions/DynamicTypeWrapper.cs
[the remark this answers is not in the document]
xuzhg: I created an issue to track on it: https://github.com/OData/WebApi/issues/1097

Issue #1085, opened 2017-09-25
OData uses wrong Column Order in ORDER BY clause for Compound Keys


bmillspaugh:
OData adds an ORDER BY clause any time you are using PageSize = X to limit the number of records returned in an API call, but when you have a compound key it adds the keys to the ORDER BY clause in alphabetical order instead of using the order defined in the [Column(Order=X)] attribute from the data model.

This causes significant performance issues when reading tables with a lot of records since it means that SQL is not able to use the primary key index and has to perform a lengthy scan operation instead.  If OData used the correct order this issue would not happen.

### Reproduce steps

Create data model with following columns:

```
[Key]
[Column(Order = 1)]
public int A { get; set; }

[Key]
[Column(Order = 2)]
public int C { get; set; }

[Key]
[Column(Order = 3)]
public int B { get; set; }
```

Query OData with no $orderby parameter

### Expected result

ORDER BY A, **C**, **B**

### Actual result

ORDER BY A, **B**, **C**

biaol-odata:
@bmillspaugh: Any sample project/code for repo? Is WebAPI involved here or you were just using the OData Core Library?

AlanWong-MS:
Addressed by PR #1106.

Galad:
Hello,
@AlanWong-MS 
I'm trying to fix some performance issue in our application and I noticed the problem mentioned in this issue.
So I tried to use a version of AspNet.OData containing the fix from PR #1106  (6.2.0-Nightly20180717112 from the maintenance-6.x branch I believe), but without success, the orderby clauses are still ordered by name in the SQL query, meaning that the Column.Order property is still not used.

I took a look at the code and I noticed something strange on [this line](https://github.com/OData/WebApi/blob/b1890153f85b07571560179a27f288f19da05930/src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs#L576)
I'm wondering how the property `IEdmStructuralProperty.DeclaringType` can ever be an instance of `PrimitivePropertyConfiguration`, because from my understanding the property `DeclaringType` contains the type that declares the property. By executing the test that was added in the PR #1106, I noticed that the variable `value` is always null, because `DeclaringType` is not a `PrimitivePropertyConfiguration`.

The other thing I noticed is that the tests that were added are not present in the master branch, and were incorrect because the automatic order by was not applied (adding `$top=10` make the test fail).

Therefore I believe that this issue should be reopened.

AlanWong-MS:
[...]

PR #1096, opened 2017-09-30
Add support for columns order in query (Column attribute).

ConstantinOr:
### Issues

*This pull request fixes issue #1085.*

### Description

- [ x ] Test cases added

### Additional work necessary

### DIFF
--- a/src/System.Web.OData/OData/Builder/StructuralTypeConfiguration.cs
+++ b/src/System.Web.OData/OData/Builder/StructuralTypeConfiguration.cs
@@ -211,9 +211,27 @@ namespace System.Web.OData.Builder
         protected internal IList<PropertyInfo> RemovedProperties { get; private set; }
 
         /// <summary>
+        /// Variable for property ExplicitProperties.
+        /// </summary>
+        protected IDictionary<PropertyInfo, PropertyConfiguration> explicitInnerProperties;
+
+        /// <summary>
         /// Gets the collection of explicitly added properties.
         /// </summary>
-        protected internal IDictionary<PropertyInfo, PropertyConfiguration> ExplicitProperties { get; private set; }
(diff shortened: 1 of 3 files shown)


src/System.Web.OData/OData/Builder/StructuralTypeConfiguration.cs
```diff
@@ -211,9 +211,27 @@ public virtual IEnumerable<NavigationPropertyConfiguration> NavigationProperties
...
+            {
+                if (explicitInnerProperties.Any(a => a.Value.QueryConfiguration != null &&
+                        a.Value.QueryConfiguration.ModelBoundQuerySettings != null))
+                {
+                    explicitInnerProperties = OrderColumn(explicitInnerProperties);
+                }
+
+                return explicitInnerProperties;
+            }
+            private set => explicitInnerProperties = value;
```
biaol-odata: change to this?
private set { explicitInnerProperties = value; }

ysmoradi: Shall we set C# version in csproj, instead of these round trips?

ConstantinOr: Hello. I  can fix property and resubmit pull request. What you can recommend?

biaol-odata: @ConstantinOr Can you please also add some tests to show that the issue is fixed?

ConstantinOr: Ok. No problem.

ConstantinOr:
Sorry i was incorrectly understood problem.

Issue #822, opened 2016-09-27
EdmLibHelpers leaks memory when under stress


brjohnstmsft:
For OData services that create an EDM model on every client request (which is necessary for some data-plane services with dynamic schema), `EdmLibHelper._concurrencyProperties` grows to unbounded size. This causes managed heap memory to be promoted to gen 2, eventually causing page file expansion, thrashing, and serious performance degradation.

### Assemblies affected
`Microsoft.AspNet.OData` 5.9.0 (although [the flaw appears to be in the latest 6.0 code](https://github.com/OData/WebApi/blob/OData60/OData/src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs) as well).

### Reproduce steps
n/a; We discovered this bug via monitoring and by examining a memory dump of a crashed `w3wp.exe` process.

### Expected result
Services based on Web API OData that create an `EdmModel` on every request should not consume unbounded managed heap memory.

### Actual result
Memory utilization grows cumulatively over time as a function of the number of incoming requests.

### Additional details
The consequences of this bug are quite serious; I urge you to treat this as a high priority.

chrischen86:
In my organization, we dynamically generate a model for a dynamic schema for every client request and also suffer from the memory leak until the server crashes.

We discovered the cause of the issue to be EdmLibHelpers's static dictionary which uses an IEdmEntitySet as a key.  As a new model is generated with each request, the dictionary indexes are always misses since the key is a difference instance.  Combined with never disposing of any of the items in the dictionary, this dictionary can grow very quickly when serving up thousands of requests.

In my organization, we implemented a work around which is to not generate any ETags (as the method to do so attempts to use this cache dictionary).

1. Implement a custom serializer inheriting from ODataEntityTypeSerializer
2. Override CreateETag(EntityInstanceContext entityInstanceContext)

After doing so, our memory leak issue was resolved.  We also experimented with defining our own EdmEntitySet class which overrides the Equals method for proper equality comparison.  This approach also solves the issue but we did not need this.

Thanks,
Chris

PR #1099, opened 2017-10-04
Fix for memory leak in EdmLibHelpers 

kosinsky:
### Issues

*This pull request fixes issue #822.*

### Description

EdmLibHelpers caches concurrency properties in private static dictionary to improve performance. When WebAPI used with dynamic models (created on each call or pretty often), that dictionary will keep references on EntityTypes that ever used in queries. As a result we have memory leak.

Fix create cache per model and attaches it as an annotation to model. As a result when model is no longer in use entity types will be GCed too.

- [x] *Test cases added.* 

### Additional work necessary

### DIFF
--- a/src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
+++ b/src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
@@ -786,7 +786,7 @@ namespace System.Web.OData.Formatter
             Contract.Assert(navigationSource != null);
 
             // Ensure that concurrency properties cache is attached to model as an annotation to avoid expensive calculations each time
-            ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>> concurrencyProperties = model.GetAnnotationValue<ConcurrencyPropertiesAnnotation>(model);
+            ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>> concurrencyProperties = model.GetAnnotationValue<ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>>(model);
             if (concurrencyProperties == null)
             {
                 concurrencyProperties = new ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>();
@@ -994,13 +994,6 @@ namespace System.Web.OData.Formatter
                     type.Name.Replace('`', '_'),
                     String.Join("_", type.GetGenericArguments().Select(t => MangleClrTypeName(t))));
             }
-        }
-
-        /// <summary>
-        /// Annotation to store cache for concurrency properties
-        /// </summary>
-        private class ConcurrencyPropertiesAnnotation : ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>
-        {
         }
     }
 }
--- a/src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
+++ b/src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
@@ -31,8 +31,6 @@ namespace System.Web.OData.Formatter
 
         private static readonly IAssembliesResolver _defaultAssemblyResolver = new DefaultAssembliesResolver();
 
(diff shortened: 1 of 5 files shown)


src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
```diff
@@ -787,8 +785,16 @@ public static IEnumerable<IEdmStructuralProperty> GetConcurrencyProperties(this
...
+            ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>> concurrencyProperties = model.GetAnnotationValue<ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>>(model);
+            if (concurrencyProperties == null)
+            {
+                concurrencyProperties = new ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>();
+                model.SetAnnotationValue(model, concurrencyProperties);
+            }
+
             IEnumerable<IEdmStructuralProperty> cachedProperties;
-            if (_concurrencyProperties != null && _concurrencyProperties.TryGetValue(navigationSource, out cachedProperties))
+            if (concurrencyProperties != null && concurrencyProperties.TryGetValue(navigationSource, out cachedProperties))
```
mikepizzo: >concurrencyProperties [](start = 16, length = 21)

This will no longer ever be null at this point, right? If it was null in line 790 you would have created it in 792.

src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
```diff
@@ -825,12 +831,15 @@ public static IEnumerable<IEdmStructuralProperty> GetConcurrencyProperties(this
                 }
             }
 
-            if (_concurrencyProperties == null)
+            if (concurrencyProperties == null)
             {
-                _concurrencyProperties = new ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>();
+                concurrencyProperties = new ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>();
             }
```
mikepizzo: You shouldn't need Lines 834-837 since you already made sure concurrencyProperties was non-null in 792.

mikepizzo:
Caching these in an annotation is clever, but might be surprising to the client if they are enumerating annotations and discover we've added this info.

kosinsky:
We are doing the same for ClrPropertyInfoAnnotation (line 691 in the EdmLibHelpers.cs). To avoid surprises I created public annotation ConcurrencyPropertiesAnnotation

src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
```diff
@@ -825,12 +831,10 @@ public static IEnumerable<IEdmStructuralProperty> GetConcurrencyProperties(this
                 }
             }
 
-            if (_concurrencyProperties == null)
+            if (results.Any())
```
mikepizzo: Don't you want to cache the results whether they are empty or not, so that you don't have to check again next time you see the type?

mikepizzo:
What happens when we try to write out the model?  How does this annotation get serialized?

---
In reply to: [334847523](https://github.com/OData/WebApi/pull/1099#issuecomment-334847523) [](ancestors = 334847523)

src/System.Web.OData/OData/Formatter/EdmLibHelpers.cs
kosinsky: Thanks. I missed that.

kosinsky:
That annotation isn't vocabulary annotation. AFAIK, it isn't serializable as well as ClrPropertyInfoAnnotation

xuzhg:
@robward-ms @mikepizzo What do you think about this PR?

mikepizzo:
As long as it doesn't affect writing, I'm fine with this.  Can we have a test that sets this and then tries to write out the CSDL?

kosinsky:
@mikepizzo done

mikepizzo:
Perfect; Thanks @Kosinsky!

PR #1110, opened 2017-10-18
Enable unit tests in features/netcore

robward-ms:
### Issues

### Description
Microsoft.Test.AspNet.OData and Microsoft.Test.AspNet.OData.TestCommon projects now compile.
WebApiOData.sln now compiles.

All UnitTests are passing.

- [   ] Test cases added

### Additional work necessary

 - Step 8: Get E2E tests to compile, run, and pass & Update performance solutions & projects


test/UnitTest/Microsoft.Test.AspNet.OData/PublicApi/PublicApiHelper.cs
```diff
@@ -402,11 +402,11 @@ private static bool AppendCustomAttributes(StringBuilder builder, object[] attri
                 }
                 foreach (object attribute in attributes)
                 {
-                    if (attribute is Runtime.InteropServices.MarshalAsAttribute)
+                    if (attribute is System.Runtime.InteropServices.MarshalAsAttribute)
```
xuzhg: Can we use "Using System;"?

robward-ms: For this issue, it's a bit more complicate since the new namespace starts with "Microsoft". The compiler searches up the namespace tree and never finds an assumes Runtime must be "Microsoft.Runtime". The file already has a "using System" but it doesn't help in this case.

That said, I can remove the partial namespace declarations and make them all usings.

PR #1115, opened 2017-10-25
Fix the bug#1113: ODataResourceDeserializer on longer supports null c…

### LINKED ISSUE #1113: ODataResourceDeserializer no longer supports null ComplexType values
v6.0.1 implements "ApplyNestedProperty" in which the foreach statement which iterates the NestedItems attempts to figure out what type the value is (ODataEntityReferenceLinkBase, ODataResourceSetWrapper, ODataResourceWrapper).

Issue is however that when the value of the ComplexProperty specified in the payload is "null" the property becomes ignored.  Consumers then have no idea the value was specified with 'null' for a value.

xuzhg:
Fix #1113

### Issues

ODataResourceDeserializer no longer supports null ComplexType values 

### Description

ODataResourceDeserializer no longer supports null ComplexType values 

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/System.Web.OData/OData/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/System.Web.OData/OData/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -308,6 +308,21 @@ namespace System.Web.OData.Formatter.Deserialization
 
             foreach (ODataItemBase childItem in resourceInfoWrapper.NestedItems)
             {
+                // it maybe null.
+                if (childItem == null)
+                {
+                    if (edmProperty == null)
+                    {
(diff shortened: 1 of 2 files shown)


src/System.Web.OData/OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -345,6 +360,7 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
                         ApplyResourceInNestedProperty(edmProperty, resource, resourceWrapper, readContext);
                     }
                 }
+
```
robward-ms: Remove this blank line to pass the build.

PR #1116, opened 2017-10-26
Fix cmdline build, update version to 7.0.0 alpha

robward-ms:
### Issues

### Description

Set version to 7.0.0 alpha.
Fix code analysis/style copy issues.
Update msbuild scripts to 2015.
Update Nuspec files.

- [   ] Test cases added

### Additional work necessary


src/Microsoft.AspNet.OData.Shared/Routing/Conventions/DynamicPropertyRoutingConvention.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
     /// </summary>
     public partial class DynamicPropertyRoutingConvention
     {
-        private const string _actionName = "DynamicProperty";
```
xuzhg: the recommended coding style is to add "_"  for the field?

robward-ms: Not for private const.

PR #1118, opened 2017-10-30
Finalize split between Shared and AspNet in preperation for AspNetCore.

robward-ms:
### Issues

### Description

Finalize split between Shared and AspNet. A few AspNet-specific bits leaked into the shared
code during refactoring of the first 8 PRs.

Additionally, this change modifies the Metadata controller to be mostly shared
with the a bit of AspNet-specific code to get the model.

- [   ] Test cases added

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.OData.Builder
                 throw Error.ArgumentNull("resourceContext");
             }
 
-            if (resourceContext.Url == null)
+            if (resourceContext.InternalUrlHelper == null)
             {
                 throw Error.Argument("resourceContext", SRResources.UrlHelperNull, typeof(ResourceContext).Name);
             }
@@ -72,7 +72,7 @@ namespace Microsoft.AspNet.OData.Builder
             {
                 throw Error.ArgumentNull("resourceContext");
             }
-            if (resourceContext.Url == null)
+            if (resourceContext.InternalUrlHelper == null)
             {
                 throw Error.Argument("resourceContext", SRResources.UrlHelperNull, typeof(ResourceContext).Name);
             }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataCollectionSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataCollectionSerializer.cs
@@ -7,7 +7,6 @@ using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Runtime.Serialization;
 using Microsoft.AspNet.OData.Common;
-using Microsoft.AspNet.OData.Extensions;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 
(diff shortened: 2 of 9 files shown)


robertmclaws:
Since you mentioned the tests in here, I'm hoping you folks are planning on improving the tests before RTM. As an outsider, I've found that the E2E tests are super-convoluted and difficult to follow. I think the previous team also used them as the single source of documentation for a bunch of features, which was not very helpful to devs... especially when the tests took a lot of set-up to get working.

To help with that, I made a testing framework called [Breakdance](https://github.com/cloudnimble/breakdance), based in part on concepts that were in the Restier unit tests. They consist of multiple components, each designed to make it trivial (one line of code) to set up tests. `Breakdance.Assemblies` tests the public API surface of an assembly by comparing the current test run against a previously checked-in run, and `Breakdance.WebApi` sets up a self-host WebAPI server to test requests against. (You can see an example [here](https://cloudnimble.github.io/Breakdance/api/Breakdance.WebApi/System.Net.Http.HttpClientExtensions.html)).

Since Breakdance traces its origins back to the OData team, it would be great if the OData team adopted it for their tests. When you get to that point, I'd be happy to help out.

Thanks, and keep up the great work!

robward-ms:
@robertmclaws  - My plan has been to migrate the Nuwa framework for AspNetCore, which requires a two major things: 
1.) Upgrade to Xunit 2 
2.) Abstract the WebApi setup to make it platform agnostic (BTW, I'm doing the same for unit tests and have been successful so far)

My goal was to make to largely leave the tests unchanged and adapt them to work on both platforms, making it easier to maintain  single set of tests to verify both platforms.

Do you think BreakDance is up to that challenge?

robward-ms:
@robertmclaws - One other thing I have not solved yet that's a test challenge: making an AspNet/AspNetCore compatible controller. Any ideas?

robertmclaws:
@robward-ms What Is the Nuwa framework? I tried to search on it but didn't find anything.

Regarding making the WebAPI setup platform agnostic, Breakdance *could* have a WebApiCore implementation that would let you have the same API surface execute on both platforms. 

I'm not sure how far down the road you'd be able to get on having the *exact* same classes run tests for both platforms, though. Once think you'd have to think about is how easy it would be for non-Microsoft developers to get their heads around. Definitely would be interested in brainstorming about it, tho.

robward-ms:
@robertmclaws - Nuwa is the framework currently used in WebApi E2E tests. It's contained in 2 projects, we have two more that provide test support and then the actually E2E project. They are all in the E2E solution, along with product code.

The first step down the Nuwa path is to convert the E2E tests top XUnit 2, which is required for a .Net Core version of the tests (XUnit 1.x does not support .Net Core). Nuwa hooks into XUnit's extensibility to setup the server, the extensibility points have change in Xunit 2.x. I've prototyped the new kook and have a few tests running under Nuwa & Xunit 2.0, my next step was to port them all. Let me stabilize that and push it, it will show up as "Step10" in my private repo. I'll have that done my Monday if you want to take a look.

robward-ms:
I took a look at Breakdance and it seems the basic idea is to spin up an ASPNet implementation (HttpServer & HttpClient) without routing it through a web server, i.e. IIS. While this approach is used in some of the unit tests, the E2E tests spin up an actual web service. 

As I'm getting Nuwa to work in Xunit2 and trying to debug the failures, I am starting to wonder what the value of an actual web service in the E2E tests. Seems like a much more slim approach would still provide *nearly* the same level of testing from an E2E perspective. 

I will readily admit to disliking debugging E2E failures due to lack of logging and the complexity of it all.

Thoughts?

robward-ms:
@robertmclaws - I've done some more work on the E2E tests. I had taken the existing code and ported it too Xunit 2.3.1 and it's new extensibility points and had it running. But as I considered porting that to AspNetCore, I realized, as you pointed out, that there is a lot of overhead. This week I removed the overhead and kept 4 files from the Nuwa framework (and copied the other projects into the existing test project) so it's down to one project with a simple base class that starts the web server. I'll likely have a PR for that next week.

PR #1121, opened 2017-11-02
Compile Shared code for NetCore 1.x & 2.x

robward-ms:
### Issues

### Description

Compile Shared code for NetCore 1.x & 2.x

- [   ] Test cases added

### Additional work necessary


src/Microsoft.AspNet.OData.Shared/Common/Error.cs
```diff
@@ -225,7 +225,7 @@ internal static OperationCanceledException OperationCanceled(string messageForma
         /// <returns>The logged <see cref="Exception"/>.</returns>
         internal static ArgumentException InvalidEnumArgument(string parameterName, int invalidValue, Type enumClass)
         {
-#if NETFX_CORE
+#if NETCORE1x
```
AlanWong-MS: So NETCORE1x doesn't have the InvalidEnumArgumentException type?

test/UnitTest/Microsoft.Test.AspNet.OData/TypeHelperTest.cs
```diff
@@ -67,12 +76,262 @@ public void ToNullable_Returns_ExpectedValue(Type type, Type expectedResult)
...
+            abstract public int Area();
         }
 
-        private class MyCustomCollection : List<int>
+        /// <summary>
+        /// Custom abstract class
+        /// </summary>
+        private class CustomConcreteClass : CustomAbstractClass
         {
+            override public int Area() { return 42; }
```
AlanWong-MS: 👍

PR #1122, opened 2017-11-03
Add fix for DateTimeOffset Day function test

mtroth-microsoft:
### Issues

*The issue address a problem in the test where the comparison fails when the day of the month goes beyond the maximum day of the month in the control months used in the test. So, on October 31st, the test failed because the results were compared to November 30th and September 30th (as well as August 31st and December 31st). These two entries were mismatched because the months did not have 31 days. This caused the test to fail.

### Description

*The broken test was modified so that the control data will be filtered when doing the comparison. Thus, any control instances that don't match with the test criteria will not be expected to exist in the query results.

- [ ] *Test cases added*

### Additional work necessary

*None needed


robward-ms:
@mtroth-microsoft - Could you either add a description of the fix to the commit or PR description?

PR #1124, opened 2017-11-06
Initial commit of AspNetCore project in buildable state.

robward-ms:
### Issues

### Description

Initial commit of AspNetCore project in buildable state.

- [   ] Test cases added

### Additional work necessary

### DIFF
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -6,10 +6,12 @@ using System.Reflection;
 using System.Resources;
 using System.Runtime.InteropServices;
 
-#if !NETCORE
-[assembly: AssemblyProduct("Microsoft OData Web API")]
-[assembly: AssemblyConfiguration("")]
+#if NETCORE
+[assembly: AssemblyProduct("Microsoft OData Web API for ASP.NET Core")]
+#else
+[assembly: AssemblyProduct("Microsoft OData Web API for ASP.NET")]
 #endif
+[assembly: AssemblyConfiguration("")]
 [assembly: AssemblyTrademark("")]
 [assembly: ComVisible(false)]
 #if !NOT_CLS_COMPLIANT
--- a/src/Microsoft.AspNetCore.OData/Adapters/WebApiUrlHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Adapters/WebApiUrlHelper.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNet.OData.Adapters
+{
+    /// <summary>
+    /// Adapter class to abstract the Asp.Net Url helper.
+    /// </summary>
+    internal class WebApiUrlHelper : IWebApiUrlHelper
+    {
+        /// <summary>
+        /// The inner helper wrapped by this instance.
+        /// </summary>
+        internal IUrlHelper innerHelper;
+
+        /// <summary>
+        /// Initializes a new instance of the WebApiUrlHelper class.
+        /// </summary>
+        /// <param name="helper">The inner helper.</param>
+        public WebApiUrlHelper(IUrlHelper helper)
+        {
+            if (helper == null)
+            {
+                throw Error.ArgumentNull("helper");
+            }
+
+            this.innerHelper = helper;
+        }
+
+        /// <summary>
+        /// Generates an OData link using the request's OData route name and path handler and given segments.
+        /// </summary>
+        /// <param name="segments">The OData path segments.</param>
+        /// <returns>The generated OData link.</returns>
+        public string CreateODataLink(IList<ODataPathSegment> segments)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Generates an OData link using the request's OData route name and path handler and given segments.
+        /// </summary>
+        /// <param name="segments">The OData path segments.</param>
+        /// <returns>The generated OData link.</returns>
+        public string CreateODataLink(params ODataPathSegment[] segments)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Generates an OData link using the given OData route name, path handler, and segments.
+        /// </summary>
+        /// <param name="routeName">The name of the OData route.</param>
+        /// <param name="pathHandler">The path handler to use for generating the link.</param>
+        /// <param name="segments">The OData path segments.</param>
+        /// <returns>The generated OData link.</returns>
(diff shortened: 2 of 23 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// Gets or sets the HTTP Request whose response is being serialized.
+        /// </summary>
+        public HttpRequest Request
+        {
+            get { return _request; }
+
+            set
+            {
+                _request = value;
+                //TODO: Set InternalRequest.
```
AlanWong-MS: Is this TODO cleared by the line preceeding it?

robward-ms: Nope, like line 42, another link is needed to set InternalRequest to a new instance of a class that does not exist in the code yet. It will be in the next PR.

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.csproj
```diff
@@ -0,0 +1,61 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
AlanWong-MS: Double-checking, we're targeting 2.0?

src/Microsoft.AspNetCore.OData/ODataController.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.OData.Edm;
+using Microsoft.OData.Edm.Csdl;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Defines a base class for OData controllers that support writing and reading data using the OData formats.
+    /// </summary>
+    //[ODataFormatting]
```
AlanWong-MS: Add todo to address later?

src/CommonAssemblyInfo.cs
```diff
@@ -6,10 +6,8 @@
 using System.Resources;
 using System.Runtime.InteropServices;
 
-#if !NETCORE
```
xuzhg: Please help me understand "Will we manually change this for Classic and Core?"

src/Microsoft.AspNetCore.OData/Adapters/WebApiUrlHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        /// <summary>
+        /// Generates an OData link using the request's OData route name and path handler and given segments.
+        /// </summary>
+        /// <param name="segments">The OData path segments.</param>
+        /// <returns>The generated OData link.</returns>
+        public string CreateODataLink(IList<ODataPathSegment> segments)
+        {
+            return null;
```
xuzhg: return null or throw NotImplemented

src/Microsoft.AspNetCore.OData/Adapters/WebApiUrlHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNet.OData.Adapters
+{
+    /// <summary>
+    /// Adapter class to convert Asp.Net WebApi Url helper to OData WebApi.
```
xuzhg: would you please resentence this line?

src/Microsoft.AspNetCore.OData/Formatter/EdmLibHelpers.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using Microsoft.AspNet.OData.Interfaces;
+
+namespace Microsoft.AspNet.OData.Formatter
+{
+    internal static partial class EdmLibHelpers
```
xuzhg: why do we need this partial? Cannot add into the original class?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataErrorSerializer.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        /// <summary>
+        /// Create an ODataError from an HttpError.
+        /// </summary>
+        /// <param name="error">The error to use.</param>
+        /// <returns>an ODataError.</returns>
+        internal static ODataError CreateODataError(object error)
+        {
+            return null;
```
xuzhg: See a lot of function need to implement? Can we throw NotImplemented

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.csproj
xuzhg: +1

xuzhg: And why?

robward-ms: I'm targeting ASP Net Core 2.x and NetStandard 2.x at the moment. We could support both ASPNet Core 1.x and 2.x but frankly the unit tests try to Mock System.Type in so many places that verification of the code would require a massive rewrite of the UTs due to reflection APIs in .Net Core 1.x vs 2.x. There is a discussion on #939 about what to target, most customers are interested in ASPNet Core 2.x.

src/Microsoft.AspNetCore.OData/ODataController.cs
robward-ms: Yes, as that class is added. Would you rather see if removed in this commit?

src/Microsoft.AspNetCore.OData/Formatter/EdmLibHelpers.cs
robward-ms: Partial classes are used in cases where a small bit of functionality is platform-specific and it's not easy to shim. This reason in this case is that how the assembly resolver is created and used.

src/CommonAssemblyInfo.cs
robward-ms: .Net Core projects want to generate assembly metadata automatically and this code lingers from a time when I was using that feature. But now it's disabled so we supply all the metadata manually.

You're right, this should account for the product difference.

src/Microsoft.AspNetCore.OData/ODataController.cs
AlanWong-MS: I'll leave it up to you as long as it's eventually removed. Just wanted to make sure you were aware of it.

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.csproj
AlanWong-MS: Gotcha, I did read about all the feedback for 2.x; I wasn't up-to-date regarding the most recent decision.

PR #1143, opened 2017-12-02
[feature/netcore] HttpRequest(Message) in OData container

atykhyy:
### Description

This PR splits out `ODataServiceCollectionExtensions.AddODataCore()` that does not add OData formatters and action selectors, and adds code to bind `HttpRequest(Message)` into the per-request instance of OData per-route container.

### DIFF
--- a/src/Microsoft.AspNet.OData/HttpRequestScope.cs
+++ b/src/Microsoft.AspNet.OData/HttpRequestScope.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Provides access to the <see cref="HttpRequestMessage"/>
+    /// to which the OData service container instance is scoped.
+    /// </summary>
+    public class HttpRequestScope
+    {
+        /// <summary>
+        /// Provides access to the <see cref="HttpRequestMessage"/>
+        /// to which the OData service container instance is scoped.
+        /// </summary>
+        public HttpRequestMessage HttpRequest { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
@@ -400,7 +400,15 @@ namespace Microsoft.AspNet.OData.Extensions
                 throw Error.ArgumentNull("routeName");
             }
 
-            return rootContainer.GetRequiredService<IServiceScopeFactory>().CreateScope();
+            IServiceScope scope = rootContainer.GetRequiredService<IServiceScopeFactory>().CreateScope();
+
+            // Bind scoping request into the OData container.
+            if (!string.IsNullOrEmpty(routeName))
+            {
+                scope.ServiceProvider.GetRequiredService<HttpRequestScope>().HttpRequest = request;
+            }
+
+            return scope;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.OData/HttpRequestScope.cs
+++ b/src/Microsoft.AspNetCore.OData/HttpRequestScope.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Provides access to the <see cref="HttpRequest"/>
+    /// to which the OData service container instance is scoped.
+    /// </summary>
+    public class HttpRequestScope
+    {
+        /// <summary>
+        /// Provides access to the <see cref="HttpRequest"/>
+        /// to which the OData service container instance is scoped.
+        /// </summary>
+        public HttpRequest HttpRequest { get; set; }
+    }
+}
--- a/src/Microsoft.AspNet.OData.Shared/Extensions/ContainerBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Extensions/ContainerBuilderExtensions.cs
@@ -10,6 +10,7 @@ using Microsoft.AspNet.OData.Query;
 using Microsoft.AspNet.OData.Query.Expressions;
 using Microsoft.AspNet.OData.Query.Validators;
 using Microsoft.AspNet.OData.Routing;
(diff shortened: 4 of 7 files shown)


src/Microsoft.AspNet.OData/HttpRequestScope.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+
+namespace Microsoft.AspNet.OData
+{
+    internal class HttpRequestScope
```
robward-ms: Seems like this should be public?

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -404,13 +404,21 @@ private static IServiceScope CreateRequestScope(this HttpRequest request, string
                 }
 
                 rootContainer = perRouteContainer.GetODataRootContainer(routeName);
-                if (perRouteContainer == null)
+                if (rootContainer == null)
```
robward-ms: Yep, good catch, thanks!

src/Microsoft.AspNet.OData/HttpRequestScope.cs
robward-ms: Should this be public?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -22,11 +22,14 @@ namespace Microsoft.AspNet.OData.Extensions
...
-        /// Adds essential OData services to the specified <see cref="IServiceCollection" />.
+        /// Adds core OData services to the specified <see cref="IServiceCollection" />.
         /// </summary>
+        /// <remarks>
+        /// Does not add OData formatters or the OData action selector.
+        /// </remarks>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
         /// <returns>An <see cref="IODataBuilder"/> that can be used to further configure the OData services.</returns>
-        public static IODataBuilder AddOData(this IServiceCollection services)
+        public static IODataBuilder AddODataCore(this IServiceCollection services)
```
robward-ms: I understand the desire for this change but I'm hoping to get containers in the pre-request container and out of the global container; as such, I'm not sure we'll need two function long term.

src/Microsoft.AspNetCore.OData/HttpRequestScope.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNet.OData
+{
+    internal class HttpRequestScope
```
robward-ms: Again, seems like it should be public.

src/Microsoft.AspNet.OData/HttpRequestScope.cs
atykhyy: I made it internal on purpose because it's just a piece of plumbing helping me to associate the request with the per-request `IServiceScope`. I'm binding HttpRequest(Message) into the container. Is there a use case for changing the HttpRequest(Message) the per-request container is associated with in the middle of request processing?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
atykhyy: You mean get rid of `IPerRequestContainer` entirely? Sure, this won't be necessary then, but wouldn't it be a bit heavy on resources to create the whole OData container for each request?

src/Microsoft.AspNet.OData/HttpRequestScope.cs
robward-ms: No, I think the request will no change in the request container, it should be immutable. I've heard feedback that it would be useful for extensibility and that would require public access to developers can use it outside of the OData dll.

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
robward-ms: Nope, I was thinking that we'd keep the per-request container but that code in what is now AddOData() would wind up inside the per-request container, which would wind up eliminating the need for the now AddOData() and leaving only AddODataCore(), at which point AddOData() would probably be a better name.

For now, I'm not pushing back on the change, just wanted to throw that out and get your thoughts on this. If we add AddODataCore() now, it will be hard to get rid of later.

src/Microsoft.AspNet.OData/HttpRequestScope.cs
atykhyy: OK.

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
atykhyy: I see. Should I delete this from the PR in the meantime?

robward-ms: How about we leave it and mark it as deprecated? If it winds up staying, we can pull the attribute out but it's a warning that it might go away.

[Obsolete("Use to avoid adding formatters and action selector; may not be needed in the future.")]

robward-ms:
Build error: HttpRequestScope.cs(9,18): error CS1591: Missing XML comment for publicly visible type or member 'HttpRequestScope'

atykhyy:
Added xmldoc comments and attribute.

src/Microsoft.AspNet.OData/HttpRequestScope.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Net.Http;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Provides access to the <see cref="HttpRequestMessage"/>
+    /// to which the OData service container instance is scoped.
+    /// </summary>
+    public class HttpRequestScope
```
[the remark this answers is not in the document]
robward-ms: @atykhyy - Super apologies, it seems this comment was waiting for me to submit a review. I didn't notice until just now, maybe 15 or 20 days after I typed the comment?

atykhyy:
Fixed.

src/Microsoft.AspNet.OData/HttpRequestScope.cs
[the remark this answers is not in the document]
atykhyy: Sure, fixed.

PR #1151, opened 2017-12-05
Switch from IConfigureOptions<> to POC for ODataQuerySettings and ODataOptions.

### LINKED ISSUE #1147: [feature/netcore] DefaultQuerySettings is required in DI container as a constructor option, currently using IOptions<DefaultQuerySettings>
DefaultQuerySettings is required in DI container as a constructor option, currently using IOptions<DefaultQuerySettings>

### Assemblies affected

AspNetCore 7.0

robward-ms:
### Issues

This pull request fixes #1147 .

### Description

Switch from IConfigureOptions<> to POC for ODataQuerySettings and ODataOptions.
Inject settings and options into both the global and pre-request containers.
Add Fluent APIs to IRouteBuilder to configure settings and options.
Add EnableContinueOnErrorHeader for batching.

- [ ] *Test cases added*

### Additional work necessary

Refactor Batching and Formatting for AspNetCore.


src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
```diff
@@ -25,6 +25,382 @@ namespace Microsoft.AspNet.OData.Extensions
...
+            }
+
+            if (defaultQuerySettings == null)
+            {
+                throw Error.ArgumentNull("defaultQuerySettings");
+            }
+
+            if (!defaultQuerySettings.MaxTop.HasValue || defaultQuerySettings.MaxTop > 0)
+            {
+                ModelBoundQuerySettings.DefaultModelBoundQuerySettings.MaxTop = defaultQuerySettings.MaxTop;
```
AlanWong-MS: Double-checking, we're setting the `ModelBoundQuerySettings.DefaultModelBoundQuerySettings.MaxTop` if `defaultQuerySettings` doesn't have a max top value or when it's greater than 0?

When `defaultQuerySettings.MaxTop` doesn't have a value, does it have a default and we're setting `ModelBoundQuerySettings.DefaultModelBoundQuerySettings.MaxTop` to it? If that's the case, is the default value 0?

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
```diff
@@ -25,6 +25,382 @@ namespace Microsoft.AspNet.OData.Extensions
     /// </summary>
     public static class ODataRouteBuilderExtensions
     {
+        /// <summary>
+        /// Sets the <see cref="DefaultQuerySettings"/> in route builder.
+        /// </summary>
+        /// <param name="builder">The <see cref="IRouteBuilder"/>.</param>
+        /// <param name="defaultQuerySettings">The default query settings.</param>
+        public static IRouteBuilder SetDefaultQuerySettings(this IRouteBuilder builder, DefaultQuerySettings defaultQuerySettings)
```
xuzhg: I am thinking Why do we need to set "Default" query setting? 

So, the public api maybe "SetQuerySettings" ?

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
AlanWong-MS: I'm wondering if you mean `&&` instead of `||`.

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
```diff
@@ -25,6 +25,382 @@ namespace Microsoft.AspNet.OData.Extensions
...
+            {
+                throw Error.ArgumentNull("defaultQuerySettings");
+            }
+
+            if (!defaultQuerySettings.MaxTop.HasValue || defaultQuerySettings.MaxTop > 0)
+            {
+                ModelBoundQuerySettings.DefaultModelBoundQuerySettings.MaxTop = defaultQuerySettings.MaxTop;
+            }
+
+            DefaultQuerySettings querySettings = builder.ServiceProvider.GetRequiredService<DefaultQuerySettings>();
```
xuzhg: It's "get" or "set"?

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
robward-ms: It's the same as the code for AspNet in HttpConfigurationExtensions.cs, basically if you have a non-0 value (null or > 0), use that instead. Yes, 0 is default.

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
robward-ms: For back-compat, I'd like to keep it the same as AspNet, which uses "Default".

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
robward-ms: Get existing object service then copy the values from the new object to the existing object.

PR #1159, opened 2017-12-12
[feature/netcore] Fix OData model builder

sperse-z:
### Description
OData model builder throws an exception in a case when a list of application parts contains duplicates.
 
### Simple code to reproduce a bug
```
public class Startup
{
    public IServiceProvider ConfigureServices(IServiceCollection services)
    {
        var assembly = typeof(ControllerInLibrary).GetTypeInfo().Assembly; 
        
        services.AddMvc()
            .AddApplicationPart(assembly)
            .AddApplicationPart(assembly);

        services.AddOData();
    }

    public void Configure(IApplicationBuilder app)
    {
        IEdmModel model = GetEdmModel(app.ApplicationServices);
        app.UseMvc(routeBuilder => routeBuilder.MapODataServiceRoute("odata", "odata", model));
    }
}
```

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
@@ -1067,7 +1067,7 @@ namespace Microsoft.AspNet.OData.Builder
         private static Dictionary<Type, List<Type>> BuildDerivedTypesMapping(IWebApiAssembliesResolver assemblyResolver)
         {
             IEnumerable<Type> allTypes = TypeHelper.GetLoadedTypes(assemblyResolver).Where(t => TypeHelper.IsVisible(t) && TypeHelper.IsClass(t) && t != typeof(object));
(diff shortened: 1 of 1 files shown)


robward-ms:
Could you describe this change a bit? The template is not filled out so it's not clear what problem you are fixing. Also, the commit message does not describe the fix.

robward-ms:
@sperse-z - Could you describe this change a bit? The template is not filled out so it's not clear what problem you are fixing. Also, the commit message does not describe the fix.

sperse-z:
Sorry for the delayed answer. I investigated that this bug happens through a code in my application.
WebApiAssembliesResolver.Assemblies property returns a list of assemblies with duplicates.
ApplicationPartManager.ApplicationParts property also returns duplicates.
I am not sure that this a bug and this a correct place to fix this. Maybe, Assembly resolver should return a correct list of assemblies.
Thanks!

robward-ms:
@sperse-z - Ah, makes sense. I agree that WebApiAssembliesResolver seems like a good place to fix it.

sperse-z:
Thanks! I canceled this pull request and created another.

PR #1160, opened 2017-12-12
[feature/netcore] Fix $count parameter

sperse-z:
### Description

*In a case when $count parameter is specified OData request processor ignores it and doesn't return count of records.*


robward-ms:
Build is failing with: 
src\Microsoft.AspNet.OData.Shared\Formatter\Serialization\ODataResourceSetSerializer.cs (213, 1) 
src\Microsoft.AspNet.OData.Shared\Formatter\Serialization\ODataResourceSetSerializer.cs(213,1): Error : SA0102: A syntax error has been discovered in file C:\BuildAgent\_work\26\s\src\Microsoft.AspNet.OData.Shared\Formatter\Serialization\ODataResourceSetSerializer.cs on line 213

The cmdline build is not yet using C# 6.0 rules so the ?? operator is not supported.

Also, this is fixed in the latest Nightly package, see: https://github.com/OData/WebApi/blob/60351499432d1426b920ff3bee488d5412521711/src/Microsoft.AspNetCore.OData/ODataFeature.cs

sperse-z:
Good to hear! I canceled this pull request. Thank you very much for your great work!

Issue #1175, opened 2017-12-22
[feature/netcore] ArgumentNullException: Value cannot be null. Parameter name: routeName


gordon-matt:
Just updated the nuget package from alpha to beta and getting an error that wasn't there before:

**ArgumentNullException: Value cannot be null. Parameter name: routeName**

**Stack Trace:**

```
at Microsoft.AspNet.OData.Extensions.HttpRequestExtensions.CreateRequestScope(HttpRequest request, String routeName)
   at Microsoft.AspNet.OData.Extensions.HttpRequestExtensions.CreateRequestContainer(HttpRequest request, String routeName)
   at Microsoft.AspNet.OData.Extensions.HttpRequestExtensions.GetRoutingConventions(HttpRequest request)
   at Microsoft.AspNet.OData.Routing.ODataActionSelector.SelectCandidates(RouteContext context)
   at Microsoft.AspNetCore.Mvc.Internal.MvcRouteHandler.RouteAsync(RouteContext context)
   at Microsoft.AspNetCore.Routing.RouteCollection.<RouteAsync>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Builder.RouterMiddleware.<Invoke>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.<Invoke>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.MigrationsEndPointMiddleware.<Invoke>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1178, opened 2017-12-22
Update release notes for netcore and #1175

robward-ms:
Add notes about #1175.
Update name of Startup.cs file as noted in #975.
Match example URLs to route configuration.

### DIFF
--- a/_posts/2017-12-21-14-01-netcore-beta1.md
+++ b/_posts/2017-12-21-14-01-netcore-beta1.md
@@ -21,6 +21,10 @@ The code for the packages can be found here:
 **Web API OData for ASP.NET Core**, has following limitations which are known issues:
 * Batching is not supported
 * Using EnableQuery in an HTTP route, i.e. non-OData route, is not fully functional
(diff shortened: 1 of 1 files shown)


_posts/2017-12-21-14-01-netcore-beta1.md
```diff
@@ -111,15 +115,21 @@ public void Configure(IApplicationBuilder app)
...
 {% endhighlight %}
 
 #### f. Start the OData service
 
 Start the OData service by running the project and open a browser to consume it. You should be able to get access to the service
-document at `http://host/service/` in which `http://host/service/` is the root path of your service. The metadata document
-can be accessed at `GET http://host:port/$metadata` and the products at `GET http://host:port/Products` where
+document at `http://host/odata/` in which `http://host/odata/` is the root path of your service. The metadata document
+can be accessed at `GET http://host:port/odata/$metadata` and the products at `GET http://host:port/odata/Products` where
 `host:port` is the host and port of your service, usually something like `localhost:xxxx`.
```
xuzhg: `something like ` sounds list an example, so Maybe it's better to give an example port.

usually something like `localhost:12345`

Issue #1194, opened 2018-01-05
[feature/netcore] FilterQueryOption.ApplyTo requires ODataQueryContext.RequestContainer


afaganga:
FilterQueryOption.ApplyTo throws on UpdateQuerySettings because ODataQueryContext.RequestContainer is null

### Assemblies affected

Microsoft.AspNet.OData 7.0.0

### Reproduce steps

```
            var model = GetEdmModel();
            var uriParser = new ODataUriParser(
                model, 
                new Uri("http://gobbldygook/"), 
                new Uri("http://gobbldygook/Vegetables?$filter=Id eq 1"));

            var path = uriParser.ParsePath();

            var context = new ODataQueryContext(model, typeof(Vegetable), null);
            var parser = new ODataQueryOptionParser(
                model, 
                path, 
                new Dictionary<string, string>(){{"$filter","Id eq 1"}});

            var filterQueryOption = new FilterQueryOption("Id eq 1", context, parser);

            var queryable = Enumerable.Empty<Vegetable>().AsQueryable();
            filterQueryOption.ApplyTo(queryable, new ODataQuerySettings()); // throws
```

### Expected result

queryable had an Expression

### Actual result

throws ArgumentNullException

### Additional detail
Exception detail

ArgumentNullException: Value cannot be null. Parameter name: provider
Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService<T>(IServiceProvider provider)
Microsoft.AspNet.OData.ODataQueryContextExtensions.UpdateQuerySettings(ODataQueryContext context, ODataQuerySettings querySettings, IQueryable query)
Microsoft.AspNet.OData.Query.FilterQueryOption.ApplyTo(IQueryable query, ODataQuerySettings querySettings)

robward-ms:
@afaganga - Yep, ODataQueryContext is public with no public way to set the container; only ODataQueryOptions sets it.

Given that this method tries to copy the settings from the passed in parameter and update the HandleNullPropagation property, maybe it makes sense to just skip the update when no request container is available. Thoughts?

afaganga:
That should do it. Also, it could make sense to have a way to set the container.
The underlying use case is to extract an Expression<Func<TSource, Boolean>>, enabling an IObservable<TSource> usage; As such, I feel parsing should not even need aspnet at all - probably this should be a part of the EdmLib bundle.

robward-ms:
[...]

Issue #1219, opened 2018-01-22
[feature/netcore] FilterBinder requires an internal interface


raffaeler:
I am trying to extract a Linq Expression from an odata query filter without using asp.net.
This was possible with odata 6.

### Assemblies affected

Using Microsoft.OData.Core 7.3.1 and Microsoft.AspNetCore.OData 7.0.0-beta1

### Reproduce steps

The class `Microsoft.AspNet.OData.Query.Expressions.FilterBinder` is public but when I use it, it requires to the service provider an instance of IWebApiAssembliesResolver which is `internal`.

robward-ms:
@raffaeler - Yep, that sure is a problem. Can you provide a small code snippet so I can understand the scenario and see what we can do to fix it?

One option I can think of, which you touch on in #1218, if to provide a non-internal version of an AssemblyResolver so you can implement one yourself to decouple from AspNetCore's ApplicationPartManager. Essentially, the IWebApiAssembliesResolver would become public and you could create one youself and return whatever assemblies you want searched.

raffaeler:
@robward-ms I am building from scratch the code that would allow a, let's say, netcore console application to extract an Expression<Func<T, bool>> for a predicate expressed with the "$filter" clause syntax.
I built several years ago for OData6 but it was definitely a mess because of the heavy dependencies on asp.net.
With the current beta, it is definitely not possible anymore with OData7, but my wish is:
- to be able to extract predicates but also projections, orderby and other clauses
- to be totally decoupled from asp.net core
- for each clause obtain a Linq Expression so that I can visit, modify or merge it with other Expressions.

The basic steps that I had in my code are:
- build the simplest possible edm model automagically (for certain clauses defining an id is not even of interes)
- create the ODataQueryContext and a ODataQueryOptionParser with the required clauses
- define the ODataQuerySettings so that I can choose to whether support null propagation generation or not
- apply the filter to my IQueryable (this last step is not really needed, it would be even better to just create the expression.

Any code solving this transformation would be good, so there is no really need to give you some intricate code that could potentially be simplified drastically.

TIA

robward-ms:
@raffaeler - Got it. My current PR, #1236, decouples ODataQueryContext and it's downstream classes from the request container. It also decouples ODataConventionModelBuilder from the request container as well. I think with #1236, you'll be able to accomplish all that. If not, let's fix it. 

Can you try out the nightly build with the code from #1236 when it is available?

xuzhg:
[...]

PR #1236, opened 2018-02-10
Fix  issue where FilterQueryOption.ApplyTo requires ODataQueryContext.RequestContainer

robward-ms:
### Issues

This pull request fixes issue #1194, #1219 

### Description

Fix issue where FilterQueryOption.ApplyTo requires ODataQueryContext.RequestContainer.

This change allows FilterQueryOption to execute without setting a
RequestContainer on ODataQueryContext, which is an internal property
set by ODataQueryOptions from the requests' container.

In this stand-alone scenario, a container cannot be injected so the uses
of the container now have a fallback option for creating objects when the
container does not exist.

However, one of those objects was an internal assembly resolver. On
.NET, the default assembly resolver can be used. However, .NET Core
does not implementation a default assembly resolver so a new one has been
added to the .NET Core version in this change. It is a simple static
list of assemblies that should be used to resolve types used in the
EDM models.

The addition of a default assembly resolver fixes a known
issue where a NotImplemented() exception was thrown when no part
manager was found to fulfill the request for assemblies. The
default assembly resolver will now provide those assemblies.

To simply injecting of custom assemblies for EDM purposes, the
output of the part manager and the default assembly resolver are
combined into a distinct list of assemblies, allow you to put EDM 
classes into a separate assembly that does not reference ASP.NET Core.
By default, the part manager will add assemblies which reference ASP.NET Core.

To add a custom assembly to the default assembly resolver, use:

`DefaultAssembliesResolver.Assemblies.Add(typeof(<type from assembly>).Assembly);`

- [X] *Test cases added*

### Additional work necessary

Need to add docs for the default assembly resolver.


robward-ms:
Turns out this little gem is supported in .NET Core 2.x (but not 1.x):

AppDomain.CurrentDomain.GetAssemblies();

That would allow the Edm model builder to have a really great fallback to getting assemblies outside of the ASP.NET Core context like the ASP.NET Classic version of WebApi did. I'm going to take advantage of that in a revision to this change, eliminating the DefaultAssemblyResolver.

src/Microsoft.AspNetCore.OData/Adapters/WebApiAssembliesResolver.cs
```diff
@@ -44,14 +45,25 @@ public IEnumerable<Assembly> Assemblies
...
             get
             {
+                new List<Assembly>();
                 if (this.innerManager != null)
                 {
                     IList<ApplicationPart> parts = this.innerManager.ApplicationParts;
-                    return parts.Where(p => p is AssemblyPart).Select(p => (p as AssemblyPart).Assembly).Distinct();
+                    IEnumerable<Assembly> assemblies = parts
+                        .Where(p => p is AssemblyPart)
+                        .Select(p => (p as AssemblyPart).Assembly);
```
robward-ms: Lost .Distinct(). Add it back.

src/Microsoft.AspNetCore.OData/Adapters/WebApiAssembliesResolver.cs
```diff
@@ -47,11 +48,22 @@ public IEnumerable<Assembly> Assemblies
                 if (this.innerManager != null)
                 {
                     IList<ApplicationPart> parts = this.innerManager.ApplicationParts;
-                    return parts.Where(p => p is AssemblyPart).Select(p => (p as AssemblyPart).Assembly).Distinct();
+                    IEnumerable<Assembly> assemblies = parts
+                        .Where(p => p is AssemblyPart)
+                        .Select(p => (p as AssemblyPart).Assembly)
+                        .Distinct();
```
AlanWong-MS: nit: use `OfType<AssemblyPart>` to save the need of doing a `p is AssemblyPart` check?

i.e.
```
IEnumerable<Assembly> assemblies = parts
    .OfType<AssemblyPart>()
    .Select(p => p.Assembly)
    .Distinct();
```

src/Microsoft.AspNet.OData.Shared/ODataQueryContext.cs
```diff
@@ -136,6 +140,10 @@ public DefaultQuerySettings DefaultQuerySettings
         /// <summary>
         /// Gets the request container.
         /// </summary>
+        /// <remarks>
+        /// The services container may not be present. See the ocnstructor in this fail for
```
AlanWong-MS: typo: ocnstructor -> constructor

robward-ms: also: fail -> file.

Issue #1228, opened 2018-02-01
[feature/netcore] ODataFormatter not using injected ODataSerializerProvider.


robward-ms:
ODataFormatter not using injected ODataSerializerProvider when injecting ODataSerializerProvider  on a per-route basis when either 2 routes or 2 servers exists in the same process.

### Assemblies affected

- `7.0.0-Nightly201801081323`

### Reproduce steps

Run CutomizeSerializerProvider and any other E2E tests that causes formatting o occur in the same process (where the server is not isolated in an AppDomain).

### Expected result

ODataFormatter uses the injected ODataSerializerProvider.

### Actual result

ODataFormatter uses the default ODataSerializerProvider.

### Additional detail

WebApi has a class called ODataSerializerProviderProxy which is used to delay-load the ODataSerializerProvider from the DI container. The proxy is used by the formatter, which is created outside of the DI container and therefore can’t use the DI container directly. At run-time, the DI container from the active request is saved on a property on the proxy before the formatter asks for the ODataSerializerProvider; once it does ask, the proxy loads the ODataSerializerProvider from DI. In most cases, all works great and as expected.

Currently, the formatters use a static instance of ODataSerializerProviderProxy:
private static readonly ODataSerializerProviderProxy _instance = new ODataSerializerProviderProxy();

Formatters are a bit odd in AspNetClassic as they are not pre-route; they are injected into the configuration via a controller configuration attribute.  However, we “support” DI of the ODataSerializerProvider on a per-route basis, i.e. we let you. So we don’t actually support ODataSerializerProvider on a per-route basis because there are only one set of formatters on the config. As to which ODataSerializerProvider is used, it’s a race condition as to which route gets a request and sets the DI container for the proxy; it could be either one.

The DI tests for ODataSerializerProvider  exposes a separate issue where if two servers, i.e. 2 HttpConfig’s, exists in the same process (and each one has a set of formatters), the pre-route DI containers from one of the requests is used for both sets of formatters; the formatters in separate HTTPConfiguration’s share the static proxy instance and its associated per-route DI container (set via the race condition). In this case, even a configuration with a single route will not use the ODataSerializerProvider injected via DI. The more I think about this, it’s probably the less interesting case; customers would be more likely to run each server instance in it’s own process (like the tests used to do when the used AppDomains for each server).

PR #1242, opened 2018-02-15
Support per-route ODataSerializerProvider/ODataDeserializerProvider.

robward-ms:
### Issues

This pull request fixes issue #1228

### Description

ODataSerializerProviderProxy is used to delay-load the ODataSerializerProvider from the
service container container. The proxy is used by the formatter, which is created outside
of the service container and therefore can't use the container directly. At run-time, the
services container from the active request is saved on a property on the proxy before the
formatter asks for the ODataSerializerProvider; once it does ask, the proxy loads the
ODataSerializerProvider from container. In most cases, all works great and as expected.

Formatters are a bit odd in that they are not per-route; they are injected into the configuration
via a controller configuration attribute or AddOdata() service container extension.  However,
the library "supports" injection of the ODataSerializerProvider on a per-route basis, i.e. you
can do it. ODataSerializerProvider on a per-route basis because there are only one set of formatters
on the config/global service container. As to which ODataSerializerProvider is used, it's a race
condition as to which route gets a request and sets the services container for the proxy.

The dependency injection tests for ODataSerializerProvider exposes a separate issue where if two
servers, i.e. 2 HttpConfig's, exists in the same process (and each one has a set of formatters),
the per-route containers from one of the requests is used for both sets of formatters; the formatters
in different HTTPConfiguration's share the static proxy instance and its associated per-route
container (set via the race condition). In this case, even a configuration with a single route may
not use the injected ODataSerializerProvider.

The first commit attempts to fix one of those problems by making the ODataSerializerProviderProxy an
instance instead of static. As a result, when multiple formatters exist in a process, the will each
its own ODataSerializerProviderProxy. This allows the formatter to get a ODataSerializerProvider from
a route associated with its instance. However, the formatters still share one instances for multiple routes. In order to fix this, we'll need to change the API of the formatter to not expose the 

The second commit eliminates ODataSerializerProviderProxy in favor of retrieving the
ODataSerializerProvider in runtime from the service container. This allows suporting
ODataSerializerProvider's on a per-route basis. A side effect of this is that if causes a
change the API of the formatter to not expose the ODataSerializerProvider as a public property.

Note: everything in the mail equally applies to AspNetCore as well as ODataDeserializerProviderProxy, etc...

- [ ] *Test cases added*


test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/ODataFormatterTests.cs
```diff
@@ -321,11 +320,17 @@ public class ODataFormatterTests
...
-                    0,
-                    ODataMediaTypeFormatters.Create(new CustomSerializerProvider(), _deserializerProvider));
+                configuration.Formatters.InsertRange(0, ODataMediaTypeFormatters.Create());
+                configuration.MapODataServiceRoute("IgnoredRouteName", null, builder =>
+                    builder.AddService(Microsoft.OData.ServiceLifetime.Singleton, sp => ODataTestUtil.GetEdmModel())
+                        .AddService<ODataSerializerProvider>(ServiceLifetime.Singleton, sp => new CustomSerializerProvider())
+                        .AddService<IEnumerable<IODataRoutingConvention>>(ServiceLifetime.Singleton, sp =>
+                            ODataRoutingConventions.CreateDefaultWithAttributeRouting("IgnoredRouteName", configuration)));
+
+
```
AlanWong-MS: nit: extra space

PR #1246, opened 2018-02-16
Merge with Official WebAPI 6.1 

kosinsky:
* Merge with Official WebAPI 6.1
* Merge all backports of VSTS forks fixes (like Stackoverflow in EF)
* Reaction work to support ODL 7.4 (for example enum to string comparison that allowed now)
* Reaction work (removed Skip on test) for proper Null elimination


kosinsky:
Sorry wrong repo

Issue #1251, opened 2018-02-27
[feature/netcore] Deriving from NavigationSourceRoutingConvention fails due to internal class


robward-ms:
NavigationSourceRoutingConvention provides and override, SelectAction, that returns an internal
class: SelectControllerResult. As a result, deriving from and override SelectionAction requires access to internals.

### Assemblies affected

- `7.0.0-Beta`

### Reproduce steps

Derive a convention from NavigationSourceRoutingConvention and attempt to customize the SelectAction method.

### Expected result

Code compiles.

### Actual result

SelectAction requires access to an internal class: SelectControllerResult

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

PR #1253, opened 2018-02-27
Make SelectControllerResult public

robward-ms:
### Issues

This pull request fixes issue #1251 

### Description

This change makes SelectControllerResult public, allowing derived classes of
NavigationSourceRoutingConvention to override SelectAction without access
to internal classes.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/SelectControllerResult.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/SelectControllerResult.cs
@@ -8,12 +8,7 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
     /// <summary>
     /// An return value for SelectController.
     /// </summary>
-    /// <remarks>
-    /// This class is not intended to be exposed publicly; it used for the internal
-    /// implementations of SelectControl(). Any design which makes this class public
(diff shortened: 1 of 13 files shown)


AlanWong-MS:
How's the existing test coverage of the class?

robward-ms:
In general coverage is quite good...in private. The tests for NetCore as waiting to be checked in. 

This change comes from my E2ECore branch, where the E2E tests for NetCore are waiting for some cleanup, PR and checkin.  There are a few derived conventions in the E2E tests and I've trying to keep E2E from access internals.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/SelectControllerResult.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
     /// implementations of SelectControl(). Any design which makes this class public
```
AlanWong-MS: We should probably modify this remark tag comment as we're making the change, or remove it and add any additional comments in the PR description.

robward-ms: I think remove is the correct action.

Issue #1252, opened 2018-02-27
[feature/netcore] ODataConventionModelBuilder requires a service container


robward-ms:
ODataConventionModelBuilder requires a service container, making it difficult to use as a stand-alone
model builder.

### Assemblies affected

- `7.0.0-Beta`

### Reproduce steps

Create an instance of ODataConventionModelBuilder.

### Expected result

Should be able to create ODataConventionModelBuilder without a service container.

### Actual result

ODataConventionModelBuilder  requires a service container.

PR #1254, opened 2018-02-27
Enable ODataConventionModelBuilder for stand-alone scenarios

robward-ms:
### Issues

This pull request fixes issue #1252 

### Description

This change adds a constructor to ODataConventionModelBuilder to allow
an instance to be constructed using only an ApplicationPartManager.

While the ApplicationPartManager is an ASP.NET Core class, it is simply
to create and populate with one or more selected assemblies, allowing
for easy use of ODataConventionModelBuilder in scoped & stand-alone scenarios:

```
  var partManager = new ApplicationPartManager();
  var part = new AssemblyPart(typeof(<class>).Assembly);
  partManager.ApplicationParts.Add(part);

  var builder = new ODataConventionModelBuilder(partManager);

```

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Builder/ODataConventionModelBuilder.cs
+++ b/src/Microsoft.AspNetCore.OData/Builder/ODataConventionModelBuilder.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
 using System;
-using System.Reflection;
 using Microsoft.AspNet.OData.Adapters;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Interfaces;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.OData/Builder/ODataConventionModelBuilder.cs
```diff
@@ -55,14 +65,42 @@ public ODataConventionModelBuilder(IServiceProvider provider)
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataConventionModelBuilder"/> class.
+        /// </summary>
+        /// <param name="provider">The service provider to use.</param>
+        /// <param name="applicationPartManager">
+        /// The application part manager to use. If null, the service
+        /// provider will be queried for the application part manager.
+        /// </param>
+        /// <param name="isQueryCompositionMode">If the model is being built for only querying.</param>
+        /// <remarks>The model built if <paramref name="isQueryCompositionMode"/> is <c>true</c> has more relaxed
```
AlanWong-MS: nit regarding consistency: the remarks tag is inline vs. the second remarks tag which is not.

robertmclaws:
So what about non-ASP.NET Core scenarios? I have plenty of situations (like for Restier, which does not run on .NET Core yet) where I want to use the ModelBuilder as a jumping-off point, so that I can modify it's output without dealing with writing the model 100% by hand. In those situations, I feel like I should be able to pass in an array of Assemblies that the ModelBuilder needs and let it do it's work.

robward-ms:
@robertmclaws - at one point in an earlier change, I considered providing a default assembly resolver that you could load up with any assembly you like and then model builder would use it by default. I wound up removing that because .NetCore can support AppDomain's GetAssemblies() but the ODataConventionModelBuilder  constructor to use it is private as I think it's better to use the ApplicationPartManager for .NetCore scenarios.

I could go back to the assembly resolver notion where you populate and/or support and interface to retrieve a collection of assemblies. Or would you rather see a ODataConventionModelBuilder constructor where you pass in assemblies? Or would you rather see the default constructor just fetch all assemblies in the AppDomain?

robertmclaws:
So I would say, if NetCore supports AppDomain.GetAssemblies(), then I think the default constructor should use that.

robward-ms:
The default constructor will work the same here as it did in 6x/AspNet Classic.

src/Microsoft.AspNetCore.OData/Builder/ODataConventionModelBuilder.cs
```diff
@@ -39,7 +38,17 @@ private ODataConventionModelBuilder()
         /// the functionality is due to the way assembly resolution is done in AspNet vs AspnetCore.
```
robertmclaws: I think these remarks are kinda vague and don't really help me understand what you're referring to here. Compatibility info here would also be useful.

src/Microsoft.AspNetCore.OData/Builder/ODataConventionModelBuilder.cs
```diff
@@ -39,7 +38,17 @@ private ODataConventionModelBuilder()
...
-            : this(provider, isQueryCompositionMode: false)
+            : this(provider, null, isQueryCompositionMode: false)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataConventionModelBuilder"/> class.
+        /// </summary>
+        /// <param name="applicationPartManager">The application part manager to use.</param>
+        /// <remarks> This function uses types that are AspNetCore-specific.</remarks>
```
robertmclaws: I would adjust these remarks to have a similar compatibility format as mentioned above. Also, a `<code>`  sample here of how to use this properly on ASP.NET Core would be amazing.

src/Microsoft.AspNetCore.OData/Builder/ODataConventionModelBuilder.cs
```diff
@@ -55,14 +64,44 @@ public ODataConventionModelBuilder(IServiceProvider provider)
...
+        /// <param name="provider">The service provider to use.</param>
+        /// <param name="applicationPartManager">
+        /// The application part manager to use. If null, the service
+        /// provider will be queried for the application part manager.
+        /// </param>
+        /// <param name="isQueryCompositionMode">If the model is being built for only querying.</param>
+        /// <remarks>
+        /// The model built if <paramref name="isQueryCompositionMode"/> is <c>true</c> has more relaxed
+        /// inference rules and also treats all types as entity types. This constructor is intended for use
+        /// by unit testing only.
```
robertmclaws: The fact that this constructor is only for unit testing should potentially be in the summary after the current text.

robward-ms:
@robertmclaws - Comments changed.

PR #1258, opened 2018-03-05
Batching support for AspNetCore.

robward-ms:
### Issues

### Description

This change adds batching support.

Batching is configured the same as it is in the ASP.NET version with
the exception of the batching middleware which is configured using the
following extensions from the Microsoft.AspNet.OData.Extensions namespace
in the StartUp.Configure() method using the IApplicationBuilder object:

`  applicationBuilder.UseODataBatching()`

Each OData route may have its own batching handler which can be passed into
a call to MapODataServiceRoute(), in the StartUp.Configure() method using
the IApplicationBuilder object:

```
  app.UseMvc(routes =>
  {
      routes.MapODataServiceRoute("OData", "odata", edmModel, new DefaultODataBatchHandler());
  });
```

Since the APIs for batching use ASP.NET constructs directly, there is very little
shared code in the batching feature. The batching APIs use the same design patterns
but different in the ASP.NET objects the reference, e.g. HttpRequestMessage vs HttpRequest.

- [ ] *Test cases added*

### Additional work necessary

Port Unit Tests to NetCore
Port E2E Tests to NetCore

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchHttpRequestMessageExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchHttpRequestMessageExtensions.cs
@@ -1,0 +1,275 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Formatter;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Http.Headers;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Mvc.Routing;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Net.Http.Headers;
+using Microsoft.OData;
+
+namespace Microsoft.AspNet.OData.Batch
+{
+    /// <summary>
+    /// Provides extension methods for the <see cref="HttpRequest"/> class.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class ODataBatchHttpRequestExtensions
+    {
+        private const string BatchIdKey = "BatchId";
+        private const string ChangeSetIdKey = "ChangesetId";
+        private const string ContentIdKey = "ContentId";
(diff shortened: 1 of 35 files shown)


src/Microsoft.AspNetCore.OData/Batch/ODataBatchHttpRequestMessageExtensions.cs
```diff
@@ -0,0 +1,277 @@
...
+        internal static Uri GetODataBatchBaseUri(this HttpRequest request, string oDataRouteName, IRouter route)
+        {
+            Contract.Assert(request != null);
+
+            if (oDataRouteName == null)
+            {
+                // Return request's base address.
+                return new Uri(request.GetDisplayUrl());
+            }
+            else
```
AlanWong-MS: nit: move the code block out of else to reduce bracelet nest level?

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -0,0 +1,233 @@
...
+            // Note that just about everything inm the HttpContext and HttpRequest is
+            // backed by one of these features. So reusing the features means the HttContext
+            // and HttpRequests are the same without needing to copy properties. To make them
+            // different, we need to avoid copying certain features to that the objects don't
+            // share the same storage/
+            IFeatureCollection features = new FeatureCollection();
+            foreach (KeyValuePair<Type, object> kvp in originalContext.Features)
+            {
+                // Don't include the OData features. They may already
+                // be present. This will get-recreated later.
```
AlanWong-MS: typo: hyphen in "get-recreated"

src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Routes the request.
+        /// </summary>
+        /// <param name="handler">The handler for processing a message.</param>
+        /// <returns>A <see cref="ODataBatchResponseItem"/>.</returns>
+        public abstract Task<ODataBatchResponseItem> SendRequestAsync(RequestDelegate handler);
+
```
AlanWong-MS: nit: empty line

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
```diff
@@ -129,7 +129,6 @@ public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterCo
 
                 Action<Exception> logErrorAction = (ex) =>
                 {
-                    throw ex;
```
AlanWong-MS: What's the effect of this?

robward-ms: This is a logging action only.  It's being used like in ODataInputFormatterHelper.ReadFromStream:

            catch (Exception e)
            {
                logErrorAction(e);
                result = defaultValue;
            }

By throwing, ODataInputFormatterHelper.ReadFromStream was not returning the default value. We should do a better job of logging here.

The AspNet version of this is below where formatterLogger is passed into the function from AspNet and is not null:

                Action<Exception> logErrorAction = (ex) =>
                {
                    if (formatterLogger == null)
                    {
                        throw ex;
                    }

                    formatterLogger.LogError(String.Empty, ex);
                };

Issue #1273, opened 2018-03-13
Service returns 406 when using a result derived from SingleResult


robward-ms:
When using a result which uses a class derived from SingleResult, the service returns 406. This scenario is used in both E2E and UnitTests.

### Assemblies affected

7.0.0-Beta2

### Reproduce steps

Create a controller which has a Get method returning a class derived from SingleResult.

### Expected result

Service returns a result.

### Actual result

Service returns 406.

PR #1276, opened 2018-03-13
Allow results to be derived from SingleResult and SingleResult<>

robward-ms:
### Issues

This pull request fixes issue #1273 

### Description

When applying formatting to a SingleResult, the formatters looked strictly
for a class of type SingleResult or SingleResult<>.

This change allows a result to be derived from SingleResult or SingleResult<>.
While not likely useful in production, this scenario is used in E2E and UnitTests.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
+++ b/src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
@@ -189,10 +189,19 @@ namespace Microsoft.AspNet.OData.Formatter
                 ODataSerializerProvider serializerProvider = Request.GetRequestContainer()
                     .GetRequiredService<ODataSerializerProvider>();
 
+                // See if this type is a SingleResult or is derived from SingleResult.
+                bool isSingleResult = false;
+                if (type.IsGenericType)
+                {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
```diff
@@ -189,10 +189,19 @@ public override bool CanWriteType(Type type)
                 ODataSerializerProvider serializerProvider = Request.GetRequestContainer()
                     .GetRequiredService<ODataSerializerProvider>();
 
+                // See if this class if a SingleType derived class.
```
xuzhg: resentence the comment

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
```diff
@@ -116,10 +116,19 @@ public override bool CanWriteResult(OutputFormatterCanWriteContext context)
             ODataSerializerProvider serializerProvider = request.ODataFeature()
                 .RequestContainer.GetRequiredService<ODataSerializerProvider>();
 
+            // See if this class if a SingleType derived class.
```
AlanWong-MS: Same with this one

PR #1281, opened 2018-03-14
Enables E2E tests for AspNetCore.

robward-ms:
### Issues

### Description

This change enables a large set of E2Es for AspNetCore. The current pass rate is:

AspNet    : 4648 pass, 0 fail - 100% enabled, 100% pass
AspNetCore: 4029 pass, 191 fail - 91% enabled, 87% pass

The E2E tests for AspNetCore compile the same test code that is used in
AspNet. To facilitate this, abstraction are used where possible to abstract
away the platform details from the test code. For instance, to create a response,
the E2Es use a test controller and its response methods:

    public class NormalOrdersController : TestODataController
    {
        ...

        [EnableQuery(PageSize = 10, MaxExpansionDepth = 5)]
        public ITestActionResult Get()
        {
            var db = new AutoExpandOrdersContext();
            return Ok(db.NormalOrders.AsQueryable());
        }

        [EnableQuery]
        public TestSingleResult<NormalOrder> Get(int key)
        {
            var db = new AutoExpandOrdersContext();
            return TestSingleResult.Create(db.NormalOrders.Where(o => o.Id == key));
        }

        ...
    }

The Get method will create either an IHttpActionResult or an IActionResult depending on which
platform is being compiled. The Get(key) method will create a SingleResult compatible
with the platform being compiled. There is also a TestNonODataController; both abstractions
and the response methods are defined in Common\Controllers\TestControllers.cs.

The web server creation occurs by deriving from WebHostTestBase, accepting a WebHostTestFixture
in the constructor of the test class, and passing it to WebHostTestBase. The fixture is responsible
for creating the appropriate web server depending on the platform and calling the test's UpdateConfiguration
method using an WebRouteConfiguration instance, which abstracts the details of web server configuration.
In AspNet, WebRouteConfiguration is an HttpConfiguration. In AspNetCore, WebRouteConfiguration is a proxy
for IRouteBuilder. These abstractions can be found in the Common\Execution folder.

Abstraction make use of conditional compilation using the NETCORE and NETFX symbols.

- [x] *Test cases added*

### Additional work necessary

Enable all unit tests for core.
Enable all E2E tests for core.


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Formatter/InheritanceTests.cs
```diff
@@ -21,21 +16,22 @@
...
+    public class DeleteAllRoutingConvention : TestEntitySetRoutingConvention
     {
-        public override string SelectAction(Microsoft.AspNet.OData.Routing.ODataPath odataPath, HttpControllerContext context,
-            ILookup<string, HttpActionDescriptor> actionMap)
+        /// <inheritdoc/>
+        protected override string SelectAction(string requestMethod, ODataPath odataPath, TestControllerContext controllerContext, IList<string> actionList)
         {
-            if (context.Request.Method == HttpMethod.Delete &&
-                odataPath.PathTemplate == "~/entityset")
+            if (requestMethod == "DELETE" &&
```
AlanWong-MS: Is there a const string defined somewhere for the DELETE operation?

robward-ms: Not a good universal one sadly. AspNetCore offers an HttpMethod class with string constants but AspNet used the System.Web verison of HttpMethods which are static class instances backed by strings. Using raw strings is the most compatible version.

The product code get's around this with an abstracted version of HttpMethod, ODataRequestMethod which is an int-based enum, and the adapter classes do the appropriate conversion. While that is possible on the test side, it seems like overkill.

PR #1287, opened 2018-03-20
Enable ASP.NET core Unit test for DateAndTimeOfDayTest

xuzhg:
### Issues

* Enable the Unit test for Asp.NET Core.

### Description

*only Enable ASP.NET core t Unit test for DateAndTimeOfDayTest for Ramp up.

- [] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/UnitTest/Microsoft.Test.AspNet.OData/DateAndTimeOfDayTest.cs
+++ b/test/UnitTest/Microsoft.Test.AspNet.OData/DateAndTimeOfDayTest.cs
@@ -1,7 +1,24 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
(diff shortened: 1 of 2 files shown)


xuzhg:
Besides, @robward-ms it seems the ASP.NET Core has the a lot of failed test cases. are you working on these?

test/UnitTest/Microsoft.Test.AspNet.OData/DateAndTimeOfDayTest.cs
```diff
@@ -190,6 +208,20 @@ public class DateAndTimeOfDayTest
             Assert.Equal(Expect, await response.Content.ReadAsStringAsync());
         }
 
+#if NETCORE
```
robward-ms: You should be able to use this code for both versions.

xuzhg: Yes. I changed. Thanks!

Issue #826, opened 2016-10-03
V6 no way to use my own ODataPathRouteConstraint


rwmcdonald:
In v6 it is impossible to provide my own ODataPathRouteConstraint.

I checked the [source](https://github.com/OData/WebApi/blob/OData60/OData/src/System.Web.OData/Extensions/HttpConfigurationExtensions.cs#L536) for MapODataServiceRoute and I see that it just uses the default:

`
            ODataPathRouteConstraint routeConstraint = new ODataPathRouteConstraint(routeName);
`

I expect this should be using GetService, or provide another way of setting my own customized ODataPathRouteConstraint.

Thanks,
Rob

rwmcdonald:
BUMP

Just wondering if this is indeed a bug, or if I'm doing it wrong.

I have overridden the ODataPathRouteConstraint in much the same way as is done in the DynamicEdmModelCreation project here: https://github.com/OData/ODataSamples/blob/master/WebApi/v4/DynamicEdmModelCreation/DynamicEdmModelCreation/CustomODataPathRouteConstraint.cs

Is this no longer possible when using the DI methods provided in v6?

Thanks for your time,

dmcallis:
Yes, I have the same problem. I think this is indeed a bug.

oucfei:
Hi @rwmcdonald  just wondering were you able to solve this issue or have any work around? We also have our own ODataPathRouteConstraint in our project, and is planning to move to V6.

rwmcdonald:
I was unable to use my own ODataPathRouteConstraint. However, I was able to workaround the issue.

It's been quite a long time, but I believe my ODataPathRouteConstraint was doing two things: Setting the model and setting the action/controller. 

I fixed it by creating a wrapper class to set the model and calling it per request as described by @VikingsFan  here: https://github.com/OData/WebApi/issues/816#issuecomment-249758634

And also built MyDynamicDataRoutingConvention implementing IODataRoutingConvention that selects the controller/action dynamically.

My final code looks something like the following:

`			IList<IODataRoutingConvention> routingConventions = ODataRoutingConventions.CreateDefault();
			routingConventions.Insert( 0, new MyDynamicDataRoutingConvention() );`
		
			
	var route = config.MapODataServiceRoute( "MyDynamic", "MyDynamic", containerBuilder =>
		containerBuilder.AddService<MyDynamicModelWrapper>( ServiceLifetime.Scoped )
[...]

PR #1289, opened 2018-03-20
Fix for issue 826. Now it is possible to inject Custom oDataRoute also

rbmanian75:
### Issues

*This pull request fixes issue #826.*

### Description

*Now it is possible to inject Custom oDataRoute. So that we can add our own odatapathconstraint*

- [ ] *Test cases added*

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -1,6 +1,6 @@
 OData-WebAPI
 
-Copyright (c) Microsoft. All rights reserved.
+Copyright (c) 2018 Microsoft. All rights reserved.
 
 Material in this repository is made available under the following terms:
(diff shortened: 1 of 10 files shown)


rbmanian75:
@visatyam when you do config.MapODataServiceRoute the odataroute is for every map. So you can define odataroute for every map.

rbmanian75:
It is been 4 months after the pull request is created. Is there someone review the pull request and merge the same.

rbmanian75:
It is been 1 year and the issue is not resolved. Microsoft you kill the features what you give and does not fix for long time. we dont feel confidence in using your products and libraries any more.

PR #1298, opened 2018-03-24
Fix E2E test failure for AspNetCore.

robward-ms:
### Issues

### Description

This change reduces the E2E failure for AspNetCore. The current pass rate is:
AspNet    : 4648 pass, 0 fail - 100% enabled, 100% pass
AspNetCore: 4123 pass, 97 fail - 91% enabled, 89% pass

This change:

1.) ModifiesEnableQuery to return a Json format similar to the HttpError format in AspNet. Many tests are expecting
    the exact json format from AspNet classic. It's a fine format not specific to any platform so I chose to
    emulate is for AspNetCore.

2.) Implements async client query pattern as the NetStandard version of the client does not support sync methods.

3.) Modifies LowerCamelCase tests to match error profile of AspNetCore for model validation failures.

- [] *Test cases added*


xuzhg:
Merged at: https://github.com/OData/WebApi/commit/bb875147fd11c639d30f879e08a03bb694a14d43

Issue #1292, opened 2018-03-21
[feature/netcore] Delta not set changed property


genusP:
Action parameter with type Delta<EntityType> not init.

### Assemblies affected

Microsoft.AspNetCore.OData: 7.0.0-b2

### Reproduce steps
Controller action
```c#
[HttpPut]
[HttpPatch]
[ODataRoute("Groups({key})")]
public async Task<IActionResult> UpdateGroupAsync(int key, Delta<ViewModels.SecurityGroup> dGroup, CancellationToken cancellationToken = default)
{
      return NoContent();
}
```
Request
```
cookie: ....
content-type: application/json;odata=verbose
Host: localhost:44346
Content-Length: 19


{"Name": "qweqwew"}
```
### Expected result
dGroup.GetChangedPropertyNames() enum with Name string

### Actual result

dGroup.GetChangedPropertyNames() return empty enumerable

xuzhg:
It seems `NonValidatingParameterBinding` attribute is not used for the Model binding  in ASP.NET Core version. @robward-ms Do you have any hints?

xuzhg:
@genusP Would you please take a look: https://github.com/OData/WebApi/pull/1300

xuzhg:
Merged.

PR #1300, opened 2018-03-25
Fix the enable the Delta with changed property

xuzhg:
### Issues

* https://github.com/OData/WebApi/issues/1292
### Description

* Delta<T> with an binding attribute named `NonValidatingParameterBindingAttribute`, which haven't any effect in ASP.NET Core model binding.

This PR is a template fix about the model binding for `Delta<T>`. We'd think more about the model binding.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/NonValidatingParameterBindingAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/NonValidatingParameterBindingAttribute.cs
@@ -3,6 +3,8 @@
 
 using System;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.OData/NonValidatingParameterBindingAttribute.cs
```diff
@@ -13,12 +15,21 @@ namespace Microsoft.AspNet.OData
     /// This is essentially a <see cref="ValidateNeverAttribute"/>.
     /// </remarks>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
-    internal sealed partial class NonValidatingParameterBindingAttribute : Attribute, IPropertyValidationFilter
```
xuzhg: @robward-ms do you know why do this class implement IPropertyValidationFilter ?

Issue #1308, opened 2018-03-27
The service returns 500 when a query with an invalid select clause


robward-ms:
An query with an invalid select clause produces a 500 error.

### Assemblies affected

7.0.0-Beta2

### Reproduce steps

Create a controller with a get method that takes a ODataQueryOptions.
Send a request with an invalid select clause.

### Expected result

The controller is called with a valid ODataQueryOptions.

### Actual result

The service returns 500.

PR #1309, opened 2018-03-27
Annotate ODataQueryOptions.cs with the non-validating attribute

robward-ms:
### Issues

This pull request fixes issue #1308 

### Description

This change adds the non-validating attribute to ODataQueryOptions.cs
to avoid having AspNetCore validate the object, which may access properties
such as SelectClause which throw if the clause is invalid.

- [ ] *Test cases added*


src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.AspNet.OData.Query
     /// This defines a composite OData query options that can be used to perform query composition.
     /// Currently this only supports $filter, $orderby, $top, $skip, and $count.
     /// </summary>
+    [NonValidatingParameterBinding]
```
xuzhg: should modify ODataQueryOption<T> ?

PR #1318, opened 2018-03-27
Fix the failing test cases, including the public api

xuzhg:
### Issues

* Fix the failing test cases, including the public api

### Description

* Fix the failing test cases, including the public api

- [ ] *Test cases added*

### Additional work necessary


test/UnitTest/Microsoft.Test.AspNet.OData/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
```diff
@@ -419,6 +419,7 @@ public class Microsoft.AspNet.OData.EnableQueryAttribute : Microsoft.AspNetCore.
 
 	public virtual System.Linq.IQueryable ApplyQuery (System.Linq.IQueryable queryable, ODataQueryOptions queryOptions)
 	public virtual object ApplyQuery (object entity, ODataQueryOptions queryOptions)
+	public static Microsoft.AspNetCore.Mvc.SerializableError CreateErrorResponse (string message, params System.Exception exception)
```
robward-ms: We should consider modifying the build script to run the public test in a separate task that fails the build.

xuzhg: That's a good idea. Once we pass/enable all the test cases, we will do it.

PR #1322, opened 2018-03-28
Change the ODataActionSelector and update the ODL version dependency

xuzhg:
### Issues

* Change the ODataActionSelector and update the ODL version dependency

### Description

* Change the ODataActionSelector and update the ODL version dependency

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -85,7 +85,7 @@ namespace Microsoft.AspNet.OData.Routing
                 // Find the action with the greatest number of matched parameters including
                 // matches with no parameters.
                 var matchedCandidates = candidates
-                    .Where(c => !c.Parameters.Any() || c.Parameters.Any(p => context.RouteData.Values.ContainsKey(p.Name)))
+                    .Where(c => !c.Parameters.Any() || c.Parameters.All(p => context.RouteData.Values.ContainsKey(p.Name)))
                     .OrderByDescending(c => c.Parameters.Count);
 
                 // Return either the best matched candidate or the first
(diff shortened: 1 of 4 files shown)


test/UnitTest/Microsoft.Test.AspNet.OData/Routing/ODataRoutingTest.cs
```diff
@@ -188,7 +188,9 @@ public ODataRoutingTest()
                     { "GET", "Products/Default.TopProductOfAll", "TopProductOfAll" },
                     { "GET", "Products/Default.TopProductOfAllByCity(city='any')", "TopProductOfAllByCity(any)" },
                     { "GET", "Products/Default.TopProductOfAllByCity(city='some%23hashes')", "TopProductOfAllByCity(some#hashes)" },
+#if NETFX
                     { "GET", "Products/Default.TopProductOfAllByCity(city='some%2fslashes')", "TopProductOfAllByCity(some/slashes)" },
```
AlanWong-MS: Is there something that NetCore doesn't support here?

xuzhg: Thanks. I added a comment in the code.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -85,7 +85,7 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
                 // Find the action with the greatest number of matched parameters including
                 // matches with no parameters.
                 var matchedCandidates = candidates
-                    .Where(c => !c.Parameters.Any() || c.Parameters.Any(p => context.RouteData.Values.ContainsKey(p.Name)))
```
robward-ms: I'm not sure you want to do this. There are cases where the signature may have more parameters than route data contains.

Specifically, you just reverted a change made earlier to fix a bug, see https://github.com/OData/WebApi/commit/46dc2c07939de075ffd6ffa7091056a7a700d4cc#diff-60f38e3eae9f9652d16f0473ff2ea189

xuzhg: Yes. The previous fix only cover certain scenarios. But, it's a bug for the following scenario:

```C#
public string DeleteRef(string navigationProperty)
{
      return String.Format(CultureInfo.InvariantCulture, "DeleteRef({0})", navigationProperty);
}

public string DeleteRef(int relatedKey, string navigationProperty)
{
     return String.Format(CultureInfo.InvariantCulture, "DeleteRef({0})ByKey({1})", navigationProperty, relatedKey);
}
```
because `navigationProperty` parameter in the two methods. So, the first method (without `relatedKey`) will never be called.

I think the logic in `SelectBestCandidate` should be refactor more. Otherwise, it can't be used in product.

PR #1327, opened 2018-03-29
revert All to Any function call in odata action selector

xuzhg:
### Issues

* The test cases failing.

### Description

*  Revert the `All to `Any` call in ODataActionSelector to make the most test case pass. 
 However, the logic in the SelectBestCandidate(...) should be fixed for the following scenario:

```C#
public string DeleteRef(string navigationProperty)
{
     return String.Format(CultureInfo.InvariantCulture, "DeleteRef({0})", navigationProperty);
}

public string DeleteRef(int relatedKey, string navigationProperty)
{
    return String.Format(CultureInfo.InvariantCulture, "DeleteRef({0})ByKey({1})", navigationProperty, relatedKey);
 }

```
The first method will never be selected.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -85,7 +85,7 @@ namespace Microsoft.AspNet.OData.Routing
                 // Find the action with the greatest number of matched parameters including
                 // matches with no parameters.
                 var matchedCandidates = candidates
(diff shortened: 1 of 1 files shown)


robward-ms:
I'm not sure "Reset" is very descriptive, it requires looking at the previous version to make sense of it all. I'd rather see the commit and PR worded to make sense by itself, i.e. explain that the change is being reverted and explain why. This helps explain why you made the change so someone know why trying to make it All() instead of Any() is problematic and what problems to consider if you try to do it again in the future.

PR #1225, opened 2018-01-30
E2e optional dollar sign

biaol-odata:
### Issues

*This pull request enables optionsDollarSign from WebAPI query parsing.*

### Description

*By setting ODataUriResolver.EnableNoDollarQueryOptions as true, WebAPI can parse OData system queries without the '$' prefixes (as those in $filter, $select, $orderby, etc) correctly and return expected response.*

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #1334, opened 2018-03-30
Fix Perf solution

biaol-odata:
### Issues

*This pull request fixes issue in PR #1225 for perf solution configuration.*

### Description

*Edit for perf solution's private build was put in wrong place. Should use notepad to update the dependencies directly instead of using VS.*

- [ ] *Test cases added*

### Additional work necessary


test/E2ETest/WebStack.QA.Test.OData/ETags/JsonETagsTests.cs
```diff
@@ -69,7 +69,6 @@ public void ModelBuilderTest()
             string expectMetadata =
                 "<EntitySet Name=\"ETagsCustomers\" EntityType=\"WebStack.QA.Test.OData.ETags.ETagsCustomer\">\r\n" +
                 "          <NavigationPropertyBinding Path=\"RelatedCustomer\" Target=\"ETagsCustomers\" />\r\n" +
-                "          <NavigationPropertyBinding Path=\"ContainedCustomer\" Target=\"ContainedCustomer\" />\r\n" +
```
xuzhg: why have this one?

biaol-odata: It is from the tip of the master branch. Should disappear after rebase. Thanks!

PR #1335, opened 2018-03-31
Make failing test cases in InheritanceTests to work in ASP.NET Core

xuzhg:
### Issues

* Make the failing test cases in InheritanceTests to work in non-odata Asp.NET Core

### Description

* Make the failing test cases in InheritanceTests to work in non-odata Asp.NET Core

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/InheritanceTests.cs
+++ b/test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/InheritanceTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+#if NETCORE
 using System;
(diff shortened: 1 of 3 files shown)


xuzhg:
@robward-ms please take a look

test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/InheritanceTests.cs
```diff
@@ -8,11 +8,21 @@
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
+#if NETFX
```
robward-ms: It' easier to manage if you make one #if NETCORE block and one under #else. With them missed like this, sorting & removing no longer works.

xuzhg: Fixed. Thanks!

PR #1348, opened 2018-04-03
fix the failing ASP.NET Core test cases in ODataResourceSerializerTests.cs

xuzhg:
### Issues

* fix the failing ASP.NET Core test cases in ODataResourceSerializerTests.cs
### Description

* fix the failing ASP.NET Core test cases in ODataResourceSerializerTests.cs

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/Serialization/ODataResourceSerializerTests.cs
+++ b/test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/Serialization/ODataResourceSerializerTests.cs
@@ -50,14 +50,10 @@ namespace Microsoft.Test.AspNet.OData.Formatter.Serialization
         {
             _model = SerializationTestsHelpers.SimpleCustomerOrderModel();
 
-            _model.SetAnnotationValue<ClrTypeAnnotation>(_model.FindType("Default.Customer"), new ClrTypeAnnotation(typeof(Customer)));
-            _model.SetAnnotationValue<ClrTypeAnnotation>(_model.FindType("Default.Order"), new ClrTypeAnnotation(typeof(Order)));
-            _model.SetAnnotationValue(
-                _model.FindType("Default.SpecialCustomer"),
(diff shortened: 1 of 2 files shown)


test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -1315,7 +1311,11 @@ public void CreateODataAction_IncludesEverything_ForFullMetadata()
             ODataAction actualAction = _serializer.CreateODataAction(action, context);
 
             // Assert
+#if NETCORE
```
robward-ms: Can you add a note as to why these should be different?

xuzhg:
@robward-ms

test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/Serialization/ODataResourceSerializerTests.cs
robward-ms: I'm unsure why we'd want to use different routes in AspNetCore vs AspNet. The more the test cases differ, the more likely we are to accidentally regress the product.

robward-ms: We can't just new up a UrlHelper like before, it needs to come from the container. That is why HttpContextExtensions.cs works like that.

I understand the change to request factory.

My questions is: Why should the AspNet test use an HttpRoute and AspNetCore use an ODataRoute?

Maybe we can find a way to make the Router in request factory use AspNetCore's Route instead of ODataRoute so the test response is the same?

Or make the AspNet test use an OData route?

xuzhg:
@robward-ms Thanks for approved this. I'd like using #1363 to cover the test cases in this PR. So, please take a look for #1363. And I close this PR.

PR #1349, opened 2018-04-03
Remove "[" and "]" as general delimiters from AspNetCore tests

robward-ms:
### Issues

### Description

Remove "[" and "]" as general delimiters from AspNetCore tests for ODataPathRouteConstraintTest.

RFC3986 specifies "[" and "]" and gen-delims in BNF notation in section  2.2 but section 3.2.2
specifies that the are only valid in a Uri as part of an IPV6 host. Therefore, they are not
valid in any part of a Uri, which is how the tests use _stringsLegalEverywhere.

- [ ] *Test cases added*


robward-ms:
The lab shows these do not reduce test failures as expected, need to investigate.

robward-ms:
I have two enlistments: the one where I did this fix, which shows 248 failures in feature/netcore (which does not match the lab) and one where the failure rate is consistent with the lab. In short, this fix is not helping anything.

PR #1351, opened 2018-04-03
Add non-ODatat test controller and Update Odata test controller

robward-ms:
### Issues

### Description

Implement a non-OData test controller similar to the one in the E2E tests
which returns abstracted results.

Add functionality to the OData test controller similar to the
one in the E2E tests.

- [ ] *Test cases added*


robward-ms:
The lab shows these do not reduce test failures as expected, need to investigate.

robward-ms:
I have two enlistments: the one where I did this fix, which shows 248 failures in feature/netcore (which does not match the lab) and one where the failure rate is consistent with the lab. In short, this fix is not helping anything.

PR #1353, opened 2018-04-03
Change OptionalParameter as Nullable

xuzhg:
### Description

* Parameters currently have a property named "Optional" that returns whether or not the property is nullable.  Nullable properties may be optional for an Action, where the value is passed in the payload,
but are required in functions where the parameter must be passed in the URL in order to resolve the correct overload.  True optional parameters were added in OData 4.01, and implemented in OData Library, but having an existing property named "Optional" on the parameter that really means "Nullable" is misleading, and makes it hard to expose true optional parameters through WebAPI OData.*

- [ ] *Test cases added*

### Additional work necessary


robward-ms:
Since the VSTS work items are not available to the community, there should be a GitHub work item to explain why this change is taking place.

xuzhg:
I created a corresponding issue and updated the description.

PR #1354, opened 2018-04-03
Remove "[" and "]" as general delimiters from AspNetCore tests

robward-ms:
### Issues

### Description

Remove "[" and "]" as general delimiters from AspNetCore tests for ODataPathRouteConstraintTest.

RFC3986 specifies "[" and "]" and gen-delims in BNF notation in section 2.2 but section 3.2.2
specifies that the are only valid in a Uri as part of an IPV6 host. Therefore, they are not
valid in any part of a Uri, which is how the tests use _stringsLegalEverywhere.

- [ ] *Test cases added*

### DIFF
--- a/test/UnitTest/Microsoft.Test.AspNet.OData/Routing/ODataPathRouteConstraintTest.cs
+++ b/test/UnitTest/Microsoft.Test.AspNet.OData/Routing/ODataPathRouteConstraintTest.cs
@@ -5,20 +5,18 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
-using Microsoft.AspNet.OData;
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Routing;
 using Microsoft.AspNet.OData.Routing.Conventions;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.OData.Edm;
-using Microsoft.Test.AspNet.OData.Factories;
-using Microsoft.Test.AspNet.OData.Common;
-using Xunit;
-using Microsoft.Test.AspNet.OData.Formatter;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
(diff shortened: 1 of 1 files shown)


robward-ms:
Fixes about 68 UTs, failure rate should be 55 UT failures with this fix.

test/UnitTest/Microsoft.Test.AspNet.OData/Routing/ODataPathRouteConstraintTest.cs
```diff
@@ -67,7 +65,15 @@ public class ODataPathRouteConstraintTest
             { "some(sub)and&other=delims" },
             { "some(delims)but%2Bupper:escaped" },              // "some(delims)but+upper:escaped"
             { "some(delims)but%2blower:escaped" },              // "some(delims)but+lower:escaped"
+#if NETCORE
+            // RFC3986 specifies "[" and "]" and gen-delims in BNF notation in section  2.2
+            // but section 3.2.2 specifies that the are only valid in a Uri as part of an IPV6
+            // host. Therefore, they are not valid in any part of a Uri.
+            { ":@" },                                           // general delimeters that work
```
xuzhg: :@  can't work in Classic?

robward-ms: ":", "@" are not escaped in either version. "[" and "]" are escaped by HttpRequest and it's support classes. They are not escaped by HttpRequestMessage. 

See (PathStringHelper)[https://github.com/aspnet/HttpAbstractions/blob/0da88e0dda36d48ae68ac24289c349605cd94828/src/Microsoft.AspNetCore.Http.Abstractions/Internal/PathStringHelper.cs] for the encoding rules in AspNetCore.

The AspNet code tests all 4 so I'm not sure I understand the question?

#if NETCORE

            // RFC3986 specifies "[" and "]" and gen-delims in BNF notation in section  2.2

            // but section 3.2.2 specifies that the are only valid in a Uri as part of an IPV6

            // host. Therefore, they are not valid in any part of a Uri.

            { ":@" },                                           // general delimeters that work

#else

            // For back-compat, AspNet still tests with "[" and "]" as gen-delims.

            { ":[]@" },                                         // general delimeters that work

#endif

PR #1362, opened 2018-04-06
Fix test case failure for model builder

AlanWong-MS:
### Description
- Update expected metadata for a test case that fails after referencing new ODL version.

- [ ] *Test cases added*


AlanWong-MS:
I ran the build on a separate test machine and it passed.

AlanWong-MS:
@xuzhg I can't seem to update the status of this build to say success but per offline discussion, the build actually passed.

Issue #1365, opened 2018-04-07
ODataActionSelector selects actions for which parameters have not been bound


robward-ms:
ODataActionSelector selects actions for which parameters have not been bound,
such as methods relating to $ref. 

### Assemblies affected

Microsoft.AspNetCore.OData.dll 7.0.0

### Reproduce steps

Create a controller with a signatures such as:

        public string DeleteRef(string navigationProperty)
        {
            return String.Format(CultureInfo.InvariantCulture, "DeleteRef({0})", navigationProperty);
        }

        public string DeleteRef(int relatedKey, string navigationProperty)
        {
            return String.Format(CultureInfo.InvariantCulture, "DeleteRef({0})ByKey({1})", navigationProperty, relatedKey);
        }

Send a DELETE request that targets the first method, i.e. entity/navigation/$ref

### Expected result

DeleteRef(string navigationProperty) is called.

### Actual result

DeleteRef(int relatedKey, string navigationProperty) is called and key is 0.

AlanWong-MS:
Looks like this issue has been addressed by #1366. @robward-ms / @xuzhg, can we close this one?

PR #1366, opened 2018-04-07
Modify ODataActionSelector to bind based on available route data

robward-ms:
### Issues

*This pull request fixes issue #1365

### Description

Modify ODataActionSelector to bind based on available route data taking
into account parameters types the bind using custom binders.

- [ ] *Test cases added*


robward-ms:
Fixes 6 failing UTs

PR #1367, opened 2018-04-07
Fix routing UTs

robward-ms:
### Issues

### Description

Fix routing UTs:

1.) Modify Match_ReturnsTrue_IfODataPathCanBeParsed to handle platform-specific differences
    in route constraint.
2.) Test ODataRoute without comparing to HttpRoute due to platform differences in HttpRoute.
3.) Modify EntityTypeFunction to return error based on ModelState instead of relying on
    exception handling which is platform dependant.

- [ ] *Test cases added*


robward-ms:
Fixes 5 failing UTs

PR #1374, opened 2018-04-10
Add Url property for resource(set) context

xuzhg:
### Issues

* Add Url property for resource(set) context

### Description

* Add Url property for resource(set) context

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/ResourceSetContext.cs
+++ b/src/Microsoft.AspNetCore.OData/ResourceSetContext.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNet.OData.Adapters;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter.Serialization;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.OData.Edm;
 
 namespace Microsoft.AspNet.OData
@@ -16,6 +17,7 @@ namespace Microsoft.AspNet.OData
     public partial class ResourceSetContext
     {
         private HttpRequest _request;
+        private IUrlHelper _urlHelper;
 
         /// <summary>
         /// Gets or sets the HTTP request that caused this instance to be generated.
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.OData/ResourceSetContext.cs
```diff
@@ -16,19 +17,20 @@ namespace Microsoft.AspNet.OData
...
+        private IUrlHelper _urlHelper;
 
         /// <summary>
         /// Gets or sets the HTTP request that caused this instance to be generated.
         /// </summary>
         /// <remarks>This signature uses types that are AspNetCore-specific.</remarks>
         public HttpRequest Request
         {
-            get { return _request; }
+            get => _request;
```
AlanWong-MS: Stylecop/Code Analysis may complain about this syntax in the build someday.

test/UnitTest/Microsoft.Test.AspNet.OData/Formatter/Serialization/ODataResourceSetSerializerTests.cs
```diff
@@ -730,8 +749,38 @@ private static void AssertEqual(Uri expected, Uri actual)
             Assert.Equal(expected.AbsoluteUri, actual.AbsoluteUri);
         }
 
-#if NETFX // Only AspNet version has UrlHelper
-        private static UrlHelper CreateMetadataLinkFactory(string metadataPath)
+#if NETCORE
+        public class MyUrlHelper : IUrlHelper
+        {
+            private string _link;
```
AlanWong-MS: Stylecop/Code Analysis may complain about starting with underscore in variable names.

src/Microsoft.AspNetCore.OData/ResourceSetContext.cs
xuzhg: AspNetCore will target the vs2017 or later. It's recommendation. But, I can revert it.

PR #1379, opened 2018-04-12
Remove non-used #if NetCore1x

xuzhg:
### Issues

* Remove non-used #if NetCore1x

### Description

* Remove non-used #if NetCore1x

- [ ] *Test cases added*

### Additional work necessary


biaol-odata:
Is this NetCore1x never used, or it has just been cleaned up? If it is the later case, would be good to have reference to where it was cleanup..

xuzhg:
@biaol-odata Never used.

PR #1383, opened 2018-04-12
Make ODataLevelsTests pass

xuzhg:
### Issues

* Make ODataLevelsTests pass

### Description

* Make ODataLevelsTests pass

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/UnitTest/Microsoft.Test.AspNet.OData/TestControllers.cs
+++ b/test/UnitTest/Microsoft.Test.AspNet.OData/TestControllers.cs
@@ -5,6 +5,7 @@
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData;
 using Microsoft.AspNet.OData.Results;
+using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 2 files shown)


test/UnitTest/Microsoft.Test.AspNet.OData/TestControllers.cs
```diff
@@ -103,6 +107,35 @@ public TestBadRequestResult(BadRequestResult innerResult)
...
+            : base(innerResult)
+        {
+            _value = innerResult.Value;
+        }
+
+        public override Task ExecuteResultAsync(ActionContext context)
+        {
+            if (_value is string)
+            {
+                context.HttpContext.Response.StatusCode = 400;
```
AlanWong-MS: nit: is there a constant that we can use from anywhere?

xuzhg: Yes, in ASP.NET Classic, but not in ASP.NET Core.

Issue #1413, opened 2018-04-28
odata webapi core: sehexception when using $apply=aggregate 


mysticdotnet:
*When i use $apply=aggregate with your sample (https://github.com/OData/WebApi/tree/master/samples/AspNetCoreODataSample.Web) it throw sehexception*

### Assemblies affected

*Microsoft.AspNetCore.OData with asp.net core

### Reproduce steps

*I clone this project (https://github.com/OData/WebApi), i open this solution the odata core solution (WebApiOData.AspNetCore.sln)
and i run the sample project (https://github.com/OData/WebApi/tree/master/samples/AspNetCoreODataSample.Web)
and after i run this query: localhost:5912/odata/Movies?$apply=groupby((ReleaseDate)) *

### Expected result

*Show the result of the odata aggregate query*

### Actual result

*sehexception*

xuzhg:
@mysticdotnet

Thanks for tring our sample. I change the sample and verify that the sehexception is not from our codes. I do think it's from efcore.

Because, when I change to use the in-memory data, the same aggregate expression can work. for example:

![image](https://user-images.githubusercontent.com/9426627/39496172-2c44209a-4d52-11e8-9dcd-523b80f4475c.png)

But, when I switch to EFCore, it throws the sehexception.

mysticdotnet:
Thanks a lot for your help.
Should i submit a ticket to the ef core team with  a reference to this issue ?

xuzhg:
@mysticdotnet That's helpful. Let's follow up the input form EFCore. Thanks.

ajcvickers:
@karelz I was unable even to repro the issue.

xuzhg:
@karelz @ajcvickers 
I think it's easy to repo using our sample project.

1. Git clone:

```json
git clone git@github.com:OData/WebApi.git
```

2. VS2017 opens solution: https://github.com/OData/WebApi/blob/master/sln/WebApiOData.AspNetCore.sln

3.  Build and run: `AspNetCoreODataSample.Web` project
4.  Postman file request at: http://localhost:5913/efcore/Movies?$apply=groupby((ReleaseDate))
5. Check the response, you will get the **sehexception**

karelz:
@xuzhg that is exactly what I tried. Attached windbg to IISExpress - no exception. Even no sehexception in VS, just 502 as I posted in https://github.com/dotnet/corefx/issues/29526#issuecomment-388932741

mysticdotnet:
[...]

PR #1424, opened 2018-05-01
modify OData sample to verify issue #1413

xuzhg:
### Issues

* modify OData sample to verify issue #1413

### Description

* modify OData sample to verify issue #1413

- [ ] *Test cases added*

### Additional work necessary


samples/AspNetCoreODataSample.Web/Models/EdmModelBuilder.cs
```diff
@@ -3,6 +3,7 @@
 
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.OData.Edm;
+using System;
```
AlanWong-MS: System namespace should come before Microsoft.

Issue #1376, opened 2018-04-11
[feature/netcore] BadRequest(ModelState) return a bad result called from an OData controller


LeandroCannizzaro:
In one controller derived from ODataController I try to do a "return a BadRequest(ModelState)".
My Expectation is to have a response at client level as the same code produce in one basic .net core controller as the following json:
```
{
"Name": ["The field Name is required"]
}
```
Instead i have the following response
{"error": {"code":"", "message":""}}
Where the response not is the same that I have in one normal controller and
more important the errors presents in the ModelState not are bounded to the returned response.

*I am using  OData WebApi 7.0.0 Beta 2*

### Reproduce steps
*Create a model with one mandatory field as*
```
    public class TestEntity : Entity
    {
        [Required]
        public string Name { get; set; }
    }

```
*Create a basic ODataController with one method Create*

```
    public class TestApiController : ODataController
    {
        [HttpPost]
        public async Task<IActionResult> Create([FromBody] TestEntity customer)
        {
            if (!ModelState.IsValid)
            {
                //This return not return the validation errors present in the ModelState object.
                return BadRequest(ModelState); 
            }

            //Some code for persist the entity.

            return Created(testEntity);
        }
     }
```
### Expected result
{"Name": ["The field Name is required"]}

### Actual result
{"error""code":"","message":""}}

Thanks in advance for the help.
Leandro

robward-ms:
See #1338 and PR #1343. There was a bug in SerializableErrorExtensions.cs where model state errors were not returned properly. This fix is not in Beta2 but you can try the nightlies to see if that fixes the issue for you.

dozer75:
I had the same issue today, retrieved the latest bits and built, there is a message there, but the message is really poor, doesn't say anything what's wrong when using the Delta type as input parameter.

This is the message I get from the service on ModelState error now:

`{"error":{"code":"","message":" : The input was not valid.\r\n","details":[{"code":"","message":":System.String[]"}]}}`

It isn't much information there WHAT's wrong. When I debug I see that the ModelState is filled by an exception looking like this:

`A null value was found for the property named '<fieldname>', which has the expected type 'Edm.String[Nullable=False]'. The expected type 'Edm.String[Nullable=False]' does not allow null values.`

The ModelState itself is actually not filled correctly either. It should honor how MVC does it otherwise, so I think this logic should be revised.

dozer75:
[...]

PR #1428, opened 2018-05-03
Fixed better ODataError generation from SerializableError

dozer75:
### Issues

*This pull request helps fixing issue #1376.*

### Description

Streamlined the error generation to both get correct and more readable errors when OData encounters a `SerializableError` instance.

Fixed an issue running unit tests on non-US regional settings

- [ ] *Test cases added*


robward-ms:
@dozer75 - Thanks!  The fix looks great!

However, it looks like it's not passing the AspNet build due to the test change:

```
018-05-03T11:45:22.9342485Z "C:\BuildAgent\_work\75\s\sln\WebApiOData.AspNet.sln" (default target) (1) ->
2018-05-03T11:45:22.9342485Z "C:\BuildAgent\_work\75\s\test\UnitTest\Microsoft.AspNet.OData.Test\Microsoft.AspNet.OData.Test.csproj" (default target) (3) ->
2018-05-03T11:45:22.9342485Z (CoreCompile target) -> 
2018-05-03T11:45:22.9342485Z   Shared\Routing\AttributeRoutingUnboundTest.cs(324,41): error CS0103: The name 'CultureInfo' does not exist in the current context [C:\BuildAgent\_work\75\s\test\UnitTest\Microsoft.AspNet.OData.Test\Microsoft.AspNet.OData.Test.csproj]
2018-05-03T11:45:22.9342485Z 
2018-05-03T11:45:22.9342485Z     0 Warning(s)
2018-05-03T11:45:22.9342485Z     1 Error(s)
```

dozer75:
Strange :| It compiled here, must be something I did wrong during commit. I'll fix it.

dozer75:
The compile error should have been fixed, missed the `#if #else` compiler condition for the using statements.

src/Microsoft.AspNetCore.OData/Extensions/SerializableErrorExtensions.cs
```diff
@@ -117,7 +176,11 @@ private static TValue GetPropertyValue<TValue>(this SerializableError error, str
         }
     }
 
-    internal static class SerializableErrorKeys
+    /// <summary>
+    ///     Different keys for adding entries to an <see cref="SerializableError" /> instance so
+    ///     that it can be parsed to a <see cref="ODataError" /> instance
+    /// </summary>
+    public static class SerializableErrorKeys
```
robward-ms: This bit of the change is causing a test failure for the PublicApi UT since this is now public. Does it need to be public? If it does, you'll need to run the TestPublicApi UT and update the baseline (.bsl file). I think the instructions are wrong at the moment and there are 2 AspNetCore .bsl files, you'll need to update the one that is not in the shared directory. The UTs all run and pass in VS.

robward-ms:
Also, looks like you need to rebase on top of master.

src/Microsoft.AspNetCore.OData/Extensions/SerializableErrorExtensions.cs
robward-ms: The test should be fine, I have it running in my latest PR. But like I said, the instructions are wrong and tell you to update the bsl file in the shared test dir. The bsl in the AspNetCore directory is the one that is actually used so you'll need to update that one.

dozer75:
As far as I can see my branch is merged with latest master?

src/Microsoft.AspNetCore.OData/Extensions/SerializableErrorExtensions.cs
dozer75: Fixed! Also added a fix to correct the wrong copy instruction.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/PublicApi/PublicApiTest.cs
```diff
@@ -40,7 +40,7 @@ public void TestPublicApi()
                 String.Format("Base line file {1} and output file {2} do not match, please check.{0}" +
                 "To update the baseline, please run:{0}{0}" +
                 "copy /y \"{2}\" \"{1}\"", Environment.NewLine,
-                @"test\UnitTest\Microsoft.AspNet.OData.Test\PublicApi\" + BaseLineFileName,
+                @"test\UnitTest\Microsoft.AspNetCore.OData.Test\PublicApi\" + BaseLineFileName,
```
robward-ms: Sadly, this is a little more complicated given that the baselines are in different folders. I also have a fix for this which creates a BaseLineFileFolder value that each test dll can set and allow the instruction to point to different locations. You can feel free to copy mine or leave this out of the change. See #1433

src/Microsoft.AspNetCore.OData/Extensions/SerializableErrorExtensions.cs
robward-ms: Looks like the public API is still failing in VSTS, did it pass locally?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/PublicApi/PublicApiTest.cs
dozer75: Ok, reverted it then, better having your fix!

robward-ms:
@xuzhg - Can you take a look as well?

xuzhg:
Merged. Thanks your contribution.

PR #1431, opened 2018-05-03
add a full solution including classic and core

xuzhg:
### Issues

* add a full solution including classic and core

### Description

*add a full solution including classic and core
Now, we can open classic and core project in one vs2017:
image

And we can run the classic and core unit test in one vs2017:

image

- [ ] *Test cases added*

### Additional work necessary


sln/WebApiOData.Full.sln
```diff
@@ -0,0 +1,85 @@
+﻿
```
robward-ms: If this one is "Full", would that imply it contained the E2E tests as well?

xuzhg: WebApiOData.E2E.Full.sln ?

xuzhg:
No needed

Issue #1247, opened 2018-02-22
Odata 7.xx spatial IConvertible  problem


premiaware:
GeographyPoint throws Object must implement IConvertible inside the Patch Method of controller.

### Assemblies affected

OData WebApi lib 7.xxx

### Reproduce steps

After upgrade my project from OData 6.xx to OData 7.xx. in the controller Patch method:

public async Task Patch([FromODataUri] long key, Delta delta)

delta always is null when the DbGeography have valid value.

How can is resolve the problem? using OData 6.x all working well.
Reading data working well

the problem is similar to old thread 

https://github.com/OData/WebApi/issues/718

but in my project the problem is only using OData 7.

xuzhg:
@premiaware are you using ASP.NET Core for OData?

premiaware:
No, I using target framework .net framework 4.6.1

xuzhg:
@premiaware sorry, I mean the Web API OData version. Is it 5.x, 6.x or 7.x?

premiaware:
Microsoft.OData.Core 7.4.1.20214
Microsoft.OData.Edm 7.4.1.20214
System.Web.OData 6.0.0.0

premiaware:
Today i have try to update to 
Microsoft.AspNet.OData 7.0.0-beta2
Microsoft.OData.Core 7.4.3
Microsoft.OData.Edm 7.4.3
Microsoft.Spatial 7.4.3

but the problem is the same, when using Spatial Data the Delta is null in Patch controller method.
how i can fix the problem?

premiaware:
I have update spatial to 7.4.4 but the problem is the same, how can fix this?

xuzhg:
Merged the PR. Please re-open it or file new issue if you have further concern. Thanks.

PR #1443, opened 2018-05-23
Fix the #1247 about spatial IConvertible problem

xuzhg:
### Issues

* #1247

### Description

* Make the post/update spatial property work

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Spatial/SpatialController.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Spatial/SpatialController.cs
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNet.OData;
+using Microsoft.Spatial;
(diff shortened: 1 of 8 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Spatial/SpatialController.cs
```diff
@@ -0,0 +1,89 @@
...
+        static SpatialCustomersController()
+        {
+            if (Customers == null)
+            {
+                string[] names = { "Mercury", "Venus", "Earth", "Mars", "Jupiter", "Saturn", "Uranus", "Neptune" };
+
+                Customers = Enumerable.Range(1, 7).Select(e => new SpatialCustomer
+                {
+                    CustomerId = e,
+                    Name = names[e - 1],
```
AlanWong-MS: Why do we take only a range of the string names?

AlanWong-MS:
Is this mostly ported code?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Spatial/SpatialController.cs
xuzhg: Just sample data.

xuzhg:
@AlanWong-MS Yes, with changes to align with the new e2e platform.

Issue #1263, opened 2018-03-06
PATCH of nested complex properties not per OData spec


catalinis:
*Short summary (3-5 sentences) describing the issue.*

Per OData Spec:
11.4.3 Update an Entity
Services SHOULD support PATCH as the preferred means of updating an entity. PATCH provides more resiliency between clients and services by directly modifying only those values specified by the client.
The semantics of PATCH, as defined in [RFC5789], is to merge the content in the request payload with the [entity’s] current state, applying the update only to those components specified in the request body. Collection properties and primitive properties provided in the payload corresponding to updatable properties MUST replace the value of the corresponding property in the entity or complex type. Missing properties of the containing entity or complex property, including dynamic properties, MUST NOT be directly altered unless as a side effect of changes resulting from the provided properties.

What we are observing is that PATCH for complex properties does not follow the spec above.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData WebApi lib 6.1.0

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*
Assume the Contact schema:
<EntityType Name="Contact" ">
      <Key>
            <PropertyRef Name="Id"/>
      </Key>
      <Property Name="Id" Type="Edm.String"/>
      <Property Name="Name" Type="Edm.String"/>
      <Property Name="HomeAddress" Type=" PhysicalAddress"/>
</EntityType>
< ComplexType Name="PhysicalAddress">
      <Property Name="Street" Type="Edm.String"/>
      <Property Name="City" Type="Edm.String"/>
</ComplexType>

and an entity instance:
{
                Id: “1”,
                Name: “Joe”
                HomeAddress:{
                                Street : “Main street”,
                                City : “Some City”
                }
}
and we want to update the Name and City using
PATCH /Contacts(‘1’)
{
                Name: “Joey”,
                HomeAddress:{
                                City : “Some other city”
                }
}

### Expected result

*What would happen if there wasn't a bug.*

Contacts(1) would be:
{
                Id: “1”,
                Name: “Joey”
                HomeAddress:{
                                Street : “Main street”,
                                City : “Some other city”
                }
}

### Actual result

*What is actually happening.*
Contacts(1) would be:
{
                Id: “1”,
                Name: “Joey”
                HomeAddress:{
                                Street : null,
                                City : “Some other city”
                }
}

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

[...]

PR #1445, opened 2018-05-24
Patch nested structural resources

biaol-odata:
### Issues

*This pull request fixes issue #1263.*

### Description

*Change Delta<T> class to represents delta payload in hierarchical structure, so that payload deserialization during request dispatching and patch functionality invoked by service controller can process the delta payload correctly, eventually resource is patched correctly without overwriting valid, unchanged properties in the resource.*

Build on PR branch: https://github.com/biaol-odata/WebApi/commits/patchNestedProperties

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Delta.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Delta.cs
@@ -44,17 +44,6 @@ namespace Microsoft.AspNet.OData
         public abstract bool TryGetPropertyValue(string name, out object value);
 
         /// <summary>
-        /// Attempts to add the delta nested resource.
-        /// </summary>
-        /// <param name="name">Name of the nested resource.</param>
-        /// <param name="deltaNestedResource">The delta object for the nested resource.</param>
-        /// <returns></returns>
-        public virtual bool TryAddNestedResource(string name, object deltaNestedResource)
-        {
-            throw new NotImplementedException();
-        }
-
-        /// <summary>
         /// Attempts to get the <see cref="Type"/> of the Property called <paramref name="name"/> from the underlying Entity.
         /// <remarks>
         /// Only properties that exist on Entity can be retrieved.
--- a/src/Microsoft.AspNet.OData.Shared/Delta.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Delta.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.OData
     /// A class the tracks changes (i.e. the Delta) for an entity.
     /// </summary>
     [NonValidatingParameterBinding]
-    public abstract class Delta : DynamicObject, IDelta 
+    public abstract class Delta : DynamicObject, IDelta
     {
         /// <summary>
         /// Clears the Delta and resets the underlying Entity.
@@ -42,6 +42,17 @@ namespace Microsoft.AspNet.OData
         /// <param name="value">The value of the Property</param>
         /// <returns>True if the Property was found</returns>
         public abstract bool TryGetPropertyValue(string name, out object value);
+
+        /// <summary>
+        /// Attempts to add the delta nested resource.
+        /// </summary>
+        /// <param name="name">Name of the nested resource.</param>
+        /// <param name="deltaNestedResource">The delta object for the nested resource.</param>
+        /// <returns></returns>
+        public virtual bool TryAddNestedResource(string name, object deltaNestedResource)
+        {
+            throw new NotImplementedException();
+        }
 
         /// <summary>
         /// Attempts to get the <see cref="Type"/> of the Property called <paramref name="name"/> from the underlying Entity.
@@ -84,14 +95,14 @@ namespace Microsoft.AspNet.OData
         }
 
         /// <summary>
-        /// Returns the Properties that have been modified through this Delta as an 
-        /// enumeration of Property Names 
+        /// Returns the Properties that have been modified through this Delta as an
+        /// enumeration of Property Names
         /// </summary>
         public abstract IEnumerable<string> GetChangedPropertyNames();
 
         /// <summary>
-        /// Returns the Properties that have not been modified through this Delta as an 
-        /// enumeration of Property Names 
+        /// Returns the Properties that have not been modified through this Delta as an
+        /// enumeration of Property Names
         /// </summary>
         public abstract IEnumerable<string> GetUnchangedPropertyNames();
     }
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.OData
         private IDictionary<string, object> _deltaNestedResources;
 
         private TStructuralType _instance;
-        private Type _entityType;
+        private Type _structuredType;
 
         private PropertyInfo _dynamicDictionaryPropertyinfo;
         private HashSet<string> _changedDynamicProperties;
@@ -92,11 +92,11 @@ namespace Microsoft.AspNet.OData
         }
 
         /// <inheritdoc/>
-        public override Type EntityType
+        public override Type StructuredType
         {
             get
             {
-                return _entityType;
+                return _structuredType;
             }
         }
 
@@ -109,11 +109,23 @@ namespace Microsoft.AspNet.OData
         /// <inheritdoc/>
         public override void Clear()
         {
-            Reset(_entityType);
+            Reset(_structuredType);
         }
 
         /// <inheritdoc/>
         public override bool TrySetPropertyValue(string name, object value)
+        {
+            if (value is IDelta)
+            {
+                return TrySetNestedResourceInternal(name, value);
+            }
+            else
+            {
+                return TrySetPropertyValueInternal(name, value);
+            }
+        }
+
+        private bool TrySetPropertyValueInternal(string name, object value)
         {
             if (name == null)
             {
@@ -161,8 +173,7 @@ namespace Microsoft.AspNet.OData
             return true;
         }
 
-        /// <inheritdoc/>
-        public override bool TryAddNestedResource(string name, object deltaNestedResource)
+        private bool TrySetNestedResourceInternal(string name, object deltaNestedResource)
         {
             if (name == null)
             {
@@ -213,11 +224,15 @@ namespace Microsoft.AspNet.OData
             if (this._deltaNestedResources.ContainsKey(name))
             {
                 // If this is a nested resource, get the value from the dictionary of nested resources.
-                TypedDelta deltaNestedResource = (TypedDelta)_deltaNestedResources[name];
-                Debug.Assert(deltaNestedResource != null, "deltaNestedResource != null");
-                Debug.Assert(IsDeltaOfT(deltaNestedResource.GetType()));
+                object deltaNestedResource = _deltaNestedResources[name];
 
-                value = deltaNestedResource.GetInstance();
+                Contract.Assert(deltaNestedResource != null, "deltaNestedResource != null");
+                Contract.Assert(IsDeltaOfT(deltaNestedResource.GetType()));
+
+                // Get the Delta<{NestedResourceType}>._instance using Reflection.
+                FieldInfo field = deltaNestedResource.GetType().GetField("_instance", BindingFlags.NonPublic | BindingFlags.Instance);
+                Contract.Assert(field != null, "field != null");
+                value = field.GetValue(deltaNestedResource);
                 return true;
             }
             else
@@ -281,7 +296,7 @@ namespace Microsoft.AspNet.OData
         /// Returns the instance that holds all the changes (and original values) being tracked by this Delta.
         /// </summary>
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Justification = "Not appropriate to be a property")]
-        internal override object GetInstance()
+        public TStructuralType GetInstance()
         {
             return _instance;
         }
@@ -321,9 +336,9 @@ namespace Microsoft.AspNet.OData
 
             // Delta parameter type cannot be derived type of original
             // to prevent unrecognizable information from being applied to original resource.
-            if (!_entityType.IsAssignableFrom(original.GetType()))
+            if (!_structuredType.IsAssignableFrom(original.GetType()))
             {
-                throw Error.Argument("original", SRResources.DeltaTypeMismatch, _entityType, original.GetType());
+                throw Error.Argument("original", SRResources.DeltaTypeMismatch, _structuredType, original.GetType());
             }
 
             RuntimeHelpers.EnsureSufficientExecutionStack();
@@ -346,7 +361,7 @@ namespace Microsoft.AspNet.OData
                 dynamic originalNestedResource = null;
                 if (!TryGetPropertyRef(original, nestedResourceName, out originalNestedResource))
                 {
-                    throw Error.Argument(nestedResourceName, "Cannot find nested resource name '{0}' in parent resource type '{1}'",
+                    throw Error.Argument(nestedResourceName, SRResources.DeltaNestedResourceNameNotFound,
                         nestedResourceName, original.GetType());
                 }
 
@@ -370,25 +385,24 @@ namespace Microsoft.AspNet.OData
         }
 
         /// <summary>
-        /// Gets the property object by the specified name.
+        /// Attempts to get the property by the specified name.
         /// </summary>
-        /// <param name="structural"></param>
-        /// <param name="propertyName"></param>
-        /// <param name="propertyRef">Output </param>
-        /// <returns></returns>
+        /// <param name="structural">The structural object.</param>
+        /// <param name="propertyName">Name of the property.</param>
+        /// <param name="propertyRef">Output for property value.</param>
+        /// <returns>true if the property is found; false otherwise.</returns>
         private static bool TryGetPropertyRef(TStructuralType structural, string propertyName,
             out dynamic propertyRef)
         {
             propertyRef = null;
-            bool found = false;
             PropertyInfo propertyInfo = structural.GetType().GetProperty(propertyName);
             if (propertyInfo != null)
             {
                 propertyRef = propertyInfo.GetValue(structural, null);
-                found = true;
+                return true;
             }
 
-            return found;
+            return false;
         }
 
         // Copy changed dynamic properties and leave the unchanged dynamic properties
@@ -449,9 +463,9 @@ namespace Microsoft.AspNet.OData
                 throw Error.ArgumentNull("original");
             }
 
-            if (!_entityType.IsInstanceOfType(original))
+            if (!_structuredType.IsInstanceOfType(original))
             {
-                throw Error.Argument("original", SRResources.DeltaTypeMismatch, _entityType, original.GetType());
+                throw Error.Argument("original", SRResources.DeltaTypeMismatch, _structuredType, original.GetType());
             }
 
             IEnumerable<PropertyAccessor<TStructuralType>> propertiesToCopy = GetUnchangedPropertyNames().Select(s => _allProperties[s]);
@@ -541,7 +555,7 @@ namespace Microsoft.AspNet.OData
             _instance = Activator.CreateInstance(structuralType) as TStructuralType;
             _changedProperties = new HashSet<string>();
             _deltaNestedResources = new Dictionary<string, object>();
-            _entityType = structuralType;
+            _structuredType = structuralType;
 
             _changedDynamicProperties = new HashSet<string>();
             _dynamicDictionaryCache = null;
@@ -550,7 +564,7 @@ namespace Microsoft.AspNet.OData
         private void InitializeProperties(IEnumerable<string> updatableProperties)
         {
             _allProperties = _propertyCache.GetOrAdd(
-                _entityType,
+                _structuredType,
                 (backingType) => backingType
                     .GetProperties(BindingFlags.Instance | BindingFlags.Public)
                     .Where(p => (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -4,10 +4,12 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Formatter;
 
@@ -28,13 +30,17 @@ namespace Microsoft.AspNet.OData
         private HashSet<string> _updatableProperties;
 
         private HashSet<string> _changedProperties;
+
+        // Nested resources or structures changed at this level.
+        private IDictionary<string, object> _deltaNestedResources;
+
         private TStructuralType _instance;
         private Type _entityType;
 
         private PropertyInfo _dynamicDictionaryPropertyinfo;
         private HashSet<string> _changedDynamicProperties;
         private IDictionary<string, object> _dynamicDictionaryCache;
- 
+
         /// <summary>
         /// Initializes a new instance of <see cref="Delta{TStructuralType}"/>.
         /// </summary>
@@ -156,6 +162,33 @@ namespace Microsoft.AspNet.OData
         }
 
         /// <inheritdoc/>
+        public override bool TryAddNestedResource(string name, object deltaNestedResource)
+        {
+            if (name == null)
+            {
+                throw Error.ArgumentNull("name");
+            }
+
+            if (!_updatableProperties.Contains(name))
+            {
+                return false;
+            }
+
+            if (_deltaNestedResources.ContainsKey(name))
+            {
+                // Ignore duplicated nested resource.
+                return false;
+            }
+
+            // Add the nested resource in the hierarchy.
+            // Note: We shouldn't add the structural properties to the <code>_changedProperties</code>, which
+            // is used for keeping track of changed non-structural properties at current level.
+            _deltaNestedResources[name] = deltaNestedResource;
+
+            return true;
+        }
+
+        /// <inheritdoc/>
         public override bool TryGetPropertyValue(string name, out object value)
         {
             if (name == null)
@@ -167,7 +200,7 @@ namespace Microsoft.AspNet.OData
             {
                 if (_dynamicDictionaryCache == null)
                 {
-                    _dynamicDictionaryCache = 
+                    _dynamicDictionaryCache =
                         GetDynamicPropertyDictionary(_dynamicDictionaryPropertyinfo, _instance, create: false);
                 }
 
@@ -177,11 +210,25 @@ namespace Microsoft.AspNet.OData
                 }
             }
 
-            PropertyAccessor<TStructuralType> cacheHit;
-            if (_allProperties.TryGetValue(name, out cacheHit))
+            if (this._deltaNestedResources.ContainsKey(name))
             {
-                value = cacheHit.GetValue(_instance);
+                // If this is a nested resource, get the value from the dictionary of nested resources.
+                TypedDelta deltaNestedResource = (TypedDelta)_deltaNestedResources[name];
+                Debug.Assert(deltaNestedResource != null, "deltaNestedResource != null");
+                Debug.Assert(IsDeltaOfT(deltaNestedResource.GetType()));
+
+                value = deltaNestedResource.GetInstance();
                 return true;
+            }
+            else
+            {
+                // try to retrieve the value as property.
+                PropertyAccessor<TStructuralType> cacheHit;
+                if (_allProperties.TryGetValue(name, out cacheHit))
+                {
+                    value = cacheHit.GetValue(_instance);
+                    return true;
+                }
             }
 
             value = null;
@@ -234,19 +281,20 @@ namespace Microsoft.AspNet.OData
         /// Returns the instance that holds all the changes (and original values) being tracked by this Delta.
         /// </summary>
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Justification = "Not appropriate to be a property")]
-        public TStructuralType GetInstance()
+        internal override object GetInstance()
         {
             return _instance;
         }
 
         /// <summary>
         /// Returns the known properties that have been modified through this <see cref="Delta"/> as an
-        /// <see cref="IEnumerable{T}" /> of property Names. Does not include the names of the changed
-        /// dynamic properties.
+        /// <see cref="IEnumerable{T}" /> of property Names.
+        /// Includes the structural properties at current level.
+        /// Does not include the names of the changed dynamic properties.
         /// </summary>
         public override IEnumerable<string> GetChangedPropertyNames()
         {
-            return _changedProperties;
+            return _changedProperties.Concat(_deltaNestedResources.Keys);
         }
 
         /// <summary>
@@ -260,8 +308,8 @@ namespace Microsoft.AspNet.OData
         }
 
         /// <summary>
-        /// Copies the changed property values from the underlying entity (accessible via <see cref="GetInstance()" />) 
-        /// to the <paramref name="original"/> entity.
+        /// Copies the changed property values from the underlying entity (accessible via <see cref="GetInstance()" />)
+        /// to the <paramref name="original"/> entity recursively.
         /// </summary>
         /// <param name="original">The entity to be updated.</param>
         public void CopyChangedValues(TStructuralType original)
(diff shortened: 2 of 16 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -481,30 +491,27 @@ private object ReadNestedResourceInline(ODataResourceWrapper resourceWrapper, IE
...
                 {
                     throw new ODataException(
                         Error.Format(SRResources.MappingDoesNotContainResourceType, structuredType.FullName()));
                 }
-
-                nestedReadContext.ResourceType = clrType;
             }
 
+            nestedReadContext.ResourceType = readContext.IsDeltaOfT
+                ? typeof(Delta<>).MakeGenericType(new Type[] { clrType })
```
xuzhg: >new Type[] { clrType } [](start = 50, length = 22)

why is it a collection?

src/Microsoft.AspNet.OData.Shared/Delta.cs
```diff
@@ -43,6 +43,17 @@ public abstract class Delta : DynamicObject, IDelta
         /// <returns>True if the Property was found</returns>
         public abstract bool TryGetPropertyValue(string name, out object value);
 
+        /// <summary>
+        /// Attempts to add the delta nested resource.
+        /// </summary>
+        /// <param name="name">Name of the nested resource.</param>
+        /// <param name="deltaNestedResource">The delta object for the nested resource.</param>
+        /// <returns></returns>
+        public virtual bool TryAddNestedResource(string name, object deltaNestedResource)
```
xuzhg: >TryAddNestedResource [](start = 28, length = 20)

I don't think this is necessary.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -28,13 +30,17 @@ public class Delta<TStructuralType> : TypedDelta, IDelta where TStructuralType :
...
+
         private TStructuralType _instance;
         private Type _entityType;
 
         private PropertyInfo _dynamicDictionaryPropertyinfo;
         private HashSet<string> _changedDynamicProperties;
         private IDictionary<string, object> _dynamicDictionaryCache;
- 
+
         /// <summary>
```
xuzhg: remove

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -155,6 +161,33 @@ public override bool TrySetPropertyValue(string name, object value)
...
+            }
+
+            // Add the nested resource in the hierarchy.
+            // Note: We shouldn't add the structural properties to the <code>_changedProperties</code>, which
+            // is used for keeping track of changed non-structural properties at current level.
+            _deltaNestedResources[name] = deltaNestedResource;
+
+            return true;
+        }
+
```
xuzhg: I don't think we should add a new public api. 

TrySetPropertyValue is enough, because they have the same signature. You can move the logic into the "TrySetPropertyValue" just compare the "object" type.

or just make it private and call from TrySetPropertyValue

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +319,76 @@ public void CopyChangedValues(TStructuralType original)
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Gets the property object by the specified name.
+        /// </summary>
+        /// <param name="structural"></param>
+        /// <param name="propertyName"></param>
+        /// <param name="propertyRef">Output </param>
```
xuzhg: complete the comments.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -234,19 +281,20 @@ public override bool TryGetPropertyType(string name, out Type type)
         /// Returns the instance that holds all the changes (and original values) being tracked by this Delta.
         /// </summary>
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Justification = "Not appropriate to be a property")]
-        public TStructuralType GetInstance()
+        internal override object GetInstance()
```
xuzhg: >internal override object [](start = 8, length = 24)

why change the return type and mark it from public to internal?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
xuzhg: If you think it's necessary, I think a corresponding method name "TryGetNestedResource(..)" is needed. 

TrySetPropertyValue vs TryGetPropertyValue

TryAddNestedResource vs TryGetNestedResource


---
In reply to: [190748771](https://github.com/OData/webapi/pull/1445#discussion_r190748771) [](ancestors = 190748771)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -177,12 +210,26 @@ public override bool TryGetPropertyValue(string name, out object value)
                 }
             }
 
-            PropertyAccessor<TStructuralType> cacheHit;
-            if (_allProperties.TryGetValue(name, out cacheHit))
+            if (this._deltaNestedResources.ContainsKey(name))
```
xuzhg: >if (this._deltaNestedResources.ContainsKey(name) [](start = 12, length = 48)

the changes here make sense, So, again, i don't think we should have a new public API named "TryAddNestedResource", Please change TrySetPropertyValue to add the nested resource.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
xuzhg: This change doesn't make sense to me.

---
In reply to: [190749152](https://github.com/OData/webapi/pull/1445#discussion_r190749152) [](ancestors = 190749152)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +319,76 @@ public void CopyChangedValues(TStructuralType original)
...
 
+            // Delta parameter type cannot be derived type of original
+            // to prevent unrecognizable information from being applied to original resource.
             if (!_entityType.IsAssignableFrom(original.GetType()))
             {
                 throw Error.Argument("original", SRResources.DeltaTypeMismatch, _entityType, original.GetType());
             }
 
-            PropertyAccessor<TStructuralType>[] propertiesToCopy = GetChangedPropertyNames().Select(s => _allProperties[s]).ToArray();
+            RuntimeHelpers.EnsureSufficientExecutionStack();
```
xuzhg: >RuntimeHelpers.EnsureSufficientExecutionStack(); [](start = 12, length = 48)

That's necessary to call?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +319,76 @@ public void CopyChangedValues(TStructuralType original)
...
+
+            // For nested resources.
+            foreach (string nestedResourceName in _deltaNestedResources.Keys)
+            {
+                // Patch for each nested resource changed under this TStructuralType.
+                dynamic deltaNestedResource = _deltaNestedResources[nestedResourceName];
+                dynamic originalNestedResource = null;
+                if (!TryGetPropertyRef(original, nestedResourceName, out originalNestedResource))
+                {
+                    throw Error.Argument(nestedResourceName, "Cannot find nested resource name '{0}' in parent resource type '{1}'",
```
xuzhg: >Cannot find nested resource name '{0}' in parent resource type '{1} [](start = 62, length = 67)

it's better to move to resources.resx

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +319,76 @@ public void CopyChangedValues(TStructuralType original)
...
+                    deltaObject.CopyChangedValues(instance);
+
+                    _allProperties[nestedResourceName].SetValue(original, instance);
+                }
+                else
+                {
+                    // Recursively patch the subtree.
+                    deltaNestedResource.CopyChangedValues(originalNestedResource);
+                }
+            }
```
xuzhg: can we just call "Patch(..)" method on the nested resource recursive?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +319,76 @@ public void CopyChangedValues(TStructuralType original)
...
+        }
+
+        /// <summary>
+        /// Gets the property object by the specified name.
```
xuzhg: >Gets the property object by the specified name. [](start = 12, length = 47)

resentence the comments because the method returns a bool?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +319,76 @@ public void CopyChangedValues(TStructuralType original)
...
+        /// </summary>
+        /// <param name="structural"></param>
+        /// <param name="propertyName"></param>
+        /// <param name="propertyRef">Output </param>
+        /// <returns></returns>
+        private static bool TryGetPropertyRef(TStructuralType structural, string propertyName,
+            out dynamic propertyRef)
+        {
+            propertyRef = null;
+            bool found = false;
```
xuzhg: > bool found = false; [](start = 11, length = 20)

found is necessary?  

line 388, can return "true" directly.
line 391, can return "false" directly

src/Microsoft.AspNet.OData/Microsoft.AspNet.OData.csproj
```diff
@@ -16,6 +16,7 @@
     <DefineConstants>$(DefineConstants);ASPNETODATA;ASPNETWEBAPI;NETFX;NETFX45</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.CSharp" />
```
xuzhg: >"Microsoft.CSharp" /> [](start = 23, length = 21)

why add this?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -236,15 +236,17 @@ protected override void UpdateConfiguration(WebRouteConfiguration configuration)
...
 
             HttpRequestMessage request = new HttpRequestMessage(new HttpMethod("PATCH"), requestUri);
+
+            // We should be able to PATCH nested resource with delta object of the same CLR type.
             var content = @"
 {
     'CurrentShape':
     {
         '@odata.type':'#Microsoft.Test.E2E.AspNet.OData.ComplexTypeInheritance.Circle',  
-        'Radius':2,
```
xuzhg: >2 [](start = 17, length = 1)

"2" can't work here?

test/UnitTest/Microsoft.AspNet.OData.Test/PublicApi/Microsoft.AspNet.OData.PublicApi.bsl
```diff
@@ -10,6 +10,7 @@ public interface Microsoft.AspNet.OData.IDelta {
 	void Clear ()
 	System.Collections.Generic.IEnumerable`1[[System.String]] GetChangedPropertyNames ()
 	System.Collections.Generic.IEnumerable`1[[System.String]] GetUnchangedPropertyNames ()
+	bool TryAddNestedResource (string name, object deltaNestedResource)
```
xuzhg: >bool TryAddNestedResource (string name, object deltaNestedResource) [](start = 1, length = 67)

I don't think we should add this public API.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
```diff
@@ -1,39 +1,27 @@
...
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Test.Abstraction;
+#if NETCORE
+using Microsoft.AspNetCore.Mvc;
+#endif
 using Microsoft.OData.Edm;
 using Newtonsoft.Json.Linq;
 using Xunit;
-#endif
```
xuzhg: keep the: 

#if NETCORe

#else

#endif

don't nest multiple #if #endif

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
```diff
@@ -161,6 +167,102 @@ public class OpenComplexTypeTest
...
+            // Arrange
+            string payload = @"{
+                ""Street"":""UpdatedStreet"",
+                ""Token@odata.type"":""#Guid"",
+                ""Token"":""9B198CA0-9546-4162-A4C0-14EAA255ACA7"",
+                ""BirthDay@odata.type"":""#Date"",
+                ""BirthDay"":""2016-01-29"",
+                ""LineB"": {""Description"": ""DescriptionB""}
+            }";
+            payload = Regex.Replace(payload, @"\s*", "", RegexOptions.Multiline);
```
xuzhg: >Regex. [](start = 22, length = 6)

I remember there's an email related to Regex.

I can't remember the detail. But, it seems this lib is not recommended to us.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
```diff
@@ -161,6 +167,102 @@ public class OpenComplexTypeTest
...
+                    ""PhoneInfo"" : {""ContactName"":""ContactNameA"", ""PhoneNumber"": 7654321}
+                },
+                ""LineB"": {
+                    ""PhoneInfo"": {""ContactName"": ""ContactNameB""}
+                }
+            }";
+            payload = Regex.Replace(payload, @"\s*", "", RegexOptions.Multiline);
+
+            const string requestUri = "http://localhost/odata/OpenCustomers(1)/Address";
+           HttpClient client = GetClient();
```
xuzhg: >ttpClie [](start = 12, length = 7)

align

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
```diff
@@ -161,6 +167,102 @@ public class OpenComplexTypeTest
...
+            request.Content.Headers.ContentType = MediaTypeWithQualityHeaderValue.Parse("application/json");
+            HttpResponseMessage response = await client.SendAsync(request);
+
+            // Assert
+            Assert.True(response.IsSuccessStatusCode);
+            Assert.Equal(HttpStatusCode.NoContent, response.StatusCode);
+        }
+
+        [Fact]
+        public async Task OpenComplexType_PatchNestedComplexTypeProperty_DeepNestedResourceOnNewSubNode()
```
xuzhg: >public async Task OpenComplexType_PatchNestedComplexTypeProperty_DeepNestedResourceOnNewSubNode() [](start = 8, length = 97)

The new added three UT has the similar function body. consider to create a common helper method, and call this method in three UTs?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
```diff
@@ -215,7 +317,9 @@ private static IList<OpenCustomer> CreateCustomers()
                     {
                         Street = "Street " + i,
                         City = "City " + i,
-                        DynamicProperties = new Dictionary<string, object> { { "IntProp", IntValues[i] } }
+                        DynamicProperties = new Dictionary<string, object> { { "IntProp", IntValues[i] } },
+                        // Leaving LineA as null
```
xuzhg: >// Leaving LineA as null [](start = 24, length = 24)

It's better to assign "null" if you explicitly want and add the comments after the assignment.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
```diff
@@ -368,6 +548,8 @@ public class OpenCustomer
 
     public class OpenAddress
     {
+        internal const int DefaultValue_Fee = 20;
```
xuzhg: >DefaultValue_Fee = 20; [](start = 27, length = 22)

why add in OpenAddress class? maybe it's better in the LineDetails class?

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
```diff
@@ -10,6 +10,7 @@ public interface Microsoft.AspNet.OData.IDelta {
 	void Clear ()
 	System.Collections.Generic.IEnumerable`1[[System.String]] GetChangedPropertyNames ()
 	System.Collections.Generic.IEnumerable`1[[System.String]] GetUnchangedPropertyNames ()
+	bool TryAddNestedResource (string name, object deltaNestedResource)
```
xuzhg: >bool TryAddNestedResource (string name, object deltaNestedResource) [](start = 1, length = 67)

here

xuzhg:
>         public ITestActionResult DeleteToAddress(int key)

So far, it seems great. Thanks your works.

However, Please correct me, I don't see the unit test for the following scenarios:

1. Old one is instance of base type, patch is instance of derived type, 

2. Old one is instance of derived type, patch is instance of base type.

3. Patch for Untyped scenario "EdmEntityObject, EdmComplexObject" 

Please point me or add the test cases for the above. 


---
Refers to: test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs:515 in 2d23190. [](commit_id = 2d231904d72e0487a258d8b14a08d3be75e35caf, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -434,8 +434,18 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
             string propertyName = EdmLibHelpers.GetClrPropertyName(nestedProperty, readContext.Model);
-            DeserializationHelpers.SetProperty(resource, propertyName, value);
+
+            if (readContext.IsDeltaOfT)
+            {
+                DeserializationHelpers.SetNestedResource(resource, propertyName, value);
+            }
+            else
+            {
+                DeserializationHelpers.SetProperty(resource, propertyName, value);
```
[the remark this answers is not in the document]
biaol-odata: good idea! Thanks.

---
In reply to: [190748107](https://github.com/OData/WebApi/pull/1445#discussion_r190748107) [](ancestors = 190748107)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -189,6 +190,14 @@ internal static void SetProperty(object resource, string propertyName, object va
             }
         }
 
+        internal static void SetNestedResource(object resource, string nestedResourceName, object deltaNestedResource)
```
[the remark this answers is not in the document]
biaol-odata: yup. Removed & consolidated under SetProperty.

---
In reply to: [190748206](https://github.com/OData/WebApi/pull/1445#discussion_r190748206) [](ancestors = 190748206)

src/Microsoft.AspNet.OData.Shared/Delta.cs
biaol-odata: Yup. Removed.

---
In reply to: [190748341](https://github.com/OData/WebApi/pull/1445#discussion_r190748341) [](ancestors = 190748341)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: Agreed. Changed as part of the consolidation related to SetProperty().
Now it is an internal method:
private bool TrySetNestedResourceInternal(string name, object deltaNestedResource)

---
In reply to: [190749752](https://github.com/OData/WebApi/pull/1445#discussion_r190749752) [](ancestors = 190749752,190748771)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: Fixed. The only difference is that, as I found out, some ppl prefer single point of return. I am OK with either way.

---
In reply to: [190750787](https://github.com/OData/WebApi/pull/1445#discussion_r190750787) [](ancestors = 190750787)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: Several other places use this method when recursion is involved.
// Recursion guard to avoid stack overflows
ODataResourceDeserializer.cs(107):            RuntimeHelpers.EnsureSufficientExecutionStack();


---
In reply to: [190750090](https://github.com/OData/WebApi/pull/1445#discussion_r190750090) [](ancestors = 190750090)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: "Attempts to get"

---
In reply to: [190750618](https://github.com/OData/WebApi/pull/1445#discussion_r190750618) [](ancestors = 190750618)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: Patch() is the wrapper method for CopyChangedValues, which becomes recursive for this change. Recursive calls of CopyChangedValues itself is more readable and just uses one call (vs. two calls if calling Patch) for each recursion?

---
In reply to: [190750483](https://github.com/OData/WebApi/pull/1445#discussion_r190750483) [](ancestors = 190750483)

src/Microsoft.AspNet.OData/Microsoft.AspNet.OData.csproj
biaol-odata: The way I found out to deal with generic of variable type argument is using dynamic, and VS requires Microsoft.CSharp.
[https://stackoverflow.com/questions/11417047/c-sharp-dynamic-compilation-and-microsoft-csharp-dll-error](https://stackoverflow.com/questions/11417047/c-sharp-dynamic-compilation-and-microsoft-csharp-dll-error)

---
In reply to: [190750826](https://github.com/OData/WebApi/pull/1445#discussion_r190750826) [](ancestors = 190750826)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
biaol-odata: Windows(1)'s Radius has original value is already 2 (set by WindowsController()). Patching a different value of Radius=1 so that we can verify value changed on L271.

---
In reply to: [190751057](https://github.com/OData/WebApi/pull/1445#discussion_r190751057) [](ancestors = 190751057)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
biaol-odata: OK. I thought we can save some duplicated lines of using if they are used by both.
I am OK with either way.

---
In reply to: [190751449](https://github.com/OData/WebApi/pull/1445#discussion_r190751449) [](ancestors = 190751449)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
biaol-odata: I think you are referring to regex's "catastrophic regex backtracking" problem that could cause CPU starvation and DOS. This is test code, and I don't think above pattern is backtracking pattern. So we should be able to use Regex here.

---
In reply to: [190751713](https://github.com/OData/WebApi/pull/1445#discussion_r190751713) [](ancestors = 190751713)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
AlanWong-MS: +1

xuzhg:
>         public abstract Type EntityType { get; }

as mike mentioned, it's better to modify this as "StructuredType".

---
Refers to: src/Microsoft.AspNet.OData.Shared/TypedDelta.cs:17 in 2d23190. [](commit_id = 2d231904d72e0487a258d8b14a08d3be75e35caf, deletion_comment = False)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
biaol-odata: I was thinking that DefaultValue_Fee is used by OpenAddress object's ctor to set the value of LineDetails.Fee, therefore it is kind of belong to OpenAddress class. But now think it probably makes more sense to co-locate this with UninitializedValue_Fee.
Moved.


---
In reply to: [190752494](https://github.com/OData/WebApi/pull/1445#discussion_r190752494) [](ancestors = 190752494)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs
biaol-odata: good idea. Create method private async Task ExecutePatchRequest(string payload); used by four tests.

---
In reply to: [190751964](https://github.com/OData/WebApi/pull/1445#discussion_r190751964) [](ancestors = 190751964)

xuzhg:
>         public ITestActionResult DeleteToAddress(int key)

check it #3. 

EdmEntityObject and EdmComplexObject can work as well.  For example, the controller should be as:

public ITestActionResult Patch(int key, EdmEntityObject customer)

So, please skip it.


---
In reply to: [391890910](https://github.com/OData/webapi/pull/1445#issuecomment-391890910) [](ancestors = 391890910)

---
Refers to: test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs:515 in 2d23190. [](commit_id = 2d231904d72e0487a258d8b14a08d3be75e35caf, deletion_comment = False)

biaol-odata:
>         public ITestActionResult DeleteToAddress(int key)

Good questions.
PATCH is typically of same type; and, for patching involved mismatching types, we don't allow using delta of derived type to patch object of base type to prevent unrecognizable data for the original resource of base-type (case #1 above), and it would result in HTTP400 (see PatchContainingEntity_MismatchedRuntimeTypeError).
case #2: Added PatchContainingEntity_DeltaIsBaseType.
case #3: Skipped as we have discussed since it is supported in a different way.

---
In reply to: [391890910](https://github.com/OData/WebApi/pull/1445#issuecomment-391890910) [](ancestors = 391890910)

---
Refers to: test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OpenComplexTypeTest.cs:515 in 2d23190. [](commit_id = 2d231904d72e0487a258d8b14a08d3be75e35caf, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: Added. Thanks for catching.

---
In reply to: [190748971](https://github.com/OData/WebApi/pull/1445#discussion_r190748971) [](ancestors = 190748971)

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
biaol-odata: It was changed to return object type due to we need to return Delta<> with variable type for nested resources.
Now the relevant code in TryGetPropertyValue has been further updated to using reflection to retrieve the instance value --> Restored to the original API.
Thanks.

---
In reply to: [190749966](https://github.com/OData/WebApi/pull/1445#discussion_r190749966) [](ancestors = 190749966,190749152)

xuzhg:
> }

Once we remove the public API change, why do we include this unchanged file in the commit?

---
Refers to: test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl:7 in f523948. [](commit_id = f523948e998cc36dbb927c06dea626b7943ed567, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/Delta.cs
xuzhg: It should be "undo", not comment them out.

Besides, if there's nothing change, please exclude it from the commit.

---
In reply to: [190772293](https://github.com/OData/webapi/pull/1445#discussion_r190772293) [](ancestors = 190772293,190748341)

src/Microsoft.AspNet.OData.Shared/TypedDelta.cs
```diff
@@ -6,19 +6,29 @@
...
 {
     /// <summary>
-    /// Represents a <see cref="Delta"/> that can be used when a backing CLR type exists for 
+    /// Represents a <see cref="Delta"/> that can be used when a backing CLR type exists for
     /// the entity type and complex type whose changes are tracked.
     /// </summary>
     public abstract class TypedDelta : Delta
     {
         /// <summary>
         /// Gets the actual type of the entity for which the changes are tracked.
```
xuzhg: >entity [](start = 40, length = 6)

please update the comments

src/Microsoft.AspNet.OData.Shared/TypedDelta.cs
```diff
@@ -6,19 +6,29 @@
...
+
+        /// <summary>
+        /// Helper method to check whether the given type is Delta generic type.
+        /// </summary>
+        /// <param name="type">The type to check.</param>
+        /// <returns>True if it is a Delta generic type; false otherwise.</returns>
+        internal static bool IsDeltaOfT(Type type)
+        {
+            return type != null && TypeHelper.IsGenericType(type) && type.GetGenericTypeDefinition() == typeof(Delta<>);
+        }
```
xuzhg: why add here? I remove we have similar function defined somewhere else.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -155,6 +173,32 @@ public override bool TrySetPropertyValue(string name, object value)
...
+                return false;
+            }
+
+            // Add the nested resource in the hierarchy.
+            // Note: We shouldn't add the structural properties to the <code>_changedProperties</code>, which
+            // is used for keeping track of changed non-structural properties at current level.
+            _deltaNestedResources[name] = deltaNestedResource;
+
+            return true;
+        }
```
xuzhg: move the private after the public methods

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -103,11 +109,23 @@ public override Type ExpectedClrType
...
+                return TrySetNestedResourceInternal(name, value);
+            }
+            else
+            {
+                return TrySetPropertyValueInternal(name, value);
+            }
+        }
+
+        private bool TrySetPropertyValueInternal(string name, object value)
         {
```
xuzhg: move private after public method

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -177,12 +221,30 @@ public override bool TryGetPropertyValue(string name, out object value)
...
+
+                // Get the Delta<{NestedResourceType}>._instance using Reflection.
+                FieldInfo field = deltaNestedResource.GetType().GetField("_instance", BindingFlags.NonPublic | BindingFlags.Instance);
+                Contract.Assert(field != null, "field != null");
+                value = field.GetValue(deltaNestedResource);
                 return true;
             }
+            else
+            {
+                // try to retrieve the value as property.
```
xuzhg: >as [](start = 45, length = 2)

of

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -271,18 +334,75 @@ public void CopyChangedValues(TStructuralType original)
...
                 propertyToCopy.Copy(_instance, original);
             }
 
             CopyChangedDynamicValues(original);
+
+            // For nested resources.
+            foreach (string nestedResourceName in _deltaNestedResources.Keys)
+            {
+                // Patch for each nested resource changed under this TStructuralType.
+                dynamic deltaNestedResource = _deltaNestedResources[nestedResourceName];
```
xuzhg: >dynamic [](start = 16, length = 7)

I think "object" is enough right?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -434,7 +434,10 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
 
             object value = ReadNestedResourceInline(resourceWrapper, nestedProperty.Type, readContext);
 
+            // First resolve Data member alias or annotation, then set the regular
+            // or delta resource accordingly.
```
xuzhg: Why add this comment?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceDataModels.cs
```diff
@@ -49,7 +49,7 @@ public class Rectangle : Polygon
         public Point TopLeft { get; set; }
         public int Width { get; set; }
         public int Height { get; set; }
-
```
xuzhg: remove the whitespace

xuzhg: It seems nothing changed in this file, please "undo" it.

---
In reply to: [191285099](https://github.com/OData/webapi/pull/1445#discussion_r191285099) [](ancestors = 191285099)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -264,16 +266,101 @@ protected override void UpdateConfiguration(WebRouteConfiguration configuration)
...
+    'OptionalShapes': [ ]
+}";
+            contentOfString = await ExecuteHTTP(request, content);
+
+            // Only 'Vertexes' is updated;  'HasBoarder' still has the correct value.
+            Assert.Contains("\"Vertexes\":[{\"X\":1,\"Y\":2},{\"X\":2,\"Y\":3},{\"X\":4,\"Y\":8}]", contentOfString);
+            Assert.Contains("\"HasBorder\":false", contentOfString);
+        }
+
+        private async Task<string> ExecuteHTTP(HttpRequestMessage request, string content)
```
xuzhg: >ExecuteHTTP [](start = 35, length = 11)

Maybe better to name it as " ExecuteRequest"

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceDataModels.cs
biaol-odata: Unintentional whitespace updates in related file only. I think what happen is that I made some edits and undid it later among local commits. Trailing whitespace were cleaned up automatically as part of this process. Let's include related trailing whitespace cleanup like this since we always cleanup trailing whitespace when a file is updated?

Here is a whitespace auto-cleanup tool you can use with VS in case you are interested: [https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer), feel free to let me know if you have any questions.

---
In reply to: [191285124](https://github.com/OData/WebApi/pull/1445#discussion_r191285124) [](ancestors = 191285124,191285099)

biaol-odata:
:clock1:

xuzhg:
I approved.

---
In reply to: [392881015](https://github.com/OData/webapi/pull/1445#issuecomment-392881015) [](ancestors = 392881015)

biaol-odata:
Merged. Thanks for the code review!

Issue #1101, opened 2017-10-06
Memort leak: ModelContainer keeps references to all IEdmModel ever used in $select/$expand clause 


kosinsky:
ModelContainer (https://github.com/OData/WebApi/blob/368278e40296e87a420f1bbffef038c3c4a852d2/OData/src/System.Web.OData/OData/Query/Expressions/ModelContainer.cs) used to get IEdmModel by ModelID for SelectExpandWrapper. And It saves in external static dictionary all IEdmModel that were ever used. 

### Assemblies affected

All

### Reproduce steps

* Create OData endpoint that creates instances of IEdmModel dynamically on each call to the service
* Access endpoint and take memory snapshot
* Access endpoint with $select or $expand clause multiple times and take memory snapshot

### Expected result

*Only one (last) copy of IEdmModel is part of snapshot*

### Actual result

*Number of IEdmModels in snapshot is equal to number of calls with $select/$expand*

### Additional detail

*Second problem with ModelContainer is that it adds ModelID as literal (not a parameter) to each query that contains $select/$expand => for SQL Server these queries are different => recompiles (affect perf) and they aren’t kept in Query Data Store (not easy to trek regressions).*

brjohnstmsft:
My team (Azure Search) is also affected by this issue.

PR #1462, opened 2018-06-05
Addressing memory leak in ModelContainer

AlanWong-MS:
### Issues
#1101

### Description
Removing model container and maintaining only one Edm model structure at a time.

- [x] *Test cases added*


AlanWong-MS:
The build was using an old build definition for some reason. The build and tests pass.

biaol-odata:
E2E tests failed. SelectExpandTests.QueryForAnEntryWithExpandNavigationPropertyExceedPageSize doesn't like a build server issue.
What is the reason for "E2E Tests except QueryForLongSelectList", "E2E Tests only QueryForLongSelectList"? Can you delete your old build def so that the PR trigger build will generate a new build def for use?

AlanWong-MS:
Looks like the errors still persist so I'll take a look at those first. Please hold off on reviewing.

AlanWong-MS:
Closing in favor of #1506.

PR #1466, opened 2018-06-07
Fixed cloning the ODataDeserializerContext

khodaie:
### Description

When cloning of the ODataDeserializerContext in the ODataResourceDeserializer, the Request and RequestContext properties should be set.

- [ ] *Test cases added*


xuzhg:
@khodaie Would you please share us more about your usage? Or Would you please file an issue to illustrate why do you want to set these two properties? By the way, your change can't work for both Classic and Core version. Would you please update your PR?

khodaie:
@xuzhg I have a customized `ODataResourceDeserilizer`  and I need to access the current request properties in the overridden `ReadEntry` method. I used the `readContext.Request.Properties`, but the `NullReferenceException` was thrown. (`readContext.Request` is `null`)

pdbruno:
Hello! I found the exact same issue. It is also happening in the method ReadNestedResourceSetInline(). Could you add this fix in the commit?

Issue #1473, opened 2018-06-11
Comparing Date or DateTime properties with date constant.


khodaie:
Comparing Date or DateTime properties with date constants throws the following exception:
> No coercion operator is defined between types 'Microsoft.OData.Edm.Date' and 'System.Nullable`1[System.DateTimeOffset]'.

### Assemblies affected

*Microsoft.AspNet.OData, Version=7.0.0.20530

### Reproduce steps

1. Declare an entity with `Date` or `DateTime` SQL type. (for example: `Operation` with `EffectiveDate` property of type `DateTime` and `Date` SQL type)
2. Create OData endpoints for the entity.
3. Execute this OData query: `~/Operations?$filter=EffectiveDate eq 2018-06-01`

### Expected result

The sample query should return HTTP response `200 OK`.

### Actual result

The HTTP response is `400 Bad Request` with this exception:

    The query specified in the URI is not valid. No coercion operator is defined between types 'Microsoft.OData.Edm.Date' and 'System.Nullable`1[System.DateTimeOffset]'.

### Additional detail

The `Microsoft.AspNet.OData.Query.Expressions.ExpressionBinderBase.CreateConvertExpression(ConvertNode convertNode, Expression source)` is responsible for this exception. Specifying the `Date` to `DateTimeOffset` conversion will fix it.

khodaie:
Proposed fix: [https://github.com/OData/WebApi/pull/1474](url)

xuzhg:
@khodaie Thank you for filing issue and sending us PR to fix. However, I am wondering what's your data source or DB?

khodaie:
@xuzhg I created a sample data source:

``` C#
    public class Operation
    {
        [Key]
        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
        public int Id { get; set; }

        public DateTime? OperationDate { get; set; }

        public DateTimeOffset? CreatedAt { get; set; }

        public TimeSpan? Duration { get; set; }
    }

    public class OperationContext : DbContext
    {
        public virtual DbSet<Operation> Operations { get; set; }
    }
```

The controller class:

``` C#
    public class OperationsController : ODataController
    {
        private readonly OperationContext _db = new OperationContext();

        [EnableQuery]
        public IQueryable<Operation> Get() => this._db.Operations;

        [EnableQuery]
        public SingleResult<Operation> Get([FromODataUri] int key) => SingleResult.Create(this._db.Operations.Where(p => p.Id == key));

[...]

PR #1474, opened 2018-06-11
Handling filtering with the Edm.Date values

khodaie:
To fix a bug when a Date or DateTime property is compared to a date constant. For example:
~/Operations?$filter=EffectiveDate eq 2018-06-11
Without this change, the above code throws the following exception:

> No coercion operator is defined between types 'Microsoft.OData.Edm.Date' and 'System.Nullable`1[System.DateTimeOffset]'.

### Issues

*This pull request fixes issue #1473.*

### Description

*Comparing the Date or DateTime fields with date constants needs to a conversion. This patch handles the conversion*

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -243,6 +243,10 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             {
                 return source;
             }
+            else if(conversionType == typeof(DateTimeOffset?) && source.Type == typeof(Date))
+            {
+                return source;
+            }
             else if (source == NullConstant)
             {
                 return source;


xuzhg:
@khodaie 

We have the following process:

|Conversion Type|Source Type|
|:--|:--|

khodaie:
@xuzhg 
I think the conversion from DateTime, or DateTimeOffset to TimeOfDay can be handled by the OData `time()` function, isn't it?

xuzhg:
@khodaie Yes. But I am confused. We also have OData `date()` function. Why can't use it in your scenario?

khodaie:
@xuzhg
I have a large existing project that is made based on the Microsoft.AspNet.OData 5.11; In that version there was no problem with comparing `DateTimeOffset` with `Date`. But, I had to upgrade because of not supporting the `group by` clause on the complex types(SQL function results) in 5.11. 
You know that changing all occurrences of this situation in all project is a really time-consuming process.

xuzhg:
@khodaie Yes. in case I am wrong. I remember that in OData.lib (Microsoft.OData.Core), we had a bug fix about "2018-06-13" de-serializing.  In the old version, "2018-06-13" is de-serialized as `DateTimeOffset` object. So, it can work at your scenario. In the new version, "2018-06-03" is de-serialized as `Date` object. When you update to new version, you should use `date()` or `time()` function.

xuzhg:
@khodaie Thank for your contribution. However, as mentioned above, it's recommendation to use `data() or `time()` built-in function. So, close it.

Issue #1407, opened 2018-04-26
BaseAddressFactory in ODataMediaTypeFormatter is not respected by GetPerRequestFormatterInstance


RalfSeidel:
With issue #645 the ODataMediaTypeFormatter got a new property that allows to override the base address for the service root. However by default the WebApi uses the IContentNegotiator "PerRequestContentNegotiator" which in turn calls  GetPerRequestFormatterInstance to create request specific instances of the formatter. The property BaseAddressFactory is not copied. Thus setting this property has no effect.

### Assemblies affected

Tested this System.Web.OData 6.1.0.10907

### Reproduce steps

Change the BaseAddressFactory of all ODataMediaTypeFormatters after the WebApi was innitialized e.g. like this

`			ServicesContainer services = GlobalConfiguration.Configuration.Services;
			IHttpControllerSelector controllerSelector = services.GetHttpControllerSelector();
			IEnumerable<HttpControllerDescriptor> controllers = controllerSelector.GetControllerMapping().Values;

			foreach ( var c in controllers ) {
				var odataFormatter = c.Configuration.Formatters.OfType<ODataMediaTypeFormatter>();
				foreach ( var f in odataFormatter ) {
					f.BaseAddressFactory = ( m ) => new Uri( "http://foo.bar/", UriKind.Absolute );
				}
			}
`
Get data from any odata controller

### Expected result

The Result should contain the modified service root url i.e. something like this:

"@context": "http://foo.bar/odata/$metadata",

### Actual result

The result contains the default url i.e. somthing like this

"@context": "http://localhost:54725/odata/$metadata",

see #645

xuzhg:
See PR #1483

xuzhg:
Merged. Thanks.

PR #1483, opened 2018-06-13
fix the BaseAddressFactory in ODataMediaTypeFormatter

xuzhg:
### Issues

* #1407 

### Description

* In Web API OData classic class "ODataMediaTypeFormatter", BaseAddressFactory property should be copy to the internal created formatter object.

This PR try to fix that and add a test to verify it.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
+++ b/src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
@@ -83,6 +83,9 @@ namespace Microsoft.AspNet.OData.Formatter
 
             // Parameter 3: request
             Request = request;
+
+            // BaseAddressFactory
+            BaseAddressFactory = formatter.BaseAddressFactory;
         }
 
         /// <summary>
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Microsoft.Test.E2E.AspNet.OData.csproj
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Microsoft.Test.E2E.AspNet.OData.csproj
@@ -1135,6 +1135,9 @@
     <Compile Include="..\ForeignKey\ForeignKeyTest.cs">
       <Link>ForeignKey\ForeignKeyTest.cs</Link>
     </Compile>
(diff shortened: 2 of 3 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Microsoft.Test.E2E.AspNet.OData.csproj
```diff
@@ -1134,6 +1134,9 @@
     </Compile>
     <Compile Include="..\ForeignKey\ForeignKeyTest.cs">
       <Link>ForeignKey\ForeignKeyTest.cs</Link>
+    </Compile>
+	<Compile Include="..\Formatter\BaseAddressFactoryTest.cs">
```
biaol-odata: Convert tab to spaces.

xuzhg: Thanks. #Resolved.

src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
```diff
@@ -117,7 +117,10 @@ public override MediaTypeFormatter GetPerRequestFormatterInstance(Type type, Htt
             else
             {
                 ODataVersion version = GetODataResponseVersion(request);
-                return new ODataMediaTypeFormatter(this, version, request);
+                return new ODataMediaTypeFormatter(this, version, request)
+                {
+                    BaseAddressFactory = this.BaseAddressFactory
```
biaol-odata: Should we let the copy cstor set this property instead of explicitly setting it here, so that when we add other copy cstor calls this is automatically included?

biaol-odata:
Looks like to me ODataMediaTypeFormatter is only applicable for classic side. If this is not the case, please check it on AspNetCore side as well. Thanks.

xuzhg:
ODataMediaTypeFormatter is only existing in classic.

biaol-odata:
:shipit:

Mortana89:
Is this also pushed for the Core side?
Any example of usage?

PR #1489, opened 2018-06-19
Change the OptionalReturn to ReturnNullable in OperationConfiguration

xuzhg:
### Issues

This is the continue on for PR #1353.

### Description

Operation (function/action) currently have a property named "OptionalReturn" that returns whether or not the return type is nullable. 

The existing property named "OptionalReturn" on the operation that really means "Nullable" is misleading, and makes it hard to expose true optional parameters through WebAPI OData.*

- [x] *Test cases added*

### Additional work necessary

1353


xuzhg:
@AlanWong-MS updated the description. Thanks.

Issue #1488, opened 2018-06-18
Add support for optional parameters to WebAPI OData stack


mikepizzo:
*We have added support for optional parameters to OData Library.  This work item is to track exposing that support through the WebAPI OData stack.*

### Assemblies affected

OData WebApi lib 6.x

### Reproduce steps

See https://github.com/OData/odata.net/issues/782 for a good example.

### Additional detail

WebAPI OData 6.x and before has an "Optional" property on a parameter; however, it really means nullable.  In WebAPI OData 7.x we renamed this property to "Nullable" so that we could introduce an actual optional property.

DavTheSuave:
Any clue how far off into the horizon this is? I don't need it for my current project but it would be really nice to have.

xuzhg:
@DavTheSuave We implemented it in the latest [7.0.1](https://www.nuget.org/packages/Microsoft.AspNetCore.OData/7.0.1) version. Thanks.

PR #1491, opened 2018-06-19
Add the optional parameter into the model builder

xuzhg:
### Issues

* https://github.com/OData/WebApi/issues/1488

### Description

* Add the optional parameter into the model builder

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
@@ -55,12 +55,12 @@ namespace Microsoft.AspNet.OData.Builder
         /// <summary>
         /// Gets or sets a value indicating whether this parameter is optional or not.
         /// </summary>
-        public bool IsOptional { get; set; }
+        public bool IsOptional { get; protected set; }
 
         /// <summary>
         /// Gets or sets a default value for optional parameter.
         /// </summary>
-        public string DefaultValue { get; set; }
+        public string DefaultValue { get; protected set; }
 
         /// <summary>
         /// Sets the optional value as true.
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
@@ -51,5 +51,44 @@ namespace Microsoft.AspNet.OData.Builder
         /// Gets or sets a value indicating whether this parameter is nullable or not.
         /// </summary>
         public bool Nullable { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this parameter is optional or not.
+        /// </summary>
+        public bool Optional { get; set; }
+
+        /// <summary>
+        /// Gets or sets a default value for optional parameter.
+        /// </summary>
+        public string DefaultValue { get; set; }
+
+        /// <summary>
+        /// Sets the optional value as true.
+        /// </summary>
+        public ParameterConfiguration IsOptional()
+        {
+            Optional = true;
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the optional value as false.
+        /// </summary>
+        public ParameterConfiguration IsRequired()
+        {
+            Optional = false;
+            return this;
+        }
+
+        /// <summary>
+        ///  Sets the optional value as true, default value as given value.
+        /// </summary>
+        /// <param name="defaultValue">The default value.</param>
+        public ParameterConfiguration HasDefaultValue(string defaultValue)
+        {
+            Optional = true;
+            DefaultValue = defaultValue;
+            return this;
+        }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
@@ -55,7 +55,7 @@ namespace Microsoft.AspNet.OData.Builder
         /// <summary>
         /// Gets or sets a value indicating whether this parameter is optional or not.
         /// </summary>
-        public bool Optional { get; set; }
+        public bool IsOptional { get; set; }
 
         /// <summary>
         /// Gets or sets a default value for optional parameter.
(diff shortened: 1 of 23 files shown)


src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -51,5 +51,44 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
...
+
+        /// <summary>
+        /// Gets or sets a default value for optional parameter.
+        /// </summary>
+        public string DefaultValue { get; set; }
+
+        /// <summary>
+        /// Sets the optional value as true.
+        /// </summary>
+        public ParameterConfiguration IsOptional()
```
AlanWong-MS: I think "SetOptional" would be more clear for the name because when I see "IsOptional", I expect it to be returning a boolean and not performing any updates.

I also think we should make the Optional.set private because we have this custom function to set the property value.

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -51,5 +51,44 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
...
+        {
+            Optional = true;
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the optional value as false.
+        /// </summary>
+        public ParameterConfiguration IsRequired()
```
AlanWong-MS: Same comment for IsRequired naming: I think something like SetRequired would be more clear.

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -51,5 +51,44 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
...
+        {
+            Optional = false;
+            return this;
+        }
+
+        /// <summary>
+        ///  Sets the optional value as true, default value as given value.
+        /// </summary>
+        /// <param name="defaultValue">The default value.</param>
+        public ParameterConfiguration HasDefaultValue(string defaultValue)
```
AlanWong-MS: I think SetDefaultValue might be more clear. Please feel free to ask the team for their opinion as well.

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -51,5 +51,44 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
...
+        /// Sets the optional value as false.
+        /// </summary>
+        public ParameterConfiguration IsRequired()
+        {
+            Optional = false;
+            return this;
+        }
+
+        /// <summary>
+        ///  Sets the optional value as true, default value as given value.
```
AlanWong-MS: nit: extra space after "///" and before "Sets"

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
AlanWong-MS: Also, should we replace the DefaultValue set method with this function? Otherwise, should the DefaultValue set method be private?

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
xuzhg: Thanks. 
I think it's better to switch both.
the property name should be "IsOptional {get, private set;}, while the fluent API should be "Option()", "Required()".  I think that's Web API OData fluent API naming convention. You maybe find a lot of such fluent APIs defined in the model part.
I will change it.

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
xuzhg: That's the Web API OData naming convention, I think. for example: there are a lot of fluent api names:

1. HasActionLink
2. HasFunctionLink
3. HasManyPath
4. HasSingleton
….

AlanWong-MS: Got it, thanks for the clarification

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
AlanWong-MS: :+1:

xuzhg: #Resolved.

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
xuzhg: #Resolved.

xuzhg:
@mikepizzo @AlanWong-MS @biaol-odata @raheph This PR is ready to review. Thanks.

xuzhg:
@mikepizzo @AlanWong-MS @biaol-odata is there any other comments from your side?

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -51,5 +51,44 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
         /// Gets or sets a value indicating whether this parameter is nullable or not.
         /// </summary>
         public bool Nullable { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this parameter is optional or not.
+        /// </summary>
+        public bool IsOptional { get; set; }
```
AlanWong-MS: Should we do a private set? It is possible to call HasDefaultValue and then set IsOptional to false. I don't think it causes any serious issues but I would be confused by the API.

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -51,5 +51,44 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
...
+
+        /// <summary>
+        /// Gets or sets a default value for optional parameter.
+        /// </summary>
+        public string DefaultValue { get; set; }
```
AlanWong-MS: Should we do a private set? HasDefaultValue performs the set for us and couples it with setting IsOptional = true.

xuzhg: Yes. Thanks. I changed it to use "protected set".

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
xuzhg: Yes. Thanks. I changed it to use "protected set".

Issue #1387, opened 2018-04-14
[feature/netcore] $select executes on client side


vsopko:
After upgrading from Microsoft.AspNetCore.OData Version=7.0.0-beta1 to Microsoft.AspNetCore.OData Version=7.0.0-beta2 $select gets all entity properties from database and produces correct result on client side. This also occurs in nightly builds.

### Assemblies affected

Microsoft.AspNetCore.OData Version=7.0.0-beta2
Microsoft.AspNetCore.OData Version=7.0.0-Nightly*

### Reproduce steps

Url: `http://localhost:61734/api/values/?$select=Age&$top=1`

Controller:
```
[Route("api/[controller]")]
public class ValuesController : Controller
{
    private readonly Context _ctx;

    public ValuesController(Context ctx)
    {
        _ctx = ctx;
    }

    [HttpGet]
    public async Task<IActionResult> Get(ODataQueryOptions<User> queryOptions)
    {
        var users = _ctx.Users.Select(x => x);
        return Ok(queryOptions.ApplyTo(users));
    }
}
```
Context model:
```
public class User
{
    [Key]
    public int Id { get; set; }
    [Required]
    public string UserName { get; set; }
    public string FirstName { get; set; }
    public string LastName { get; set; }
    public int Age { get; set; }
    public virtual ICollection<UserRole> UserRoles { get; set; }
}
```

### Expected result
Return value:
[{"Age": 50}]

Sql profiler:
exec sp_executesql N'SELECT TOP(@__TypedProperty_0)
[x].[Age] AS [Value]
FROM [User] AS [x] N'@__TypedProperty_0 int',@__TypedProperty_0=1
### Actual result
Return value:
[{"Age": 50}]

Sql profiler:
exec sp_executesql N'SELECT TOP(@__TypedProperty_0)
**[x].[Id],**
[x].[Age],
**[x].[FirstName],
[x].[LastName],
[x].[UserName]**
FROM [User] AS [x] N'@__TypedProperty_0 int',@__TypedProperty_0=1

robward-ms:
@vsopko - What backend DB are you using?

vsopko:
@robward-ms Microsoft SQL Server Express (64-bit) 14.0.1000.169 on Windows 10 Pro

robward-ms:
@vsopko - What is ".Select(x => x);" doing for you? I think this might be forcing the evaluation of _ctx.Users before the application of the query.

vsopko:
[...]

PR #1493, opened 2018-06-20
Only add Instance for the select all binder for perf

xuzhg:
### Issues

*  #1387.*

### Description

* Without the changes, when we use $select=xxx, the SQL statement will retrieve all the properties, that's not expected.
* With this change, when we use $select=xxx, only the xxx property will be retrieved.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandBinderTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandBinderTest.cs
@@ -78,7 +78,8 @@ namespace Microsoft.AspNet.OData.Test.Query.Expressions
             Assert.True(enumerator.MoveNext());
             var partialCustomer = Assert.IsAssignableFrom<SelectExpandWrapper<Customer>>(enumerator.Current);
             Assert.False(enumerator.MoveNext());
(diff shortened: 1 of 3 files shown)


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -344,9 +345,11 @@ public void ProjectAsWrapper_Element_ProjectedValueDoesNotContainInstance_IfSele
 
             // Assert
             Assert.Equal(ExpressionType.MemberInit, projection.NodeType);
-            Assert.NotEmpty((projection as MemberInitExpression).Bindings.Where(p => p.Member.Name == "Instance"));
+            Assert.Empty((projection as MemberInitExpression).Bindings.Where(p => p.Member.Name == "Instance"));
+            Assert.NotEmpty((projection as MemberInitExpression).Bindings.Where(p => p.Member.Name == "InstanceType"));
```
AlanWong-MS: Is it possible to verify the properties retrieved at ID and Orders? (From `$select=ID,Orders`)

AlanWong-MS: Or if the change has been verified in debugger, then that's fine.

Issue #1498, opened 2018-06-25
ASP.NET Core 2.1 - Task-returning methods do not return OData envelope


aelij:
After upgrading to ASP.NET Core 2.1 (from 2.0), the OData endpoints do not return the envelope for task-returning methods.

### Assemblies affected

* OData WebApi 7 Beta 4

### Reproduce steps

* Create a task-returning method in an OData controller:
```csharp
[HttpGet, ODataRoute("X")]
public Task<X> GetX() { return ...; }
```

### Expected result
Returned JSON:

```js
{
    "@odata.context": "http://localhost:30006/api/odata/$metadata#Microsoft.Foo.Server.Provider.Model.X",
    "Field1": "xxxxx",
    "Field2": "xxxxxx"
}
```

### Actual result

```js
{
    "Field1": "xxxxx",
    "Field2": "xxxxxx"
}
```

aelij:
cc: @davidmorissette

aelij:
FYI I've verified this is fixed in ASP.NET 2.1.1: https://github.com/aspnet/Mvc/pull/7834

xuzhg:
@aelij Thank for your information. However, i think we should:
1. do what i have done in the PR #1501 to make sure it can work in 2.0 and 2.1 version.
2. upgrade to v2.1?

JohnGalt1717:
I'm using .NET Core 2.1.2 (latest) and the release version of oData for .NET core.

Using the documentation, I've set it up but the output is always simple json.

Adding these mvc options:  

options.RespectBrowserAcceptHeader = true;
options.ReturnHttpNotAcceptable = true;

results in a 406 error when requesting odata formats as if it isn't registering the formatters properly.

PR #1501, opened 2018-06-26
Return Task<> from method of controller doesn't pick up odata output formatter

xuzhg:
### Issues

*This pull request fixes issue #1498.*

### Description

* In ASP.NET Core version, the type in `CanWriteResult` is the `Task<T>`, it can't pass the `ODataOutputFormatter` 's verification.

* While, in Classic version, the `type` in `CanWriteType(Type)` of `ODataMediaTypeFormatter` is the T of Take<T>.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
@@ -456,6 +456,16 @@ namespace Microsoft.AspNet.OData
             return result;
         }
 
+        internal static Type GetTaskInnerTypeOrSelf(Type type)
+        {
+            if (IsGenericType(type) && type.GetGenericTypeDefinition() == typeof(Task<>))
+            {
+                return type.GetGenericArguments().First();
+            }
+
+            return type;
+        }
+
         private static Type GetInnerGenericType(Type interfaceType)
         {
             // Getting the type T definition if the returning type implements IEnumerable<T>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -456,6 +456,16 @@ internal static IEnumerable<Type> GetLoadedTypes(IWebApiAssembliesResolver assem
             return result;
         }
 
+        internal static Type GetTaskInnerTypeOrSelf(Type type)
+        {
```
aelij: I think this doesn't cover all cases. See [here](https://github.com/aspnet/Mvc/blob/624a5ed522e8e32a97fb1ba61e25d2302a5abfed/src/Microsoft.AspNetCore.Mvc.Core/Internal/ActionMethodExecutor.cs) (for example, `ValueTask`). Perhaps you could use the [same method](https://github.com/aspnet/Common/blob/57fdf3b2206edda51d7481d036c26fea16578100/shared/Microsoft.Extensions.ObjectMethodExecutor.Sources/CoercedAwaitableInfo.cs#L30) to infer the awaitable type?

xuzhg: @aelij Thanks. So far, I think it's enough. 

`ValueTask` struct is added in C# 7.0, it's shipped in "System.Threading.Tasks.Extensions", while this package is not referenced in our products. 

by the way, CoercedAwaitableInfo or AwaitableInfo is internal class.

Issue #1503, opened 2018-06-26
Enable support for batch in Json format in WebApi V7


biaol-odata:
*ODL 7.4.4+ supports [batch in Json format](http://docs.oasis-open.org/odata/odata-json-format/v4.01/csprd04/odata-json-format-v4.01-csprd04.html#_Toc499716905), and it needs to be enabled in WebApi layer.*

### Assemblies affected

*OData WebApi lib 7.0*

### Reproduce steps

*N/A(new feature).*

### Expected result

*N/A(new feature).*

### Actual result

*N/A(new feature).*

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

biaol-odata:
Close as merged.

PR #1504, opened 2018-06-26
Enable support for Json Batch on WebApi layer

biaol-odata:
### Issues

*This pull request fixes issue #1503.*

### Description

*Add support for application/json media type so that batch processing flow for json format can be initiated.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Common/SRResources.resx
+++ b/src/Microsoft.AspNet.OData.Shared/Common/SRResources.resx
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
     Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
-    various data types are done through the TypeConverter classes 
+
+    The primary goals of this format is to allow a simple XML format
+    that is mostly human readable. The generation and parsing of the
+    various data types are done through the TypeConverter classes
     associated with the data types.
-    
+
     Example:
-    
+
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.OData/Batch/ODataBatchHttpRequestMessageExtensions.cs
```diff
@@ -217,24 +219,35 @@ internal static Task CreateODataBatchResponseAsync(this HttpRequest request, IEn
...
                 await response.WriteAsync(SRResources.BatchRequestMissingContentType);
                 return false;
             }
-            if (!String.Equals(contentType.MediaType.ToString(), BatchMediaType, StringComparison.OrdinalIgnoreCase))
+
+            string mediaType = contentType.MediaType.ToString();
+            bool isMimeBatch = String.Equals(mediaType, BatchMediaTypeMime, StringComparison.OrdinalIgnoreCase);
+            bool isJsonBatch = String.Equals(mediaType, BatchMediaTypeJson, StringComparison.OrdinalIgnoreCase);
+
+            if ( !isMimeBatch && !isJsonBatch)
```
AlanWong-MS: nit: extra space between ( and !isMimeBatch

src/Microsoft.AspNet.OData.Shared/Common/SRResources.resx
```diff
@@ -548,7 +548,7 @@
     <value>The property 'EdmObject' of {0} cannot be null.</value>
   </data>
   <data name="BatchRequestInvalidMediaType" xml:space="preserve">
-    <value>The batch request must have '{0}' as the media type.</value>
+    <value>The batch request must have '{0}' or '{1}' as the media type.</value>
```
AlanWong-MS: Should we just hardcode the types?

biaol-odata: I think it should be good to keep the message and data separate.

---
In reply to: [198253383](https://github.com/OData/WebApi/pull/1504#discussion_r198253383) [](ancestors = 198253383)

Issue #1434, opened 2018-05-04
OData version-Header missing on Delete


wurst0815:
According the the [specification](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part1-protocol/odata-v4.0-errata03-os-part1-protocol-complete.html#_Toc453752219), an OData Service must respond with the OData Version Header.

But this Header is missing on Delete.
![image](https://user-images.githubusercontent.com/13557494/39618103-6a4f2d8a-4f82-11e8-8db1-3a8cbca48344.png)

Microsoft.AspNet.WebAPI.ODATA Version 5.7.0
Microsoft.Data.Odata Version 5.8.3

xuzhg:
@wurst0815 the link you provided for the spec is targeting to the OData spec v4. However, the libraries you're using are targeting to OData spec v3. Would you please take a look?  Besides, we are working on the OData spec v4 and its .NET implementation, meanwhile the spec V3 and its implementations are in maintenance status.

wurst0815:
Hi,
you are right, the used libraries are 
Microsoft.AspNet.OData  v6.1.0
Microsoft.OData.Core 7.4.4

The issue seems to affect all actions that respond with an 'No Content' payload, not only DELETE actions.

PR #1509, opened 2018-06-28
issue #1434: add the OData-Version into the No-Content response

xuzhg:
### Issues

* https://github.com/OData/WebApi/issues/1434

### Description

* Add the `OData-Version` into No-Content response.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Results/ResultHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/ResultHelpers.cs
@@ -7,7 +7,9 @@ using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter;
 using Microsoft.AspNet.OData.Formatter.Serialization;
+using Microsoft.AspNet.OData.Routing;
 using Microsoft.AspNetCore.Http;
+using Microsoft.OData;
 using Microsoft.OData.Edm;
 using ODataPath = Microsoft.AspNet.OData.Routing.ODataPath;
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.OData/Results/ResultHelpers.cs
```diff
@@ -59,5 +60,18 @@ public static void AddEntityId(HttpResponse response, Func<Uri> entityId)
                 response.Headers.Add(EntityIdHeaderName, entityId().ToString());
             }
         }
+
+        public static void AddServiceVersion(HttpResponse response, Func<string> version)
```
biaol-odata: Rename to "AddServiceVersionForNoContentResponse" to indicate that it only handle the case of NoContent response code.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -112,6 +112,8 @@ protected override void UpdateConfiguration(WebRouteConfiguration configuration)
                 response.StatusCode,
                 requestUri,
                 contentOfString));
+
+            Assert.Equal("4.0", response.Headers.GetValues("OData-Version").Single());
```
biaol-odata: This is a POST request with response code 201, so it is not related to the case of 204 here, right?

src/Microsoft.AspNetCore.OData/Results/ResultHelpers.cs
xuzhg: Thanks. That's same syntax of the above method “AddEntityId". 
Besides, that's a public method in an internal class. It's not mandatory to add "comments" in an internal class.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
xuzhg: Thanks. That's not added from my change. That's added in the OData Formatter, for example in Core version, it's added from "https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs#L183"

I added this code to verify the header.

xuzhg:
All tests passed. @mikepizzo @AlanWong-MS @biaol-odata

Issue #1518, opened 2018-07-05
Microsoft.AspNet.OData 7.0.0 Has Undeclared NuGet Dependencies


commonsensesoftware:
Version 7.0.0 of the **Microsoft.AspNet.OData** package has undeclared NuGet dependencies. The offending dependency is at least **Newtonsoft.Json**, but there may be others.

For reasons that are unclear, the [source project](../blob/master/src/Microsoft.AspNet.OData/Microsoft.AspNet.OData.csproj#L40) references **Newtonsoft.Json 10.0.0**, but this reference is not declared in the [\*.nuspec](../blob/master/src/Microsoft.AspNet.OData/Microsoft.AspNet.OData.Release.nuspec
).

### Assemblies affected

Microsoft.AspNet.OData 7.0.0

### Reproduce steps

```powershell
Install-Package Microsoft.AspNet.OData -Version 7.0.0
```

### Expected result

Successful compilation without warnings.

### Actual result

**MSB3277** - _Found conflicts between different versions of "Newtonsoft.Json" that could not be resolved._

### Additional detail

**Newtonsoft.Json** is a transitive dependency. It's unclear if the latest OData package actually requires a newer version of the package. If it doesn't, then the project should reference **6.0.4**, which is the lowest version required by the lowest, supported version of ASP.NET Web API. If a higher version is required, then it needs to be explicitly defined in the \*.nuspec file.

I would also **strongly** recommend converting the source project to the new _SDK_ format just like the **Microsoft.AspNetCore.OData** project. This will avoid this type of issue in the future.

xuzhg:
@commonsensesoftware I filed a PR to downgrade the reference version. See #1519.

commonsensesoftware:
Great! Thanks for the quick turnaround. Pending the test run, the changes looks good to me.

xuzhg:
Merged. Thanks.

PR #1519, opened 2018-07-05
downgrade the newtonsoft.json version to 6.0.4 for classic

xuzhg:
### Issues

*This pull request fixes issue https://github.com/OData/WebApi/issues/1518.*

### Description

Microsoft.AspNet.OData has a reference to Microsoft.AspNet.WebApi.Client package, which has a reference to newtonsoft.json as version >= 6.0.4

image

However, Microsoft.AspNet.OData itself has a reference to newtonsoft.json as version = 10.0.3.

So, there is a conflict. We should downgrade the reference to 6.0.4.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/samples/AspNetODataSample.Web/Web.config
+++ b/samples/AspNetODataSample.Web/Web.config
@@ -44,10 +44,6 @@
         <bindingRedirect oldVersion="1.0.0.0-3.0.0.0" newVersion="3.0.0.0" />
       </dependentAssembly>
       <dependentAssembly>
-        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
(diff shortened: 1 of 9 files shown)


robertmclaws:
Out of curiosity, what was the reason for downgrading this library vs upgrading Microsoft.AspNet.WebApi.Client? Version 10 is *significantly* faster than 6, and adding a specific reference to version 10 in either project should not break anything. The problem was that the code was built with 10, but the 10 reference was not in the NuSpec file. Adding the 10 reference to the NuSpec file should have fixed it.

xuzhg:
@robertmclaws We should make the package working at customer side who are still using the lower version of Microsoft.AspNet.WebApi.Client.  Besides, It's easy to update the reference at customer side without any conflict error/warning message.

PR #1527, opened 2018-07-09
Add UseOData() extension methods for IApplicationBuilder

xuzhg:
### Issues

* Add UseOData() extension methods for IApplicationBuilder. (Only for ASP.NET Core)

### Description

* 
We have `AddOData()` to add the service, but it shows that we should call `UseMvc(..)` and set up an action to register the OData route. This PR to combine it into an extension method `UseOData()` for IApplicationBuilder.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
@@ -4,6 +4,7 @@
 using Microsoft.AspNet.OData.Batch;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.OData.Edm;
 
 namespace Microsoft.AspNet.OData.Extensions
 {
@@ -12,8 +13,20 @@ namespace Microsoft.AspNet.OData.Extensions
     /// </summary>
     public static class ODataApplicationBuilderExtensions
     {
+        private static readonly int MaxTop = 100;
+
         /// <summary>
-        /// USe OData batching middleware.
+        /// The default OData route name.
+        /// </summary>
+        public readonly static string DefaultRouteName = "odata";
+
+        /// <summary>
+        /// The default OData route prefix.
+        /// </summary>
+        public readonly static string DefaultRoutePrefix = "odata";
+
+        /// <summary>
+        /// Use OData batching middleware.
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder "/> to use.</param>
         /// <returns>The <see cref="IApplicationBuilder "/>.</returns>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
```diff
@@ -26,5 +37,50 @@ public static IApplicationBuilder UseODataBatching(this IApplicationBuilder app)
...
+                b.MapODataServiceRoute(routeName, routePrerix, model);
+            });
+        }
+
+        private static void VerifyODataIsRegistered(IApplicationBuilder app)
+        {
+            // We use the IPerRouteContainer to verify if AddOData() was called before calling UseOData
+            if (app.ApplicationServices.GetService(typeof(IPerRouteContainer)) == null)
+            {
+                throw Error.InvalidOperation(SRResources.MissingODataServices, "OData services");
```
biaol-odata: change to "nameof(IPerRouteContainer)"?

xuzhg: As the comment (line 79) saying, I use "IPerRouteContainer" to test the OData services, not only the IPerRouteContainer. So, I input the "OData Services".

biaol-odata: I understand that the code is using IPerRouteCotnainer to check whether OData Services has been setup properly. But the usage of "OData Services" string is duplicated to what is already contained in MissingOdataServices, which already says 
```
"Cannot find '{0}'. OData services have not been configured. Are you missing a call to AddOData()?"
```

xuzhg:
@mikepizzo @AlanWong-MS @biaol-odata is there any other comments from your side?

src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
```diff
@@ -26,5 +37,50 @@ public static IApplicationBuilder UseODataBatching(this IApplicationBuilder app)
...
+            if (app == null)
+            {
+                throw Error.ArgumentNull("app");
+            }
+
+            VerifyODataIsRegistered(app);
+
+            return app.UseMvc(b =>
+            {
+                b.Select().Expand().Filter().OrderBy().MaxTop(100).Count();
```
AlanWong-MS: nit: is there a default top variable used somewhere instead of hardcoding 100?

src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
AlanWong-MS: +1, if we can use SRResources.MissingODataServices, that would be great

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/ODataApplicationBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,46 @@
...
+            // Arrange
+            var applicationBuilderMock = new Mock<IApplicationBuilder>();
+            applicationBuilderMock.Setup(s => s.ApplicationServices).Returns(Mock.Of<IServiceProvider>());
+            var edmModel = new Mock<IEdmModel>().Object;
+
+            // Act & Assert
+            var exception = Assert.Throws<InvalidOperationException>(() => applicationBuilderMock.Object.UseOData(edmModel));
+
+            Assert.Equal("Cannot find 'OData services'. The OData services have not been configured. Are you missing a call to AddOData()?",
+                exception.Message);
```
AlanWong-MS: Can we use SRResources.MissingODataServices here?

src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
xuzhg: #Changed.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/ODataApplicationBuilderExtensionsTest.cs
xuzhg: Changed.

src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
xuzhg: It seems no.

src/Microsoft.AspNetCore.OData/Extensions/ODataApplicationBuilderExtensions.cs
```diff
@@ -12,8 +13,20 @@ namespace Microsoft.AspNet.OData.Extensions
     /// </summary>
     public static class ODataApplicationBuilderExtensions
     {
+        private static readonly int MaxTop = 100;
```
AlanWong-MS: Is there a variable that is shared with Microsoft.AspNet side? Should it be in the Shared project?

xuzhg: It looks no. Thx

Issue #1523, opened 2018-07-09
PageSize on EnableQuery attr does not respect EnableConstantParameterization


adzhiljano:
**First of all, great lib ! : )**

Now, when `PageSize` is set on the `EnableQuery` attribute and `EnableConstantParameterization` is set to true (the default value), there is no SQL parameter in the resulting query. Which might pollute the SQL Server query plan cache with unwanted query plans (containing constants).

### Assemblies affected

OData WebApi lib 6.1.0

### Reproduce steps
```c#
[EnableQuery(PageSize = 10)]
public IQueryable<Order> Get() { return db.Orders; }
```

### Expected result
 Notice the **TOP (@p__linq__0)**
```sql
SELECT TOP (@p__linq__0) 
    [Extent1].[Id] AS [Id], 
    [Extent1].[Name] AS [Name],
    ...
    FROM [dbo].[Orders] AS [Extent1]
    ORDER BY [Extent1].[Id] ASC
```
### Actual result
 Notice the **TOP (11)**
```sql
SELECT TOP (11) 
    [Extent1].[Id] AS [Id], 
    [Extent1].[Name] AS [Name],
    ...	
    FROM [dbo].[Orders] AS [Extent1]
    ORDER BY [Extent1].[Id] ASC
```
### Proposed solution

The system query option `$top` respects the `EnableConstantParameterization` flag already, so maybe something similar should be done for the `PageSize` handling.

In the ``ODataQueryOptions.ApplyTo`` method where the query options are applied, pass down the
`querySettings` or `querySettings.EnableConstantParameterization` to `LimitResults` and then to `TruncatedCollection<T>` and the constructor (all constructors accepting IQueryable) should now look like:

```c#
public TruncatedCollection(IQueryable<T> source, int pageSize, bool parameterize)
	: base(ExpressionHelpers.Take(source, checked(pageSize + 1), typeof(T), parameterize) as IQueryable<T>)
{
	Initialize(pageSize);
}
```

instead of the current:
```c#
public TruncatedCollection(IQueryable<T> source, int pageSize)
	: base(source.Take(checked(pageSize + 1)))
{
	Initialize(pageSize);
}
```

Tell me what you think about this solution, and I can make a PR.

Thanks!

biaol-odata:
[...]

PR #1535, opened 2018-07-16
Fix EnableQuery attribute to take into account constant parameterization when PageSize is set (#1523)

adzhiljano:
### Issues

*This pull request fixes issue #1523.*

### Description

*When PageSize is being handled in the EnableQuery attribute it should respect the ODataQuerySettings.EnableConstantParameterization property. Otherwise, there is no way to force a generation of a SQL parameter in the resulting query. Which can result in polluting the SQL Server query plan cache with unwanted query plans (containing constants).*

- [x] *Test cases added*


biaol-odata:
Hi @adzhiljano , please run the test shown below and update the publicAPI.bsl per test output. 
```
2018-07-16T08:06:39.1611387Z Failed   Microsoft.AspNet.OData.Test.PublicApi.PublicApiTest.TestPublicApi
```

adzhiljano:
Hey @biaol-odata,

I updated both
`test/UnitTest/Microsoft.AspNet.OData.Test/PublicApi/Microsoft.AspNet.OData.PublicApi.bsl`
and
`test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl`

Cheers!

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
```diff
@@ -2487,7 +2487,7 @@ public class Microsoft.AspNet.OData.Query.ODataQueryOptions {
 	public bool IsSupportedQueryOption (string queryOptionName)
 	public static bool IsSystemQueryOption (string queryOptionName)
 	public static bool IsSystemQueryOption (string queryOptionName, bool isDollarSignOptional)
-	public static IQueryable`1 LimitResults (IQueryable`1 queryable, int limit, out System.Boolean& resultsLimited)
+	public static IQueryable`1 LimitResults (IQueryable`1 queryable, int limit, bool parameterize, out System.Boolean& resultsLimited)
```
biaol-odata: We cannot break the public API by adding the bool.
Please consider adding an overload with the original method invoking the new overload with default value.

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
```diff
@@ -2667,9 +2667,9 @@ public class Microsoft.AspNet.OData.Query.TopQueryOption {
 
 public class Microsoft.AspNet.OData.Query.TruncatedCollection`1 : List`1, ICollection`1, IEnumerable`1, IList`1, IReadOnlyCollection`1, IReadOnlyList`1, ICollection, IEnumerable, IList, ICountOptionCollection, ITruncatedCollection {
 	public TruncatedCollection`1 (IEnumerable`1 source, int pageSize)
-	public TruncatedCollection`1 (IQueryable`1 source, int pageSize)
 	public TruncatedCollection`1 (IEnumerable`1 source, int pageSize, System.Nullable`1[[System.Int64]] totalCount)
-	public TruncatedCollection`1 (IQueryable`1 source, int pageSize, System.Nullable`1[[System.Int64]] totalCount)
+	public TruncatedCollection`1 (IQueryable`1 source, int pageSize, bool parameterize)
```
biaol-odata: same here. We cannot break public API.

adzhiljano:
Fixed the issues as you suggested. I've used `false` as a default value to match the current behaviour. Please take a look at the way I'm getting the `private static readonly MethodInfo _limitResultsGenericMethod` in `ODataQueryOptions` and if that is ok with your conventions. Thanks!

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
```diff
@@ -436,8 +436,8 @@ internal IQueryable ApplyPaging(IQueryable result, ODataQuerySettings querySetti
             if (pageSize > 0)
             {
                 bool resultsLimited;
-                result = LimitResults(result, pageSize, querySettings.EnableConstantParameterization, out resultsLimited);
-                if (resultsLimited && InternalRequest.RequestUri != null && InternalRequest.RequestUri.IsAbsoluteUri &&
+                result = LimitResults(result, pageSize, out resultsLimited);
```
mikepizzo: @KanishManuja-MS:Don't you want to retain the querySettings.EnableConstantParameterization parameter?

Issue #1529, opened 2018-07-12
ETag missing on derived entities


jpiquot:
The ETag is missing in the json query result on derived entities.

### Assemblies affected

OData WebApi lib 7 lastest from master branch.
.NET Core SDK 2.1.300
EF Core 2.1

### Reproduce steps
```

public class Organization
{
    public int Id {get;set;}
    public byte[] RowVersion {get;set;}
    public string Name {get;set;}
}
public class LegalEntity: Organization 
{
    public string LegalName {get;set;}
   ...
}
builder.EntitySet<Organization>("Organizations").EntityType.Property(p => p.RowVersion).IsConcurrencyToken();

builder.EntitySet<LegalEntity>("LegalEntity").EntityType.Property(p => p.RowVersion).IsConcurrencyToken();


```
### Expected result

> http://localhost:5000/odata/LegalEntities?$top=2&$select=Id,Name

```
{
    "@odata.context": "http://localhost:5000/odata/$metadata#LegalEntities(Id,Name)",
    "value": [
        {
            "@odata.type": "#WarehouseManagement.LogisticPlatform",
            "@odata.etag": "W/\"YmluYXJ5J0FBQUFBQUFBRXEwPSc=\"",
            "Id": 1,
            "Name": "Garonor 318305 318305"
        },
        {
            "@odata.type": "#WarehouseManagement.LogisticPlatform",
            "@odata.etag": "W/\"YmluYXJ5J0FBQUFBQUFBRXE0PSc=\"",
            "Id": 2,
            "Name": "Maurepas 318306 318306"
        }
    ]
}

```
### Actual result

> http://localhost:5000/odata/LegalEntities?$top=2&$select=Id,Name

```

{
    "@odata.context": "http://localhost:5000/odata/$metadata#LegalEntities(Id,Name)",
    "value": [
        {
            "@odata.type": "#WarehouseManagement.LogisticPlatform",
            "Id": 1,
            "Name": "Garonor 318305 318305"
        },
        {
            "@odata.type": "#WarehouseManagement.LogisticPlatform",
            "Id": 2,
            "Name": "Maurepas 318306 318306"
        }
    ]
}

```
### Additional detail
If I query the base class, everything is OK : 

> http://localhost:5000/odata/Organizations?$top=2&$select=Id,Name

```

{
    "@odata.context": "http://localhost:5000/odata/$metadata#Organizations(Id,Name)",
    "value": [
        {
[...]

PR #1537, opened 2018-07-17
Fix issue #1529: ETag missing on derived entity set

xuzhg:
### Issues

* #1529 .*

### Description

* When customer defined entity set on derived type, while the concurrency property defined on base type, the etag can't present on the payload. See detail in the #1529.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
@@ -598,6 +598,10 @@ namespace Microsoft.AspNet.OData.Builder
 
             IEnumerable<StructuralPropertyConfiguration> concurrencyPropertyies =
                 entityTypeConfig.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken);
+            foreach (var baseType in entityTypeConfig.BaseTypes())
+            {
+                concurrencyPropertyies = concurrencyPropertyies.Concat(baseType.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken));
+            }
 
             IList<IEdmStructuralProperty> edmProperties = new List<IEdmStructuralProperty>();
 
--- a/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
@@ -600,7 +600,8 @@ namespace Microsoft.AspNet.OData.Builder
                 entityTypeConfig.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken);
             foreach (var baseType in entityTypeConfig.BaseTypes())
             {
-                concurrencyPropertyies = concurrencyPropertyies.Concat(baseType.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken));
+                concurrencyPropertyies = concurrencyPropertyies.Concat(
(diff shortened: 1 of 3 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ETags/DerivedETagTests.cs
```diff
@@ -91,5 +104,19 @@ private static IEdmModel GetEdmModel(WebRouteConfiguration configuration)
...
+        {
+            string requestUri = this.BaseAddress + "/derivedEtag/ETagsDerivedCustomers?$select=Id";
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+            request.Headers.Accept.ParseAdd("application/json");
+            HttpResponseMessage response = await this.Client.SendAsync(request);
+            Assert.True(response.IsSuccessStatusCode);
+            var jsonResult = await response.Content.ReadAsObject<JObject>();
+            var derivedEtags = jsonResult.GetValue("value").Select(e => e["@odata.etag"].ToString());
+
+            Assert.StartsWith("W/\"bnVsbA==", String.Concat(derivedEtags));
```
biaol-odata: Can we do derivedETags.Single() here?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -598,6 +598,10 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
 
             IEnumerable<StructuralPropertyConfiguration> concurrencyPropertyies =
                 entityTypeConfig.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken);
+            foreach (var baseType in entityTypeConfig.BaseTypes())
+            {
+                concurrencyPropertyies = concurrencyPropertyies.Concat(baseType.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken));
+            }
```
biaol-odata: nit: split this into two lines?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ETags/DerivedETagTests.cs
xuzhg: the response payload will include multiple etag's value. So, I can use First() to check.  #resolved.

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
xuzhg: #resolved.

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -598,6 +598,11 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
 
             IEnumerable<StructuralPropertyConfiguration> concurrencyPropertyies =
                 entityTypeConfig.Properties.OfType<StructuralPropertyConfiguration>().Where(property => property.ConcurrencyToken);
+            foreach (var baseType in entityTypeConfig.BaseTypes())
+            {
+                concurrencyPropertyies = concurrencyPropertyies.Concat(
```
AlanWong-MS: typo: concurrencyPropertyies ->concurrencyProperties

xuzhg: #fixed.

Issue #1541, opened 2018-07-18
ModelID does not get parameterized in resulting SQL queries


adzhiljano:
Correct me if I'm wrong, but as what I saw the ability to have multiple models (IEdmModel) in a single project/service/etc. resulted in mapping a _Guid_ to every model, so that when having a select/expand you know which model should be used (I guess for serialization). I'm referring to the [following line](https://github.com/OData/WebApi/blob/b1890153f85b07571560179a27f288f19da05930/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs#L45) in the `SelectExpandBinder` . Actually it's not that important what it's used for, I'm just curious. 

The problem is that the value of this _Guid_ exists in the resulting SQL query and it should respect the `ODataQuerySettings.EnableConstantParameterization` so that it is parameterized. Otherwise the SQL Server query plan cache might get polluted with unwanted query plans (containing constants). Note that even a recycle of the IIS App pool will result in creating a new Guid and that will change the text of the SQL query => new query plan.

### Assemblies affected

OData WebApi lib 6.1.0 and 7.0.0

### Reproduce steps

~/Orders?$select=Id

### Expected result

**Notice the @p__linq__0 AS [C1]**

```sql
SELECT
    [Project1].[Id] AS [Id], 
    [Project1].[C1] AS [C1],
    [Project1].[C2] AS [C2]
    FROM ( SELECT 
        [Extent1].[Id] AS [Id],
        @p__linq__0 AS [C1], 
        N''Id'' AS [C2]
        FROM [dbo].[Orders] AS [Extent1]
    )  AS [Project1]
    ORDER BY [Project1].[Id] ASC
```

### Actual result
**Notice the N'8abc0b0c-f672-4f56-b686-494206860c13' AS [C1]**

```sql
SELECT
    [Project1].[Id] AS [Id], 
    [Project1].[C1] AS [C1], 
    [Project1].[C2] AS [C2]
    FROM ( SELECT 
        [Extent1].[Id] AS [Id], 
        N'8abc0b0c-f672-4f56-b686-494206860c13' AS [C1], 
        N'Id' AS [C2]
        FROM [dbo].[Orders] AS [Extent1]
    )  AS [Project1]
    ORDER BY [Project1].[Id] ASC
```

PR #1542, opened 2018-07-18
Fix SelectExpandBinder to take into account constant parameterization when using ModelId as part of a query(#1541)

adzhiljano:
### Issues

This pull request fixes issue #1541.

### Description

When ModelId is used as part of a query in the SelectExpandBinder, it should respect the ODataQuerySettings.EnableConstantParameterization property. Otherwise, there is no way to force a generation of a SQL parameter in the resulting query. Which can result in polluting the SQL Server query plan cache with unwanted query plans (containing constants).

- [ ] *Test cases added*


AlanWong-MS:
Partly related: #1101

Issue #850, opened 2016-11-08
ODataConventionModelBuilder takes into account [EnumMember] on enum members, but ODataPrimitiveSerializer (?) does not, causing mismatch in the payload.


fpintos:
In trying to workaround the issue of enum members not being made lower camel case (see #849), I've tried tagging my enums with `[DataContract]` and `[EnumMember(Value="...")]`. This worked to generate models with enum members with different case, but when payloads are written, the values still come up with the name as they appear in the CLR type.

### Assemblies affected
From NuGet package Microsoft.AspNet.OData v6.0.0

### Reproduce steps
- Build a model with ODataConventionModelBuilder that includes an enum
- Add `[DataContract]` and `[EnumMember(Value="...")]` to the enum to override the names.
- Verify that the resulting IEdmEnumType contains members with the new names.
- Make a request to an entity that contains such enum as a property and notice the returned value matches the CLR enum member name, not the EDM name.

### Expected result
Enum values to be emitted with their EDM names.

### Actual result
Enum values are emitted with their CLR names.

### Additional details
I believe the problem might be in ODataPrimitiveSerializer. ConvertUnsupportedPrimitives, where it does this:
```c#
if (type.IsEnum)
{
    // Enums are treated as strings
    return value.ToString();
}
```

Attached is a LinqPad sample that builds such as EDM model and then writes a message with an entity; we can see in the output the EDM names of the enum members corresponding to the [EnumMember] attribute, but the payload of the message contains CLR names.
[EnumValuesAreWrittenWithClrNamesInsteadOfEdmNames.zip](https://github.com/OData/WebApi/files/577206/EnumValuesAreWrittenWithClrNamesInsteadOfEdmNames.zip)

```c#
[DataContract(Name = "myEnum1")]
public enum Enum1
{
	[EnumMember(Value = "myValue")]
	HelloWorld,
}

public class Type1
{
	public string Id { get; set; }
	public Enum1 HelloWorld { get; set; }
}

void Main()
{
	var builder = new ODataConventionModelBuilder();
	builder.EnableLowerCamelCase();
	builder.EnumType<Enum1>();
	builder.EntitySet<Type1>("a");

	var model = builder.GetEdmModel();

	// this shows the EDM names as using the [EnumMember(Value)]
	model.SchemaElements.OfType<IEdmEnumType>().First().Members.Select(x => new { x.Name, x.Value.Value }).Dump("enum type members");

	var entitySet = model.FindDeclaredEntitySet("a");
[...]

PR #1558, opened 2018-07-31
Fix the issue #850, (de)serialize using the Edm Enum member string

xuzhg:
### Issues

*This pull request fixes issue ##850.*

### Description

* Enum supports camel case, however when serialize the enum member, it's using the old case, doesn't support the camel case. This fix is a quick fix for this.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/samples/AspNetCoreODataSample.Web/Models/Level.cs
+++ b/samples/AspNetCoreODataSample.Web/Models/Level.cs
@@ -14,7 +14,7 @@ namespace AspNetCoreODataSample.Web.Models
         [EnumMember(Value = "medium")]
         Medium,
 
-        [EnumMember(Value = "high")]
+        [EnumMember(Value = "veryhigh")]
         High
     }
 }
--- a/samples/AspNetCoreODataSample.Web/Models/Level.cs
+++ b/samples/AspNetCoreODataSample.Web/Models/Level.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Runtime.Serialization;
+
+namespace AspNetCoreODataSample.Web.Models
+{
+    [DataContract(Name = "level")]
+    public enum Level
+    {
+        [EnumMember(Value = "low")]
+        Low,
+
+        [EnumMember(Value = "medium")]
+        Medium,
+
+        [EnumMember(Value = "high")]
+        High
+    }
+}
--- a/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
@@ -440,13 +440,7 @@ namespace Microsoft.AspNet.OData.Builder
             // add annotation for properties
             Dictionary<PropertyInfo, IEdmProperty> edmProperties = edmTypeMap.EdmProperties;
             model.AddClrPropertyInfoAnnotations(edmProperties);
-
-            // add Enum mapping annotation
-            if (edmTypeMap.EnumMembers != null && edmTypeMap.EnumMembers.Any())
-            {
-                model.SetAnnotationValue(model, new ClrEnumMemberAnnotation(edmTypeMap.EnumMembers));
-            }
-
+            model.AddClrEnumMemberInfoAnnotations(edmTypeMap);
             model.AddPropertyRestrictionsAnnotations(edmTypeMap.EdmPropertiesRestrictions);
             model.AddPropertiesQuerySettings(edmTypeMap.EdmPropertiesQuerySettings);
             model.AddStructuredTypeQuerySettings(edmTypeMap.EdmStructuredTypeQuerySettings);
@@ -519,6 +513,21 @@ namespace Microsoft.AspNet.OData.Builder
                 {
                     model.SetAnnotationValue(edmProperty, new ClrPropertyInfoAnnotation(clrProperty));
                 }
+            }
+        }
+
+        private static void AddClrEnumMemberInfoAnnotations(this EdmModel model, EdmTypeMap edmTypeMap)
+        {
+            if (edmTypeMap.EnumMembers == null || !edmTypeMap.EnumMembers.Any())
+            {
+                return;
+            }
+
+            var enumGroupBy = edmTypeMap.EnumMembers.GroupBy(e => e.Key.GetType(), e => e);
+            foreach (var enumGroup in enumGroupBy)
+            {
+                IEdmType edmType = edmTypeMap.EdmTypes[enumGroup.Key];
+                model.SetAnnotationValue(edmType, new ClrEnumMemberAnnotation(enumGroup.ToDictionary(e => e.Key, e => e.Value)));
             }
         }
 
--- a/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
@@ -440,6 +440,13 @@ namespace Microsoft.AspNet.OData.Builder
             // add annotation for properties
             Dictionary<PropertyInfo, IEdmProperty> edmProperties = edmTypeMap.EdmProperties;
             model.AddClrPropertyInfoAnnotations(edmProperties);
+
+            // add Enum mapping annotation
+            if (edmTypeMap.EnumMembers != null && edmTypeMap.EnumMembers.Any())
+            {
+                model.SetAnnotationValue(model, new ClrEnumMemberAnnotation(edmTypeMap.EnumMembers));
+            }
+
             model.AddPropertyRestrictionsAnnotations(edmTypeMap.EdmPropertiesRestrictions);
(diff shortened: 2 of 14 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataEnumSerializer.cs
```diff
@@ -89,6 +90,14 @@ public sealed override ODataValue CreateODataValue(object graph, IEdmTypeReferen
                 }
             }
 
+            // Enum member maybe support camel case. So, try to use the Edm member name to create Enum value.
+            var member = enumType.EnumDefinition().Members
```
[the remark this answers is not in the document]
xuzhg: I think my comments maybe confusing. It's not only for camel case, it's also for model alias because it's allowed to input any string in the [EnumMember] attribute, "high", or "High", or "hIgH", etc. 

I updated my comments to refresh this point.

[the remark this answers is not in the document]
biaol-odata: Then, does the string in the [EnumMember] attribute require case-insensitive (StringComparison.OrdinalIgnoreCase)? Alias can be a completely different string from the Enum (not only case-insensitive), right?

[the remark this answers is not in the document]
xuzhg: En… It seems you're right. It can be any valid string. Let me update the PR later.

biaol-odata:
https://github.com/OData/WebApi/pull/1558

src/Microsoft.AspNet.OData.Shared/ClrEnumMemberAnnotation.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Represents a mapping from an <see cref="IEdmEnumMember"/> to a CLR Enum member.
+    /// </summary>
+    public class ClrEnumMemberAnnotation
+    {
+        private IDictionary<Enum, IEdmEnumMember> _maps;
```
biaol-odata: nit: naming, "_map"?

src/Microsoft.AspNet.OData.Shared/ClrEnumMemberAnnotation.cs
```diff
@@ -0,0 +1,70 @@
...
+        private IDictionary<IEdmEnumMember, Enum> _reverseMaps;
```
biaol-odata: >_reverseMaps [](start = 50, length = 12)

"_reverseMap"?

src/Microsoft.AspNet.OData.Shared/ClrEnumMemberAnnotation.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNet.OData.Common;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Represents a mapping from an <see cref="IEdmEnumMember"/> to a CLR Enum member.
+    /// </summary>
+    public class ClrEnumMemberAnnotation
```
biaol-odata: >ClrEnumMemberAnnotation [](start = 17, length = 23)

Looks like there are two maps used, for regular lookup and reversed lookup. Not sure what is the perf gain for this? Look like the original _maps is only used in one place for serialization.
Also please update the summary to reflect the direction(s) of the lookup.

src/Microsoft.AspNet.OData.Shared/ClrEnumMemberAnnotation.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>The Edm <see cref="IEdmEnumMember"/>.</returns>
+        public IEdmEnumMember GetEdmEnumMember(Enum clrEnumMemberInfo)
+        {
+            IEdmEnumMember value;
+            if (_maps.TryGetValue(clrEnumMemberInfo, out value))
+            {
+                return value;
+            }
+
+            return null;
```
biaol-odata: >return null; [](start = 12, length = 12)

nit: Does the TryGetValue call output null for value already, so that we can just return the value directly no match whether the key is found?

src/Microsoft.AspNet.OData.Shared/ClrEnumMemberAnnotation.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>The backing CLR Enum member info.</returns>
+        public Enum GetClrEnumMember(IEdmEnumMember edmEnumMember)
+        {
+            Enum value;
+            if (_reverseMaps.TryGetValue(edmEnumMember, out value))
+            {
+                return value;
+            }
+
+            return null;
```
biaol-odata: >return null [](start = 12, length = 11)

same as above.

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -440,6 +440,13 @@ private static void ValidateActionOverload(IEnumerable<ActionConfiguration> conf
             // add annotation for properties
             Dictionary<PropertyInfo, IEdmProperty> edmProperties = edmTypeMap.EdmProperties;
             model.AddClrPropertyInfoAnnotations(edmProperties);
+
+            // add Enum mapping annotation
+            if (edmTypeMap.EnumMembers != null && edmTypeMap.EnumMembers.Any())
+            {
+                model.SetAnnotationValue(model, new ClrEnumMemberAnnotation(edmTypeMap.EnumMembers));
```
biaol-odata: >model [](start = 41, length = 5)

So we are adding the enum annotations to the model as top level, instead of associating them with the actual Enum type in which they are defined? Annotations for a specific enum type should have a local scope instead of global (model-wide) scope, right?

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -704,6 +704,22 @@ public static string GetClrPropertyName(IEdmProperty edmProperty, IEdmModel edmM
...
+                throw Error.ArgumentNull("edmModel");
+            }
+
+            ClrEnumMemberAnnotation annotation = edmModel.GetAnnotationValue<ClrEnumMemberAnnotation>(edmModel);
+            if (annotation != null)
+            {
+                return annotation;
+            }
+
+            return null;
```
biaol-odata: Can we just directly return result from the edmModel call?

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -704,6 +704,22 @@ public static string GetClrPropertyName(IEdmProperty edmProperty, IEdmModel edmM
             return propertyName;
         }
 
+        public static ClrEnumMemberAnnotation GetClrEnumMemberAnnotation(this IEdmModel edmModel)
```
biaol-odata: >public static ClrEnumMemberAnnotation GetClrEnumMemberAnnotation [](start = 8, length = 64)

Wondering what is the purpose of the "public" modifier here? Looks like we don't need <summary> doc for all public methods in this internal helper class? This is up to you.

biaol-odata:
Can we add test that is included in the CI build? I am seeing the updates in the samples of enum member annotation, but without exercising the annotations. Samples are not part of CI build eithere, right?

biaol-odata:
Also, please check the build log for build error.

samples/AspNetCoreODataSample.Web/Models/Level.cs
```diff
@@ -0,0 +1,20 @@
...
+    [DataContract(Name = "level")]
+    public enum Level
+    {
+        [EnumMember(Value = "low")]
+        Low,
+
+        [EnumMember(Value = "medium")]
+        Medium,
+
+        [EnumMember(Value = "veryhigh")]
```
AlanWong-MS: nit: "veryhigh" -> "high"?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/EnumTypeTest.cs
```diff
@@ -1192,10 +1201,16 @@ public enum Life
...
         [EnumMember(Value = "sad")]
         Sad = 2,
+
         JustSoSo = 3,
+
         [EnumMember]
-        KeepDefaultName
+        KeepDefaultName = 4,
+
+        [EnumMember(Value = "soso")]
```
AlanWong-MS: Is the mismatched value part of the test case?

xuzhg: That's intentionally. I want to test the model alias.

samples/AspNetCoreODataSample.Web/Models/Level.cs
xuzhg: That's intentionally. I want to test the model alias.

xuzhg:
Ping...

Issue #1559, opened 2018-08-01
Don't assume port 80 for nextLink when request was HTTPS


andrewlock:
I just spent half a day debugging this issue, where Power BI could connect to an ASP.NET Core OData api over HTTP, but not over HTTPS. I finally figured out that the `@odata.nextLink` was returning an invalid port in the URL when accessed over HTTPS - port 80 instead of port 443.

### Assemblies affected

 OData WebApi lib 7.0.0

### Reproduce steps

The issue only appeared when the app was behind a reverse proxy (HAproxy and / or NGINX). In these cases, the ASP.NET Core `ForwardedHeadersMiddleware` wasn't explicitly setting the forwarded port (due to [this issue](https://github.com/aspnet/BasicMiddleware/issues/204)), so `request.Host.Port == null`.

That causes the `UriBuilder` in the `GetNextPageLink` function to use the default port of 80, even though the scheme is `"https"`. Requests to the next page link will consequently fail.

https://github.com/OData/WebApi/blob/8cf204777d183f909bc16c7743d5c058c52a6a78/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs#L268-L279

### Expected result

`@odata.nextLink` should use the correct port, e.g. https://example.com:80/odata/MyEntities?$skip=10
### Actual result

`@odata.nextLink` should use the correct port, e.g. https://example.com:443/odata/MyEntities?$skip=10 or https://example.com/odata/MyEntities?$skip=10 

### Additional detail

I believe this issue could be fixed by allowing `UriBuilder` to resolve the port itself if the port isn't set, e.g:

```c#
UriBuilder uriBuilder = new UriBuilder(request.Scheme, request.Host.Host) { Path = (request.PathBase + request.Path).ToUriComponent() };
if (request.Host.Port.HasValue)
{
    uriBuilder.Port = request.Host.Port.Value;
}
```

I'm happy to raise a PR for this if this sounds acceptable

biaol-odata:
Ideally, we should relying on the ASP.NET Core to handle the reverse proxy scenario correctly, instead of trying to patch up in WebAPI?

biaol-odata:
Adding ASP.NET Core issue link: https://github.com/aspnet/BasicMiddleware/issues/204

biaol-odata:
[...]

PR #1560, opened 2018-08-03
Don't assume port 80 when generating nextLink in ASP.NET Core #1559

andrewlock:
### Issues

*This pull request fixes issue #1559*

### Description

Don't explicitly set a port when generating the nextLink

If the port isn't set on `request.Host`, port 80 was assumed previously. If the app is running behind a reverse proxy, and the port was not forwarded, this can result in invalid URLs - for example an HTTPS URL using port 80. Instead, we just omit the port completely if it's not explicitly set.

- [x] *Test cases added*

### Additional work necessary

Unfortunately, I couldn't build the repo after cloning (`Could not load file or assembly 'Microsoft.Build, Version=14.0.0.0`), and there are no existing tests for `HttpRequestExtensions` as far as I can see, so I haven't tried to add any.

### DIFF
--- a/test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/HttpRequestExtensionsTest.cs
+++ b/test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/HttpRequestExtensionsTest.cs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Test.Abstraction;
 using Microsoft.AspNetCore.Http;
--- a/test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/HttpRequestExtensionsTest.cs
+++ b/test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/HttpRequestExtensionsTest.cs
@@ -1,0 +1,83 @@
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNetCore.Http;
+using System;
+using System.Net.Http;
+using Xunit;
+
+namespace Microsoft.AspNet.OData.Test
+{
+    public class HttpRequestMessageExtensionsTest
+    {
+        [Theory]
+        [InlineData("http://localhost/Customers", 10, "http://localhost/Customers?$skip=10")]
+        [InlineData("http://localhost/Customers?$filter=Age ge 18", 10, "http://localhost/Customers?$filter=Age%20ge%2018&$skip=10")]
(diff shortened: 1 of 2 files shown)


biaol-odata:
LGTM

biaol-odata:
The underlying ASP.NET Core issue https://github.com/aspnet/BasicMiddleware/issues/204 is being re-triaged.

andrewlock:
Just in my opinion, even if the underlying issue is fixed, I feel like the code is more-correct after the PR, as it does not assume any specific port to use for link generation.

biaol-odata:
@andrewlock Thanks for your contribution. We looked at the PR in today's team meeting, the code looks good. As general, we usually require test coverage for code. Can you please add some simple test? Thanks!

andrewlock:
I'll do my best, could take a while as I can't build the solution or run any tests (`Could not load file or assembly 'Microsoft.Build, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.`, so will be trial and error with the CI...

biaol-odata:
@andrewlock I think ASP.NET Core version of WebApi requires VS 2017. Are you using VS2017?

andrewlock:
@biaol-odata , yes I am - I found the solution files eventually! It's the `build.cmd` that's failing to run on my machine. 

Given that all your readme states under "Build" is `build.cmd`, it's not very friendly to newcomers unfortunately 🙂 

I've added some unit tests that cover general link generation (similar to the tests of the non-core ASP.NET) and some tests for this specific case.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/HttpRequestExtensionsTest.cs
```diff
@@ -0,0 +1,83 @@
...
+            Assert.Equal(nextPageUri, nextPageLink.AbsoluteUri);
+        }
+
+        [Theory]
+        [InlineData("http://example.com:80/Customers", 10, "http://example.com/Customers?$skip=10")]
+        [InlineData("https://example.com:443/Customers", 10, "https://example.com/Customers?$skip=10")]
+        public void GetNextPageLink_UsesPortWhenDefaultPortIsProvided(string requestUri, int pageSize, string nextPageUri)
+        {
+            // Arrange
+            // The RequestFactory removes the default port, so we explicitly add it
```
biaol-odata: For this case, when the port specified in uri is same as the default value. request.Host is formalized, so we add it back, and the generated linked is still the one w/o port number?

andrewlock: Yeah - `UriBuilder` formalizes the uri inside `GetNextPageLink()` too. Which is absolutely fine, the problem before was that in these cases, port 80 would be enforced, even if it was https!

andrewlock:
Copyright added 👍

biaol-odata:
Merged. Closed.

Issue #1579, opened 2018-08-20
Star select ($select=*) not returning dynamic properties


maximpashuk:
Hi I am using **Microsoft.AspNetCore.OData 7.0.1**

I always think that call to "**/movies**" should be equals to call to "**/movies?$select=***".
But this is not true in a case of dynamic properties.

This behavior can be simply reproduced in sample app in this repo **([AspNetCoreODataSample.Web](https://github.com/OData/WebApi/tree/master/samples/AspNetCoreODataSample.Web)).**

If you add to Movie model

```csharp
public class Movie
{
    public int ID { get; set; }
    ...
    [NotMapped] // attribute required by ef core
    public IDictionary<string, object> DynamicProperties { get; set; }
}
```

and later in MoviesController you add some example data
```csharp
Movie m = new Movie
{
    Title = "Conan",
    ...
    DynamicProperties = new Dictionary<string, object>
    {
        {"key1", "value1" },
        {"key2", "value2" },
    }
};
```

call to "efcore/Movies" returns dynamic properties, while call to "efcore/Movies?$select=*" does not return dynamic properties.

[ODATA specs](http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_SystemQueryOptionselect) said about that

> It is also possible to request all structural properties, **including any dynamic properties, using the star operator**

So I think this behavior is a bug that should be fixed.

xuzhg:
Merged.

PR #1581, opened 2018-08-20
Fix $select=* not returning dynamic properties

xuzhg:
### Issues

* #1579 
### Description

* Fix the problem when using $select=* can't return the dynamic properties.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/samples/AspNetCoreODataSample.Web/Models/Person.cs
+++ b/samples/AspNetCoreODataSample.Web/Models/Person.cs
@@ -1,11 +1,15 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+using System.Collections.Generic;
+
 namespace AspNetCoreODataSample.Web.Models
 {
     public class Person
     {
         public string FirstName { get; set; }
         public string LastName { get; set; }
+
+        public IDictionary<string, object> Dynamics { get; set; }
     }
 }
--- a/samples/AspNetCoreODataSample.Web/Models/Person.cs
+++ b/samples/AspNetCoreODataSample.Web/Models/Person.cs
@@ -10,6 +10,6 @@ namespace AspNetCoreODataSample.Web.Models
         public string FirstName { get; set; }
         public string LastName { get; set; }
 
-        public IDictionary<string, object> Dynamics { get; set; }
+        public IDictionary<string, object> DynamicProperties { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
@@ -129,6 +129,8 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                 }
                 else
                 {
+                    // Explicitly set SelectAllDynamicProperties as false, while the BuildSelections method will set it as true
+                    // if it meets the select all condition.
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -129,8 +129,8 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
                 }
                 else
                 {
-                    BuildSelections(selectExpandClause, allStructuralProperties, allComplexStructuralProperties, allNavigationProperties, allActions, allFunctions);
                     SelectAllDynamicProperties = false;
+                    BuildSelections(selectExpandClause, allStructuralProperties, allComplexStructuralProperties, allNavigationProperties, allActions, allFunctions);
```
AlanWong-MS: Does BuildSelections modify SelectAllDynamicProperties? I am wondering why the function call is moved after

xuzhg: Yes. In the BuildSelections(…), we should set it as "true" if we have `$select=*`. 

I have to move the initialization ahead of "BuildSelections(...)" method call.

AlanWong-MS: Okay, I think the change you have is the quickest and easiest way to fix it, but I'm not comfortable with the way we are updating SelectAllDynamicProperties since we are setting it to false and then _maybe_ BuildSelections will update it to true. Would it be a lot of code changes to add an additional `out` boolean parameter to BuildSelections? And then set SelectAllDynamicProperties to the boolean? I think that would be an easier code pattern to maintain and read.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -375,7 +378,7 @@ private static bool GetSelectsOpenTypeSegments(SelectExpandClause selectExpandCl
                 return false;
             }
 
-            if (selectExpandClause.AllSelected)
+            if (IsSelectAll(selectExpandClause))
```
AlanWong-MS: What is the difference?

samples/AspNetCoreODataSample.Web/Models/Person.cs
```diff
@@ -1,11 +1,15 @@
...
+using System.Collections.Generic;
+
 namespace AspNetCoreODataSample.Web.Models
 {
     public class Person
     {
         public string FirstName { get; set; }
         public string LastName { get; set; }
+
+        public IDictionary<string, object> Dynamics { get; set; }
```
AlanWong-MS: Dynamics->DynamicProperties?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
xuzhg: Emm, well, it's not i think. It looks to initialize a default value to "SelectAllDynamicProperties" property. It should move to the head of this constructor or in its property initilizer. But we put here it to make sure we initialize it as a correct value.

AlanWong-MS: As we discussed, perhaps we can comment that BuildSelections modifies the SelectAllDynamicProperties to true in some conditions.

PR #1592, opened 2018-08-24
Fix #1591 -  Honor namespace value when registering Types using Builder

### LINKED ISSUE #1591: OData builder does not honor namespace
OData EDM builder does not add new entity and complex types automatically to the namespace registered in ODataConventionModelBuilder.

### Assemblies affected

OData 7.X assemblies

### Reproduce steps

ODataConventionModelBuilder builder = new ODataConventionModelBuilder
            {
                Namespace = "MyCustomNamespace",
            };

builder.EntitySet<MyEntity>("myentities");

### Expected result

MyEntity should be registered with MyCustomNamespace namespace

### Actual result

MyEntity is registered with Default namespace

### Additional detail
One way to fix this is to do something like
builder.EntitySet<MyEntity>("myentities").EntityType.Namespace = "MyCustomNamespace";
but this does not work with automatic registration, hence not a really usable workaround.

RamjotSingh:
### Issues

*This pull request fixes issue #1591*

### Description
- Honoring Namespace value passed to ModelBuilder for all classes registered. If user passed a value that value is used as the namespace for the registered types by default. If no value is specified by the user fallback to CLR namespace. If CLR namespace is null, we use 'Default' as the namespace.
This PR also brings a parity between the namespace value used on Actions\Functions and the actual namespace of the entity.
- Fixed the build to run on machines which are running just VS2017, also fixed for machines running Preview version of VS2017.

- [x] *Test cases added*

### Additional work necessary
This might be regarded as a breaking change (it was an acting bug imo tho.) So based on that a Doc change might be required. Since users are expected to use namespace in front of action\function name and that behavior has not changed maybe no documentation will be needed. However $metadata document generation will change with this PR.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -63,19 +63,47 @@ $LOGDIR = $ENLISTMENT_ROOT + "\bin"
 # Default to use Visual Studio 2015
 $VS14MSBUILD=$PROGRAMFILESX86 + "\MSBuild\14.0\Bin\MSBuild.exe"
 $VSTEST = $PROGRAMFILESX86 + "\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe"
-$SN = $PROGRAMFILESX86 + "\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.5.1 Tools\sn.exe"
-$SNx64 = $PROGRAMFILESX86 + "\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.5.1 Tools\x64\sn.exe"
+
+# Figure out the directory and path for SN.exe
+$SN = $null
+$SNx64 = $null
+ForEach ($directory in [System.IO.Directory]::EnumerateDirectories($PROGRAMFILESX86 + "\Microsoft SDKs\Windows"))
+{
+    ForEach($sdkVersionDirectory in [System.IO.Directory]::EnumerateDirectories($directory + "\bin"))
+    {
+        if([System.IO.File]::Exists($sdkVersionDirectory + "\sn.exe") -and [System.IO.File]::Exists($sdkVersionDirectory + "\x64\sn.exe"))
+        {
+            $SN = $sdkVersionDirectory + "\sn.exe"
+            $SNx64 = $sdkVersionDirectory + "\x64\sn.exe"
+            break
+        }
+    }
+}
 
 # Use Visual Studio 2017 compiler for .NET Core and .NET Standard. Because VS2017 has different paths for different
 # versions, we have to check for each version. Meanwhile, the dotnet CLI is required to run the .NET Core unit tests in this script.
+# Furthurmore, Visual Studio 2017 has a Preview version as well which uses Microsoft Visual Studio\Preview as path instead of \2017
+$VS15VARIANTS = "2017", "Preview"
+$VS15VARIANTPATH = $null
+ForEach ($variant in $VS15VARIANTS)
+{
+    $tempVSPath = ($PROGRAMFILESX86 + "\Microsoft Visual Studio\{0}") -f $variant
+	if([System.IO.Directory]::Exists($tempVSPath))
+    {
+        $VS15VARIANTPATH = $tempVSPath
+        break
+    }
+}
+
 $VS15VERSIONS = "Enterprise", "Professional", "Community"
 $VS15MSBUILD = $null
 ForEach ($version in $VS15VERSIONS)
 {
-    $tempMSBuildPath = ($PROGRAMFILESX86 + "\Microsoft Visual Studio\2017\{0}\MSBuild\15.0\Bin\MSBuild.exe") -f $version
+    $tempMSBuildPath = ($VS15VARIANTPATH + "\{0}\MSBuild\15.0\Bin\MSBuild.exe") -f $version
     if([System.IO.File]::Exists($tempMSBuildPath))
     {
         $VS15MSBUILD = $tempMSBuildPath
+        $VSTEST = ($VS15VARIANTPATH + "\{0}\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe") -f $version
         break
     }
 }
@@ -311,6 +339,12 @@ Function RunBuild ($sln, $vsToolVersion)
     if($vsToolVersion -eq '15.0')
     {
         $MSBUILD=$VS15MSBUILD
+    }
+
+    # If VS2015 is not present, try to use VS2017
+    if(![System.IO.File]::Exists($MSBUILD))
+    {
+        $MSBUILD = $VS15MSBUILD
     }
 
     & $MSBUILD $slnpath /t:$Build /m /nr:false /fl "/p:Platform=Any CPU" $Conf /p:Desktop=true `
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
@@ -53,7 +53,7 @@ namespace Microsoft.AspNet.OData.Builder
             }
             set
             {
-                if (string.IsNullOrEmpty(value))
+                if (String.IsNullOrEmpty(value))
                 {
                     throw Error.PropertyNull();
                 }
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.OData.Builder
         /// </summary>
         public ODataModelBuilder()
         {
-            Namespace = "Default";
+            _namespace = "Default";
             ContainerName = "Container";
             DataServiceVersion = _defaultDataServiceVersion;
             MaxDataServiceVersion = _defaultMaxDataServiceVersion;
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
@@ -16,6 +16,8 @@ namespace Microsoft.AspNet.OData.Builder
     /// </summary>
     public class ODataModelBuilder
     {
+        private const string DefaultNamespace = "Default";
+
         private static readonly Version _defaultDataServiceVersion = EdmConstants.EdmVersion4;
         private static readonly Version _defaultMaxDataServiceVersion = EdmConstants.EdmVersion4;
 
@@ -35,11 +37,12 @@ namespace Microsoft.AspNet.OData.Builder
         /// </summary>
         public ODataModelBuilder()
         {
-            _namespace = "Default";
+            _namespace = DefaultNamespace;
             ContainerName = "Container";
             DataServiceVersion = _defaultDataServiceVersion;
             MaxDataServiceVersion = _defaultMaxDataServiceVersion;
             BindingOptions = NavigationPropertyBindingOption.None;
+            HasAssignedNamespace = false;
         }
 
         /// <summary>
@@ -53,13 +56,17 @@ namespace Microsoft.AspNet.OData.Builder
             }
             set
             {
+                // If user chooses to set the namespace to null, we revert back to 'Default' namespace.
                 if (String.IsNullOrEmpty(value))
                 {
-                    throw Error.PropertyNull();
+                    this.HasAssignedNamespace = false;
+                    _namespace = DefaultNamespace;
                 }
-
-                this.HasAssignedNamespace = true;
-                _namespace = value;
+                else
+                {
+                    this.HasAssignedNamespace = true;
+                    _namespace = value;
+                }
             }
         }
 
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
@@ -28,6 +28,7 @@ namespace Microsoft.AspNet.OData.Builder
 
         private Version _dataServiceVersion;
         private Version _maxDataServiceVersion;
+        private string _namespace;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ODataModelBuilder"/> class.
@@ -44,7 +45,23 @@ namespace Microsoft.AspNet.OData.Builder
         /// <summary>
         /// Gets or sets the namespace that will be used for the resulting model
         /// </summary>
-        public string Namespace { get; set; }
+        public string Namespace
+        {
+            get
+            {
+                return _namespace;
+            }
+            set
+            {
+                if (string.IsNullOrEmpty(value))
+                {
+                    throw Error.PropertyNull();
+                }
+
+                this.HasAssignedNamespace = true;
+                _namespace = value;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the name of the container that will hold all the navigation sources, actions and functions
@@ -141,6 +158,14 @@ namespace Microsoft.AspNet.OData.Builder
         /// Gets or sets the navigation property binding options.
         /// </summary>
         public NavigationPropertyBindingOption BindingOptions { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the <see cref="Namespace"/> was auto assigned or is using a value assigned by user.
+        /// </summary>
+        /// <value>
+        ///   <c>true</c> if the <see cref="Namespace"/> is using user assigned value; otherwise, <c>false</c>.
+        /// </value>
+        internal bool HasAssignedNamespace { get; private set; }
 
         /// <summary>
         /// Registers an entity type as part of the model and returns an object that can be used to configure the entity type.
--- a/src/Microsoft.AspNet.OData.Shared/Builder/StructuralTypeConfiguration.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/StructuralTypeConfiguration.cs
@@ -58,7 +58,9 @@ namespace Microsoft.AspNet.OData.Builder
             ModelBuilder = modelBuilder;
             _name = clrType.EdmName();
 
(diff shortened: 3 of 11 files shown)


src/Microsoft.AspNet.OData.Shared/Builder/StructuralTypeConfiguration.cs
```diff
@@ -58,7 +57,9 @@ protected StructuralTypeConfiguration(ODataModelBuilder modelBuilder, Type clrTy
             ClrType = clrType;
             ModelBuilder = modelBuilder;
             _name = clrType.EdmName();
-            _namespace = clrType.Namespace ?? DefaultNamespace;
+
+            // Use the namespace is uder specified one, otherwise use CLR Namespace. If CLR Namespace is null we fallback to "Default"
```
xuzhg: What's the scenario about "CLR namespace is null?"

src/Microsoft.AspNet.OData.Shared/Builder/StructuralTypeConfiguration.cs
```diff
@@ -58,7 +57,9 @@ protected StructuralTypeConfiguration(ODataModelBuilder modelBuilder, Type clrTy
...
+            _namespace = modelBuilder.HasAssignedNamespace ? modelBuilder.Namespace : clrType.Namespace ?? modelBuilder.Namespace;
```
xuzhg: It seems it changes the previous logic for the namespace. It will change all the structural type namespace. 

By the way, do you want to change the EnumType namespace?

src/Microsoft.AspNet.OData.Shared/Builder/StructuralTypeConfiguration.cs
RamjotSingh: No realistic scenario tbh. I left it there thinking there might have been scenario. According to documentation it can be null for anonymous types and generic params both of which I don't think are allowed by OData. If you think this is safe to remove I can remove. 

https://docs.microsoft.com/en-us/dotnet/api/system.type.namespace?redirectedfrom=MSDN&view=netframework-4.7.2#System_Type_Namespace

src/Microsoft.AspNet.OData.Shared/Builder/StructuralTypeConfiguration.cs
RamjotSingh: I thought Enums were also going through Structural namespace I will make the appropriate change there as well.

Yes the intent indeed is to change the logic. As explained in the issue, today's approach messes with the auto builder and there is no simple way to put everything into same namespace. Builder's namespace has no value unless you are adding functions or actions. This PR is aiming to change that.

RamjotSingh:
@AlanWong-MS , @biaol-odata , @KanishManuja-MS , @madansr7 Can you please review this :)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using Microsoft.AspNet.OData.Builder;
+using Microsoft.AspNet.OData.Test.Builder.TestModels;
+using Microsoft.AspNet.OData.Test.Builder.TestModelss;
```
xuzhg: Why do we have these two namespace?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -54,5 +56,73 @@ public void AddDynamicPropertyDictionary_ThrowsIfTypeIsNotDictionary()
...
                 string.Format("The argument must be of type '{0}'.", "IDictionary<string, object>"));
         }
+
+        /// <summary>
+        /// Tests the namespace assignment logic to ensure that user assigned namespaces are honored during registration.
+        /// </summary>
+        [Fact]
+        public void NamespaceAssignment_AutoAssignsNamespaceToStructuralType_AssignedNamespace()
+        {
+            // Act.
```
xuzhg: >// Act. [](start = 12, length = 7)

// Arrange & Act

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -54,5 +56,73 @@ public void AddDynamicPropertyDictionary_ThrowsIfTypeIsNotDictionary()
...
+            Assert.Equal(expectedNamespace, modelBuilder.ComplexType<ZipCode>().Namespace);
+        }
+
+        /// <summary>
+        /// Tests the namespace assignment logic to ensure that default CLR namespace is used if user does not assign one.
+        /// </summary>
+        [Fact]
+        public void NamespaceAssignment_AutoAssignsNamespaceToStructuralType_DefaultNamespace()
+        {
+            // Act.
```
xuzhg: >// Act. [](start = 12, length = 7)

// Arrange & Act

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -54,5 +56,73 @@ public void AddDynamicPropertyDictionary_ThrowsIfTypeIsNotDictionary()
...
+            Assert.Equal(typeof(ZipCode).Namespace, modelBuilder.ComplexType<ZipCode>().Namespace);
+        }
+
+        /// <summary>
+        /// Tests the namespace assignment logic to check the order in which namespace is assigned matters.
+        /// </summary>
+        [Fact]
+        public void NamespaceAssignment_AutoAssignsNamespaceToStructuralType_NamespaceAssignedAfterAddingEntities()
+        {
+            // Act.
```
xuzhg: >// Act. [](start = 12, length = 7)

// Arrange & Act

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -54,5 +56,73 @@ public void AddDynamicPropertyDictionary_ThrowsIfTypeIsNotDictionary()
...
+            string expectedNamespace = "TestingNamespace";
+            ODataConventionModelBuilder modelBuilder = new ODataConventionModelBuilder();
+
+            modelBuilder.EntitySet<Client>("clients");
+
+            modelBuilder.Namespace = expectedNamespace;
+            modelBuilder.ComplexType<ZipCode>();
```
xuzhg: >modelBuilder.ComplexType<ZipCode>(); [](start = 12, length = 36)

Can we add another test case, for example:

1. Set the namespace at builder,
2. but set another namespace for certain type?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/EnumTypeTest.cs
```diff
@@ -1127,6 +1127,38 @@ public void ODataConventionModelBuilder_DataContractAttribute_WithAddedExplicitl
             Assert.Contains(enumType.Members, (m) => m.Name.Equals("KeepDefaultName"));
         }
 
+        /// <summary>
+        /// Tests the namespace assignment logic to ensure that user assigned namespaces are honored during registration.
+        /// </summary>
+        [Fact]
+        public void NamespaceAssignment_AutoAssignsNamespaceToEnumType_AssignedNamespace()
+        {
+            // Act.
```
xuzhg: >// Act. [](start = 12, length = 7)

// Arrange & Act

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/EnumTypeTest.cs
```diff
@@ -1127,6 +1127,38 @@ public void ODataConventionModelBuilder_DataContractAttribute_WithAddedExplicitl
...
+            Assert.Equal("Test", modelBuilder.EnumType<Life>().Namespace);
+        }
+
+        /// <summary>
+        /// Tests the namespace assignment logic to ensure that user assigned namespaces are honored during registration.
+        /// </summary>
+        [Fact]
+        public void NamespaceAssignment_AutoAssignsNamespaceToEnumType_DefaultNamespace()
+        {
+            // Act.
```
xuzhg: >// Act. [](start = 12, length = 7)

// Arrange & Act

xuzhg:
@ramjsing@microsoft.com What's the change for "Nuget.exe"? It looks you update the "nuget.exe", which version are you using?

---
In reply to: [415674856](https://github.com/OData/webapi/pull/1592#issuecomment-415674856) [](ancestors = 415674856)

xuzhg:
@ramjsing@microsoft.com Would you please run the "Build.cmd" at your local to make sure all test cases are passed? Thanks.

---
In reply to: [425166242](https://github.com/OData/webapi/pull/1592#issuecomment-425166242) [](ancestors = 425166242,415674856)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
xuzhg: I think that's a typo. Would you please help fix it?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
RamjotSingh: Added a new test NamespaceAssignment_AutoAssignsNamespaceToStructuralType_AssignedNamespaceAndClassNamespace

RamjotSingh:
@xuzhg I am updating the nuget.exe to the latest version 4.7.1. The existing one did not have support for newer MSBuild (and VS2017)

RamjotSingh:
I ran the script and ensured all the tests passed
```[Success:] Pull Nuget Packages Success
[Info:] Building Asp.Net OData Projects...
[Info:] Building WebApiOData.AspNet.sln ...
[Success:] Build WebApiOData.AspNet.sln SUCCESS
[Info:] Building WebApiOData.AspNetCore.sln ...
[Success:] Build WebApiOData.AspNetCore.sln SUCCESS
[Info:] Building WebApiOData.E2E.AspNet.sln ...
[Success:] Build WebApiOData.E2E.AspNet.sln SUCCESS
[Info:] Building WebApiOData.E2E.AspNetCore.sln ...
[Success:] Build WebApiOData.E2E.AspNetCore.sln SUCCESS
[Success:] Build Done!

[Info:] Skip strong name validations for assemblies...
[Success:] SkipStrongName Done
[Info:] Testing Asp.Net OData Projects...
[Info:] Running test AspNetClassic UnitTest...
[Info:] Launching E:\Code\OData-WebApi\bin\Release\UnitTest\AspNet\Microsoft.AspNet.OData.Test.dll...
[Info:] Running test AspNetCore UnitTest...
[Info:] Launching E:\Code\OData-WebApi\test\UnitTest\Microsoft.AspNetCore.OData.Test\Microsoft.AspNetCore.OData.Test.csproj...
[Info:] Running test AspNetClassic E2ETests...
[Info:] Launching E:\Code\OData-WebApi\bin\Release\E2ETest\AspNet\Microsoft.Test.E2E.AspNet.OData.dll...
[Info:] Running test AspNetCore E2ETests...
[Info:] Launching E:\Code\OData-WebApi\bin\Release\E2ETest\AspNetCore\Microsoft.Test.E2E.AspNetCore.OData.dll...
[Info:] Test Done.
Collecting test results
Test summary:
Passed :        31975
Skipped:        2
Failed :        0
---------------
Total : 31975
Congratulation! All of the tests passed!
Build time:     00:00:28.8700421
Test time:      00:26:10.7634029
```

xuzhg:
@mikepizzo @KanishManuja-MS @biaol-odata Would you please take a look?

build.ps1
```diff
@@ -63,19 +63,47 @@ $LOGDIR = $ENLISTMENT_ROOT + "\bin"
...
+            $SN = $sdkVersionDirectory + "\sn.exe"
+            $SNx64 = $sdkVersionDirectory + "\x64\sn.exe"
+            break
+        }
+    }
+}
 
 # Use Visual Studio 2017 compiler for .NET Core and .NET Standard. Because VS2017 has different paths for different
 # versions, we have to check for each version. Meanwhile, the dotnet CLI is required to run the .NET Core unit tests in this script.
+# Furthurmore, Visual Studio 2017 has a Preview version as well which uses Microsoft Visual Studio\Preview as path instead of \2017
```
biaol-odata: Not sure why you need to update build.ps1 for this task. The dev environment typically requires VS2015 and VS2017 installed. Is this the case for your local dev environment as well?

RamjotSingh: I am running VS2017Preview version. Since we are asking people to run VS2017, dev can choose to run preview or stable version.

src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
```diff
@@ -44,7 +45,23 @@ public ODataModelBuilder()
...
+            get
+            {
+                return _namespace;
+            }
+            set
+            {
+                if (String.IsNullOrEmpty(value))
+                {
+                    throw Error.PropertyNull();
+                }
```
mikepizzo: I generally dislike having an initial state that can't be returned to.

Instead of throwing, perhaps we should set the namespace back to "Default" (and set HasAssignedNamespace back to false)?

src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
```diff
@@ -28,13 +28,14 @@ public class ODataModelBuilder
 
         private Version _dataServiceVersion;
         private Version _maxDataServiceVersion;
+        private string _namespace;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ODataModelBuilder"/> class.
         /// </summary>
         public ODataModelBuilder()
```
mikepizzo: >ODataModelBuilder [](start = 15, length = 17)

For code clarity, should initialize HasDefaultNamespace to false in the constructor.

src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
RamjotSingh: Default namespace is being set to "Default". This check is just to ensure that you can't set the value to null.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/ODataActionTests.cs
```diff
@@ -74,10 +74,10 @@ public async Task CanDispatch_ActionPayload_ToBoundAction()
         }
 
         private const string EntityPayload = @"{ 
-                ""Customer"": {""@odata.type"":""#Microsoft.AspNet.OData.Test.Formatter.Customer"", ""ID"":101,""Name"":""Avatar"" } , 
+                ""Customer"": {""@odata.type"":""#org.odata.Customer"", ""ID"":101,""Name"":""Avatar"" } ,
```
mikepizzo: >org.odata [](start = 50, length = 9)

Would prefer to use a namespace besides org.odata, which sounds official. maybe something like My.Schema?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ODataActionParametersTest.cs
```diff
@@ -23,9 +23,9 @@ public class ODataActionParametersTest
 
         [Theory]
         [InlineData("Drive", "Vehicles(Model=6,Name='6')/org.odata.Drive")]
-        [InlineData("Drive", "Vehicles(Model=6,Name='6')/Microsoft.AspNet.OData.Test.Builder.TestModels.Car/org.odata.Drive")]
+        [InlineData("Drive", "Vehicles(Model=6,Name='6')/org.odata.Car/org.odata.Drive")]
```
mikepizzo: >org.odata [](start = 57, length = 9)

Would prefer to use a namespace other than "org.odata", which sounds official. Maybe something like "My.Schema", or "TestNamespace" or "ns", similar to what is used in some of the other tests?

RamjotSingh: The org.data was not something I introduced. I just changed it to fix the test to actually honor the namespace.

src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
mikepizzo: I understand that the user can set the namespace to "Default", but, from the code perspective, this is (subtly) different (though currently functionality the same) than never having set it in the first place. I was wondering if setting the namespace to null shouldn't reset to the default state (set HasAssignedNamespace to false and namespace to "Default", rather than throw an error.

This is more of a philosophical thing; we have an initial state (HasAssignedNamespace) that can't be reset.  You can mimic the behavior, by knowing that the default namespace is "Default", and explicitly setting that, but there's no way to fully "undo" the fact that you touched this property.  So, for example, if we added other behavior based on whether the namespace was explicitly set, the developer would have no way to get back to that state.

---
In reply to: [234746322](https://github.com/OData/WebApi/pull/1592#discussion_r234746322) [](ancestors = 234746322)

RamjotSingh: Sure makes sense. Removing the error.

RamjotSingh: Done.

RamjotSingh:
Is 7.1 release after 7.0.2? I see the PR was marked for 7.0.2. Is it being moved to 7.1?

mikepizzo:
Since this could technically be a breaking change, we wanted to make the release that we are planning on doing this week 7.1.0 rather than 7.0.2, and include this change.

PR #1601, opened 2018-08-31
Fixes #1600 - ActionDescriptorExtensions is not thread safe

### LINKED ISSUE #1600: ActionDescriptorExtensions is not thread safe
GetEdmModel adds an entry to actionDescriptor.Properties if it does not exist but is not thread safe.
In the time between Properties.TryGetValue and Properties.Add, the property may have been added already and you get: `ArgumentException: An item with the same key has already been added` 

### Assemblies affected

Microsoft.AspNetCore.OData 7.0.1

### Reproduce steps

This will throw the expected exception
```
using System;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Http;
using Microsoft.AspNetCore.Http.Internal;
using Microsoft.AspNetCore.Mvc.Abstractions;
using Microsoft.AspNetCore.Mvc.ApplicationParts;
using Moq;
using Xunit;
using ActionDescriptorExtensions = Microsoft.AspNet.OData.Extensions.ActionDescriptorExtensions;

...

var services = new Mock<IServiceProvider>();
services.Setup(s => s.GetService(typeof(ApplicationPartManager))).Returns(new ApplicationPartManager());
var request = new DefaultHttpRequest(new DefaultHttpContext { RequestServices = services.Object });
var ad = new ActionDescriptor();
Parallel.For(0, 10, i => { ActionDescriptorExtensions.GetEdmModel(ad, request, typeof(string)); });
```

### Expected result

No error would be thrown

### Actual result

```
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Microsoft.AspNet.OData.Extensions.ActionDescriptorExtensions.GetEdmModel(ActionDescriptor actionDescriptor, HttpRequest request, Type entityClrType) in D:\dev\OData.AspNetCore\src\Microsoft.AspNetCore.OData\Extensions\ActionDescriptorExtensions.cs:line 54
```

### Additional detail

The example code is a minimal artificial example of the problem.  This surfaced when running NUnit tests in parallel using the [Parallelizable] attribute against the same API method.

freeranger:
### Issues

*This pull request fixes issue #1600.*

### Description

Adds a sync lock to avoid multiple instances adding the same key to the actionDescriptor.Properties dictionary

- [X] *Test cases added*
- [X ] *Build and test with one-click build and test script passed*

### Additional work necessary

Note that the tests pass if I correct a syntax error in PeopleController.cs, (missing comma) but that is outside the scope of this PR and should be fixed separately

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/ActionDescriptorExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/ActionDescriptorExtensions.cs
@@ -19,6 +19,8 @@ namespace Microsoft.AspNet.OData.Extensions
         // also clearly ties the property to code in this assembly.
         private const string ModelKeyPrefix = "Microsoft.AspNet.OData.Model+";
 
+        private static readonly object SyncLock = new object();
+
         internal static IEdmModel GetEdmModel(this ActionDescriptor actionDescriptor, HttpRequest request, Type entityClrType)
         {
             if (actionDescriptor == null)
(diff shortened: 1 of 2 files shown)


freeranger:
I see that https://github.com/OData/WebApi/pull/1615 has fixed the issue I mentioned re: PeopleController.

freeranger:
@xuzhg Does this PR need additional work?  Thanks

Issue #1597, opened 2018-08-30
Urls targeting collections of EdmComplexObjects fail to serialize.


mtroth-microsoft:
if I target the entity set:
/Entities/key
and the entity type contains a Collection(ComplexType) property, the result works as expected. But when I navigate directly to that property, the service errors out because the media formatter is unable to find a qualifying serializer to handle List<EdmComplexObject>.
/Entities/key/ListOfComplexObjects

this was tested against master.

xuzhg:
@mtroth-microsoft I don't know why did you get such error. It should work for the current ASP.NET OData library.

![image](https://user-images.githubusercontent.com/9426627/44885597-8dcde700-ac76-11e8-8f19-ee743cb3b544.png)

Did you implement the correct method in your entityset controller? In the convention routing, the method name should be : `Get{ComplexPropertyName}`

mtroth-microsoft:
Yes. Are you sure yours is a typeless model? I have no Clr classes anywhere in my component.

xuzhg:
Yes. It should work for untyped (typeless) model.
For example, I have the following untyped model. (Without CLR type).
```C#
        public static IEdmModel GetUntypedModel()
        {
            var model = new EdmModel();
            var customer = new EdmEntityType("NS", "Customer");
            customer.AddKeys(customer.AddStructuralProperty("Id", EdmPrimitiveTypeKind.Int32));
            model.AddElement(customer);

            var address = new EdmComplexType("NS", "Address");
            address.AddStructuralProperty("City", EdmPrimitiveTypeKind.String);
            model.AddElement(address);
            customer.AddStructuralProperty("Addresses", new EdmCollectionTypeReference(new EdmCollectionType(new EdmComplexTypeReference(address, true))));
            var container = new EdmEntityContainer("NS", "Default");
            var customers = new EdmEntitySet(container, "Customers", customer);
            model.AddElement(container);
            container.AddElement(customers);

            return model;
        }
```

In the controller, I have:


```C#

IEdmCollectionTypeReference edmType;
var model = Request.GetModel();
var address = model.SchemaElements.OfType<IEdmComplexType>().First();
edmType = new EdmCollectionTypeReference(new EdmCollectionType(new EdmComplexTypeReference(address, true)));
EdmComplexObjectCollection collection = new EdmComplexObjectCollection(edmType);

EdmComplexObject complex = new EdmComplexObject(address);
complex.TrySetPropertyValue("City", "RedmondUntyped");
[...]

PR #1607, opened 2018-09-04
Handle untyped collections in context's getedmtype method

mtroth-microsoft:
*This pull request fixes issue #1597

### Description

I could find no alternative to resolve a query using the ODataResourceSetSerializer than including the block of code that I have added to the ODataSerializerContext.GetEdmType method. Given that the method is called by the WriteObject method in the ResourceSet Serializer, please explain how to make that call succeed when a collection of EdmEntityObject or EdmComplexObject is provided given the remainder of the logic in the method. Note that I am producing the IEdmModel by directly consuming the csdl generated by a tool.


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -161,6 +163,22 @@ internal IEdmTypeReference GetEdmType(object instance, Type type)
                         typeof(IEdmObject).Name);
                 }
             }
+            else if (typeof(IEnumerable).IsAssignableFrom(type) &&
+                typeof(IEdmObject).IsAssignableFrom(type.GetGenericArguments().FirstOrDefault()))
```
xuzhg: Web API recommends to use "EdmEnittyObjectCollection, EdmComplexObjectCollection, EdmEnumObjectCollection" for the collection.

mtroth-microsoft: my approach is consistent with the approaches used in this class and others in webapi.

---
In reply to: [215081845](https://github.com/OData/WebApi/pull/1607#discussion_r215081845) [](ancestors = 215081845)

Issue #1617, opened 2018-09-20
Take and Top query use wrong Provider.CreateQuery Method


HRasch:
I use OData together with a AutoMapper expression conversion. When i use a query like 

http://localhost:54445/Article?$select=ArticleId,Name&$skip=1000&$top=10

an Exception is thrown.

The Reason for this exception is that the select creates a projection type, and the Take / Skip methods use the non-generic overload of CreateQuery from the inner Linq Provider. This Provider creates a new Query with the type of the original entity instead of the projection-type. When the top-parameter is applied, the query type doesn't match anymore.

### Assemblies affected

Microsoft.AspNet.OData v7.0.1
AutoMapper 6.2.2

### Reproduce steps

Create a OData-Service, 

use as Queryable a List of Elements, 
create a DTO for this Element, 
create a AutoMapper mapping, 
create a "DTO-Queryable" by using list.UseAsDataSource().For<Dto>(), 

use the Dto-Queryable as Source for a EnableQuery-Get Action

Execute a Query against this Controller with Select, Skip and Top parameter

### Expected result

Return the requested columns with count of top

### Actual result

[...]

PR #1618, opened 2018-09-20
Take and Top query use wrong Provider.CreateQuery Method #1617

HRasch:
### Issues

*This pull request fixes issue #1617*

### Description

IQueryableProvider.CreateQuery has changed the Element Type to the "native" Provider Element type. The generic overload with the current element type must be used instead.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/ExpressionHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/ExpressionHelpers.cs
@@ -27,13 +27,22 @@ namespace Microsoft.AspNet.OData
             Expression skipValueExpression = parameterize ? LinqParameterContainer.Parameterize(typeof(int), count) : Expression.Constant(count);
 
             Expression skipQuery = Expression.Call(null, skipMethod, new[] { query.Expression, skipValueExpression });
-            return query.Provider.CreateQuery(skipQuery);
+
+            var createMethod = query.Provider.GetType().GetMethods()
+                .First(x => x.Name == "CreateQuery" && x.IsGenericMethod)
+                .MakeGenericMethod(type);
+
+            return createMethod.Invoke(query.Provider, new[] { skipQuery }) as IQueryable;
         }
 
         public static IQueryable Take(IQueryable query, int count, Type type, bool parameterize)
         {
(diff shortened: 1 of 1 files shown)


HRasch:
Hello, is it possible to get any feedback on this PR? Can it be merged to the next version? Are there any open tasks I have to do?

src/Microsoft.AspNet.OData.Shared/ExpressionHelpers.cs
```diff
@@ -27,13 +27,26 @@ public static IQueryable Skip(IQueryable query, int count, Type type, bool param
             Expression skipValueExpression = parameterize ? LinqParameterContainer.Parameterize(typeof(int), count) : Expression.Constant(count);
 
             Expression skipQuery = Expression.Call(null, skipMethod, new[] { query.Expression, skipValueExpression });
-            return query.Provider.CreateQuery(skipQuery);
+
+            var createMethod = query.Provider.GetType().GetInterface(nameof(IQueryProvider))
+                .GetTypeInfo()
+                .GetDeclaredMethods("CreateQuery")
+                .First(x => x.IsGenericMethod)
```
xuzhg: Is it possible to cache the method searching?

biaol-odata:
@HRasch Looks good to me. But, as you mentioned above, this change is already covered by multiple tests? Should we enhance the test to verify the current element type is now used?

HRasch:
I said "is already covered by multiple tests" because there are many tests that cover the skip and take operator itself. This tests succeed even without my change, cause they are using the default List-IQueryable. In my case I have to use a converting IQueryable from AutoMapper (IQueryable.UseAsDataSource().For<T>()). The Provider here is generic, it creates the query with the ElementType from the underlying provider. I think the difference to a default list is, that if a normal list is projected, it also changes the type IQueryProvider returns on the non-generic createquery. So any later created query-expression automatically gets the elementtype of the projection. In case of the Automapper-QueryProvider is just returns an expression with the original elementtype. I think it's very complex to create a test-case to simulate the exact behavior, cause IQueryProvider must be mocked, to simulate the exact AutoMapper-Behavior.

PR #1641, opened 2018-10-11
Fix issue #1631, Singleton and single navigation property possible without key defined on its type

xuzhg:
…thout key defined on its type

### Issues

*This pull request fixes issue #1631.*

### Description

* Enable Singleton defined without key 
* Enable Single-Value navigation property defined without key.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Formatter;
 using Microsoft.OData.Edm;
+using Microsoft.OData.Edm.Validation;
 
 namespace Microsoft.AspNet.OData.Builder
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
```diff
@@ -608,20 +608,30 @@ public virtual void ValidateModel(IEdmModel model)
...
-                    throw Error.InvalidOperation(SRResources.EntityTypeDoesntHaveKeyDefined, entity.Name);
+                    throw Error.InvalidOperation(SRResources.EntitySetTypeHasNoKeys, entitySet.Name,
+                        entitySet.EntityType().FullName());
                 }
             }
 
-            foreach (IEdmNavigationSource navigationSource in model.EntityContainer.Elements.OfType<IEdmNavigationSource>())
+            // The type of collection navigation property should have key(s) defined.
+            HashSet<IEdmStructuredType> visited = new HashSet<IEdmStructuredType>();
+            foreach (IEdmStructuredType structuredType in model.SchemaElementsAcrossModels().OfType<IEdmStructuredType>())
```
mikepizzo: Is this used anywhere?

mikepizzo: Comment showed up on wrong line; is the HashSet "visited" defined on line 622 used anywhere?

---
In reply to: [227550096](https://github.com/OData/WebApi/pull/1641#discussion_r227550096) [](ancestors = 227550096)

mikepizzo:
>         public virtual void ValidateModel(IEdmModel model)

Is there a reason we do separate model validation in ModelBuilder, rather than just using the EdmValidator?

---
Refers to: src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs:604 in e5a844b. [](commit_id = e5a844be04cfdb747304d7665523bc7c7184687b, deletion_comment = False)

xuzhg:
@mikepizzo I think the main reason is to allow error tolerance. The validation rules by default maybe too heavy for some scenarios. However, when the customer get Edm model by calling `GetEdmModel()`, he can call `model.Validate(out errors)` to verify the model by himself. 
In Web API, we add the minimal validation rules for the model by introducing this separate model validation.

src/Microsoft.AspNet.OData.Shared/Builder/ODataModelBuilder.cs
xuzhg: #resolved.

Issue #1463, opened 2018-06-05
expand causes N +1 query problem


smasherprog:
[Table("Tracker", Schema = "HR")]
    public class Tracker 
    { 
       [System.ComponentModel.DataAnnotations.Key]
        public int Id { get; set; }
        [Required, MinLength(6), MaxLength(48), Column(TypeName = "varchar(48)")]
        public string EmployeeSid { get; set; }
        public virtual ICollection<TrackerInfo> TrackerInfos { get; set; }
    }
    [Table("TrackerInfo", Schema = "HR")]
    public class TrackerInfo 
    {       
        [System.ComponentModel.DataAnnotations.Key]
        public int Id { get; set; }
        public virtual Tracker Tracker { get; set; }
        [Required]
        public int TrackerId { get; set; }
    }

When expanding multiple queries are executed
Trackers?$expand=TrackerInfos

### Assemblies affected

7.0 Beta 4, .net 2.1 ef core 2.1

### Reproduce steps

Create two classes with one having a collection of the other. Try to do a Get with expand on the collection and the SQL generated will not be good. The query's that are generated are as below, https://github.com/aspnet/EntityFrameworkCore/issues/8531

### Expected result

The query should be a join on the collection and finish but it executes a where for each child row eventhough the data has already been retrieved. 

### Actual result

SELECT [a.TrackerInfos].[Id], [a.TrackerInfos].[CompletedBy], [a.TrackerInfos].[CompletedDate], [a.TrackerInfos].[CreatedDate], [a.TrackerInfos].[Key_Id], [a.TrackerInfos].[SupportType], [a.TrackerInfos].[TrackerId]
FROM [HR].[TrackerInfo] AS [a.TrackerInfos]
INNER JOIN (
    SELECT [a0].[Id]
    FROM [HR].[Tracker] AS [a0]
) AS [t0] ON [a.TrackerInfos].[TrackerId] = [t0].[Id]
ORDER BY [t0].[Id]
[07:11:07 INF] Executed DbCommand (3ms) [Parameters=[@_outer_Id='1'], CommandType='"Text"', CommandTimeout='30']
SELECT [t].[Id], [t].[CompletedBy], [t].[CompletedDate], [t].[CreatedDate], [t].[Key_Id], [t].[SupportType], [t].[TrackerId]
FROM [HR].[TrackerInfo] AS [t]
WHERE @_outer_Id = [t].[TrackerId]
[...]

PR #1653, opened 2018-10-21
Fixing N+1 query bug on expands

joshcomley:
### Issues

BUG: OData ignoring EntityFrameworkCore included data, forces a requery and N+1
https://github.com/OData/WebApi/issues/1463

### Description

The way OData wraps queries to allow for the dynamic nature of what OData can return is invoking the lazy loading capabilities of EF Core.

Take a simple example:

    http://host/odata/Customers?$expand=Orders

First the customers are fetched:

    SELECT ... FROM Customers

Then as the JSON serializer evaluates the results, for each customer a separate query is made to lazy load the orders:

    SELECT ... FROM Orders WHERE CustomerId = @Id

Thus for 100 customers returned, there will be 101 queries (one to get the customers, and another 100 to look up the orders for each customer), yielding an exponential performance decrease.

The solution is to force EF Core to employ eager loading and evaluate the query fully in advance (which is always preferable as if we are returning the JSON we know we need all the data requested in one go).

There are two areas to address this.

The expression generated for `/Customers?$expand=Orders` looks like the below, however this PR adds the missing `ToList()` (as indicated) to trigger eager loading of the child collection:

	{
	value(Microsoft.EntityFrameworkCore.Query.Internal.EntityQueryable`1[MsODataTest21.Customer])
		.Select
		(
			Param_0 => new SelectAllAndExpand`1() 
			{
				ModelID = "d1eb9bd5-df72-4404-9192-7f44cbc341e3", 
				Instance = Param_0, 
				UseInstanceForProperties = True, 
				Container = new NamedProperty`1() 
				{
					Name = "Orders", 
					Value = Param_0.Orders.Select(
						Param_1 => 
							new SelectAll`1() 
							{
								ModelID = "d1eb9bd5-df72-4404-9192-7f44cbc341e3", 
								Instance = Param_1, 
								UseInstanceForProperties = True
							})
							.ToList() // < ** This needs to be added **
				}
			}
		)
	}

Now we get two queries no matter the volume of customers returned:

    SELECT ... FROM Customers

Followed by:

	SELECT ...
	FROM Orders o
	INNER JOIN (
		SELECT Id
		FROM Customer
	) AS t ON o.CustomerId = t.Id

The second area of concern is when querying expanded child collectons:

    http://host/odata/Customers?$expand=Orders($filter=contains(DeliveryAddress,'b'))

This results in the below query (with the added `ToList()` from above):

	{
	value(Microsoft.EntityFrameworkCore.Query.Internal.EntityQueryable`1[MsODataTest21.Customer])
		.Select
		(
		Param_0 => new SelectAllAndExpand`1() 
			{
				ModelID = "d1eb9bd5-df72-4404-9192-7f44cbc341e3", 
				Instance = Param_0, 
				UseInstanceForProperties = True, 
				Container = new NamedProperty`1() 
				{
					Name = "Orders", 
					Value = 
						Param_0
							.Orders
							.AsQueryable() // < We need to get rid of this
							.Where($it =>
								$it.DeliveryAddress.Contains(
									value(Microsoft.AspNet.OData.Query.Expressions.LinqParameterContainer+TypedLinqParameterContainer`1[System.String]).TypedProperty))
										.Select
											(
											Param_1 => 
												new SelectAll`1() 
												{
													ModelID = "d1eb9bd5-df72-4404-9192-7f44cbc341e3", 
													Instance = Param_1, 
													UseInstanceForProperties = True
												}
											)
											.ToList()
				}
			}
		)
	}

Again, the unnecessary call to `AsQueryable()` (indicated) triggers lazy loading from EF Core, and even with the initial `ToList()` implemented will still result in an inefficient query. However, removing the `AsQueryable()` yields as-expected eagerly loaded results.

Removing the `AsQueryable()` results in two queries as before, with the second query honouring the filter.

- [x] *Test cases added* // Tests have been modified as necessary

### Additional work necessary

There may be other areas where lazy loading is triggered, as the `AsQueryable` method is still employed elsewhere.


src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -647,7 +642,9 @@ private Expression ProjectCollection(Expression source, Type elementType, Select
 
             // expression
             //      source.Select((ElementType element) => new Wrapper { })
-            Expression selectedExpresion = Expression.Call(GetSelectMethod(elementType, projection.Type), source, selector);
+            var selectMethod = GetSelectMethod(elementType, projection.Type);
```
biaol-odata: Creating this "selectMethod" variable is not strictly needed, right?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -647,7 +642,9 @@ private Expression ProjectCollection(Expression source, Type elementType, Select
...
+            var methodCallExpression = Expression.Call(selectMethod, source, selector);
+            Expression selectedExpresion = Expression.Call(ExpressionHelperMethods.QueryableToList.MakeGenericMethod(projection.Type), methodCallExpression);
```
biaol-odata: Do we have corresponding test for this addition of "ToList()"?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -210,17 +210,12 @@ internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IE
 
                 if (isCollection)
                 {
-                    Expression filterSource =
-                        typeof(IEnumerable).IsAssignableFrom(source.Type.GetProperty(propertyName).PropertyType)
-                            ? Expression.Call(
-                                ExpressionHelperMethods.QueryableAsQueryable.MakeGenericMethod(clrElementType),
-                                nullablePropertyValue)
-                            : nullablePropertyValue;
+                    Expression filterSource = nullablePropertyValue;
```
biaol-odata: Would removing the ternary operation cause some side effect?

biaol-odata:
@joshcomley Thanks for the PR. I am not very familiar with this area, but your description and summary looks good to me.

As for what you mentioned above, "There may be other areas where lazy loading is triggered, as the AsQueryable method is still employed elsewhere.", it would be nice if you can do a through check and see where-else such change should be applicable. Reducing N+1 to 2 is going to be a very good improvement on expanded query.

biaol-odata:
Previous build on your PR has a lot of errors that don't seem related to you changes. I just kicked off another build on your PR and see whether we can get a cleaner build.

biaol-odata:
@joshcomley The current build is passing. Can you please take a look at the comments, especially doing a through check on where-else this is also applicable? We would love to see N+1 --> 2.

joewashington75:
Hi - do we know when this will fix will be released?

mikeoconnor1980:
Hey @joshcomley , @biaol-odata @xuzhg,

Did you have an update on this? Sounds like the fix has been made, but needs reviewing.


Cheers,
Mike

Mhirji:
Hello,

Just wondering if this is a PR which will be merged (and when)?  If you could please let us know it would be great as there seem to be a few of us whom are suffering a bit from this the issues this PR fixes.

Thanks!!

Mhirji:
Hello again . . . Just wondering if this PR will be merged?

mikepizzo:
Calling ToList() will populate the entire results in memory which, for large results, could be problematic. Enumerating an IQueryable allows us to more efficiently stream the result.

Not sure why creating enumerating the queryable with an expand triggers lazy loading in EF; I would have expected that the $expand would result in eager loading.

floyd-may:
Is there a way to work around this issue in the meantime?

JoshZA:
👍

olakaldestad:
Four months ago this PR was moved from "under-review" to "Ready for review". Is it completely lost in limbo, or is there any chance it might get pulled for 7.2? @raheph @mikepizzo @xuzhg @KanishManuja-MS

NullMDR:
> Calling ToList() will populate the entire results in memory which, for large results, could be problematic. Enumerating an IQueryable allows us to more efficiently stream the result.
> 
> Not sure why creating enumerating the queryable with an expand triggers lazy loading in EF; I would have expected that the $expand would result in eager loading.

https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-2.1#optimization-of-correlated-subqueries

floyd-may:
@NullMDR not helpful.

olakaldestad:
@mikepizzo
> Calling ToList() will populate the entire results in memory

This is only true if you call it directly on the IQueryable, and is not at all applicable here... See the link provided by @NullMDR for documentation.

And again, is there a chance this will be included in 7.2? It adds a pretty massive optimization that is both well documented, extremely simple and (I presume) a complete dealbreaker for anyone dealing with something more complex than creating a To-Do list for CS101.

mikepizzo:
Thanks for the pointer @NullMDR.  We will talk with the EF folks to figure out the best path forward here, and consider a fix for 7.2. I am concerned that the "hint" of  using ToList() for EF may cause issues with other LINQ providers, so we may have to make this specific to EF (which I hate to do, but am willing to be pragmatic).  Perhaps the EF folks can help us find a cleaner solution.

bdebaere:
@mikepizzo Be very careful with this PR. Using `[EnableQuery(PageSize = 100)]` breaks this and reverts to the previous behavior of multiple queries being executed.

floyd-may:
I think it may be important to clarify what "multiple queries" means, especially in light of the link @NullMDR posted. Given N entities all being `$expand`ed in some way in the same call, expected behavior per EF Core may be up to N different queries. Just want to make sure that's clear in case anyone's expectation might be exactly one query. For N = 4 and 100 rows, say, that means 4 queries in direct EF code (not via OData). That's still a whole lot better than possibly up 301 queries.

albertwoo:
When can we expect the 7.2 to be released?

btull89:
I would also like to see this PR merged. I'm having to work off this branch specifically to avoid the n+1 issues with `$expand`.

PaulGilchrist:
This issue is dramatically impacting one of our companies largest projects.  Can we get an ETA on when this fix will be available?

MYoucef:
Could you let us know when this PR will be merged? Lazy loading dramatically impacts the performance of our application.

mikepizzo:
Engaging with the EF team to verify the approach for inclusion in our next release.

frblondin:
> Engaging with the EF team to verify the approach for inclusion in our next release.

Nice: as far as I understand there are concerns related to the amount of data being fetched by the single request (instead of N +1). In any case the data will be retrieved while serializing the data. If there are really performance ramifications, you could provide this mechanism through an option? Looks like some people are expecting this :)

Nice to hear that you are working on it.

mikepizzo:
@bdebaere: I didn't see anything in our code that would change how ToList() was appended based on whether or not paging was specified. Are you suggesting that we don't generate the ToList() if paging is specified, or that EF Core doesn't perform the optimization, even if ToList() is specific, if there is a take (or, presumably, any other options) specified on the nested query?

philewis:
Is there any update on this? This is having a huge impact on a large project at my company.

mikepizzo:
@philewis - updated PR passes all tests and will be included in our upcoming (7.2) release, due out in a couple of weeks.  Thanks for your patience!

philewis:
@mikepizzo I appreciate the reply, thank you! I look forward to the fix.

RMHR:
> @philewis - updated PR passes all tests and will be included in our upcoming (7.2) release, due out in a couple of weeks. Thanks for your patience!

Hi, its been 2 weeks, do you know when its going out?

PaulGilchrist:
I see that 7.2 has been released, but still seeing the N+1 behavior.  Is there a configuration that is needed to enable the fixed behavior?

mikepizzo:
@Paul: Yes, since the change may not always be the best plan, depending on your data, and works well for EF but could have different behavior against other LINQ providers, you have to enable the new behavior by setting EnableCorrelatedSubqueryBuffering = true on the EnableQueryAttribute.

erjicles:
I set EnableCorrelatedSubqueryBuffering = true on the EnableQueryAttribute, but I am still seeing the N+1 behavior.

mikepizzo:
@erjicles: can you share your query? Note that if you have any filters, top, etc. on the expand, EF will not be able to optimize and will revert to the N+1 behavior.

erjicles:
Sure thing. First just to make sure I'm using the latest, I updated Microsoft.AspNetCore.OData from 7.1.0 to 7.2.1. I'm using the following OData query to test this behavior:

localhost:5011/api/v1/customers?$top=10&$expand=addresses

This results in 11 queries - the first to select the top 10 customers, and one per customer to select their addresses. This happens with or without EnableCorrelatedSubqueryBuffering = true.

I tried various permutations on this query, including removing the $top, adding a $filter, and so on, but the result is the same in all cases.

PaulGilchrist:
It is unrealistic to expect that filters would not be used with expands.  Filters are the very thing that keeps the final size of the expanded data set within a reasonable size and returning only relevant data to the requester..

erjicles:
Is there any special setup that has to happen with the edm builder to get OData to recognize the navigation properties? I've tried a few things there as well but nothing has changed the behavior.

@PaulGilchrist - are you still experiencing the issue as well, or did EnableCorrelatedSubqueryBuffering = true solve your issue?

luanle1106:
I had same problem.

mikepizzo:
@PaulGilchrist; I agree with you, but there's nothing we can do from the OData side; EF is the one that is deciding how to execute the queries and has the N+1 behavior. The change we did provides a hint to EF to do two queries and then correlate the results locally, but I believe that EF is only able to perform that optimization if there is no filter on the expanded property (else they would have to evaluate that filter locally).

We'll have to work with the EF team to see if there is a pattern we can use to fetch the expanded content all in one request.

webbie70:
I ran into a problem with the new `EnableCorrelatedSubqueryBuffering` property.  The `ODataQuerySetting.CopyFrom(ODataQuerySettings settings)` method misses to copy the value over to the new settings object. This leads to the flag being reset to false by the `UpdateQuerySettings()` extension method.

dsidedp:
At least destination object gets resolved from DI and registered as scoped.
So, as a temporary solution this can be used
```
((ODataQuerySettings)query.Context.RequestContainer.GetService(typeof(ODataQuerySettings))).EnableCorrelatedSubqueryBuffering = true;
```

PaulGilchrist:
> At least destination object gets resolved from DI and registered as scoped.
> So, as a temporary solution this can be used somewhere earlier in a request call chain
> 
> ```
> ((ODataQuerySettings)query.Context.RequestContainer
>    .GetService(typeof(ODataQuerySettings))).EnableCorrelatedSubqueryBuffering = true;
> ```

Can this be done globally?

dsidedp:
Sure, you just need to override default factory for ODataQuerySettings like this:
```
app.UseMvc(b =>
    b.EnableDependencyInjection(cfg =>
        cfg.AddService(
            Microsoft.OData.ServiceLifetime.Scoped,
            typeof(ODataQuerySettings),
            _ => new ODataQuerySettings { EnableCorrelatedSubqueryBuffering = true })));
```

ankitpatel5989:
Still not working. I have defined mentioned attribute(EnableCorrelatedSubqueryBuffering ) in get request and also updated setting in startup.cs (EnableDependencyInjection)

Please help me to resolve it.

PR #1658, opened 2018-10-24
Fix the security vulnerabilities in project dependencies

xuzhg:
### Issues

*This pull request fixes issue the security vulnerabilities*

### Description

1. Updated the Microsoft.Data.OData to 5.8.4
2. Updated the Microsoft.AspNetCore.All to 2.0.9
3. Updated the Microsoft.AspNetCore.Server.Kestrel.* to 2.0.4

image

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Microsoft.Test.E2E.AspNet.OData.csproj
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Microsoft.Test.E2E.AspNet.OData.csproj
@@ -27,11 +27,13 @@
       <HintPath>..\..\..\..\sln\packages\EntityFramework.6.1.3\lib\net45\EntityFramework.SqlServer.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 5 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/App.config
```diff
@@ -54,6 +54,18 @@
         <bindingRedirect oldVersion="0.0.0.0-2.7.41101.299" newVersion="2.7.41101.299" />
       </dependentAssembly>
 
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.Data.Edm" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-5.8.4.0" newVersion="5.8.4.0" />
```
biaol-odata: I would expect these ODL-5.x dependencies already exist in the config and we just need to bump the versions to 5.8.4, instead of adding them?

xuzhg: Removed. Thanks.

Issue #1659, opened 2018-10-29
Odata V4 group by with Top and skip not working


SachinKotnala:
I am not able to get groupby to work with top and skip which should be most simplest thing. I am fetching data using OData and able to get output using the query below

https://localhost:6523/api/OData/AssetUsage/?$apply=groupby((assetId,Asset/name),aggregate(interactions with sum as totalInteractions, uniqueInteractions with sum as totalUniqueInteractions))&$orderBy=totalInteractions asc

Output of the above query

However when I try to apply skip or top on the above generated output, I get an error mentioned below

https://localhost:6523/api/OData/AssetUsage/?$apply=groupby((assetId,Asset/name),aggregate(interactions with sum as totalInteractions, uniqueInteractions with sum as totalUniqueInteractions))&$orderBy=totalInteractions asc&$top=1

Error -

message": "The query specified in the URI is not valid. Could not find a property named 'Asset' on type 'Portal.Models.Entities.AssetUsage'.",

"innererror":

-"message": "Could not find a property named 'Asset' on type 'Portal.Models.Entities.AssetUsage'.",

-"type": "Microsoft.OData.ODataException",

Entity class structure

public class AssetUsage
{
    public int Id { get; set; }
    [ForeignKey("Asset")]
    public int? AssetId { get; set; }
    public int YearMonthId { get; set; }
    public int Interactions { get; set; }
    public int UniqueInteractions { get; set; }
    public DateTimeOffset Recency { get; set; }
    public virtual Asset Asset { get; set; }

}

pielegacy:
We're needing to implement pagination powered by OData and this limitation is affecting our application as well. This seems like a pretty standard use case for OData and if this cannot be fixed we'll be forced to look into GraphQL which seems to [handle this kind of logic in a much cleaner fashion](https://graphql.org/learn/pagination/).

biaol-odata:
@SachinKotnala Error is related to query option parsing. Not sure why the extra $top (or $skip as you mentioned) triggered the error. Do yo have more detail information such as inner stack trace available?

SachinKotnala:
@biaol-odata Thanks for investigating this issue. I have attached inner stack trace below.

{
    "error": {
        "code": "",
        "message": "The query specified in the URI is not valid. Could not find a property named 'Asset' on type 'Portal.Models.Entities.AssetUsage'.",
        
"innererror": {
            
[...]

PR #1664, opened 2018-11-02
Individual query options to utilize DI config

biaol-odata:
…nfig is available.

### Issues

*This pull request fixes issue #1659.*

### Description

*ODataQueryOptions already use dependency-injected config settings from container. We should make such injected config settings available for sub-query option classes such as OrderByQueryOption, FilterQueryOption, SelectExpandQueryOption; the rest such as Top, Skip are not needed for now since value types are int-type (case-insensitive), but could be added if desired.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/OrderByQueryOption.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/OrderByQueryOption.cs
@@ -9,7 +9,6 @@ using System.Linq.Expressions;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Query.Expressions;
 using Microsoft.AspNet.OData.Query.Validators;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNet.OData.Shared/Query/OrderByQueryOption.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Query.Expressions;
 using Microsoft.AspNet.OData.Query.Validators;
+using Microsoft.Extensions.DependencyInjection;
```
KanishManuja-MS: I don't think you need this.

biaol-odata: Yup, removed.

biaol-odata:
@mikepizzo and @KanishManuja-MS : Thanks for the CR!

PR #1667, opened 2018-11-02
Allow ODataFeature.TotalCount set to null.

biaol-odata:
### Issues

*This pull request fixes issue #1637.*

### Description

*Removed the null check, this will align with IODataFeature which already has a nullable type for TotalCount.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageProperties.cs
+++ b/src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageProperties.cs
@@ -104,9 +104,9 @@ namespace Microsoft.AspNet.OData.Extensions
                 object totalCount;
                 if (_request.Properties.TryGetValue(TotalCountKey, out totalCount))
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageProperties.cs
```diff
@@ -104,9 +104,9 @@ public ODataPath Path
                 object totalCount;
                 if (_request.Properties.TryGetValue(TotalCountKey, out totalCount))
                 {
-                    // Fairly big problem if following cast fails. Indicates something else is writing properties with
-                    // names we've chosen. Do not silently return null because that will hide the problem.
-                    return (long)totalCount;
+                    // TotalCount is defined as null-able long, allowed to be set as null, so it should be
```
KanishManuja-MS: It seems like we restricted setting total count as null intentionally. I think things can get messy if some uses the same property name which can also be set to null.

biaol-odata:
Thanks Mike!

Issue #1657, opened 2018-10-23
OData $batch queries do not honor/preserve prefer headers.


preston-m-price:
I am researching an issue affecting our service where **prefer** headers are not preserved in $batch queries.

### Assemblies affected

OData WebApi lib 6.2.2

### Reproduce steps
For example this GET works as expected:
```
GET https://myservice/_odata/Entity?$select=Field1,Field2 HTTP/1.1
prefer: odata.maxpagesize=5
```
But this $batch request body does not:
```
--batch_e17f47a9-41f9-45ad-b59b-98282d06975c
Content-Type: application/http
Content-Transfer-Encoding: binary
prefer: odata.maxpagesize=5

GET https://myservice/_odata/Entity?$select=Field1,Field2 HTTP/1.1

--batch_e17f47a9-41f9-45ad-b59b-98282d06975c
```
### Expected result

The $batch response should honor the **prefer** header and respond with a limited page size and nextLink.

### Actual result

The $batch request does not honor the **prefer** header.

### Additional detail

The GET request honors the prefer header and we get a page size of 5 with a nextLink. However, the response from the $batch request does not honor the page size. Digging into the code I see the entire prefer header is missing from the ODataBatchRequestItem returned by the DefaultODataBatchHandler.ParseBatchRequestsAsync method. The [spec for multipart media types](https://tools.ietf.org/html/rfc2046#page-17) suggests that any header that does not start with “Content-“ MAY be ignored but SHOULD be retained. Is this behavior by design, or should these headers in $batch operations be retained?

I have not tested with other headers but I suspect **prefer** is not the only header affected by this behavior.

KanishManuja-MS:
@preston-m-price According to the [standard](http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#_Toc505771136), a batch request should not support this preference header. However, we did find a regression in our 7.x version of the library for the single GET request. Thanks for writing this issue, please let us know if having this with every GET request is too painful.

preston-m-price:
I am confused by your response. The spec you referenced says: 

> The maxpagesize preference SHOULD NOT be applied to a batch request, but MAY be applied to individual requests within a batch.

In my $batch example above the header is being set in the POST body for the GET sub-request, not on the POST $batch request itself. So it seems like this should be in spec, right? Or maybe I'm misunderstanding. Does the regression you reference affect the general GET scenario or the GET sub-request within a $batch?

[...]

PR #1673, opened 2018-11-07
Respect preference header for paging

KanishManuja-MS:
### Issues

*This pull request fixes issue #1657*

### Description
As mentioned in the issue, a client should be able to specify a page size preference in the header but WebAPI did not respect the header. 

With this PR-

- The methods using the enable query attribute will respect odata.maxpagesize and maxpagesize in the preference header.
- Customers can use public static methods on the helper class to get the value of the prefer header for the page size. 

- [x] *Test cases added*

### Additional work necessary
The next step should be for the framework to send preference-applied header in the response but since it is optional, I am keeping the size of the fix for this release small due to time constraint. 
For the initial issue to be fully resolved for the batch queries, changes in ODL will be required.

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/RequestPreferenceHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/RequestPreferenceHelpers.cs
@@ -66,8 +66,7 @@ namespace Microsoft.AspNet.OData
             }
             else
             {
-                maxPageSize = maxPageSize.ToLower();
-                int index = maxPageSize.IndexOf(preferenceHeaderName) + preferenceHeaderName.Length;
+                int index = maxPageSize.IndexOf(preferenceHeaderName,System.StringComparison.OrdinalIgnoreCase) + preferenceHeaderName.Length;
                 string value = "";
                 if(maxPageSize[index++]=='=')
                 {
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNet.OData.Shared/RequestPreferenceHelpers.cs
```diff
@@ -33,6 +36,54 @@ internal static bool RequestPrefersReturnNoContent(IWebApiHeaders headers)
...
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        private static int GetMaxPageSize(IEnumerable<string> preferences, string preferenceHeaderName)
+        {
+            const int Failed = -1;
+            string maxPageSize = preferences.FirstOrDefault(s => s.Contains(preferenceHeaderName));
```
mikepizzo: >preferenceHeaderName [](start = 76, length = 20)

Should we do case insensitive comparison here?

KanishManuja-MS: Thanks for the catch. There is no overload available for Contains that would ignore the case hence using the IndexOf instead of Contains.

mikepizzo: Should probably apply case-insensitivity to other preferences as well.

---
In reply to: [231698281](https://github.com/OData/WebApi/pull/1673#discussion_r231698281) [](ancestors = 231698281)

Issue #15, opened 2015-01-12
Consider automated server paging based on ResultLimit with next page links that use $skiptoken instead of $skip


LianwMS:
Next page links right now:  

[ http://localhost/Customers?$skip=10 Skiptoken is more performant and robust
](http://localhost/Customers?$skip=10

Skiptoken is more performant and robust) .  
## Work Item Details

**Original CodePlex Issue:**    [Issue 500](http://aspnetwebstack.codeplex.com/workitem/500)
**Status:** Active
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Oct 3, 2012 at 11:54 PM
**Reported by:**    youssefm
**Updated on:** Nov 24, 2014 at 11:54 PM
**Updated by:** goldenfire6

chuanboz:
Any reason this is marked as low priority?

Issue #1639, opened 2018-10-09
Incorrect `NextLink` generated when '$skip' and '$top' operators are not all lowercase


julealgon:
When sending requests with `$skip` and `$top` not all on lowercase, the `nextLink` generated by `GetNextPageLink(int)` incorrectly fails to detect them and creates invalid URLs.

### Assemblies affected

Microsoft.AspNetCore.OData 7.0.1

### Reproduce steps

Call an OData API with uppercase `$skip` or `$top` and manually write the `@odata.nextlink` value by using `GetNextPageLink(int)`.

Sample:
- Request: `http://localhost/api/randomdata?$SKIP=1`
- PageSize: 3

### Expected result

`"@odata.nextLink":"http://localhost/api/randomdata?$skip=4"`

### Actual result

`"@odata.nextLink":"http://localhost/api/randomdata?$SKIP=1&$skip=2"`

(notice the 2 "skips" in the final URL)

### Additional detail

https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs#L27-L43

As can be seen here, there is no case handling at all on the query string keys, meaning that only completely lowercase will match and apply the logic.

Since OData operators can have any casing, there needs to be some handling here.

Suggestion:
- Change `switch(key)` to `switch(key.ToLower())` here:

https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs#L31

KanishManuja-MS:
@julealgon Thanks for catching this. I will fix it with my next PR.

Issue #1638, opened 2018-10-09
Incorrect '$top' value generated when using 'Request.GetNextPageLink(int)'


julealgon:
When both server-driven paging and client-driven paging are involved (that is, there is a max pagesize configured on the server, and a `$top=X` is specified in the query), the generated `nextLink` value when using `GetNextPageLink(int)` is incorrect and sometimes invalid.

### Assemblies affected

Microsoft.AspNetCore.OData 7.0.1

### Reproduce steps

1. Create an OData API with server-driven paging. Set paging size to 3 (for example)
2. Configure the API to allow `$top`, max value 2 (for example)
3. Manually handle Skip/Top via `ODataQuerySettings`, and set NextLink value using `GetNextPageLink` method
3. Run query by specifying a `$top` value and observe returned `@odata.nextlink` value

Sample:
```csharp
[Page(MaxTop = 2, PageSize = 3)]
public class SampleModel
{
    public int Id { get; set; }

    public string Text { get; set; }
}

[ODataRoute("randomdata")]
public IActionResult GetData(ODataQueryOptions<SampleModel> oDataQueryOptions)
{
    var data = new[]
    {
        new SampleModel{Id = 1, Text = "First" },
        new SampleModel{Id = 2, Text = "Second" },
        new SampleModel{Id = 3, Text = "Third" },
        new SampleModel{Id = 4, Text = "Fourth" },
        new SampleModel{Id = 5, Text = "Fifth" },
        new SampleModel{Id = 6, Text = "Sixth" },
        new SampleModel{Id = 7, Text = "Seventh" },
        new SampleModel{Id = 8, Text = "Eight" },
        new SampleModel{Id = 9, Text = "Ninth" }
    };

    var pageSize = 3;
    var results = data
        .Skip(oDataQueryOptions.Skip?.Value ?? 0)
        .Take(oDataQueryOptions.Top?.Value ?? pageSize);

    Request.ODataFeature().NextLink = Request.GetNextPageLink(pageSize);

    return new OkObjectResult(results);
}
```

Request:
```http://localhost/api/randomdata?$top=2```

### Expected result

```json
{
	"@odata.context": "http://localhost/api/$metadata#randomdata",
	"value": [{
			"Id": 1,
			"Text": "First"
		}, {
			"Id": 2,
			"Text": "Second"
		}
	],
	"@odata.nextLink": "http://localhost/api/randomdata?$top=2&$skip=3"
}
```

### Actual result

[...]

Issue #1633, opened 2018-10-03
`NextLink` and `TotalCount` are always `null` when using `ODataQueryOptions` without `ApplyTo`?


julealgon:
When using an explicit `ODataQueryOptions` without applying it to a `IQueryable` instance (that is, manually handling the various options), it appears that `NextLink` and `TotalCount` properties are always set to `null` in the request.

How do I make the OData framework understand the values I'm using to generate a proper link and total count values?

### Assemblies affected

Microsoft.AspNetCore.OData 7.0.1

### Reproduce steps

```csharp
public IEnumerable<char> GetData(ODataQueryOptions oDataQueryOptions)
{
    var data = "123456789".ToCharArray();

    oDataQueryOptions.Validate(new OdataValidationSettings
    {
        AllowedQueryOptions = AllowedQueryOptions.Skip | AllowedQueryOptions.Top | AllowedQueryOptions.Count,
        MaxTop = 5
    };

    return new PageResult<char>(
        data.Skip(oDataQueryOptions.Skip?.Value ?? 0)
            .Take(oDataQueryOptions.Top?.Value ?? 3),
        Request.ODataFeature().NextLink,
        Request.ODataFeature().TotalCount);
}
```

### Expected result

NextLink and Count properties returned to the client when more data is available.

### Actual result

Raw results without NextLink and Count.

### Additional detail

While I understand the framework wouldn't magically know the values for those, I'm also unsure about how to "inform" it about the values. The method above is a sample, but in the real code I won't be able to call this on top of a `IQueryable` because we are using Dapper: I need to pass the skip/top values into out stored procedures which will do the paging.

I was unable to find any documentation whatsoever about using those extensions with manual filtering.

Am I supposed to not use the `ODataFeature` extension for those and create them by myself somehow?

julealgon:
This appears to be directly related to https://github.com/OData/WebApi/issues/377

KanishManuja-MS:
@julealgon Please feel free to review the following design- https://github.com/OData/WebApi/blob/f0baa4f8dc37be3500c5d8602fd633d1235b2205/_posts/2018-10-15-12-2-Use-SkipToken-for-Server-Side-Paging.md

We want to make it easier to use, your input will be appreciated. I am planning on adding sample code for a use case similar to yours.

KanishManuja-MS:
[...]

PR #1689, opened 2018-11-21
Implement $skiptoken for paging

KanishManuja-MS:
### Issues
This is a feature which also fixes the following issues - 
*This pull request fixes issue #15 #986 #1633 #1639 #1638 *

### Description

For more information about Skiptoken and the high-level overview of the implementation, you can check the [design](http://odata.github.io/WebApi/#12-2-Use-SkipToken-for-Server-Side-Paging) out. 

- [x] *Test cases added*

### Additional work necessary
Add more test cases, add more validation checks, add the sample. 
These work items will be on hold until the first pass of code review.


src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.AspNet.OData
     /// </summary>
     internal static partial class GetNextPageHelper
     {
-        internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<string, string>> queryParameters, int pageSize)
+        internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<string, string>> queryParameters, int pageSize, string skipTokenValue = "")
```
xuzhg: >= "" [](start = 155, length = 4)

why can't use 'null"?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableQueryTests.cs
```diff
@@ -43,7 +43,7 @@ public class EnableQueryTests
             {
                 return new TheoryDataSet<string, string>
                 {
-                    {"?$skiptoken=5", "SkipToken"},
```
xuzhg: >{"?$skiptoken=5", "SkipToken"}, [](start = 20, length = 31)

should add it somewhereelse?

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        /// <summary>
+        /// Bad Documentation
+        /// </summary>
+        /// <param name="result"></param>
+        /// <param name="model"></param>
+        /// <param name="orderByQueryOption"></param>
+        /// <returns></returns>
+        public static string GetSkipTokenValue(IQueryable result, IEdmModel model, OrderByQueryOption orderByQueryOption)
```
xuzhg: >GetSkipTokenValue [](start = 29, length = 17)

We should provide a DI for customer to create their own Skiptokenvalue generating logic.

src/Microsoft.AspNet.OData.Shared/Query/Validators/SkipTokenQueryValidator.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// Represents a validator used to validate a <see cref="SkipTokenQueryOption"/> based on the <see cref="ODataValidationSettings"/>.
+    /// </summary>
+    public class SkipTokenQueryValidator
+    {
+        /// <summary>
+        /// Validates a <see cref="SkipTokenQueryOption" />.
+        /// </summary>
+        /// <param name="skipToken">The $skip query.</param>
```
xuzhg: >$skip [](start = 40, length = 5)

$skiptoken

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
```diff
@@ -58,6 +58,18 @@ public IEdmTypeReference GetEdmType()
             return model.GetEdmTypeReference(elementType);
         }
 
+        public IEdmType GetEdmType(IEdmModel model)
```
xuzhg: >GetEdmType [](start = 24, length = 10)

Why can't use above "GetEdmType()" method?

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
```diff
@@ -116,14 +116,27 @@ public ODataDeserializerProvider DeserializerProvider
...
+
         /// <summary>
         /// Get the next page link for a given page size.
         /// </summary>
         /// <param name="pageSize">The page size.</param>
         /// <returns></returns>
         public Uri GetNextPageLink(int pageSize)
         {
-            return this.innerRequest.GetNextPageLink(pageSize);
+            this.Context.NextLink = this.innerRequest.GetNextPageLink(pageSize);
```
mikepizzo: Help me understand this code; it looks like calling GetNextPageLink has the side-effect of setting the NextLink on the context, but if I don't call GetNextPageLink then the NextLink doesn't get set on the context?

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
mikepizzo: +1

---
In reply to: [238494040](https://github.com/OData/WebApi/pull/1689#discussion_r238494040) [](ancestors = 238494040)

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -23,35 +23,54 @@ internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<str
             StringBuilder queryBuilder = new StringBuilder();
 
             int nextPageSkip = pageSize;
+            
+            //If no value for skiptoken is provided; revert to using skip 
+            bool useSkipToken = true;
+            if (String.IsNullOrWhiteSpace(skipTokenValue))
+            {
+                useSkipToken = false;
+            }
```
mikepizzo: bool useSkipToken = !String.IsNullOrWhiteSpace(skipTokenValue);

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -23,35 +23,54 @@ internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<str
...
                         int top;
                         if (Int32.TryParse(value, out top))
                         {
-                            // There is no next page if the $top query option's value is less than or equal to the page size.
+                            // There is no next page if the $top query option's value is less than or equal to the page size. You should not call this API if top <= pagesize.
                             Contract.Assert(top > pageSize);
                             // We decrease top by the pageSize because that's the number of results we're returning in the current page
-                            value = (top - pageSize).ToString(CultureInfo.InvariantCulture);
+                            if (top > pageSize)
+                            {
```
mikepizzo: Do you need this, since you assert top > pageSize above?

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -23,35 +23,54 @@ internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<str
...
+                                // We increase skip by the pageSize because that's the number of results we're returning in the current page
+                                nextPageSkip += skip;
+                            }
                         }
                         continue;
                     default:
                         break;
                 }
 
+                if ((key == "$skip" && useSkipToken) || (key == "$skiptoken" && useSkipToken))
```
mikepizzo: > && useSkipToken [](start = 76, length = 16)

so it's valid to pass $skiptoken in as a query parameter (and it's written with its current value) as long as I don't provide a separate skipTokenValue parameter when calling the GetNextPageLink?

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
KanishManuja-MS: Prior to my change, Context.NextLink did not get set if GetNextPageLink method got called. The consumer of the API had to explicitly set that after making a call to this public method. To make this paging API easier to consume, I decided to move the setting of the Context.NextLink inside the public method as I thought it is not intuitive for consumers to set the context.nextlink explicitly.

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
KanishManuja-MS: Not for the [EnableQuery] codeflow but if someone tries to use request.GetNextPageLink while they are applying query options without the framework, it makes sense to throw an error in debug mode if they are not checking the top value and pagesize before calling this API.

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
KanishManuja-MS: Yes, it does sound weird. The reason for this are following -
1. Issue #986 requested a behavior like this.
2. The request for last page of results may have $skiptoken as a query parameter but no skiptoken value, which is a valid case that would require the $skiptoken to be omitted from the url.

Disclaimer for 2nd point - If EnableQuery is used, this code will not be called for the last page. The point is valid only for using the API explicitly.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableQueryTests.cs
KanishManuja-MS: Added to otherSupportedTestData

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
KanishManuja-MS: I agree with that in the longer term vision but we can discuss if that has to be part of this PR or we will develop it later once people start using SkipToken

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -756,6 +756,12 @@ public static bool HasPrecision(EdmPrimitiveTypeKind primitiveTypeKind)
                     primitiveTypeKind == EdmPrimitiveTypeKind.TimeOfDay);
         }
 
+        public static IEdmTypeReference GetTypeReferenceOfProperty(IEdmModel model, Type clrType, string propertyName)
+        {
+            PropertyInfo propertyInfo = clrType.GetProperty(propertyName);
```
mikepizzo: This is a very expensive operation.  What are the scenarios where we call GetTypeReferenceOfProperty? is there another way to get the same information? Can the information be cached?

mikepizzo:
>                  querySettings.PageSize.HasValue))

don't we need to take skiptoken into account here?

---
Refers to: src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs:371 in 555cae6. [](commit_id = 555cae69a7cf1f8e87f86e84b6d96378ab958800, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// </summary>
+        public ODataQueryContext Context { get; private set; }
+
+        /// <summary>
+        /// Gets the raw $skiptoken value.
+        /// </summary>
+        public string RawValue { get; private set; }
+
+        /// <summary>
+        /// Gets the value of the $skiptoken as a parsed integer.
```
mikepizzo: >integer [](start = 57, length = 7)

why integer?

KanishManuja-MS: Incorrect documentation, leftover from the prototype implementation.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+                else
+                {
+                    compare = binder.CreateBinaryExpression(BinaryOperatorKind.GreaterThan, property, constant, true);
+                }
+
+                Expression condition = Expression.AndAlso(lastEquality, compare);
+                where = where == null ? condition : Expression.OrElse(where, condition);
+
+                lastEquality = Expression.AndAlso(lastEquality, binder.CreateBinaryExpression(BinaryOperatorKind.Equal, property, constant, true));
+                count++;
```
mikepizzo: >count++; [](start = 16, length = 8)

what is count used for? I see it is initialized, then incremented in the loop, but don't see where its value is used.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+
+        /// <summary>
+        /// XML doc, make it public to avoid security issues
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="records"></param>
+        /// <returns></returns>
+        public static T FetchLastItemInResults<T>(IQueryable<T> records)
+        {
+            T[] arr = records.AsArray();
```
mikepizzo: Ouch! This will execute the query and enumerate through all of the results; is there any way to defer getting the results until the results have been serialized?

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+            return arr[arr.Length - 1];
```
mikepizzo: Can you use records.LastOrDefault(), so that you don't have to materialize the entire result as an array?

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+            return results;
+        }
+
+        /// <summary>
+        /// XML doc, make it public to avoid security issues
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="records"></param>
+        /// <returns></returns>
+        public static T FetchLastItemInResults<T>(IQueryable<T> records)
```
mikepizzo: >records [](start = 64, length = 7)

what happens if records is an empty result?

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
```diff
@@ -416,12 +426,20 @@ public virtual IQueryable ApplyTo(IQueryable query, ODataQuerySettings querySett
...
                     InternalRequest.Context.NextLink == null)
                 {
-                    Uri nextPageLink = InternalRequest.GetNextPageLink(pageSize);
-                    InternalRequest.Context.NextLink = nextPageLink;
+                    DefaultQuerySettings settings = Context.RequestContainer.GetRequiredService<DefaultQuerySettings>();
+                    if (settings.EnableSkipToken)
+                    {
+                        string skipTokenValue = SkipTokenQueryOption.GetSkipTokenValue(result, this.Context.Model, OrderBy);
+                        InternalRequest.GetNextPageLink(pageSize, skipTokenValue);
+                    }
```
xuzhg: We should move this to serialization instead of trying to get the last element here.

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -225,6 +225,30 @@ public static Uri GetNextPageLink(this HttpRequestMessage request, int pageSize)
...
+        {
+            if (request == null || request.RequestUri == null)
+            {
+                throw Error.ArgumentNull("request");
+            }
+
+            Uri requestUri = request.RequestUri;
+
+            if (!requestUri.IsAbsoluteUri)
+            {
```
xuzhg: Should this allow for relative urls?

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.AspNet.OData
     /// </summary>
     internal static partial class GetNextPageHelper
     {
-        internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<string, string>> queryParameters, int pageSize)
+        internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<string, string>> queryParameters, int pageSize, string skipTokenValue = null)
```
xuzhg: consider generating the $skiptoken value here.

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -24,23 +24,36 @@ internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<str
...
+            //If no value for skiptoken is provided; revert to using skip 
+            bool useSkipToken = !String.IsNullOrWhiteSpace(skipTokenValue);
+
             foreach (KeyValuePair<string, string> kvp in queryParameters)
             {
-                string key = kvp.Key;
+                string key = kvp.Key.ToLowerInvariant();
                 string value = kvp.Value;
+
+                if ((key == "$skip" && useSkipToken) || (key == "$skiptoken"))
```
xuzhg: move $skiptoken to switch; and the skip to the switch statement as well.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,352 @@
...
+        private void ParseKeyValuePairs(string rawValue)
+        {
+            _propertyValuePairs = new Dictionary<string, object>();
+            string[] keyValues = rawValue.Split(',');
+            foreach (string keyAndValue in keyValues)
+            {
+                string[] pieces = keyAndValue.Split(new char[] { ':' }, 2);
+                if (pieces.Length > 1 && !String.IsNullOrWhiteSpace(pieces[1]))
+                {
+                    IEdmTypeReference type = EdmLibHelpers.GetTypeReferenceOfProperty(Context.Model, Context.ElementClrType, pieces[0]);
```
xuzhg: >ElementClrType [](start = 109, length = 14)

Consider Untyped properties, how will we handle those?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1025,5 +1045,24 @@ internal static Expression ConvertNull(Expression expression, Type type)
...
             }
         }
+
+        /// <summary>
+        /// Compares two guids
+        /// </summary>
+        /// <param name="firstValue"></param>
+        /// <param name="secondValue"></param>
+        /// <returns>An integer value based on the Guid's CompareTo method</returns>
+        public static int GuidCompare(Guid firstValue, Guid secondValue)
```
xuzhg: Verify how guids compare to null.

src/Microsoft.AspNet.OData.Shared/Query/Validators/SkipTokenQueryValidator.cs
```diff
@@ -0,0 +1,43 @@
...
+            if (skipToken == null)
+            {
+                throw Error.ArgumentNull("skipQueryOption");
+            }
+
+            if (validationSettings == null)
+            {
+                throw Error.ArgumentNull("validationSettings");
+            }
+            //See what additional checks need to be in place given the validation settings
```
xuzhg: Add the checks.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -26,5 +26,10 @@ public class ODataOptions
         /// Gets or Sets a value indicating if batch requests should continue on error.
         /// </summary>
         public bool EnableContinueOnErrorHeader { get; set; }
+
+        /// <summary>
+        /// Gets or Sets a value which determines if skip or skipToken gets used for server-side paging.
+        /// </summary>
+        public bool UseSkipTokenForServerSidePaging { get; set; }
```
xuzhg: >UseSkipTokenForServerSidePaging [](start = 20, length = 31)

This does not get used.

KanishManuja-MS: Removed.

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
KanishManuja-MS: fixed.

src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
KanishManuja-MS: Moved the skiptoken generation code to this.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
KanishManuja-MS: We would default back to $skip. Since, we are guaranteed that there are more objects than the page size. However, we don't call this anymore.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
KanishManuja-MS: Changed the approach.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
KanishManuja-MS: Fixed. Now, we will defer the generation of $skiptoken value till the serialization.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
KanishManuja-MS: Leftover from prototype, deleted.

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
KanishManuja-MS: We needed this only for enum types until ODL can start parsing enums without the type reference being passed.
Now, we append the value with 'enumType' to indicate the type and only in those cases try to obtain the typereference.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
KanishManuja-MS: Changed it to use the above API.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
KanishManuja-MS: We will use $skip if there is a dynamic property used in the orderby clause.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -328,12 +334,30 @@ private IEnumerable<ODataOperation> CreateODataOperations(IEnumerable<IEdmOperat
...
+           // newUri.SelectAndExpand = writeContext.SelectExpandClause;
+           // if (writeContext.SelectExpandClause != null)
+           // {
+           //     foreach(ExpandedNavigationSelectItem item in writeContext.SelectExpandClause.SelectedItems)
+           //     {
+           //         newUri.OrderBy = item.OrderByOption;
+           //         newUri.Filter = item.FilterOption;
+           //     }
+           // }
+           //// Uri finalNestedUri = newUri.BuildUri(ODataUrlKeyDelimiter.Slash);
```
xuzhg: If that's a case, file an issue and copy the link here?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -36,6 +37,20 @@ public ODataSerializerContext()
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataSerializerContext"/> class.
+        /// </summary>
+        /// <param name="resource">The resource whose navigation property is being expanded.</param>
+        /// <param name="selectExpandClause">The <see cref="SelectExpandClause"/> for the navigation property being expanded.</param>
+        /// <param name="edmProperty">The complex property being nested or the navigation property being expanded.
+        /// If the resource property is the dynamic complex, the resource property is null.
+        /// </param>
+        /// <param name="queryOptions">The <see cref="ODataQueryOptions"/> for the navigation property being expanded.</param>
+        public ODataSerializerContext(ResourceContext resource, SelectExpandClause selectExpandClause, IEdmProperty edmProperty, ODataQueryOptions queryOptions)
```
xuzhg: >ODataQueryOptions [](start = 129, length = 17)

Why do you want to save the whole (ODataQueryOption?)

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
```diff
@@ -0,0 +1,160 @@
...
+        /// <summary>
+        /// Apply the $skiptoken query to the given IQueryable.
+        /// </summary>
+        /// <param name="query">The original <see cref="IQueryable"/>.</param>
+        /// <param name="querySettings">The query settings to use while applying this query option.</param>
+        /// <param name="orderByNodes">Information about the orderby query option.</param>
+        /// <returns>The new <see cref="IQueryable"/> after the skiptoken query has been applied to.</returns>
+        public IQueryable<T> ApplyTo<T>(IQueryable<T> query, ODataQuerySettings querySettings, IList<OrderByNode> orderByNodes)
+        {
+            return skipToken.ApplyTo<T>(query, querySettings, orderByNodes) as IOrderedQueryable<T>;
```
xuzhg: >skipToken [](start = 19, length = 9)

rename it

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -264,8 +264,10 @@ public static ETag<TEntity> GetETag<TEntity>(this HttpRequest request, EntityTag
         /// </summary>
         /// <param name="request">The request on which to base the next page link.</param>
         /// <param name="pageSize">The number of results allowed per page.</param>
+        /// <param name="lastValue"></param>
```
xuzhg: >"lastValue"></param> [](start = 24, length = 20)

make it complete

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
KanishManuja-MS: This was done as per Mike's suggestion. We do this to avoid creating OrderByNodes again from the OrderByClause and with a breaking change for a major release we may remove SelectExpandClause as a separate property as well.

Since this should only be a reference, I don't think it has much performance impact.

src/Microsoft.AspNet.OData.Shared/Query/SkipTokenQueryOption.cs
Galad: Hello,
For what it is worth we have such a use case where we need to generate a different skiptoken and apply it to the query in a different way than currently implemented.

KanishManuja-MS: @Galad Thanks for the feedback. This PR is outdated now, I would be pleased if you can take a look at #1723 . You would be able to inject your own implementation that derives from SkipTokenHandler class. Please let me know if you find any shortcomings with that.

Galad: @KanishManuja-MS Thanks a lot, I will take a look

Galad:
Hello,
So I was able to try extending the SkipTokenHandler and I was able to do what I needed to do with it.
I just have a concern about concurrent requests which cannot be handled properly with the current design. I will explicit it as a comment in in the code.

src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
```diff
@@ -0,0 +1,301 @@
...
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNet.OData.Query
+{
+    /// <summary>
+    /// Default implementation of SkipTokenHandler for the service. 
+    /// </summary>
+    public class DefaultSkipTokenHandler : SkipTokenHandler
+    {
+        private IDictionary<string, object> _propertyValuePairs;
```
Galad: I have a concern about storing the values of the current request in the skip token handler. This, in conjunction with the scope of the skip token handler being Singleton, will prevent concurrent requests to work properly, because concurrent requests will write in the same dictionary.

My suggestion is to use the return value of the method `ProcessSkipTokenValue` (which is currently dismissed by the caller), and pass it to the `Apply` methods.

Galad: Sorry don't mind the comment I mixed up the pull requests.

Issue #1565, opened 2018-08-09
Odata Count Incorrect values on Odata Function [.Net Core]


albertoVieiraNeto:
### Short Description

After creating a custom Function on odata if one the navigation properties is expanded and a count added inside this expand, the numbers returned are completely wrong.

### Assemblies affected

Version affected/Tested :
Microsoft.AspNetCore.OData @7.0.1

### Reproduce steps

register a custom function on the model builder:
```csharp
   builder.EntitySet<Post>("Post");
            builder
                .EntityType<Post>()
                .Collection
                .Function("Mine")
                .ReturnsCollectionFromEntitySet<Post>(nameof(Post));
```

Create a function on the controller:
```csharp
        [EnableQuery]
        [HttpGet]
        public IQueryable<Post> Mine(/*int? group = null*/)
        {

            var SubjectId = User.Claims.GetSubjectId();
            IQueryable<Post> result;

            result = (from postSubs in _DbContext.PostGroupSubscribe
                      join post in _DbContext.Post on postSubs.PostGroup.Id equals post.Group.Id
                      where postSubs.User.Id == SubjectId
                      select post);

            return result;
        }
```

### Models Used

```csharp
    public class Post : ModelBaseWithId
    {
        [StringLength(4000)]
        public string Content { get; set; }

        [StringLength(4000)]
        public string NormalizedText { get; set; }

        public DateTimeOffset ExpireDate { get; set; } = DateTimeOffset.Now;

        public virtual PostGroupCategory Category { get; set; }

        public virtual ICollection<PostLikes> Likes { get; set; }

        public virtual UserAccount Owner { get; set; }

        public virtual Post Parent { get; set; }

        public virtual ICollection<Post> Children { get; set; }

        public virtual ICollection<PostFiles> Files { get; set; }

        public virtual PostGroup Group { get; set; }

    }
```

```csharp
    public class PostGroupSubscribe: ModelBaseWithId
    {
        public int PostGroupId { get; set; }
        public virtual PostGroup PostGroup { get; set; }

        public Guid UserId { get; set; }
        public virtual UserAccount User { get; set; }
    }
```

```csharp
    public abstract class ModelBaseWithId
    {
        [Required]
        [Key]
        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
        public int Id { get; set; }

        [Required]
        [DatabaseGenerated(DatabaseGeneratedOption.Computed)]
        public virtual Guid Guid { get; set; }

[...]

PR #1693, opened 2018-11-28
Fix assignment of CollectionExpandedProperty.TotalCount when using EF

gunars-skangals:
### Issues

This pull request fixes issue #1565 

### Description

As the expression contained in NamedPropertyExpression.TotalCount can be either Int64 or Nullable<Int64> depending on ODataQuerySettings.HandleNullPropagation, cast result of expression to the nullable type before assigning, otherwise the property is set to some random number.

- [X] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
@@ -118,7 +118,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
                 if (property.CountOption != null && property.CountOption.Value)
                 {
-                    memberBindings.Add(Expression.Bind(namedPropertyType.GetProperty("TotalCount"), property.TotalCount));
+                    memberBindings.Add(Expression.Bind(namedPropertyType.GetProperty("TotalCount"), ExpressionHelpers.ToNullable(property.TotalCount)));
                 }
             }
             else
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/SelectExpandNestedDollarCountTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/SelectExpandNestedDollarCountTest.cs
@@ -112,9 +112,31 @@ namespace Microsoft.AspNet.OData.Test
             Assert.Contains("\"Categories@odata.count\":9", payload); // Categories
         }
 
-        private Task<HttpResponseMessage> GetResponse(string uri, string acceptHeader)
+        [Fact]
+        public async Task SelectExpand_WithOneLevelNestedDollarCount_HandleNullPropagationFalse_Works()
         {
-            var controllers = new[] { typeof(MsCustomersController), typeof(MetadataController) };
+            // Arrange
+            string uri = "/odata/MsOrders?$expand=Categories($count=true)";
+
+            // Act
+            HttpResponseMessage response = await GetResponse(uri, AcceptJson);
+
+            // Assert
+            Assert.NotNull(response);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            string payload = await response.Content.ReadAsStringAsync();
+
+            Assert.DoesNotContain("\"Orders@odata.count\":5,", payload); // Top (Orders)
+            Assert.Contains("\"Categories@odata.count\":9", payload); // Categories
+        }
+
+        private Task<HttpResponseMessage> GetResponse(string uri, string acceptHeader, bool handleNullPropagation = true)
+        {
+            var controllers = new[] {
+                typeof(MsCustomersController),
+                typeof(MsOrdersController),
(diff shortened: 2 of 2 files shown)


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/SelectExpandNestedDollarCountTest.cs
```diff
@@ -112,9 +112,31 @@ public async Task SelectExpand_WithAllDollarCount_Works()
...
+            Assert.NotNull(response);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            string payload = await response.Content.ReadAsStringAsync();
+
+            Assert.DoesNotContain("\"Orders@odata.count\":5,", payload); // Top (Orders)
+            Assert.Contains("\"Categories@odata.count\":9", payload); // Categories
+        }
+
+        private Task<HttpResponseMessage> GetResponse(string uri, string acceptHeader, bool handleNullPropagation = true)
```
mikepizzo: >bool handleNullPropagation = true [](start = 87, length = 33)

Please add variations to the test for handle null propagation equal to both true and false (i.e., turn this method into [Theory] with [InlineData] for handleNullPropagation = true and handleNullPropagation=false). Better yet, make all of the tests that assume handleNullPropagation=false into [Theory] that run with handleNullPropagation equal to both true and false.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
```diff
@@ -118,7 +118,7 @@ private static Expression CreateNamedPropertyCreationExpression(NamedPropertyExp
 
                 if (property.CountOption != null && property.CountOption.Value)
                 {
-                    memberBindings.Add(Expression.Bind(namedPropertyType.GetProperty("TotalCount"), property.TotalCount));
+                    memberBindings.Add(Expression.Bind(namedPropertyType.GetProperty("TotalCount"), ExpressionHelpers.ToNullable(property.TotalCount)));
```
mikepizzo: >ExpressionHelpers.ToNullable [](start = 100, length = 28)

Should this be dependent upon whether or not null propagation handling is true or false? i.e., should we only call ExpressionHelpers.ToNullable if null propagation handling is true?

mikepizzo:
@gunars-skangals - thanks for the pull request to fix this issue!

See my comments inline. Should we only call ExpressionHelpers.ToNullable if null propagation handling is true?

If you can follow up, I'd love to get this fix integrated into our next WebAPI OData build.

Thanks!

src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
gunars-skangals: it has been a while and had almost forgotten about this

from what i can recall, the property that is being set is of type long? (class CollectionExpandedProperty? ), no matter what the settings are, and for assignment to work correctly, the assigned expression should be converted to nullable anyway

i don't remember if the expression type was correct in HandleNullPropagationOption.True case, but ExpressionHelpers.ToNullable does not do anything if the expression is of nullable type already
i don't see a straightforward way of getting that setting value inside of the PropertyContainer class anyway so imho this should stay as is

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/SelectExpandNestedDollarCountTest.cs
gunars-skangals: removed the redundant argument - HandleNullPropagationOption is set on controller action level anyway

do not quite understand the point: SelectExpand_WithOneLevelNestedDollarCount_HandleNullPropagationFalse_Works is already a variation on test SelectExpand_WithOneLevelNestedDollarCount_Works which tests the same thing (count field for a nested collection) but with the default HandleNullPropagationOption value of True, which depends on the namespace of query provider (class HandleNullPropagationOptionHelper, method GetDefaultHandleNullPropagationOption)

mikepizzo: Right. My preference would be to have unit tests that explicitly tests both the null propagation true case as well as the null propagation false case. I realize that the existing test on ln 38 doesn't specify null propagation, which picks up the default, which for the current query provider happens to be null propagation true, but that's not explicit in the test so a future change to that test, such as using a different query provider, could inadvertently invalidate the intention that we have coverage for both null propagation false and null propagation true. 

One way to address this would be to specifically set HandleNullPropagation to OData.Query.HandleNullPropagationOption.True in the MsCustomersController, so that we are explicit that we are testing a handle null propagation value of true, rather than relying on a query provider-specific default value.

---
In reply to: [300698007](https://github.com/OData/WebApi/pull/1693#discussion_r300698007) [](ancestors = 300698007)

Issue #1714, opened 2018-12-17
Nav/any(t:false) converted to Source.Any() instead of Source.Any(_=>false) or just false


kosinsky:
Nav/any(t:false) converted to Source.Any() instead of Source.Any(_=>false) or just false

### Assemblies affected

OData WebApi lib 6.0+

### Reproduce steps

*Try following expression Category/QueryableProducts/any(P: false)*

### Expected result

*$it => false*

### Actual result

*$it => $it.Category.QueryableProducts.Any()*

PR #1715, opened 2018-12-17
Fix for Nav/any(f:false) conversion

kosinsky:
### Issues

*This pull request fixes issue #1714.*

### Description

*Any expression with contradiction is converted to false.*

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -1369,11 +1369,18 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
             Expression body = null;
             // uri parser places an Constant node with value true for empty any() body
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -1375,6 +1375,13 @@ public virtual Expression BindAnyNode(AnyNode anyNode)
                 body = ApplyNullPropagationForFilterBody(body);
                 body = Expression.Lambda(body, anyIt);
             }
+            else if (anyNode.Body != null && anyNode.Body.Kind == QueryNodeKind.Constant
+                && (bool)(anyNode.Body as ConstantNode).Value == false)
```
xuzhg: >(bool) [](start = 19, length = 6)

is this cast safe?

kosinsky: Yes. We are checking anyNode.Body.Kind == QueryNodeKind.Constant first and casting only after that

Issue #1154, opened 2017-12-09
[feature/netcore] Support aggregate on EF Core


techniq:
It appears using `$apply=aggregate(...)` on the alpha release of Microsoft.AspNetCore.OData is not supported.

For example, the following works on WebApi 5.10.0 on .NET Framework:

`/odata/Employees?$apply=aggregate(HoursPerWeek%20with%20average%20as%20AvgHoursPerWeek)`

but returns a 406 (Not Acceptable) on Microsoft.AspNetCore.OData.dll, 7.0.0-Nightly201712061320 using .NET Core 2.0.

techniq:
I just pulled the latest from @robward-ms's [NetCore-Latest](https://github.com/robward-ms/WebApi/tree/NetCore-Latest) branch (7.0.0-...) and I am no longer receiving a 406 response but instead receive a 200 response with only a partial body:

```
{"@odata.context":"http://localhost:5000/odata/$metadata#Employees(AvgHoursPerWeek)","value":[
```

It looks to be due to the following exception

<details>
  <summary>
System.NotSupportedException: Could not parse expression '$it.AsQueryable()': This overload of the method 'System.Linq.Queryable.AsQueryable' is currently not supported.
</summary>

```
[40m[32minfo[39m[22m[49m: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:5000/odata/Employees?$apply=aggregate(HoursPerWeek%20with%20average%20as%20AvgHoursPerWeek)  
Microsoft.AspNetCore.Hosting.Internal.WebHost:Information: Request starting HTTP/1.1 GET http://localhost:5000/odata/Employees?$apply=aggregate(HoursPerWeek%20with%20average%20as%20AvgHoursPerWeek)  
[40m[32minfo[39m[22m[49m: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[1]
      Executing action method odatatest.Controllers.EmployeesController.Get (odata-test) with arguments ((null)) - ModelState is Valid
Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:Information: Executing action method odatatest.Controllers.EmployeesController.Get (odata-test) with arguments ((null)) - ModelState is Valid
[40m[32minfo[39m[22m[49m: Microsoft.EntityFrameworkCore.Infrastructure[10403]
[...]

Issue #1257, opened 2018-03-02
[feature/netcore] Ability to modify query after applying queryOptions


techniq:
### Assemblies affected

- `7.0.0-Nightly201803011314`

### Reproduce steps

In OData 6.x, it used to be possible to cast the result of `queryOptions.ApplyTo(...)` as an `IQueryable<T>` where `T` is the type of the `DbContext.Set<T>()` and make further changes.

For example, I would use this to add additional fields, apply additional filtering, etc before returning a result.  A use case would be to provide an `Export` capability, such as this trivial example.

```c#
[HttpGet]
public IActionResult Export(ODataQueryOptions<Employee> queryOptions)
{
    var query = queryOptions.ApplyTo(DbContext.Set<Employee>()) as IQueryable<Employee>;
    var exportQuery = query.Select(e => new EmployeeExport
    {
        Name = e.Name,
        // ...
    });
    return Ok(exportQuery);
}
```

### Expected result

Ability to continue to modify query after calling `queryOptions.ApplyTo(...)`

### Actual result

If the odata query contains anything except for `$filter` (for example, `$top`, `$select`, or `$expand`), the result is not castable to `IQueryable<T>` (returns null)

techniq:
@robward-ms Any thoughts on this one?

I'm currently using `beta-2`(as I can no longer install recent [nightlies](https://github.com/OData/WebApi/issues/1321)).  There are some [E2E](https://github.com/OData/WebApi/blob/feature/netcore/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/QueryComposition/ODataQueryOptionsTests.cs#L59-L75) tests that appear to cast to `IQueryable<T>` but they are not using `$select` or `$expand`.

This is currently a big blocker for my project and was hoping to find some solution (and is a regression from 6.x).

robward-ms:
@techniq - What is the backend you are using? I think it's EFCore but I want to verify.

techniq:
Correct, EFCore (SQL Server)

techniq:
@robward-ms any thoughts on this one?

robward-ms:
@techniq - My thoughts are the query expression building in OData does not play well with EFCore in general :).

[...]

PR #1728, opened 2019-01-09
Support $apply for EF Core

kosinsky:
### Issues

*This pull request fixes issues #1154 and #1257. Contributes to #669*

### Description

We are generating expressions that look like
```
Next = new AggregationPropertyContainer() {
       Name = "Prop2",
       Value = $it.Prop2,
       Next = …
}                    

```
Value type is object, as a result proper .NET expression must look like as 
```
Next = new AggregationPropertyContainer() {
       Name = "Prop2",
       Value = (object) $it.Prop2,
       Next = …
}    
``` 
or AccessViolationException will be thrown when executing on real objects. On other hand, casts to object isn't translatable by EF6 as a result skipping (object) in that case.
We used to do that adjustment for Linq to Objects, EF Core could execute some parts client side and uses linq to object as result.

Also had to tune how aggregation on top of IGrouping works. Linq to Objects/EF Core and EF 6 has different limitations

- [x] *Test cases added*


jannikbuschke:
@kosinsky  @robward-ms is there anything stopping from releasing this fix? Im eagerly waiting for $apply 👍

madansr7:
@jannikbuschke, we have been working to release a few important updates in ODL so haven't reviewed this one yet. WebAPI's next release would be in march and i am hoping this PR will be part of that release.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Aggregation/AggregationContextCore.cs
```diff
@@ -0,0 +1,33 @@
...
+            base.OnConfiguring(optionsBuilder);
+        }
+    }
+
+
+    public class AggregationContextCoreSql : AggregationContextCoreBase
+    {
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder.UseSqlServer(@"Data Source=(LocalDb)\MSSQLLocalDB;Integrated Security=True;Persist Security Info = True;Database = AggregationTestCore1");
```
xuzhg: >AggregationTestCore1 [](start = 142, length = 20)

maybe rename it.

xuzhg:
It seems good to me.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/AggregationBinder.cs
```diff
@@ -161,6 +161,18 @@ public IQueryable Bind(IQueryable query)
             return result;
         }
 
+        /// <summary>
+        /// Checks IQueryable provider for need of EF6 oprimization
```
mikepizzo: >oprimization [](start = 55, length = 12)

nit: spelling "optimization"

src/Microsoft.AspNet.OData.Shared/Query/Expressions/AggregationBinder.cs
```diff
@@ -30,7 +30,7 @@ internal class AggregationBinder : ExpressionBinderBase
 
         private Type _groupByClrType;
 
-        private bool _linqToObjectMode = false;
+        private bool _classicEF = false;
```
mikepizzo: Note that we effectively switch the default here, since _linqToObjectMode has the same default value, but the opposite behavior, of _classicEF. That's fine, as long as bind is called to explicitly set this value. I assume there are no scenarios where this value is used without first calling bind?

src/Microsoft.AspNet.OData.Shared/Query/HandleNullPropagationOptionHelper.cs
```diff
@@ -9,11 +9,11 @@ namespace Microsoft.AspNet.OData.Query
...
+        internal const string EntityFrameworkQueryProviderNamespace = "System.Data.Entity.Internal.Linq";
 
-        private const string ObjectContextQueryProviderNamespaceEF5 = "System.Data.Objects.ELinq";
-        private const string ObjectContextQueryProviderNamespaceEF6 = "System.Data.Entity.Core.Objects.ELinq";
-        private const string ObjectContextQueryProviderNamespaceEFCore2 = "Microsoft.EntityFrameworkCore.Query.Internal";
+        internal const string ObjectContextQueryProviderNamespaceEF5 = "System.Data.Objects.ELinq";
+        internal const string ObjectContextQueryProviderNamespaceEF6 = "System.Data.Entity.Core.Objects.ELinq";
+        internal const string ObjectContextQueryProviderNamespaceEFCore2 = "Microsoft.EntityFrameworkCore.Query.Internal";
 
         private const string Linq2SqlQueryProviderNamespace = "System.Data.Linq";
```
mikepizzo: >System.Data.Linq [](start = 63, length = 16)

Are we correctly handling Linq2SqlQueryProviderNamespace, or did we change defaults for this (and other) non-ef linq providers? Should non-ef linq providers have the _classicEF behavior? Or would the previous code (likely) not have worked for these providers?

kosinsky: Previous version was never tested with LINQ 2 SQL. I don't see any LINQ2SQL tests in the repo. I could try to add them.

kosinsky: I did some digging and Linq2SQL doesn't work with aggregations even without my changes. I managed to support only simple case like` $apply=aggregate($count as Count)`, but as soon as I'm adding groupby or more expressions to aggregate I'm getting exceptions like:

```
Unhandled Exception: System.NotSupportedException: The member 'GroupByWrapper.Container' has no supported translation to SQL.
   at System.Data.Linq.SqlClient.PostBindDotNetConverter.Visitor.VisitMember(SqlMember m)
   at System.Data.Linq.SqlClient.SqlVisitor.Visit(SqlNode node)
   at System.Data.Linq.SqlClient.SqlVisitor.VisitExpression(SqlExpression exp)
   at System.Data.Linq.SqlClient.SqlVisitor.VisitUnaryOperator(SqlUnary uo)
   at System.Data.Linq.SqlClient.PostBindDotNetConverter.Visitor.VisitUnaryOperator(SqlUnary uo)
```

Also other query options isn't 100$ supported. For example, `$select=Id,Name` works, but $select=Id,Name&$expand=Orders fails with:


```
Could not translate expression 'Table(Customer).Select(Param_0 => new SelectSome`1() {ModelID = \"8bdc9a3a-5ff6-486e-bbe3-53c69ce10dcf\", Container = new SingleExpandedPropertyWithNext1`1() {Name = \"Order\", Value = new SelectAll`1() {ModelID = \"8bdc9a3a-5ff6-486e-bbe3-53c69ce10dcf\", Instance = Param_0.Order, UseInstanceForProperties = True}, Next0 = new NamedProperty`1() {Name = \"Id\", Value = Convert(Param_0.Id)}, Next1 = new NamedProperty`1() {Name = \"Name\", Value = Param_0.Name}, IsNull = (Param_0.Order == null)}})' into SQL and could not treat it as a local expression.
```

@mikepizzo  Is it acceptable to resolve that comment by throwing not supported exception when Linq2SQL is used and creating an issue to eventually support Linq2SQL?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/AggregationBinder.cs
kosinsky: New default behavior is correct form .NET point of view. Expression like: object a = 15; must look like   object a = (object) 15; on MSIL to do proper boxing or AccessViolationException will be thrown. Unfortunately, EF6 can't translate that to SQL. So we have to do hack for EF6. The same for IEnumerable and IGroupingBy

src/Microsoft.AspNet.OData.Shared/Query/HandleNullPropagationOptionHelper.cs
kosinsky: @mikepizzo Is it ok for me to block Linq2SQL due to issue mentioned above?

mikepizzo: Yes; that's fine with me, as long as you throw a clear error for the new scenarios and aren't blocking scenarios that worked previously.

kosinsky:
@mikepizzo do you have any other feedback to this PR?

TehWardy:
Is this part of the current version of Microsoft.AspNetCore.OData ?
I'm using version 7.4.1 at the moment and I can't seem to get any form aggregation / apply operation working.

TehWardy:
It looks like the limitation here may be in EF in my case ...

based on the above my query:
> ?$apply=groupby((DebtorName),aggregate($count as Count))
should be fine within odata, the exception given (as shown below) implies that this is a known issue with EF ... 

```
2020-08-07 20:08:29,196 [9] ERROR  Api.IAppBuilderExtensions - Processing of the LINQ expression '(GroupByShaperExpression:
KeySelector: new GroupByWrapper{ GroupByContainer = new LastInChain{ 
        Name = (N'DebtorName'), 
        Value = (a.DebtorName) 
    }
     }
, 
ElementSelector:(EntityShaperExpression: 
    EntityType: ActiveTransaction
    ValueBufferExpression: 
        (ProjectionBindingExpression: EmptyProjectionMember)
    IsNullable: False
)
)' by 'RelationalProjectionBindingExpressionVisitor' failed. This may indicate either a bug or a limitation in EF Core. See https://go.microsoft.com/fwlink/?linkid=2101433 for more detailed information.
```

Issue #1725, opened 2019-01-04
No coercion operator is defined between types 'Microsoft.OData.Edm.Date' and 'System.DateTimeOffset'.


svyme:
We decided to finally update move from oData 6.14 to 7.5 to enjoy new features, fixes, and improvements.
Went through all required code changes and all looked good until we got to testing endpoints where $filter contains a date field ($filter=AsOfDate eq 2018-01-04)

Getting the below exception when applying query options:
No coercion operator is defined between types 'Microsoft.OData.Edm.Date' and 'System.DateTimeOffset'.

I see two more issues were already opened to report this bug. Both were closed without any solution. How come?
A workaround to require API consumers to change just because we update internal library is not a workable approach.

We were able to reproduce this with versions 7.5, 7. But not version 6.14.

madansr7:
@svyme , could you please your metadata with us? Are you using EF or EF core?

svyme:
Stack trace:
   at System.Linq.Expressions.Expression.GetUserDefinedCoercionOrThrow(ExpressionType coercionType, Expression expression, Type convertToType)
   at System.Linq.Expressions.Expression.Convert(Expression expression, Type type, MethodInfo method)
   at System.Linq.Expressions.Expression.Convert(Expression expression, Type type)
   at Microsoft.AspNet.OData.Query.Expressions.ExpressionBinderBase.CreateConvertExpression(ConvertNode convertNode, Expression source)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.BindConvertNode(ConvertNode convertNode)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.BindSingleValueNode(SingleValueNode node)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.Bind(QueryNode node)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.BindBinaryOperatorNode(BinaryOperatorNode binaryOperatorNode)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.BindSingleValueNode(SingleValueNode node)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.Bind(QueryNode node)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.BindConvertNode(ConvertNode convertNode)
   at Microsoft.AspNet.OData.Query.Expressions.FilterBinder.BindSingleValueNode(SingleValueNode node)
[...]

PR #1733, opened 2019-01-11
Fix issue #1725 and related: make filter DateTimeOffset with Date work

xuzhg:
### Issues

*This pull request fixes issue #1725 and related.*
Similar: 

OData/odata.net#1205
#1473

It looks many customers complain about this regression. So, this PR is to fix the regression.

Now, with this change, you can run the "AspNewCoreODataSample.Web" and file requests:

```C#
http://localhost:5913/efcore/Movies?$filter=date(ReleaseDate) eq 2017-03-03
http://localhost:5913/efcore/Movies?$filter=ReleaseDate eq 2017-03-03
```
All can work.

### Description

*Assumed a property named "ReleaseDate" as "Edm.DateTimeOffset", this change is to make the filters (as below) to work:
1. $filter=ReleaseDate eq 2018-09-09
2. $filter=date(ReleaseDate) eq 2018-09-09

- [x ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -243,6 +243,10 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             {
                 return source;
             }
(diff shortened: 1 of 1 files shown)


mikepizzo:
Change looks good; would like to see some test cases.

PR #1738, opened 2019-01-17
Support navigation properties on complex types

### LINKED ISSUE #1145: AutoExpand in sub child
I have a problem with AutoExpand attribute when it's used in a sub child.
I need to expand a child in an AutoExpand entity.

Assemblies affected

Microsoft.AspNet.OData v6.1.0

Reproduce steps

I use 4 entites
class Entity1
{
int identite;
string libelle;
EntityType entityType;
}
class EntityType
{
int idEntityType;
string libelle;
EntityAutoExpand entityAutoExpand;
}
[AutoExpand]
class EntityAutoExpand
{
int idEntityAutoExpand;
string libelle;
List listSubItems;
}
class EntityAutoExpandSubItems
{
int idEntityAutoExpandSubItems;
string libelle;
}

When I call my WebService on EntityType with
https://localhost:9443/odata/EntityType?$expand=entityAutoExpand($expand=listSubItems)
Or
https://localhost:9443/odata/EntityType?$expand=entityAutoExpand
that's work would expected.

But if I call on Entity1
https://localhost:9443/odata/Entity1?&$expand=EntityType($expand=EntityAutoExpand($expand=EntityAutoExpandSubItems))
Or
https://localhost:9443/odata/Entity1?&$expand=EntityType($expand=EntityAutoExpand)
I have an error 500 with message "An element with the same key have already be added"

Additional detail

When I remove AutoExpand attribute, that's work.

Thanks in advance

KanishManuja-MS:
### Issues

This pull request enables expanding on navigation properties on complex types and fixes #1145 .

### Description

1. SelectExpand classes assumed that the navigation property will always be expanded on an entity type but with this PR we changed it to be structured type instead of the entity type. 
2. This PR enabled having a path to a navigation property in the expand query option.
3. Enabled SelectExpandClause to propagate to nested properties on complex types. 
4. Fixed the issue with AutoExpand and expanding on a nested resource. 

- [x] *Test cases added*

### Additional work necessary
Documentation to include this feature. 
More test cases to be added with Code review. 
Select and Expand with a path serialize all properties on complex types. This to be fixed with an upcoming feature.


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -100,6 +100,12 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
                 allActions = new HashSet<IEdmAction>(model.GetAvailableActions(entityType));
                 allFunctions = new HashSet<IEdmFunction>(model.GetAvailableFunctions(entityType));
             }
+            else if (structuredType != null)
```
xuzhg: so, why do we still keep the if(...) clause.

src/Microsoft.AspNet.OData.Shared/Query/SelectExpandQueryOption.cs
```diff
@@ -51,10 +51,6 @@ public class SelectExpandQueryOption
             }
 
             IEdmEntityType entityType = context.ElementType as IEdmEntityType;
-            if (entityType == null)
```
xuzhg: >entityType [](start = 16, length = 10)

either to remove the check, maybe it's better to check the structure type

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -119,12 +119,10 @@ private LambdaExpression GetProjectionLambda()
 
         internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEdmProperty property, Expression source)
```
xuzhg: >IEdmEntityType [](start = 57, length = 14)

either to remove this check, do we need to change it as "structure type"

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -154,7 +152,6 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
 
         internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IEdmProperty property, Expression source)
```
xuzhg: >IEdmEntityType [](start = 58, length = 14)

either to remove this check, do we need to change it as "structure type"

src/Microsoft.AspNet.OData.Shared/Routing/ODataPathSegmentHandler.cs
```diff
@@ -235,8 +235,6 @@ public override void Handle(PathTemplateSegment segment)
         /// <param name="segment">the segment to handle</param>
         public override void Handle(PropertySegment segment)
         {
-            _navigationSource = null;
-
```
xuzhg: leave a comment to illustrate what happened to "_navigaitonSource"

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore/NavigationPropertyOnComplexType/NavigationPropertyOnComplexTypeTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            configuration.AddControllers(typeof(PeopleController));
+            configuration.JsonReferenceLoopHandling =
+                Newtonsoft.Json.ReferenceLoopHandling.Ignore;
+            configuration.MaxTop(2).Expand().Select().OrderBy().Filter();
+            configuration.MapODataServiceRoute("odata", "odata", ModelGenerator.GetConventionalEdmModel());
+        }
+
+        [Theory]
+        [InlineData(PeopleBaseUrl + "(1)/Location/ZipCode", "{\"Zip\":98052,\"City\":\"Redmond\",\"State\":\"Washington\"}")]
```
xuzhg: >Location/ZipCode [](start = 41, length = 16)

Make sure "People(1)?$expand=Location/ZipCode "  work

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
KanishManuja-MS: We populate the actions and function only if it is an entity type. The second if indirectly implies it is only for the complex type. 

---
In reply to: [250464529](https://github.com/OData/WebApi/pull/1738#discussion_r250464529) [](ancestors = 250464529)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore/NavigationPropertyOnComplexType/NavigationPropertyOnComplexTypeTests.cs
KanishManuja-MS: Currently, it does not work. Having a path in expand throws a validation error. I believe it is a slightly different feature to enable that. Once path is supported, I don't think there should be other problems.

---
In reply to: [250465447](https://github.com/OData/WebApi/pull/1738#discussion_r250465447) [](ancestors = 250465447)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -755,15 +770,15 @@ private void WriteDynamicComplexProperty(object propertyValue, IEdmTypeReference
...
+                    WriteComplexAndExpandedNavigationProperty(navPropertyToExpand.Key, navPropertyToExpand.Value, resourceContext, writer, null);
                     writer.WriteEnd();
                 }
             }
         }
 
         private void WriteComplexAndExpandedNavigationProperty(IEdmProperty edmProperty, SelectExpandClause selectExpandClause,
             ResourceContext resourceContext,
-            ODataWriter writer)
+            ODataWriter writer, IList<ExpandedNavigationSelectItem> selectExpandItems)
```
xuzhg: > IList<ExpandedNavigationSelectItem> selectExpandItems [](start = 31, length = 54)

we can't get it from "selectExpandClause" ?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore/NavigationPropertyOnComplexType/PeopleController.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNet.OData;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Microsoft.Test.E2E.AspNet.OData.NavigationPropertyOnComplexType
+{
+    [Route("api/[controller]")]
```
xuzhg: >[Route("api/[controller]")] [](start = 4, length = 27)

why do we still have "Web API" route?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -308,18 +410,18 @@ internal static void ValidatePathIsSupported(ODataPath path)
...
+            //    throw new ODataException(SRResources.UnsupportedSelectExpandPath);
+            //}
+
+            //if (segmentCount == 2)
+            //{
+            //    if (!(path.FirstSegment is TypeSegment))
+            //    {
+            //        throw new ODataException(SRResources.UnsupportedSelectExpandPath);
+            //    }
+            //}
```
xuzhg: We still need to validate the path. Right?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -40,6 +43,7 @@ public SelectExpandNode()
         public SelectExpandNode(SelectExpandNode selectExpandNodeToCopy)
```
xuzhg: >selectExpandNodeToCopy [](start = 49, length = 22)

We don't copy the "_selectItems and _edmProperty" ?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -57,8 +61,11 @@ public SelectExpandNode(SelectExpandNode selectExpandNodeToCopy)
         /// <param name="writeContext">The serializer context to be used while creating the collection.</param>
         /// <remarks>The default constructor is for unit testing only.</remarks>
         public SelectExpandNode(IEdmStructuredType structuredType, ODataSerializerContext writeContext)
-            : this(writeContext.SelectExpandClause, structuredType, writeContext.Model)
+            : this()
         {
+            _selectItems = writeContext.ExpandedNavigationSelectItems;
+            _edmProperty = writeContext.EdmProperty;
```
xuzhg: >_edmProperty = writeContext.EdmProperty; [](start = 12, length = 40)

I don't understand why we should record the property?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
xuzhg: So, it seems if we only change the return for "GetAvailableActions() and GetAvailableFunctions()", we can use "structuredType" directly, don't need to compare whether it's entity or complex?

---
In reply to: [250717218](https://github.com/OData/WebApi/pull/1738#discussion_r250717218) [](ancestors = 250717218,250464529)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -34,8 +35,9 @@ public ODataSerializerContext()
         /// <param name="edmProperty">The complex property being nested or the navigation property being expanded.
         /// If the resource property is the dynamic complex, the resource property is null.
         /// </param>
+        /// <param name="expandedItems">The <see cref="ExpandedNavigationSelectItem"/> for the navigation property being expanded.</param>
         /// <remarks>This constructor is used to construct the serializer context for writing nested and expanded properties.</remarks>
-        public ODataSerializerContext(ResourceContext resource, SelectExpandClause selectExpandClause, IEdmProperty edmProperty)
+        public ODataSerializerContext(ResourceContext resource, SelectExpandClause selectExpandClause, IEdmProperty edmProperty, IList<ExpandedNavigationSelectItem> expandedItems = null)
```
xuzhg: >ExpandedNavigationSelectItem [](start = 135, length = 28)

I want to implement $expand=property/navProperty/$ref, so, it's better to consider to use "ExpandedNavigationReferenceItem".

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore/NavigationPropertyOnComplexType/NavigationPropertyOnComplexTypeTests.cs
```diff
@@ -0,0 +1,99 @@
...
```
xuzhg: >PeopleBaseUrl + "(1) [](start = 20, length = 20)

It's better to move the prefix into the function body.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KanishManuja-MS: I don't think so. We need the query options inside the expand as well.

KanishManuja-MS: No. 

---
In reply to: [258238418](https://github.com/OData/WebApi/pull/1738#discussion_r258238418) [](ancestors = 258238418)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +151,71 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
+                        IEdmProperty propertyInPath = propertyAccessPathSegment.Property;
+                        currentProperty = EdmLibHelpers.GetClrPropertyName(propertyInPath, _model);
+                    }
+                    if (String.Compare(currentProperty, propertyName, StringComparison.Ordinal) == 0)
+                    {
+                        break;
+                    }
+                    if (!String.IsNullOrEmpty(currentProperty))
+                    {
+                        left = Expression.Property(left, currentProperty);
```
xuzhg: >left = Expression.Property(left, currentProperty) [](start = 24, length = 49)

add nullable check at each level

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +151,71 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
 
                 source = Expression.TypeAs(source, castType);
             }
-
+            Expression left = source;
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            Expression propertyValue = Expression.Property(source, propertyName);
+            if (expandItem != null)
+            {
+                for (int i = 0; i < expandItem.PathToNavigationProperty.Count; i++)
```
xuzhg: > for (int i = 0; i < expandItem.PathToNavigationProperty.Count; i++) [](start = 15, length = 68)

what about if the segment is the type cast segment.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -671,8 +682,16 @@ public virtual string CreateETag(ResourceContext resourceContext)
                     Name = complexProperty.Name
                 };
 
+                IList<ExpandedNavigationSelectItem> selectItems = null;
+                if (selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes != null)
```
xuzhg: >if (selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes != null) [](start = 16, length = 72)

Let's discuss the followings with @mike

1. $select=Name&$expand=Np
2. $select=Name,NP&$expand=Np
3. $select=Name&$expand=Cp/Np
4. $select=Cp&$expand=Cp/Np
5. $select=Cp/Np&$expand=Cp/Np

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -184,19 +206,103 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+                            {
+                                ODataPathSegment segment = expandItem.PathToNavigationProperty.ElementAt(i);
+                                PropertySegment propertySegment = segment as PropertySegment;
+                                if (propertySegment != null)
+                                {
+                                    complexProperty = propertySegment.Property;
+                                    break;
+                                }
+                            }
+                            if (complexProperty != null && ExpandedNavigationPropertiesOnComplexTypes.ContainsKey(complexProperty))
```
xuzhg: >if (complexProperty != null && ExpandedNavigationPropertiesOnComplexTypes.ContainsKey(complexProperty)) [](start = 28, length = 103)

Do we support this:

~/Customers(1)/Address?$expand=NS.UsAddress/NavigationProperty

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore/NavigationPropertyOnComplexType/NavigationPropertyOnComplexTypeTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            configuration.AddControllers(typeof(PeopleController));
+            configuration.JsonReferenceLoopHandling =
+                Newtonsoft.Json.ReferenceLoopHandling.Ignore;
+            configuration.MaxTop(2).Expand().Select().OrderBy().Filter();
+            configuration.MapODataServiceRoute("odata", "odata", ModelGenerator.GetConventionalEdmModel());
+        }
+
+        [Theory]
+        [InlineData("/Location/ZipCode", "{\"Zip\":98052,\"City\":\"Redmond\",\"State\":\"Washington\"}")]
+        [InlineData("/Location?$expand=ZipCode", "{\"Zip\":98052,\"City\":\"Redmond\",\"State\":\"Washington\"}")]
```
xuzhg: >{\"Zip\":98052,\"City\":\"Redmond\",\"State\":\"Washington\"} [](start = 50, length = 61)

it's better to test the property name expanded

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
KanishManuja-MS: They are consumed into the public properties once a selectexpandnode is initialized. We don't need to copy. 

---
In reply to: [262739928](https://github.com/OData/WebApi/pull/1738#discussion_r262739928) [](ancestors = 262739928)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KanishManuja-MS: First 4 are supported but a path in select is not supported. 

---
In reply to: [274692314](https://github.com/OData/WebApi/pull/1738#discussion_r274692314) [](ancestors = 274692314)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
KanishManuja-MS: Yes, we support this. 

---
In reply to: [274694855](https://github.com/OData/WebApi/pull/1738#discussion_r274694855) [](ancestors = 274694855)

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -513,7 +513,11 @@ internal static IList<ODataPathSegment> GenerateBaseODataPathSegments(this Resou
                 }
                 else
                 {
-                    odataPath.Add(new EntitySetSegment((IEdmEntitySet)navigationSource));
+                    IEdmEntitySet entitySet = navigationSource as IEdmEntitySet;
+                    if (entitySet != null)
+                    {
+                        odataPath.Add(new EntitySetSegment(entitySet));
+                    }
```
xuzhg: Is there any scenario that the input navigation source is null?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -110,6 +146,11 @@ public ODataSerializerContext(ResourceContext resource, SelectExpandClause selec
         /// </summary>
         public SelectExpandClause SelectExpandClause { get; set; }
 
+        /// <summary>
+        /// Gets or sets the list of <see cref="ExpandedNavigationSelectItem"/>.
+        /// </summary>
+        public IList<ExpandedReferenceSelectItem> ExpandedNavigationSelectItems { get; set; }
```
xuzhg: >ExpandedNavigationSelectItems [](start = 50, length = 29)

maybe rename it to refresh the type. => ExpandedReferenceSelectItem ?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -110,6 +146,11 @@ public ODataSerializerContext(ResourceContext resource, SelectExpandClause selec
         /// </summary>
         public SelectExpandClause SelectExpandClause { get; set; }
 
+        /// <summary>
+        /// Gets or sets the list of <see cref="ExpandedNavigationSelectItem"/>.
```
xuzhg: >ExpandedNavigationSelectItem [](start = 48, length = 28)

ExpandedReferenceSelectItem

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -53,17 +69,37 @@ public ODataSerializerContext(ResourceContext resource, SelectExpandClause selec
...
                 if (navigationProperty != null)
                 {
-                    NavigationSource = context.NavigationSource.FindNavigationTarget(NavigationProperty);
+                    bool isDeclaringTypeEntity = navigationProperty.DeclaringType is IEdmEntityType;
+                    if (!isDeclaringTypeEntity)
+                    {
+                        IEnumerable<IEdmNavigationPropertyBinding> bindings = context.NavigationSource.FindNavigationPropertyBindings(navigationProperty);
+                        foreach (IEdmNavigationPropertyBinding binding in bindings)
+                        {
+                            if (binding.NavigationProperty == navigationProperty)
```
xuzhg: >if (binding.NavigationProperty == navigationProperty) [](start = 28, length = 53)

It's not correct, right?

For example:

ComplexType
{
      EntityType NavigationProperty {get;set;}
}

EntityType
{
       ComplexType complexProperty1 {get;set;}
       ComplexType complexProperty2 {get;set;}
}

<EntitySet Name="Set" EntityType="NS.EntityType" >
     <NPB Path="complexProperty1/NavigationProperty" Target="SetA" />
     <NPB Path="complexProperty2/NavigationProperty" Target="SetB" />
</EntitySet>


So, for the same navigation property, only "SetA" is returned, and "SetB" never be returned?

src/Microsoft.AspNet.OData.Shared/Query/SelectExpandQueryOption.cs
```diff
@@ -50,10 +50,10 @@ public class SelectExpandQueryOption
                 throw Error.ArgumentNull("queryOptionParser");
             }
 
-            IEdmEntityType entityType = context.ElementType as IEdmEntityType;
-            if (entityType == null)
+            if (!(context.ElementType is IEdmStructuredType))
             {
-                throw Error.Argument("context", SRResources.SelectNonEntity, context.ElementType.ToTraceString());
+                // TODO: localize the error string.
```
xuzhg: >// TODO: localize the error string. [](start = 16, length = 35)

don't forget. :)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/DefaultODataPathHandlerTest.cs
```diff
@@ -659,7 +659,6 @@ public void CanParsePropertySegment(string odataPath)
             PropertySegment propertySegment = Assert.IsType<PropertySegment>(segment);
 
             Assert.Equal(expectedText, propertySegment.ToUriLiteral());
-            Assert.Null(path.NavigationSource);
```
xuzhg: >Assert.Null(path.NavigationSource); [](start = 12, length = 35)

not only remove it, try to verify it with the previous navigation source.

xuzhg: Same for others?

---
In reply to: [279473187](https://github.com/OData/WebApi/pull/1738#discussion_r279473187) [](ancestors = 279473187)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -194,10 +237,85 @@ private void BuildExpansions(SelectExpandClause selectExpandClause, HashSet<IEdm
...
+        /// </summary>
+        private IEdmStructuralProperty FindNextPropertySegment(ODataPath path)
+        {
+            IEdmStructuralProperty complexProperty = null;
+            bool breakNext = false;
+
+            // If the current SelectExpandNode is not top-level and has a property associated with it then return the next property segment from the path.
+            if (Property != null)
+            {
+                foreach (ODataPathSegment segment in path)
```
xuzhg: >foreach (ODataPathSegment segment in path) [](start = 16, length = 42)

how about duplicated property in the path?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -665,14 +677,48 @@ public virtual string CreateETag(ResourceContext resourceContext)
...
+                IList<ExpandedReferenceSelectItem> selectItems = null;
+                if (selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes != null)
+                {
+                    if (selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes.ContainsKey(complexProperty))
+                    {
+                        selectItems = selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes[complexProperty];
+                    }
+                }
+
+                if (!resourceContext.StructuredType.DeclaredProperties.Contains(complexProperty))
```
xuzhg: >DeclaredProperties [](start = 52, length = 18)

should use `Properties()`, otherwise the complexProperty on the baseType can't be found

xuzhg: Add a test case to verify this scenario:

complexType (BaseType)
{
    Property1 (complexType)
}

ComplexType(SubType)
{
   Property2 (complexType)
}

EntityType
{
    Property (SubType)
}

~/....$expand=Property/Property1/...
~/...$expand=Property/Property2/...


---
In reply to: [280163679](https://github.com/OData/WebApi/pull/1738#discussion_r280163679) [](ancestors = 280163679)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -117,17 +117,15 @@ private LambdaExpression GetProjectionLambda()
...
         }
 
-        internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEdmProperty property, Expression source)
+        internal Expression CreatePropertyNameExpression(IEdmStructuredType elementType, IEdmProperty property, Expression source)
         {
-            Contract.Assert(elementType != null);
             Contract.Assert(property != null);
             Contract.Assert(source != null);
 
             IEdmEntityType declaringType = property.DeclaringType as IEdmEntityType;
```
xuzhg: >property.DeclaringType as IEdmEntityType; [](start = 43, length = 41)

The property maybe the structural property, the cast will give null, let's verify these codes working?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +150,97 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
             return Expression.Constant(property.Name);
         }
 
-        internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IEdmProperty property, Expression source)
+        internal Expression CreatePropertyValueExpression(IEdmStructuredType elementType, IEdmProperty property, Expression source)
         {
-            Contract.Assert(elementType != null);
```
xuzhg: >Contract.Assert(elementType != null); [](start = 12, length = 37)

this line should be always valid.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +150,97 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
-            IEdmEntityType declaringType = property.DeclaringType as IEdmEntityType;
-            Contract.Assert(declaringType != null, "only entity types are projected.");
+            FilterClause filterClause = expandItem != null ? expandItem.FilterOption : null;
+
+            IEdmStructuredType declaringType = property.DeclaringType;
+
+            //if there is a path in $expand then find the right declaring type;
+            if (expandItem != null)
+            {
+                PropertySegment propertyAccessPathSegment = expandItem.PathToNavigationProperty.ElementAt(0) as PropertySegment;
```
xuzhg: >as PropertySegment [](start = 109, length = 18)

First segment in the pathToNavProperty can be type cast segment?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +150,97 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
 
                 source = Expression.TypeAs(source, castType);
             }
 
+            Expression left = source;
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            Expression propertyValue = Expression.Property(source, propertyName);
+            if (expandItem != null)
+            {
+                for (int i = 0; i < expandItem.PathToNavigationProperty.Count; i++)
```
xuzhg: >for (int i = 0; i < expandItem.PathToNavigationProperty.Count; i++) [](start = 16, length = 67)

foreach ()

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +150,97 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
+                    string currentProperty = String.Empty;
+                    PropertySegment propertyAccessPathSegment =
+                        expandItem.PathToNavigationProperty.ElementAt(i) as PropertySegment;
+                    if (propertyAccessPathSegment != null)
+                    {
+                        IEdmProperty propertyInPath = propertyAccessPathSegment.Property;
+                        currentProperty = EdmLibHelpers.GetClrPropertyName(propertyInPath, _model);
+                    }
+
+                    TypeSegment typeSegment = expandItem.PathToNavigationProperty.ElementAt(i) as TypeSegment;
```
xuzhg: Move line 210 - 215 into else { }

src/Microsoft.AspNet.OData.Shared/Query/SelectExpandQueryOption.cs
```diff
@@ -50,10 +50,10 @@ public class SelectExpandQueryOption
...
+                throw Error.Argument("Element type should be a structured type to select or expand on the property.");
```
xuzhg: >"Element type should be a structured type to select or expand on the property." [](start = 37, length = 79)

Move it to resource

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -185,6 +185,18 @@ internal static bool TryGetCharSet(MediaTypeHeaderValue mediaType, IEnumerable<s
...
+                {
+                    foreach (var item in writeContext.SelectExpandClause.SelectedItems)
+                    {
+                        ExpandedNavigationSelectItem selectItem = item as ExpandedNavigationSelectItem;
+                        if (selectItem != null)
+                        {
+                            writeContext.ExpandedReferenceSelectItems.Add(selectItem);
+                        }
+                    }
+                }
```
xuzhg: You can use Linq to combine the codes

xuzhg:
>                 SelectAllDynamicProperties = true;

These codes is wrong for:

$expand=Address/City 

because, the above query option output payload should not include any other structural properties.

Fix it maybe not in this PR, but, we should fix that before 7.2 release. Consider to fix it with the $select=Address/City issue.

---
Refers to: src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs:138 in 6f829f3. [](commit_id = 6f829f3201c1b3200c791b0b944c83e67f6da3c6, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -194,12 +243,93 @@ private void BuildExpansions(SelectExpandClause selectExpandClause, HashSet<IEdm
...
+                // Return the first property if top-level resource
+                int count = path.Count;
+                for (int i = 0; i < count; i++)
+                {
+                    ODataPathSegment segment = path.ElementAt(i);
+                    PropertySegment propertySegment = segment as PropertySegment;
+                    if (propertySegment != null)
+                    {
+                        complexProperty = propertySegment.Property;
+                        break;
```
xuzhg: Combine them using First(c => c is PropertySegment);

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -665,14 +677,48 @@ public virtual string CreateETag(ResourceContext resourceContext)
...
+
+            foreach (var item in selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes)
+            {
+                IEdmProperty complexProperty = item.Key;
+                IList<ExpandedReferenceSelectItem> selectItems = null;
+                if (selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes != null)
+                {
+                    if (selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes.ContainsKey(complexProperty))
+                    {
+                        selectItems = selectExpandNode.ExpandedNavigationPropertiesOnComplexTypes[complexProperty];
```
xuzhg: do code clean

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -665,14 +677,48 @@ public virtual string CreateETag(ResourceContext resourceContext)
...
+                    }
+                }
+
+                if (!resourceContext.StructuredType.Properties().Contains(complexProperty))
+                {
+                    continue;
+                }
```
xuzhg: it seems the "structuredType" always contain the properyt?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -194,12 +243,93 @@ private void BuildExpansions(SelectExpandClause selectExpandClause, HashSet<IEdm
...
+                    {
+                        complexProperty = propertySegment.Property;
+                        if (breakNext)
+                        {
+                            break;
+                        }
+                        if (pathSegments[pathIndex] == complexProperty.Name)
+                        {
+                            pathIndex++;
+                        }
```
xuzhg: how about the insert type cast segment

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -303,24 +433,9 @@ private void AddOperations(HashSet<IEdmAction> allActions, HashSet<IEdmFunction>
...
-            }
-
-            if (segmentCount == 2)
-            {
-                if (!(path.FirstSegment is TypeSegment))
-                {
-                    throw new ODataException(SRResources.UnsupportedSelectExpandPath);
-                }
-            }
-
```
xuzhg: shall we verify the middle segment is "type cast" or "property" segment only?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/NavigationPropertyOnComplexType/PeopleRepository.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Test.E2E.AspNet.OData.NavigationPropertyOnComplexType
+{
+    public class PeopleRepository
```
xuzhg: >PeopleRepository [](start = 17, length = 16)

maybe add test for EF & EFCore

KanishManuja-MS: EF and EFCore do not support navigation properties on complex types. 

---
In reply to: [295043430](https://github.com/OData/WebApi/pull/1738#discussion_r295043430) [](ancestors = 295043430)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/NavigationPropertyOnComplexType/PeopleRepository.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public List<Person> _people { get; set; }
+        public static List<ZipCode> _zipCodes;
```
xuzhg: why is "_people" instance meanwhile _zipCodes static?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/NavigationPropertyOnComplexType/PeopleController.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNet.OData;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.Test.E2E.AspNet.OData.Common.Controllers;
+
+namespace Microsoft.Test.E2E.AspNet.OData.NavigationPropertyOnComplexType
+{
+   public class PeopleController : TestODataController
```
xuzhg: align

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/NavigationPropertyOnComplexType/NavigationPropertyOnComplexTypeModel.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using Microsoft.AspNet.OData.Builder;
+using Microsoft.OData.Edm;
+using System.ComponentModel.DataAnnotations;
+using Microsoft.Test.E2E.AspNet.OData.Aggregation;
```
xuzhg: sort

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Serialization/SelectExpandNodeTest.cs
```diff
@@ -214,12 +214,22 @@ public void BuildSelectExpandNode_ThrowsODataException_IfUnknownSelectItemPresen
...
             ODataPath path = new ODataPath(new ValueSegment(previousType: null));
 
             ExceptionAssert.Throws<ODataException>(
-                () => SelectExpandNode.ValidatePathIsSupported(path),
+                () => SelectExpandNode.ValidatePathIsSupportedForSelect(path),
+                "A path within the select or expand query option is not supported.");
+        }
+
+        [Fact]
+        public void ValidatePathIsSupported_ThrowsForUnsupportedPathForExpand()
```
xuzhg: >ValidatePathIsSupported_ThrowsForUnsupportedPathForExpand [](start = 20, length = 57)

is there positive test case?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -110,6 +126,11 @@ public ODataSerializerContext(ResourceContext resource, SelectExpandClause selec
         /// </summary>
         public SelectExpandClause SelectExpandClause { get; set; }
 
+        /// <summary>
+        /// Gets of sets the relative path to the resouce being serialized
+        /// </summary>
+        public IEdmPathExpression PathToProperty { get; set; }
```
xuzhg: >IEdmPathExpression [](start = 15, length = 18)

maybe to use a Stack<string> is a better choice?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -182,24 +206,115 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+                            SelectExpandClause expandItems = new SelectExpandClause(expandedItemsForClause, false);
+                            ExpandedNavigationPropertiesOnComplexTypes.Add(complexProperty, expandItems);
+                        }
                     }
                 }
             }
         }
 
+        /// <summary>
+        /// Finds the appropriate property segment which should be responsible for propagating the expand item.
```
xuzhg: >Finds the appropriate property segment which should be responsible for propagating the expand item. [](start = 12, length = 99)

can you add an example into the summary to illustrate the functionality of this method for better maintenance later.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -182,24 +206,115 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
         /// </summary>
         public ISet<IEdmFunction> SelectedFunctions { get; private set; }
 
-        private void BuildExpansions(SelectExpandClause selectExpandClause, HashSet<IEdmNavigationProperty> allNavigationProperties)
+        /// <summary>
+        /// Gets the path to property corresponding to the SelectExpandNode. Null for a top-level select expand.
+        /// </summary>
+        public IEdmPathExpression PathToProperty { get; set; }
```
xuzhg: >set [](start = 56, length = 3)

shall we public set for this property?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -182,24 +206,115 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+
+        /// <summary>
+        /// Gets the property corresponding to the SelectExpandNode. Null for a top-level select expand.
+        /// </summary>
+        public IEdmProperty Property { get; set; }
```
xuzhg: >set [](start = 44, length = 3)

shall we public set for this property?

xuzhg:
I signed off because the most codes look good to me, however, please resolve all the comments.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -199,6 +215,11 @@ public SelectExpandClause SelectExpandClause
         /// </summary>
         public ODataQueryOptions QueryOptions { get; internal set; }
 
+        /// <summary>
+        /// Gets of sets the relative path to the resouce being serialized
```
mikepizzo: >of [](start = 17, length = 2)

nit "or"

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -84,15 +85,30 @@ internal ODataSerializerContext(ResourceContext resource, IEdmProperty edmProper
...
 
+            IList<string> pathSegments = context.PathToProperty != null ? context.PathToProperty.PathSegments.ToList() : new List<string>();
+            if (edmProperty != null)
+            {
+                bool isDerived = !resource.StructuredType.AsElementType().IsOrInheritsFrom(edmProperty.DeclaringType);
+                if (isDerived)
+                {
+                    pathSegments.Add(edmProperty.DeclaringType.FullTypeName());
+                }
+                pathSegments.Add(edmProperty.Name);
```
mikepizzo: >pathSegments.Add(edmProperty.Name); [](start = 16, length = 35)

Rather than calling ToList<> on the path segments in order to call Add(), would it be more efficient to keep it an IEnumerable<string> and call Concat?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -32,6 +32,7 @@ public SelectExpandNode()
             SelectedStructuralProperties = new HashSet<IEdmStructuralProperty>();
             SelectedComplexProperties = new HashSet<IEdmStructuralProperty>();
             SelectedNavigationProperties = new HashSet<IEdmNavigationProperty>();
+            ExpandedNavigationPropertiesOnComplexTypes = new Dictionary<IEdmProperty, ExpandedNavigationSelectItem>();
```
mikepizzo: >IEdmProperty [](start = 72, length = 12)

This will always be IEdmStructuralProperty, no?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -32,6 +32,7 @@ public SelectExpandNode()
...
             ExpandedProperties = new Dictionary<IEdmNavigationProperty, ExpandedNavigationSelectItem>();
```
mikepizzo: >ExpandedProperties [](start = 12, length = 18)

In V8, will we combine ExpandedNavigationPropertiesOnComplexTypes and ExpandedProperties?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -126,13 +123,20 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
                 {
                     allNavigationProperties = new HashSet<IEdmNavigationProperty>(entityType.NavigationProperties());
                     allActions = new HashSet<IEdmAction>(model.GetAvailableActions(entityType));
                     allFunctions = new HashSet<IEdmFunction>(model.GetAvailableFunctions(entityType));
                 }
+                else if (structuredType != null)
+                {
+                    allNavigationProperties = new HashSet<IEdmNavigationProperty>(structuredType.NavigationProperties());
+                    allActions = new HashSet<IEdmAction>();
+                    allFunctions = new HashSet<IEdmFunction>();
```
mikepizzo: why would these not be populated with available actions/functions, as in the entitytype case?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -242,30 +253,125 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+        /// Gets the path to property corresponding to the SelectExpandNode. Null for a top-level select expand.
+        /// </summary>
+        public IEdmPathExpression PathToProperty { get; set; }
+
+        /// <summary>
+        /// Gets the property corresponding to the SelectExpandNode. Null for a top-level select expand.
+        /// </summary>
+        public IEdmProperty Property { get; set; }
+
+        private void BuildExpansions(IList<SelectItem> selectedItems, HashSet<IEdmNavigationProperty> allNavigationProperties)
```
mikepizzo: >IList<SelectItem> [](start = 37, length = 17)

Why is this an IList<SelectItem> rather than an IEnumerable<SelectItem>? Do we ever modify the list?  If not, we should make this an IEnumerable<SelectItem> and not call ToList() on Ln 175.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -242,30 +253,125 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+        /// </summary>
+        private IEdmStructuralProperty FindNextPropertySegment(ODataPath path)
+        {
+            IEdmStructuralProperty complexProperty = null;
+            // If the current SelectExpandNode is not top-level and has a property associated with it then return the next property segment from the path.
+            if (Property != null)
+            {
+                int pathIndex = 0;
+                bool endOfPathToProperty = false;
+                IList<string> pathSegments = PathToProperty.PathSegments.ToList();
```
mikepizzo: >PathToProperty [](start = 45, length = 14)

Any chance that PathToProperty might be null here?  What if a user sets Property but not PathToProperty? Should probably handle to avoid a null ref exception.

mikepizzo: >PathToProperty [](start = 45, length = 14)

Somewhere (maybe in the setter for PathToProperty) should we validate that PathToProperty terminates in a PropertySegment?  What will this code do if it does not?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -242,30 +253,125 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Finds the appropriate property segment which should be responsible for propagating the expand item.
+        /// For instance, if we are creating the SelectExpandNode for property p2 for the query ~/EnitySet/Key/p1/p2/p3?$expand=NP, we want to return property p3 here. 
+        /// </summary>
+        private IEdmStructuralProperty FindNextPropertySegment(ODataPath path)
```
mikepizzo: >FindNextPropertySegment [](start = 39, length = 23)

Would like to have you walk me through this and explain your assumptions. i.e., it looks like it will ignore properties in the path that inconveniently don't match what's in the PathToProperty?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -242,30 +253,125 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
                     }
+                    else
+                    {
+                        //This is the case where the navigation property is not on the current type. We need to propagate the expand item to deeper SelectExpandNode.
+                        IEdmProperty complexProperty = FindNextPropertySegment(expandReferenceItem.PathToNavigationProperty);
+
+                        if (complexProperty != null && ExpandedNavigationPropertiesOnComplexTypes.ContainsKey(complexProperty))
+                        {
+                            SelectExpandClause oldClause = ExpandedNavigationPropertiesOnComplexTypes[complexProperty].SelectAndExpand;
+                            SelectExpandClause newClause = new SelectExpandClause(oldClause.SelectedItems.Concat(new List<SelectItem>() { expandReferenceItem }), false);
```
mikepizzo: >List<SelectItem>() { expandReferenceItem } [](start = 117, length = 42)

probably cheaper to use SelectItem[]{expandReferenceItem]

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -242,30 +253,125 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+                            ExpandedNavigationSelectItem newItem = new ExpandedNavigationSelectItem(expandReferenceItem.PathToNavigationProperty, navigationSegment.NavigationSource, newClause);
+                            ExpandedNavigationPropertiesOnComplexTypes.Add(complexProperty, newItem);
+                        }
+                        else if (complexProperty != null)
```
mikepizzo: wouldn't it be better to do if(complexProperty !=null) on 301, and then nest if(ExpandedNavigationPropertiesOnComplexType.ContainsKey() .. else) so that you don't have to check complexProperty != null twice?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -242,30 +253,125 @@ public SelectExpandNode(SelectExpandClause selectExpandClause, IEdmStructuredTyp
...
+                        {
+                            List<SelectItem> expandedItemsForClause = new List<SelectItem>();
+                            expandedItemsForClause.Add(expandReferenceItem);
+                            SelectExpandClause clause = new SelectExpandClause(expandedItemsForClause, false);
```
mikepizzo: >expandedItemsForClause [](start = 79, length = 22)

delete lines 310-311 and just insert new SelectItem[]{expandReferenceItem}

src/Microsoft.AspNet.OData.Shared/Query/SelectExpandQueryOption.cs
```diff
@@ -88,10 +87,11 @@ internal SelectExpandQueryOption(string select, string expand, ODataQueryContext
                 throw Error.Argument(SRResources.SelectExpandEmptyOrNull);
             }
 
-            IEdmEntityType entityType = context.ElementType as IEdmEntityType;
-            if (entityType == null)
+            IEdmStructuredType structuredType = context.ElementType as IEdmStructuredType;
+
+            if (structuredType == null)
```
mikepizzo: >structuredType == null [](start = 16, length = 22)

You don't ever use structuredType; why not remove ln 90 and just do if (!(context.ElementType is IEdmStructuredType)) as you did in ln 55?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +153,116 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
             return Expression.Constant(property.Name);
         }
 
-        internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IEdmProperty property, Expression source)
+        internal Expression CreatePropertyValueExpression(IEdmStructuredType elementType, IEdmProperty property, Expression source)
         {
-            Contract.Assert(elementType != null);
```
mikepizzo: Is this assert no longer valid? can elementType be null?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +153,116 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
+            if (expandItem != null)
+            {
+                foreach (ODataPathSegment segment in expandItem.PathToNavigationProperty)
+                {
+                    PropertySegment propertyAccessPathSegment = segment as PropertySegment;
+                    if (propertyAccessPathSegment != null)
+                    {
+                        IEdmProperty newProperty = propertyAccessPathSegment.Property;
+                        declaringType = newProperty.DeclaringType;
+                        break;
```
mikepizzo: >break [](start = 24, length = 5)

You'll have to explain this logic to me.  If my path has multiple complex types, won't this stop at the first one? Is that what I want?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +153,116 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
+                    }
+                    else
+                    {
+                        TypeSegment castSegment = segment as TypeSegment;
+                        if (castSegment != null)
+                        {
+                            continue;
+                        }
+
+                        break;
```
mikepizzo: you don't actually use castSegment; could you replace ln 189-195 with if(!(segment is TypeSegment)) {break;} ?

mikepizzo: >break [](start = 24, length = 5)

so, if we find a segment other than a PropertySegment or a TypeSegment, we just ignore the entire path?  This would be an error, right?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -152,40 +153,116 @@ internal Expression CreatePropertyNameExpression(IEdmEntityType elementType, IEd
...
+                        {
+                            Type castType = EdmLibHelpers.GetClrType(typeSegment.EdmType, _model);
+                            left = Expression.TypeAs(left, castType);
+                        }
+                    }
+
+                    if (String.Compare(currentProperty, propertyName, StringComparison.Ordinal) == 0)
+                    {
+                        break;
+                    }
```
mikepizzo: Really?  what if the same property name is defined in my parent and nested type?  Don't you always want to build the expression for the entire PathToNavigationProperty?  Seems like better logic would be to handle the expandItem=null case as an else, so you don't have to try and stop one before the final property.

mikepizzo:
>                 Expression typeName = CreateTypeNameExpression(source, entityType, _model);

This will always be an entity type?  or it's valid to pass null for a complex type?

---
Refers to: src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs:412 in 2c245ea. [](commit_id = 2c245eafe9105ad1dce04ea458a174e975193e74, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -347,13 +425,13 @@ private Expression ProjectElement(Expression source, SelectExpandClause selectEx
                 Dictionary<IEdmNavigationProperty, ExpandedReferenceSelectItem> propertiesToExpand = GetPropertiesToExpandInQuery(selectExpandClause);
                 ISet<IEdmStructuralProperty> autoSelectedProperties;
 
-                ISet<IEdmStructuralProperty> propertiesToInclude = GetPropertiesToIncludeInQuery(selectExpandClause, entityType, navigationSource, _model, out autoSelectedProperties);
+                ISet<IEdmStructuralProperty> propertiesToInclude = GetPropertiesToIncludeInQuery(selectExpandClause, structuredType, navigationSource, _model, out autoSelectedProperties);
                 bool isSelectingOpenTypeSegments = GetSelectsOpenTypeSegments(selectExpandClause, entityType);
```
mikepizzo: >entityType [](start = 98, length = 10)

is it valid to pass null here for a complex type?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -373,7 +451,7 @@ private Expression ProjectElement(Expression source, SelectExpandClause selectEx
 
         private static bool GetSelectsOpenTypeSegments(SelectExpandClause selectExpandClause, IEdmEntityType entityType)
         {
-            if (!entityType.IsOpen)
+            if (entityType == null || !entityType.IsOpen)
```
mikepizzo: Why not support open complex types?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -795,20 +875,38 @@ private static MethodInfo GetSelectMethod(Type elementType, Type resultType)
...
-                // add keys
-                foreach (IEdmStructuralProperty keyProperty in entityType.Key())
+                foreach (ExpandedNavigationSelectItem expandedNavigationSelectItem in selectedItems.OfType<ExpandedNavigationSelectItem>())
+                {
+                    // Include the entire property in the selection. This should be changed when we start supporting enhanced select option.
+                    foreach (var segment in expandedNavigationSelectItem.PathToNavigationProperty)
+                    {
+                        PropertySegment propertySegment = segment as PropertySegment;
+                        if (propertySegment != null &&
+                            entityType.Properties().Contains(propertySegment.Property))
```
mikepizzo: >entityType [](start = 28, length = 10)

Why does this not apply to complex types?

test/UnitTest/Microsoft.AspNet.OData.Test/PublicApi/Microsoft.AspNet.OData.PublicApi.bsl
```diff
@@ -3312,14 +3313,16 @@ public class Microsoft.AspNet.OData.Formatter.Serialization.SelectExpandNode {
 	public SelectExpandNode (SelectExpandNode selectExpandNodeToCopy)
 	public SelectExpandNode (Microsoft.OData.Edm.IEdmStructuredType structuredType, ODataSerializerContext writeContext)
 	public SelectExpandNode (Microsoft.OData.UriParser.SelectExpandClause selectExpandClause, Microsoft.OData.Edm.IEdmStructuredType structuredType, Microsoft.OData.Edm.IEdmModel model)
-	public SelectExpandNode (Microsoft.OData.UriParser.SelectExpandClause selectExpandClause, Microsoft.OData.Edm.IEdmStructuredType structuredType, Microsoft.OData.Edm.IEdmModel model, bool expandedReference)
```
mikepizzo: Isn't removing this overload a breaking change? Why are we convinced it's okay?

KanishManuja-MS: It wouldn't be as the above constructor is not released yet. Expand on $ref PR introduced that constructor which I refactored.

---
In reply to: [308507235](https://github.com/OData/WebApi/pull/1738#discussion_r308507235) [](ancestors = 308507235)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
KanishManuja-MS: Can operations be bound to complex types? EdmModelHelper method to get available functions only took in entity types. 

---
In reply to: [308488199](https://github.com/OData/WebApi/pull/1738#discussion_r308488199) [](ancestors = 308488199)

mikepizzo: Yes, protocol allows bounding operations to complex types. Would have to check to see what we do/do not support in the libraries.

---
In reply to: [308848817](https://github.com/OData/WebApi/pull/1738#discussion_r308848817) [](ancestors = 308848817,308488199)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
KanishManuja-MS: Maybe, I think there are some fundamental differences. The complex type dictionary can contain navigation select items even for it's nested properties where as ExpandedProperties only contains select items for navigation properties on the current structured type. If we are okay with that loss of information, we can probably merge these.

---
In reply to: [308487591](https://github.com/OData/WebApi/pull/1738#discussion_r308487591) [](ancestors = 308487591)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
KanishManuja-MS: Checked in code and verified with Sam that we do not support bound actions/functions on complex type currently. 

---
In reply to: [308873937](https://github.com/OData/WebApi/pull/1738#discussion_r308873937) [](ancestors = 308873937,308848817,308488199)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
KanishManuja-MS: I wanted to limit the scope but I guess I will now. TODO: Add a test case. 

---
In reply to: [308506219](https://github.com/OData/WebApi/pull/1738#discussion_r308506219) [](ancestors = 308506219)

test/UnitTest/Microsoft.AspNet.OData.Test/PublicApi/Microsoft.AspNet.OData.PublicApi.bsl
KanishManuja-MS: I believe that I had it fixed in the $ref PR. I will know when I rebase. 

---
In reply to: [308843120](https://github.com/OData/WebApi/pull/1738#discussion_r308843120) [](ancestors = 308843120,308507235)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerContext.cs
KanishManuja-MS: Used queue instead. 

---
In reply to: [300159016](https://github.com/OData/WebApi/pull/1738#discussion_r300159016) [](ancestors = 300159016)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Serialization/SelectExpandNodeTest.cs
KanishManuja-MS: A lot of E2E positive tests. 

---
In reply to: [299586374](https://github.com/OData/WebApi/pull/1738#discussion_r299586374) [](ancestors = 299586374)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
mikepizzo: Did you add the test case?

---
In reply to: [326413640](https://github.com/OData/WebApi/pull/1738#discussion_r326413640) [](ancestors = 326413640,308506219)

test/UnitTest/Microsoft.AspNet.OData.Test/PublicApi/Microsoft.AspNet.OData.PublicApi.bsl
mikepizzo: Please verify after rebase.

---
In reply to: [326414044](https://github.com/OData/WebApi/pull/1738#discussion_r326414044) [](ancestors = 326414044,308843120,308507235)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
mikepizzo: Can you open an issue to track adding support for bound operations to complex types?

---
In reply to: [326287637](https://github.com/OData/WebApi/pull/1738#discussion_r326287637) [](ancestors = 326287637,308873937,308848817,308488199)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -370,9 +473,8 @@ private void AddOperations(HashSet<IEdmAction> allActions, HashSet<IEdmFunction>
                 }
             }
         }
-
         // we only support paths of type 'cast/structuralOrNavPropertyOrAction' and 'structuralOrNavPropertyOrAction'.
-        internal static void ValidatePathIsSupported(ODataPath path)
+        internal static void ValidatePathIsSupportedForSelect(ODataPath path)
```
mikepizzo: what about nav props on a complex type nested under another complex type?

mikepizzo: This is existing code for testing path for $select. It should support nested complex type, but does not. Removing this check would expose issues downstream where the code assuming at most 2 segments.  Will need to be fixed as part of bigger fix.

---
In reply to: [327307928](https://github.com/OData/WebApi/pull/1738#discussion_r327307928) [](ancestors = 327307928)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/SelectExpandNode.cs
mikepizzo: Understand the logic after walking through it with you, and it's fine given the current design. We will probably need to modify a bit once we do select options on complex types, at which time we can revisit how properties of complex types in $select are handled in general and ideally make it more similar to how we handle nested entities.

---
In reply to: [308497764](https://github.com/OData/WebApi/pull/1738#discussion_r308497764) [](ancestors = 308497764)

PR #1762, opened 2019-02-13
Fix ordering the keys according to the ColumnAttribute.Order property

Galad:
### Issues

This pull request fixes issue #1085 

### Description

In order to ensure a stable ordering of the results with different request, the library adds a default order when required.
When an entity has a composite key, the order of the properties matters, and if not specified correctly, the query may have bad performances.

The fix ensures that the order specified in the model via the ColumnAttribute.Order property is respected when the stable order is ensured by the library.
It does it by configuring the keys in the correct order in the EDM model, and not in the function applying the default order.

Also, previously the order of the key in the model was arbitrarely chosen, which could be suprising for the user when the default order was applied because it would not match what it in the model. Now the order in the model will be the same as the default order.

#### Side effect
A side effect is that the fix will may change the order to the key in the EDM model.

- [x] Test cases added
- [ ] Build and test with one-click build and test script passed

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/Conventions/Attributes/ColumnAttributeEdmPropertyConvention.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/Conventions/Attributes/ColumnAttributeEdmPropertyConvention.cs
@@ -37,16 +37,17 @@ namespace Microsoft.AspNet.OData.Builder.Conventions.Attributes
                 return;
             }
 
+            // Respect order in the column attribute
+            var columnAttribute = attribute as ColumnAttribute;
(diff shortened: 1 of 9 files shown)


Galad:
Hello,
I am not able to see why the build fails. Could I get the details about the errors ?
Thanks

raheph:
@Galad Thanks for the PR, we will review this and try to ship it with the next release if things look good.

KanishManuja-MS:
@Galad 
image

Galad:
@KanishManuja-MS That's weird, the solution WebApiOData.sln does not exist in the repository as far as I know.
I surely would appreciate some help on that.

PR #1770, opened 2019-02-23
Add insert, update, delete restrictions

bdebaere:
Add insert, update, delete restrictions based on https://www.odata.org/blog/introducing-a-capabilities-vocabulary/. It is now possible to call IsNotInsertable(), IsNotUpdatable(), IsNotDeletable() on the NavigationSourceConfiguration and NavigationSourceConfiguration<T>. It is now also possible to call IsNotInsertable(), IsNotUpdatable(), IsNotDeletable() on the PropertyConfiguration. This results in additional annotations added where necessary.


bdebaere:
I don’t know why the build fails. I don’t have access to check the log. It builds successfully on my local branch.

xuzhg:
it seems the TestPublicApi failed, please update that in both projects.

bdebaere:
Sorry about that. Resolved.

xuzhg:
@bdebaere Thanks a lot for your contribution.

However, what i am thinking is that "do we need to build these functionality in Web API OData? Should be these at use side? " And it seems Capabilities changes a lot, see https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Capabilities.V1.md.
I'd invite @mikepizzo share his thoughts about it.

raheph:
@bdebaere It seems it's better to add such extension methods into OData.Edm library and use OData model builder pattern to call those extension methods.

mikepizzo:
Would really like to see this functionality added at the right level, and using the latest OASIS vocabulary.

PR #1786, opened 2019-03-21
handle string literal constant nodes in template parser

hippiehunter:
### Description
Fixes issue #1791 .

The purpose of this PR is to handle prevent failure during parsing of odata templates when driven by https://github.com/Microsoft/aspnet-api-versioning currently the parser produces operation segments and subsequently fails to build the parameter mapping when parsing something like the following "Availability/Default.FindAvailability(Region='{Region}',Unit='{Unit}',Member={Member},MemberArea={MemberArea},FunctionType='{FunctionType}',Adults={Adults},Children={Children},ReservationType={ReservationType},RateCode='{RateCode}')"

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RoutingConventionHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RoutingConventionHelpers.cs
@@ -320,6 +320,10 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
                     {
                         nameInRouteData = uriTemplateExpression.LiteralText.Trim();
                     }
+                    else
+                    {
+                        nameInRouteData = node.Value as string;
+                    }
                 }
                 else
                 {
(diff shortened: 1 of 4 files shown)


raheph:
@hippiehunter Would you please create an issue to mention the problem? And please add test cases to cover your changes?

hippiehunter:
@raheph, ok I've added my unit test, It looks like I put it in the right place, but please let me know if it should have gone somewhere else.

hippiehunter:
@raheph, I cant see the result of the failing CI run. This works locally for me.

mikepizzo:
It appears from https://github.com/microsoft/aspnet-api-versioning/issues/502 that this was an issue that will be addressed in the https://github.com/Microsoft/aspnet-api-versioning, and that we should not merge this PR, but will leave it open for now until we get confirmation.

commonsensesoftware:
@mikepizzo , @hippiehunter, I can confirm there is a bug in the OData route template generation in API Versioning.

The server side must generate templates in the same form you'd use with `[ODataRoute]`, but the API Explorer must generate templates with any additional characters such as quotes because clients that are not _OData-aware_ (ex: Swagger tools) will not know to include them in requests. It appears I have the issue resolved and the fix should be published in the near future.

mikepizzo:
Chris confirmed this is an issue in aspnet-api-versioning, and I believe has already fixed it, so closing this PR.  Thanks for your help getting this resolved!

PR #1742, opened 2019-01-21
[Regression] Fixes and coverage for primitives in URL parameters.

robertmclaws:
### Issues

*This pull request fixes issue #1741 .*

### Description

*Briefly describe the changes of this pull request.*

### Checklist (Uncheck if it is not completed)

- [X] *Test cases added*
- [X] *Build and test with one-click build and test script passed*

### Additional work necessary

Add any other primitive test cases as needed.

PR #1793, opened 2019-04-16
[Regression] Fix ODataModelBinderConverterTests in debug mode

kosinsky:
### Description

PR #1742 introduced regression when building in Debug mode (Visual Studio or .\build.ps -q)
Test for converting from string to Boolean will fail in EdmPrimitiveHelpers.ConvertPrimitiveValue on following Contart.Assert in debug mode 
```
 else
                {
                    Contract.Assert(type == typeof(uint) || type == typeof(ushort) || type == typeof(ulong));

                    // Note that we are not casting the return value to nullable<T> as even if we do it
                    // CLR would unbox it back to T.
                    return Convert.ChangeType(value, type, CultureInfo.InvariantCulture);
                }
```
however will work in Release mode, because "true" is convertible to bool.true
That affect development + for incorrect Boolean string will return internal .NET error instead of nice ODataException

To solved that case, I added special treatment for Boolean data type.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Common/SRResources.Designer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Common/SRResources.Designer.cs
@@ -1803,7 +1803,18 @@ namespace Microsoft.AspNet.OData.Common
                 return ResourceManager.GetString("PropertyMustBeDateTimeOffsetOrDate", resourceCulture);
             }
         }
-        
+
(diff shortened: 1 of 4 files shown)


xuzhg:
Merged at: https://github.com/OData/WebApi/commit/9a908285e9e235e95b46ffaf34f3d0efba9c020a

PR #1839, opened 2019-06-27
401 support

mikepizzo:
### Description

Adds support for serializing 4.01 compliant responses.

- [x] *Test cases added*


mikepizzo:
Note: test failure was an intermittent test issue due to random value generation, and has nothing to do with this PR. I re-queued the build trigger and everything passed.

Issue #194, opened 2015-01-21
Count with filter doesn't work in ODATA queries


LianwMS:
According to ODATA v4 Spec Protocol Part 

Section 11.2.5.1 System Query Option $filter

http://docs.oasis-open.org/odata/odata/v4.0/errata02/os/complete/part1-protocol/odata-v4.0-errata02-os-part1-protocol-complete.html#_Toc406398301 

The following query is supported in ODATA

http://host/service/Categories?$filter=Products/$count lt 10

However when I try this against a Web API 2.2 with OData4, I am getting the following error:

> "code":"","message":"The query specified in the URI is not valid. The parent value for a property access of a property '$count' is not a single value. Property access can only be applied to a single value.","innererror":{ "message":"The parent value for a property access of a property '$count' is not a single value. Property access can only be applied to a single value.","type":"Microsoft.OData.Core.ODataException","stacktrace":"   at Microsoft.OData.Core.UriParser.Parsers.EndPathBinder.BindEndPath....

Is this the expected behavior?

AndriyKuznetsov:
I have the same issue.

rpawlaszek:
Same here.

fazzamar:
According to the OData 4 standard this should work (http://docs.oasis-open.org/odata/odata/v4.0/errata02/os/complete/part2-url-conventions/odata-v4.0-errata02-os-part2-url-conventions-complete.html#_Toc406398087).
Any word on this?

mikila85:
Should work.. same here just not with filter but with orderby for my example:

http://localhost:60760/odata/CurrentlyAdmitted?$top=50&$orderby=admissionsLastYear/$count

getting:
<pre>
{
  "error":{
    "code":"","message":"The query specified in the URI is not valid. The parent value for a property access of a property '$count' is not a single value. Property access can only be applied to a single value.","innererror":{
[...]

PR #738, opened 2016-05-20
Issue #194 Count with filter support

VikingsFan:
### Issues

This pull request fixes #194

PR #1844, opened 2019-07-03
Support count in orderby and filter query options

genusP:
### Issues

fix issue #194 

### Description

This PR based on #738 

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -203,6 +203,43 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             }
         }
 
+        private Expression BindCountNode(CountNode node)
+        {
+            Expression source = Bind(node.Source);
+            Expression countExpression = Expression.Constant(null, typeof(long?));
+            Type elementType;
+            if (!TypeHelper.IsCollection(source.Type, out elementType))
+            {
+                return countExpression;
+            }
+
+            MethodInfo countMethod;
+            if (typeof(IQueryable).IsAssignableFrom(source.Type))
+            {
+                countMethod = ExpressionHelperMethods.QueryableCountGeneric.MakeGenericMethod(elementType);
+            }
+            else
+            {
+                countMethod = ExpressionHelperMethods.EnumerableCountGeneric.MakeGenericMethod(elementType);
+            }
+
+            // call Count() method. 
+            countExpression = Expression.Call(null, countMethod, new[] { source });
+
+            if (QuerySettings.HandleNullPropagation == HandleNullPropagationOption.True)
+            {
+                // source == null ? null : countExpression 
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -1425,6 +1462,12 @@ private Expression BindCustomMethodExpressionOrNull(SingleValueFunctionCallNode
         /// <returns>The LINQ <see cref="Expression"/> created.</returns>
         private Expression BindSingleValueNode(SingleValueNode node)
         {
+            CountNode countNode = node as CountNode;
+            if (countNode != null)
+            {
+                return BindCountNode(countNode);
+            }
+
             switch (node.Kind)
```
mikepizzo: Any reason this is not in the switch statement for case: QueryNodeKind.Count?

src/Microsoft.AspNet.OData.Shared/Query/OrderByNode.cs
```diff
@@ -60,6 +60,11 @@ public static IList<OrderByNode> CreateCollection(OrderByClause orderByClause)
             List<OrderByNode> result = new List<OrderByNode>();
             for (OrderByClause clause = orderByClause; clause != null; clause = clause.ThenBy)
             {
+                if (clause.Expression is CountNode)
+                {
+                    result.Add(new OrderByCountNode(clause));
+                    continue;
+                }
                 if (clause.Expression is NonResourceRangeVariableReferenceNode ||
```
mikepizzo: According to style guidelines, should insert a blank line after closing brace.

mikepizzo:
@genusP - Thanks for the contribution! This should make it into our upcoming 7.2 release.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
genusP: No reason. Fixed.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/FunctionConfigurationTest.cs
```diff
@@ -830,6 +830,27 @@ public void GetEdmModel_SetsDateTimeAsParameterType_WorksForDefaultConverter()
             Assert.True(parameter.Type.IsNullable);
         }
 
+        [Fact]
+        public void GetEdmModel_SetDecimalAsParameterType()
```
KanishManuja-MS: >GetEdmModel_SetDecimalAsParameterType [](start = 20, length = 37)

I wanted to make sure that I am not missing something here. Is this relevant to the change?

genusP: You are right. I remove this changes.

PR #1855, opened 2019-07-17
Flatten properties before using in aggregation to avoid nested queries.

kosinsky:
### Description

For queries like:  `$apply=groupby((Name), aggregate(Order/Price with sum as TotalPrice))` with aggregation on properties of related entity EF6 and EF Core generate suboptimal queries that lead for really bad performance.

**EF 6**
Before change SQL query looked like:
```SQL
SELECT 
    [Project3].[C1] AS [C1], 
    [Project3].[C2] AS [C2], 
    [Project3].[Name] AS [Name], 
    [Project3].[C3] AS [C3], 
    [Project3].[C4] AS [C4]
    FROM ( SELECT 
        [Project2].[Name] AS [Name], 
        [Project2].[C1] AS [C1], 
        [Project2].[C2] AS [C2], 
        N'TotalPrice' AS [C3], 
        [Project2].[C3] AS [C4]
        FROM ( SELECT 
            [Distinct1].[Name] AS [Name], 
            [Distinct1].[C1] AS [C1], 
            [Distinct1].[C2] AS [C2], 
            (SELECT 
                SUM([Extent3].[Price]) AS [A1]
                FROM  [dbo].[Customers] AS [Extent2]
                LEFT OUTER JOIN [dbo].[Orders] AS [Extent3] ON [Extent2].[Order_Id] = [Extent3].[Id]
                WHERE ([Distinct1].[C2] = N'Name') AND (([Distinct1].[Name] = [Extent2].[Name]) OR (([Distinct1].[Name] IS NULL) AND ([Extent2].[Name] IS NULL)))) AS [C3]
            FROM ( SELECT DISTINCT 
                [Extent1].[Name] AS [Name], 
                1 AS [C1], 
                N'Name' AS [C2]
                FROM [dbo].[Customers] AS [Extent1]
            )  AS [Distinct1]
        )  AS [Project2]
    )  AS [Project3]
    ORDER BY [Project3].[C4] ASC
```
Nested query with DISTINCT to get group by on Customers and than joins with Orders and Customers. That lead for double scan of customers table.

After change:
```SQL
SELECT 
    [Project1].[C2] AS [C1], 
    [Project1].[C3] AS [C2], 
    [Project1].[Name] AS [Name], 
    [Project1].[C4] AS [C3], 
    [Project1].[C1] AS [C4]
    FROM ( SELECT 
        [GroupBy1].[A1] AS [C1], 
        [GroupBy1].[K1] AS [Name], 
        [GroupBy1].[K2] AS [C2], 
        [GroupBy1].[K3] AS [C3], 
        N'TotalPrice' AS [C4]
        FROM ( SELECT 
            [Join1].[K1] AS [K1], 
            [Join1].[K2] AS [K2], 
            [Join1].[K3] AS [K3], 
            SUM([Join1].[A1]) AS [A1]
            FROM ( SELECT 
                [Extent1].[Name] AS [K1], 
                1 AS [K2], 
                N'Name' AS [K3], 
                [Extent2].[Price] AS [A1]
                FROM  [dbo].[Customers] AS [Extent1]
                LEFT OUTER JOIN [dbo].[Orders] AS [Extent2] ON [Extent1].[Order_Id] = [Extent2].[Id]
            )  AS [Join1]
            GROUP BY [K1], [K2], [K3]
        )  AS [GroupBy1]
    )  AS [Project1]
    ORDER BY [Project1].[C1] ASC
```
Just join Customers and Orders than aggregate.

**EF Core**
EF Core doing most of aggregations client-side anyway (EF Core 3.0 will change that). However, before change it first gets all orders and than joins with Customers client-side and only after that it will do aggregation
```SQL
SELECT [$it.Order0].[Id], [$it.Order0].[Price]
FROM [Order] AS [$it.Order0]
```
After change, group by will still happen client-side, but at least we will use join to get Customers and related orders in one query:
```SQL
SELECT [$it0].[Id], [$it0].[AddressId], [$it0].[Bucket], [$it0].[Name], [$it0].[OrderId], [$it.Order0].[Price]
FROM [Customers] AS [$it0]
LEFT JOIN [Order] AS [$it.Order0] ON [$it0].[OrderId] = [$it.Order0].[Id]
```

- [x] *Test cases added*

### Additional work necessary

After EF Core 3.0 done with support group by  we will need to validate and might by do some adjestments


src/Microsoft.AspNet.OData.Shared/Query/Expressions/AggregationBinder.cs
```diff
@@ -416,7 +511,7 @@ private Expression CreatePropertyAggregateExpression(ParameterExpression accum,
                         aggregationExpression = Expression.Call(null, sumMethod, asQuerableExpression, propertyLambda);
 
                         // For Dynamic properties cast back to object
-                        if (propertyAccessor.Type == typeof(object))
+                        if (propertyLambda.Type == typeof(object))
```
mikepizzo: >propertyLambda [](start = 28, length = 14)

Please verify that this should be propertyLambda.Type, and not body.Type.

PR #1868, opened 2019-07-31
Remove constructor from public surface; upcoming PR will remove it

KanishManuja-MS:
### Issues

*This pull request removes a public constructor introduced by the expand on $ref PR. An upcoming PR for the next release removes that constructor and therefore to avoid a breaking change, I makes sense to make the constructor internal for the time being.*

### Additional work necessary


robertmclaws:
Just so you folks are aware, taking a constructor from public to private is STILL a breaking change, and under the rules of semantic versioning, you would need to increase the major version of the WebAPI release in order to communicate that change to consumers.

mikepizzo:
Yep; this constructor was added in a previous *unreleased* PR. We are making it internal now, before releasing, so that we don't have a breaking change to the released code in the future.

KanishManuja-MS:
> Just so you folks are aware, taking a constructor from public to private is STILL a breaking change, and under the rules of semantic versioning, you would need to increase the major version of the WebAPI release in order to communicate that change to consumers.

I updated the description as I realized it was a little misleading. Anyway, thanks for looking out!

Issue #793, opened 2016-08-15
Support build in Function in aggregation


VikingsFan:
currently we can't parse:
    
    $apply=groupby((day(DateTime),month(DateTime),year(DateTime)),aggregate(Id with countdistinct as Count))

kosinsky:
Doing it in that way will violate specification. Spec says (http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016583 ): In its simplest form the first parameter of groupby specifies the grouping properties, a comma-separated list of one or more single-valued property paths (paths ending in a single-valued primitive, complex, or navigation property) that is enclosed in parentheses.

Following syntax is correct way to answer that question (it's not implemented yet too): 
`$apply=compute(day(DateTime) as day, month(DateTime) as month, year(DateTime) as year)/groupby((day,month,year),aggregate(Id with countdistinct as Count)
`

smlynch:
@kosinsky Any workaround for this in the meantime?

kosinsky:
Unfortunately, I don't have good way, however, grouping way by unique field (so no real aggregation) and introducing in the aggregate section and then using one more groupby could help.
`$apply=groupby((Id), aggregate(year(DateTime) with max as year, month(DateTime) with max as month), Id with max as Id)/groupby((year, month), ....).
`
I suspect performance on executing that query could be not very good depended on implementation of IQueryable

smlynch:
Thanks @kosinsky.  I meant to reply back earlier and let you know I worked around this by creating a view in the database and added `DateTimeYear`, `DateTimeMonth`, ... columns when I need to be able to group by this metric.

Seteh:
> Following syntax is correct way to answer that question (it's not implemented yet too):
`$apply=compute(day(DateTime) as day, month(DateTime) as month, year(DateTime) as year)/groupby((day,month,year),aggregate(Id with countdistinct as Count)`

Hi @kosinsky, is there any progress on it?

PR #1877, opened 2019-08-02
Added support for compute() transformation in $apply

### LINKED ISSUE #1370: OData Core Throws null reference exception when trying to execute
I'm working .Net Core using on the 7.0.0 Beta-2
http://localhost:59943/Api/Values/OData?$filter=substring(LastName, 1) eq 'u'
When trying to execute a simple filter on a query some functions throw a null reference exception

```        
        CustomContext _context;
   
        public ValuesController(CustomContext context)
        {
            this._context = context;
        }

        [HttpGet("OData")]
        public async Task<User[]> ODataFailingTestings(ODataQueryOptions<User> queryOptions)
        {
            var baseQuery = this._context.Users.AsQueryable();
            if (queryOptions != null)
            {
                baseQuery = queryOptions.ApplyTo(baseQuery, new ODataQuerySettings()) as IQueryable<User>;
            }

            var values = await baseQuery.ToArrayAsync();
            return values;
        }
```

kosinsky:
### Description
* Support for compute() transformation in $apply (http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016588 )
* Support for functions in aggregate()
* Changed Null value propagation to be true by default for EF Core. It's needed because EF Core InMemory and Client-side evaluation behave in the same way as Linq2Objects. I also fixes #1370 

Known limitation: compute() transformation must be used with groupby or aggregate. So we couldn't do `$apply=compute(1 as Test)` that is the same as `$compute=1 as Test`.  Limitation caused by the same problem as not supporting $select after $apply. 

### Issues
Fixes scenario from issue #793

- [x] *Test cases added*


src/Microsoft.AspNet.OData.Shared/Query/Expressions/ComputeBinder.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.OData.Edm;
+using Microsoft.OData.UriParser;
+using Microsoft.OData.UriParser.Aggregation;
+
+namespace Microsoft.AspNet.OData.Query.Expressions
+{
+    internal class ComputeBinder : TransformationBinderBase
+    {
+        private const string GroupByContainerProperty = "GroupByContainer";
```
xuzhg: where is it used?

xuzhg:
@kosinsky  i re-based the PR on the latest master. It seems some test cases are failing. Would you please take a look if you think it's better to release in 7.2.2 release.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ComputeBinder.cs
kosinsky: Removed

SenyaMur:
When it will be merged and when available for getting from nuget?
It will be available in Net Core 3.0 (#1748 ) ?

kosinsky:
These features (compute() and Core 3.0) aren't related. And compute() implementation doesn't do anything that could affect Core 3.0

KanishManuja-MS:
@SenyaMur It should be available with the next release. .Net Core support may come after the next release. As @kosinsky pointed out, this feature will carry forward to the .Net Core support release.

SenyaMur:
@KanishManuja-MS When will be available next release or pre-release?

VenuMadhavPerumalla:
@KanishManuja-MS - I see that issue is still open. Can you please let me know when this will be available

KanishManuja-MS:
@VenuMadhavPerumalla We will try to get this in our next WebAPI release. Thanks!

GitMonkey007:
@KanishManuja-MS - could you advise when the next WebAPI release will be? We’d also find this functionality useful...

KanishManuja-MS:
@GitMonkey007 We don't have date put down as of now but I would imagine towards the end of the month or early February. In the meanwhile, I encourage you to review this PR to see if it meets all your needs.

src/Microsoft.AspNet.OData.Shared/Query/HandleNullPropagationOptionHelper.cs
```diff
@@ -46,11 +46,13 @@ public static HandleNullPropagationOption GetDefaultHandleNullPropagationOption(
...
                 case ObjectContextQueryProviderNamespaceEF6:
-                case ObjectContextQueryProviderNamespaceEFCore2:
+             
                     options = HandleNullPropagationOption.False;
                     break;
-
+                
                 case Linq2ObjectsQueryProviderNamespace:
+                // EF Core before 3.0 does a lot of client evaluations and has InMemory support which is the same as Linq2Objects
+                case ObjectContextQueryProviderNamespaceEFCore2:
```
mikepizzo: Isn't this a breaking change?

kosinsky: Potentially. However, if you try to create expressions like $filter=length(Field) gt 10 when backed by EF InMemory it will throw with version of code that in master now. That's a bug.


When we generate IQueryable  provided by EF  and we have something like 
```C#
x => x.Field.Length > 10
```
 we assumed that EF will pass that operation to the SQL in a form
```SQL
LEN(Field) > 10
```
If Field is null it will; work in SQL, but executed directly C# expression throw exception and we have to convert if to
```C#
x => x.Field == null ? false : x.Field.Length > 10
```
We are doing it for linq to objects, but not EF InMemory

kosinsky: @mikepizzo Looks like we don't have tests for $filter and EF Core. Do you want me to:
* Create separate PR to add these tests
* Add these test to this PR
* Create an issue to work on extra EF Core tests after this PR is megred

mikepizzo: I'm fine with a separate PR for EF Core tests so that we don't hold up this PR any longer -- thanks!

src/Microsoft.AspNet.OData.Shared/Query/Expressions/DynamicTypeWrapper.cs
```diff
@@ -35,7 +40,7 @@ public bool TryGetPropertyValue(string propertyName, out object value)
     [JsonConverter(typeof(DynamicTypeWrapperConverter))]
     internal class GroupByWrapper : DynamicTypeWrapper
     {
-        private Dictionary<string, object> _values;
+        protected Dictionary<string, object> _values;
```
mikepizzo: >_values [](start = 45, length = 7)

If this is protected (and not private) it should follow standard (i.e., pascal) naming conventions.  Really, fields should not be protected (or public); the preferred mechanism is to provide getters/setters, or other methods for accessing/mutating.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/TransformationBinderBase.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            Contract.Assert(elementType != null);
+            _elementType = elementType;
+
+            this._lambdaParameter = Expression.Parameter(this._elementType, "$it");
+        }
+
+        protected Type _elementType;
+        protected ParameterExpression _lambdaParameter;
+        protected bool _classicEF = false;
```
mikepizzo: shouldn't really expose fields as protected/public but, if we do, should expose with standard (i.e., Pascal) naming.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/DynamicTypeWrapper.cs
```diff
@@ -123,5 +128,65 @@ internal class NoGroupByAggregationWrapper : GroupByWrapper
...
+        /// </summary>
+        public string ModelID { get; set; }
+
+        private bool _merged;
+        protected override void EnsureValues()
+        {
+            base.EnsureValues();
+            if (!this._merged)
+            {
+                // Base properties availbale via Instance can be real OData properties or generated in previous transformations
```
mikepizzo: >availbale [](start = 35, length = 9)

nit: typo

src/Microsoft.AspNet.OData.Shared/Query/Expressions/DynamicTypeWrapper.cs
```diff
@@ -123,5 +128,65 @@ internal class NoGroupByAggregationWrapper : GroupByWrapper
...
+                var instanceContainer = this.Instance as DynamicTypeWrapper;
+                if (instanceContainer != null)
+                {
+                    // Add proeprties generated in previous transformations to the collection
+                    _values.MergeWithReplace(instanceContainer.Values);
+                }
+                else
+                {
+                    // Add real OData properties to the collection
+                    // We need to use injected Model to real proeprty names
```
mikepizzo: >proeprty [](start = 61, length = 8)

nit: typo

src/Microsoft.AspNet.OData.Shared/Query/Expressions/DynamicTypeWrapper.cs
```diff
@@ -123,5 +128,65 @@ internal class NoGroupByAggregationWrapper : GroupByWrapper
...
+                    var edmType = GetEdmType() as IEdmEntityTypeReference;
```
mikepizzo: >IEdmEntityTypeReference [](start = 50, length = 23)

Will this cast always succeed, or might GetEdmType return something besides an IEdmEntityTypeReference?
What about complex types?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -648,25 +656,55 @@ protected void EnsureFlattenedPropertyContainer(ParameterExpression source)
...
+        {
+            while (expression.Method.Name == "Where")
+            {
+                expression = expression.Arguments.FirstOrDefault() as MethodCallExpression;
+            }
+
+            return expression;
+        }
+
+        private static void CollectContainerAssugments(Expression source, MethodCallExpression expression, Dictionary<string, Expression> result)
```
mikepizzo: >CollectContainerAssugments [](start = 28, length = 26)

spelling -- "CollectContainerAssignments"?

mikepizzo:
A few minor comments, mostly regarding:
1) Ability to apply Compute() to a complex type
2) Would prefer to provide a cleaner mechanism for mutating state than exposing fields from base types

Also a few minor spelling issues, but other than that it looks good -- thanks, and apologies for taking so long to get this reviewed!

mikepizzo:
Hoping to get a beta release out next week, with RTM a couple of weeks after, depending on feedback.

---
In reply to: [571371458](https://github.com/OData/WebApi/pull/1877#issuecomment-571371458) [](ancestors = 571371458)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/DynamicTypeWrapper.cs
kosinsky: For this PR (adding compute() to $apply), it should be true all the time, because we are using only ```EntitySet?$apply=....```

When we will work on $apply inside $expand (I have that feature in my fork). For queries like:
```EntitySet?$expand=ComplexTypeCollection($apply=...)``` it will change.  I think it will be really difficult to validate that ComputeWrapper works for complex types new before implementing $expand($apply) and/or $expand($compute)

File: src/Microsoft.AspNet.OData.Shared/Query/Expressions/TransformationBinderBase.cs [hunk 9]
```diff
@@ -0,0 +1,180 @@
...
+        internal virtual bool IsClassicEF(IQueryable query)
+        {
+            var providerNS = query.Provider.GetType().Namespace;
+            return (providerNS == HandleNullPropagationOptionHelper.ObjectContextQueryProviderNamespaceEF6
+                || providerNS == HandleNullPropagationOptionHelper.EntityFrameworkQueryProviderNamespace);
+        }
+
+        protected void SetElementType(Type elementType, ParameterExpression lambdaParameter)
+        {
+            Contract.Assert(elementType == lambdaParameter.Type);
```
mikepizzo: If this is the assert, why not just have a SetLambdaParameter and use it to set both?  Or even have ElementType getter return this.LambdaParameter.Type ?

kosinsky:
I'm not planning any additional changes. It's good to go as soon as CI completes. However, CI stuck in "waiting for status to be reported". Looks like build wasn't even started.

mikepizzo:
Was able to run CI and verify, although pipeline was unable to post results back to GitHub.

computrengineer:
Working great for us so far! Thanks everyone!! :)

SenyaMur:
How to get this feature ?

Issue #1894, opened 2019-09-08
Skip token does not respect EnableLowerCamelCase convention


FolkCoder:
When an `ODataConventionModelBuilder` enables camel-case formatting of property names via the `EnableLowerCamelCase` extension, generated skip tokens use the model's original property name rather than the formatted camel-case name. In models with non-integer key properties, the difference in casing between the skip token and the EDM property causes an exception to be throw when navigating to the next link.

### Assemblies affected

This issue is present in Microsoft.AspNetCore.OData v7.2.1.

### Reproduce steps

Create an API project that uses an ODataConventionModelBuilder with the EnableLowerCamelCase convention and request a paged collection of an entity containing a decimal key property. Navigate to the next page link.

A sample repository is available [here](https://github.com/FolkCoder/TestDecimalSkipToken).

### Expected result

When using camel-case formatting, the skip token identifier should match the formatting of its formatted EDM property.

### Actual result

A generated skip token contains the original property name rather than its formatted API property format (e.g., $skiptoken=Id-1 rather than @skiptoken=id-1). 

When using a decimal key, the following error is thrown:

> System.InvalidOperationException: \'The binary operator GreaterThan is not defined for the types 'System.Nullable\`1[System.Decimal]\' and '\System.Nullable\`1[System.Int32]\'.\'

### Additional detail

The `DefaultSkipHandler` class uses the CLR rather than the EDM property name to generate a skip token. When applying the paging specified in the next page link, the `DefaultSkipHandler` class fails to resolve the property associated with the skip token because it cannot match the formatted EDM property name against the original CLR property name. As a fallback, the `DefaultSkipHandler` determines the property type based on the skip token's value, with low-value numbers resolving to integers. When attempting to filter the collection, the expression cannot resolve a comparison between the decimal key defined by the CLR and the int key resolved by the skip handler.

mikepizzo:
Addressed  PR #1895

PR #1895, opened 2019-09-08
Ensure generated skip token respects casing conventions

FolkCoder:
### Issues

*This pull request fixes issue #1894.*

### Description

When an ODataConventionModelBuilder enables camel-case formatting of property names via the EnableLowerCamelCase extension, generated skip tokens use the model's original property name rather than the formatted camel-case name. This can cause unexpected errors when the skip token handler attempts to resolve the property during filtering operations.

This pull request modifies the default skip token handler to use formatted property names on the EDM rather than the original model.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
@@ -116,24 +116,24 @@ namespace Microsoft.AspNet.OData.Query
             int lastIndex = propertiesForSkipToken.Count() - 1;
             IEdmStructuredObject obj = lastMember as IEdmStructuredObject;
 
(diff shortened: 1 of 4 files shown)


KanishManuja-MS:
@FolkCoder Thanks for catching that and the PR. The changes look good to me.

mikepizzo:
Does this mean that the skip token handler expects the property names to be in camelCase, not in the casing specified in $metadata, if EnableCamelCase is true?

FolkCoder:
> Does this mean that the skip token handler expects the property names to be in camelCase, not in the casing specified in $metadata, if EnableCamelCase is true?

@mikepizzo `EnableCamelCase` seems to change the casing specified in $metadata already, which is what one would expect. So this pull request essentially aligns the skip token casing with the metadata casing.

mikepizzo:
Got it; thanks @FolkCoder!

Issue #1899, opened 2019-09-12
AmbiguousMatchException in ApplyTo method


komdil:
As you know in C# we can create the same properties in different cases. For example, I can create `Title `and `TITLE `properties. When I create properties like that in $select query of these properties I'm getting `AmbiguousMatchException `

### Reproduce steps
Create the same properties with different cases: 
`public class Product
{
    [Key]
    public Guid Id { get; set; }
    public string Name { get; set; }
    public decimal Price { get; set; }
    public string Category { get; set; }

    [NotMapped]
    public virtual string Title { get; set; }

    [NotMapped]
    public virtual string TITLE { get; set; }
}`

Include them in builder:
` builder.EntityType<Product>().Property(a => a.Title);
  builder.EntityType<Product>().Property(a => a.TITLE);`

Create controller with action: 
` public IHttpActionResult Get(ODataQueryOptions<Product> queryOptions, CancellationToken cancellationToken)
  {
     Context = GetContext();
     var products = Context.GetEntities<Product>();
     var result = queryOptions.ApplyTo(products);
     return Ok(result);
  }` 
Start service and send request: https://localhost:*****/Products?$select=Id,Title

### Expected result

I should be able to get Title and TITLE with $select query

### Actual result
Exception:
`System.Reflection.AmbiguousMatchException: 'Ambiguous match found.'`

xuzhg:
@komdil We should first search the property using exact naming, then using the case insensitive. It seems it's a bug in the code. Are you interested in contributing it?

madansr7:
closing since the fix has been merged to master.

PR #1907, opened 2019-09-23
AmbiguousMatchException in ApplyTo method

komdil:
### Issues

*This pull request fixes issue #1899 .*

### Description
As you know in C# we can create the same properties in different cases. For example, I can create Title and TITLE properties. When I create properties like that in ApplyTo method I was getting **Ambiguous match found exception.** The issue was coming from **Expression.Property(source, propertyName);** in **CreatePropertyValueExpressionWithFilter** method. It seems **Expression.Property** method throws this exception if the class has the same properties with different case sensitivity. So, I changed **propertyName** to **propertyInfo**

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -185,7 +185,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             }
 
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            var propertyInfo = source.Type.GetProperty(propertyName);
+            PropertyInfo propertyInfo = source.Type.GetProperty(propertyName);
             Expression propertyValue = Expression.Property(source, propertyInfo);
             Type nullablePropertyType = TypeHelper.ToNullable(propertyValue.Type);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -185,7 +185,8 @@ internal Expression CreatePropertyValueExpression(IEdmEntityType elementType, IE
             }
 
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            Expression propertyValue = Expression.Property(source, propertyName);
+            var propertyInfo = source.Type.GetProperty(propertyName);
```
mikepizzo: Note that, as a general coding practice, we try not to use "var" in our production code.

komdil: It is fixed. I found a lot of "var"s in codes:
image

Is anyone fixing them?

mikepizzo: Thanks for the quick follow-up @komdil!

We haven't gone back through the code to remove var, but we are fixing them opportunistically as we update the code and are trying not to add any new uses of var in the product code.

Issue #1676, opened 2018-11-09
The IN operator not working with Enum


flibustier7seas:
When filtering values of Enum by the IN operator, api returns 500 (Internal Server Error)..

### Assemblies affected

Microsoft.AspNet.OData 7.0.1

### Reproduce steps

```csharp
   public class User
    {
        public long Id { get; set; }
        public string Name { get; set; }
        public UserRole Role { get; set; }
    }

    public enum UserRole
    {
        Administrator,
        Manager,
    }
```
Request: `http://localhost:63000/v1/Users?$filter=Role in ('Administrator')`

### Expected result

```json
{
    "@odata.context": "http://localhost:63000/v1/$metadata#Users",
    "value": [
        {
            "Id": 1,
            "Name": "User_1",
            "Role": "Administrator"
        }
    ]
}
```

### Actual result

```json
{
    "error": {
        "code": "",
        "message": "An error has occurred.",
        "innererror": {
            "message": "The value \"Microsoft.OData.ODataEnumValue\" is not of type \"ODataExample.Model.UserRole\" and cannot be used in this generic collection.\r\nParameter name: value",
            "type": "System.ArgumentException",
[...]

PR #1910, opened 2019-09-24
Allow IN operator for enums

Danielku15:
### Issues

*This pull request fixes issue #1676 .*

### Description

The query parser for OData returns for enums not the direct enum value but a `ODataEnumValue`. This must be considered when building the collection within the `FilterBinder.BindCollectionConstantNode`

- [x] *Test cases added*

### Additional work necessary

For me already some tests are failing on the master branch. My changes did not introduce new failing tests on my environment.

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -13,6 +13,7 @@ using System.Runtime.CompilerServices;
 using Microsoft.AspNet.OData.Adapters;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Formatter;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -399,7 +400,7 @@ public virtual Expression BindCollectionConstantNode(CollectionConstantNode node
             // so using a foreach loop and doing an implicit cast from object to the CLR type of ItemType.
             foreach (ConstantNode item in node.Collection)
             {
-                castedList.Add(item.Value);
+                castedList.Add(EnumDeserializationHelpers.ConvertEnumValue(item.Value, constantType));
```
mikepizzo: For readability, I would prefer to break this into two lines; one that tries to convert the enum value and the second that does the add.

PR #1920, opened 2019-10-03
Add a compatibility flag for generating next link with negative top v…

KanishManuja-MS:
…alues as well

### Issues

*This pull request fixing a paging issue in 7.2.1.  *

### Description

This PR introduces a new flag for backward compatibility. We started returning null for nextlink with OData WebAPI 7.2.0 if the page size value was larger than the $top value. However, although it was a bug fix but it was a breaking change. The flag should now give the flexibility to generate a next link for such scenarios as well. 

- [x] *Test cases added*


src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -293,8 +293,11 @@ public static Uri GetNextPageLink(this HttpRequest request, int pageSize, object
                 uriBuilder.Port = request.Host.Port.Value;
             }
 
+            ODataOptions odataOptions = request.HttpContext.RequestServices.GetRequiredService<ODataOptions>();
+            ODataCompatibilityOptions compatibilityOptions = odataOptions.CompatibilityOptions;
+
```
mikepizzo: Consider a(n internal?) extension method on HttpRequest for getting compatibility options. Then shared code can call this method and we don't have to have different logic each place we get compatibility options.

src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+using System;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Contains bitmasks for features that need backward compatibilty.
+    /// </summary>
+    [Flags]
+    public enum CompatibilityOptions
```
xuzhg: >CompatibilityOptions [](start = 16, length = 20)

Compatibility naming sounds like to compatibility to a certain previous version, Right? But, The main of this options is for the propose?

Issue #1937, opened 2019-10-23
string.Replace is missing


bdebaere:
The replace functionality on string properties is missing. For example /odata/MyEntity?$filter=replace(string, 'query', 'query2') eq 'query2' throws with exception "Unknown function 'replace'".

mikepizzo:
We actually intentionally omit string.Replace because it was identified as a possible DOS attack surface. Using string.Replace, a bad actor could potentially create a request that would overflow the server.

Life was some much easier before we really had security...

bdebaere:
@mikepizzo Interesting, would you consider adding Regex.Match?

PR #1938, opened 2019-10-23
Implement string.Replace method

bdebaere:
This pull request adds the string.Replace method and fixes issue https://github.com/OData/WebApi/issues/1937.


mikepizzo:
Thanks for the contribution.  Unfortunately, we cannot take it as implementation of string.Replace was identified in a security review as a potential DOS attack surface.

Issue #1947, opened 2019-11-01
ODataValueProvider is not used


bdebaere:
The `ODataValueProvider` which is being added is not being used due to an ASP.NET implementation taking precedence.

The problem lies in `ODataServiceCollectionExtensions.AddOData(this IServiceCollection services)` which performs an add instead of an insert for

```csharp
                // Add the value provider.
                options.ValueProviderFactories.Add(new ODataValueProviderFactory());
```

PR #1948, opened 2019-11-01
Use ODataValueProvider

bdebaere:
### Issues

This pull request fixes issue https://github.com/OData/WebApi/issues/1947.

### Description

Inserts the ODataValueProviderFactory instead of adding it, which causes ODataValueProvider to take precedence over the default ASP.NET implementation.
A conversion for double values to string with a culture invariant approach was added. Before this, some tests (RoutesCorrectly) were failing because 1.2D was converted to "12". The default ASP.NET implementation which was used does not cover this use case. It seems that the conversion which is currently implemented by the ODataValueProvider is what ASP.NET used i.e. ASP.NET used `var stringValue = value as string ?? value?.ToString() ?? string.Empty;`, instead now they use `string stringValue = value as string ?? Convert.ToString(value, this.culture) ?? string.Empty;` which has a consistent result across cultures instead.

Either way, depending on whether ODataValueProvider should be used, ODataValueProvider should either be used or removed.


mikepizzo:
Thanks for the contribution. Change looks good to me. Are there any relevant tests that could be added to verify that the change addresses the original issue and any edge cases?

xuzhg:
Merged at c62e8b17fb0c22dbadb51d6db3aa3de19e2ba1ee

PR #1949, opened 2019-11-05
Fix EdmStructuredObject default value getter

adzhilyanov:
### Issues

`Activator.CreateInstance(clrType)` does not work when `clrType` is `System.String` or `System.Byte[]` (_System.MissingMethodException : No parameterless constructor defined for this object._) which are mapped to `Edm.String` and `Edm.Binary`.

### Description

Return empty string when the type is `Edm.String[Nullable=False]` and empty array when the type is `Edm.Binary[Nullable=False]`.

- [x] *Test cases added*

### Additional work necessary
Not sure what to do about `Edm.Stream`, `Edm.Geography**` and `Edm.Geometry**` properties.


src/Microsoft.AspNet.OData.Shared/EdmStructuredObject.cs
```diff
@@ -221,6 +221,14 @@ internal static object GetDefaultValue(IEdmTypeReference propertyType)
                     (isCollection && propertyType.AsCollection().ElementType().IsPrimitive()))
                 {
                     // primitive or primitive collection
+                    if (clrType == typeof(string))
```
xuzhg: Edm.Primitive (System.String) is a built-in Primitive type.

EdmStructuredObject is used for complex or entity instance, the type of them is structured type. So, I don't know why should we add "primitive type" into "structured type"?

adzhilyanov: Hey, maybe the title is a bit misleading. It should be something like 'Fix EdmStructuredObject's properties default value getter'. 
You are correct, but the `EdmStructuredObject` has properties, which can be primitives. So when you try to get a property from the `EdmStructuredObject`, if the property does not exist, a default value is returned. This PR aimed to fix how the default value for a property is obtained - fixing the `GetDefaultValue` method (which is used from `TryGetPropertyValue`). I don't know if this is the right place for the `GetDefaultValue` though.

PR #1963, opened 2019-11-20
Enable Route-based routing in ASP.NET Core 3.0

xuzhg:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/samples/AspNetCore3xODataSample.Web/Program.cs
+++ b/samples/AspNetCore3xODataSample.Web/Program.cs
@@ -17,10 +17,6 @@ namespace AspNetCore3xODataSample.Web
             Host.CreateDefaultBuilder(args)
                 .ConfigureWebHostDefaults(webBuilder =>
                 {
-                    webBuilder.ConfigureKestrel(options =>
-                    {
-                        options.AllowSynchronousIO = true;
-                    });
                     webBuilder.UseStartup<Startup>();
                 });
     }
--- a/samples/AspNetCore3xODataSample.Web/Program.cs
+++ b/samples/AspNetCore3xODataSample.Web/Program.cs
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
+
+namespace AspNetCore3xODataSample.Web
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            CreateHostBuilder(args).Build().Run();
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.ConfigureKestrel(options =>
+                    {
+                        options.AllowSynchronousIO = true;
+                    });
+                    webBuilder.UseStartup<Startup>();
+                });
+    }
+}
--- a/src/Microsoft.AspNetCore.OData/ODataValueProviderFactory.cs
+++ b/src/Microsoft.AspNetCore.OData/ODataValueProviderFactory.cs
@@ -5,7 +5,10 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
-using Microsoft.AspNetCore.Mvc.Internal;
+#if NETCOREAPP3_0
+#else
+    using Microsoft.AspNetCore.Mvc.Internal;
+#endif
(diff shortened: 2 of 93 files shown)


samples/AspNetCore3xODataSample.Web/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.ConfigureKestrel(options =>
+                    {
+                        options.AllowSynchronousIO = true;
+                    });
```
rynowak: This should go closer to where you do I/O. Don't expect users to understand this detail or how to work around it.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
```diff
@@ -74,12 +77,18 @@ public static IRouteBuilder Create()
 
             // Create a route build with a default path handler.
             IRouteBuilder routeBuilder = new RouteBuilder(appBuilder);
+
+#if NETCOREAPP3_0
+            routeBuilder.DefaultHandler = null;
+#else
```
rynowak: A few options to work around this:
- Write your own route handler for test purpose (mock)
- Use reflection to retrieve the type and get it from the service provider
- Call `UseMvc` and then use that route builder
- Come up with a different strategy for testing

I don't have the full context on what you're trying to do here.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Aggregation/AggregationController.cs
```diff
@@ -87,6 +87,11 @@ public virtual bool CleanCommands()
         {
             return true;
         }
+
+        public void Dispose()
+        {
+           // _db.Dispose();
```
vladimir-shmidt: Once this is not used then better to remove disposable interface

samples/AspNetCore3xODataSample.Web/Program.cs
mikepizzo: @saxu@microsoft.com - is this line still needed? We do now set AllowSynchronousIO in the Input and Output formatters.

---
In reply to: [348247017](https://github.com/OData/WebApi/pull/1963#discussion_r348247017) [](ancestors = 348247017)

xuzhg: Yes. I will remove that later.

---
In reply to: [356331132](https://github.com/OData/WebApi/pull/1963#discussion_r356331132) [](ancestors = 356331132,348247017)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Aggregation/AggregationController.cs
xuzhg: Hmm... yes.

---
In reply to: [352288339](https://github.com/OData/WebApi/pull/1963#discussion_r352288339) [](ancestors = 352288339)

mikepizzo: Added the interface because CA was complaining that the controller was not disposable.  However, if we actually call dispose on the context here, the context gets disposed after the controller returns but before serialization, so we get a null ref exception during serialization.

---
In reply to: [356372476](https://github.com/OData/WebApi/pull/1963#discussion_r356372476) [](ancestors = 356372476,352288339)

mikepizzo:
>     <title>Microsoft ASP.NET Core 2.0 for OData v4.0</title>

Sholdn't the title change based on target framework?

---
Refers to: src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.Release.nuspec:5 in 0ec9a04. [](commit_id = 0ec9a04f807d18fc328620adb49cffda33ba69c6, deletion_comment = False)

src/Microsoft.AspNetCore.OData/ODataValueProviderFactory.cs
```diff
@@ -5,7 +5,10 @@
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
-using Microsoft.AspNetCore.Mvc.Internal;
+#if NETCOREAPP3_0
+#else
```
mikepizzo: Rather than check for NETCOREAPP3_0 (which would break when targeting 3_1 or 5, would prefer:

#if NETSTANDARD2_0
   using Microsoft.AspNetCore.Mvc.Internal
#endif

xuzhg:
merged at 5b7145ab7a31bf6dd91651f925caf893502ba8a1, Thanks.

Issue #1965, opened 2019-11-21
Inconsistent behavior on 'key' parameter convention for OData routes


julealgon:
The OData standard route convention for entity sets with a single ID property does not behave consistently when using composite key convention: the entityset route is matched, but binding does not occur and arguments have default values.

My team spent a couple hours trying to figure out why the value was 0 on the controller even though they were passing the correct value in the URL.

### Assemblies affected

- Microsoft.AspNetCore.OData 7.2.2

### Reproduce steps

1. Create an entity with a single key property.
2. Create a get endpoint by key, and use `key{PropertyName}` as the argument (instead of `key`)
3. Call into the route, passing a value for the key: eg `GET /odata/entities(10)`
4. Notice how the route matches and hits the controller, but `key{PropertyName}` is 0

Use the following repo as a minimal repro:
- https://github.com/julealgon/ODataBindingIssue.git

`GET http://localhost:33616/odata/entities(10)`

![image](https://user-images.githubusercontent.com/461579/69378059-4d4bf600-0c8c-11ea-9b53-b2e3678cbaf8.png)

VS

![image](https://user-images.githubusercontent.com/461579/69378567-64d7ae80-0c8d-11ea-8b23-01b4229a8fe2.png)

### Expected result

2 possibilities:
1. The route should NOT be matched when the key convention is incorrect. Follow existing convention: `key` for single key, `keyX`, `keyY` for composite X/Y key
2. The route should match and the value should be BOUND. Allow for composite key convention when not using composite keys: bind either `key` OR `keyX` for entity X key property

The second seems slightly more intuitive to me, but the first also makes sense if we want to follow strict conventions in the library.

### Actual result

Route matches, but key value is not bound.

### Additional detail

The most puzzling aspect of this problem is how inconsistent it behaves, and leads people into incorrect assumptions.

On scenario 3 below, if I change from `keyOnly` to `keyThisDoesNotExist`, the route does not match anymore. It's clearly using a separate process to define the route match and then to perform argument binding.

These processes should use the same logic to avoid this problem.

Scenario 1: Composite Key + Composite Key convention
✔️  CORRECT

Matches route: `GET /server/entities(10,'a')`

```
class Entity
{
    [Key]
    int First { get; set; }

    [Key]
    string Second { get; set; }
}

class EntitiesController
{
    IQueryable<Entity> Get(int keyFirst, string keySecond)
    ...
}

```
```
keyFirst = 10
keySecond = "a"
```

---

[...]

PR #1975, opened 2019-12-09
Pass keyId parameter to controller actions of single-key entities

habbes:
### Issues

This pull request fixes issue #1965

### Description

Assuming an entity has a single key property called `ID`.

A convention controller action for that entity will not support having `keyID` as the parameter name, in addition to `key` for routes that contain a key segment . That is, either `key` or `keyID` will work and receive the bound value.

i.e. this will still work as it currently does:
```cs
// GET /odata/Customers(10)

public IActionResult<Customer> Get(int key)
{
    // key == 10
}
```

and this will now also work:
```cs
// GET /odata/Customers(10)

public IActionResult<Customer> Get(int keyID)
{
    // key == 10
}
```

**Note**: Both `key` and `keyID` are added to the route data regardless of the actual name of the parameter used on the controller action. So this fix relies on the fact that values that do not correspond to action parameters are ignored.

- [x] *Test cases added*

### Additional work necessary

Documentation will need an update mentioning that it's now possible to bind route key segments to `key{KeyPropertyName}` action parameters by default using convention-based routing.


xuzhg:
Merged at 94764eea985bde8f8b4fa154513eb9afc2800677

PR #1977, opened 2019-12-09
Add support for derived type constraint annotation

KanishManuja-MS:
### Issues

This pull request enables support for derived type constraint annotation. 

### Description
Non-convention model builder now has APIs and thus support for adding out of line annotations for derived type constraint. 

- [x] *Test cases added*


src/Microsoft.AspNet.OData.Shared/Builder/ActionConfiguration.cs
```diff
@@ -233,6 +233,15 @@ public ActionConfiguration SetBindingParameter(string name, IEdmTypeConfiguratio
             return this;
         }
 
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public ActionConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
+        {
+            base.AddDerivedTypeConstraintToReturnTypeImplementation(typeof(TSubClass));
```
xuzhg: >base [](start = 12, length = 4)

"base. " is unnecessary?

src/Microsoft.AspNet.OData.Shared/Builder/CollectionPropertyConfiguration.cs
```diff
@@ -63,5 +63,14 @@ public CollectionPropertyConfiguration IsRequired()
             OptionalProperty = false;
             return this;
         }
+
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public CollectionPropertyConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
+        {
+            this.AddDerivedTypeConstraintImpl(typeof(TSubClass));
```
xuzhg: >this [](start = 12, length = 4)

"this." is unnecessary?

xuzhg: same for others

---
In reply to: [358414860](https://github.com/OData/WebApi/pull/1977#discussion_r358414860) [](ancestors = 358414860)

src/Microsoft.AspNet.OData.Shared/Builder/SingletonConfiguration.cs
```diff
@@ -41,5 +41,14 @@ public SingletonConfiguration(ODataModelBuilder modelBuilder, EntityTypeConfigur
             : base(modelBuilder, entityType, name)
         {
         }
+
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public SingletonConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
```
xuzhg: >AddDerivedTypeConstraint<TSubClass>() [](start = 38, length = 37)

For design perspective, add a non-generic method to support adding TypeConstraint.

Call the non-generic method in the generic method.

xuzhg: Same for others.

---
In reply to: [358416790](https://github.com/OData/WebApi/pull/1977#discussion_r358416790) [](ancestors = 358416790)

xuzhg:
1) Add the non-generic mothed to support runtime configuration.
2) Call non-generic method in your generic method
3) Where's conventional model build supporting?
4) Add the E2E test cases to verify the type constraint working, includes but not limits
  a)  to verify ODL throw correct exception for non-supported type cast
  b) to verify the ODL throw correct exception for non-supported instance serialization

src/Microsoft.AspNet.OData.Shared/Builder/EdmTypeBuilder.cs
```diff
@@ -26,6 +26,7 @@ internal class EdmTypeBuilder
         private readonly Dictionary<PropertyInfo, IEdmProperty> _properties = new Dictionary<PropertyInfo, IEdmProperty>();
         private readonly Dictionary<IEdmProperty, QueryableRestrictions> _propertiesRestrictions = new Dictionary<IEdmProperty, QueryableRestrictions>();
         private readonly Dictionary<IEdmProperty, ModelBoundQuerySettings> _propertiesQuerySettings = new Dictionary<IEdmProperty, ModelBoundQuerySettings>();
+        private readonly Dictionary<IEdmProperty, PropertyConfiguration> _propertyConfigurations = new Dictionary<IEdmProperty, PropertyConfiguration>();
```
xuzhg: >_propertyConfigurations [](start = 73, length = 23)

We can't use the "_properties" field?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -638,13 +652,64 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
...
+        }
+
+        internal static void AddDerivedTypeConstraintAnnotation(this EdmModel model, IEdmVocabularyAnnotatable target, 
+            Dictionary<Type, IEdmType> edmTypeMap, List<Type> types)
+        {
+            IEdmTerm term = ValidationVocabularyModel.DerivedTypeConstraintTerm;
+            if (types.Count == 0)
+            {
+                return;
+            }
```
xuzhg: switch line 689 and ling 690

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -638,13 +652,64 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
...
+            List<EdmStringConstant> collectionConstants = new List<EdmStringConstant>();
+            foreach (var type in types)
+            {
+                if (edmTypeMap.ContainsKey(type))
+                {
+                    collectionConstants.Add(new EdmStringConstant(edmTypeMap[type].FullTypeName()));
+                }
+                else
+                {
+                    throw Error.InvalidOperation("Cannot find EdmType corresponding to {0}", type);
```
xuzhg: >Cannot find EdmType corresponding to {0} [](start = 50, length = 40)

use the SRSources for the error message?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -638,13 +652,64 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
...
+        internal static void AddDerivedTypeConstraintAnnotation(this EdmModel model, IEdmVocabularyAnnotatable target, 
+            Dictionary<Type, IEdmType> edmTypeMap, List<Type> types)
+        {
+            IEdmTerm term = ValidationVocabularyModel.DerivedTypeConstraintTerm;
+            if (types.Count == 0)
+            {
+                return;
+            }
+
+            List<EdmStringConstant> collectionConstants = new List<EdmStringConstant>();
```
xuzhg: >List<EdmStringConstant> collectionConstants = new List<EdmStringConstant>(); [](start = 12, length = 76)

shall we make sure no duplicated existed?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -1,18 +1,19 @@
...
 using Microsoft.OData.Edm.Csdl;
 using Microsoft.OData.Edm.Validation;
 using Microsoft.OData.Edm.Vocabularies;
+using Microsoft.OData.Edm.Vocabularies.V1;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Reflection;
```
xuzhg: move System.* ahead

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
```diff
@@ -240,6 +246,25 @@ internal void SetBindingParameterImplementation(string name, IEdmTypeConfigurati
             _bindingParameter = new BindingParameterConfiguration(name, bindingParameterType);
         }
 
+        internal virtual void AddDerivedTypeConstraintToReturnTypeImplementation(Type t)
```
xuzhg: > t [](start = 85, length = 2)

rename the parameter name

xuzhg: >virtual [](start = 17, length = 7)

I don't think it's virtual method, right?

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
```diff
@@ -240,6 +246,25 @@ internal void SetBindingParameterImplementation(string name, IEdmTypeConfigurati
...
+        {
+            if (ReturnType.ClrType == null)
+            {
+                throw Error.InvalidOperation("No CLR Type specified for this configuration.");
```
xuzhg: >No CLR Type specified for this configuration. [](start = 46, length = 45)

centralize the error messages

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
```diff
@@ -240,6 +246,25 @@ internal void SetBindingParameterImplementation(string name, IEdmTypeConfigurati
             _bindingParameter = new BindingParameterConfiguration(name, bindingParameterType);
         }
 
+        internal virtual void AddDerivedTypeConstraintToReturnTypeImplementation(Type t)
+        {
+            if (ReturnType.ClrType == null)
```
xuzhg: >ReturnType [](start = 16, length = 10)

ReturnType maybe "null", right?

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
xuzhg: >Type t [](start = 81, length = 6)

t maybe null

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -61,6 +64,37 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
...
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public ParameterConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
+        {
+            Type t = typeof(TSubClass);
+
+            if (TypeConfiguration.ClrType == null)
+            {
+                throw Error.InvalidOperation("No CLR Type specified for this configuration.");
```
xuzhg: >No CLR Type specified for this configuration. [](start = 46, length = 45)

centralize the error messages

src/Microsoft.AspNet.OData.Shared/Builder/PropertyConfiguration.cs
```diff
@@ -157,6 +164,29 @@ public bool Unsortable
...
+            }
+
+            if (DerivedTypeConstraints.Contains(t))
+            {
+                throw Error.InvalidOperation("Type {0} already added as derived type constraint", t.Name);
+            }
+
+            DerivedTypeConstraints.Add(t);
+            return this;
+        }
```
xuzhg: It seems similar method repeated a lot, maybe consider to create a helper method

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBuilder/ExplicitModelBuilderTests.cs
```diff
@@ -135,56 +144,60 @@ private static IEdmModel GetExplicitEdmModel()
...
+
+            var action = modelBuilder.Action("GetAddress").Returns<Address>().AddDerivedTypeConstraint<Address>();
+            action.Parameter<Supplier>("supplier").AddDerivedTypeConstraint<ToiletPaperSupplier>();
 
             return modelBuilder.GetEdmModel();
         }
 
         [Fact]
-
         public async Task VerifyMetaDataIsGeneratedCorrectly()
```
xuzhg: >VerifyMetaDataIsGeneratedCorrectly() [](start = 26, length = 36)

Add test cases in "MetadataControllerTest.cs" to verify the annotation showing in the xml

xuzhg: need E2E test cases to verify:
1) Uri parsing
2) Serialization

---
In reply to: [358425347](https://github.com/OData/WebApi/pull/1977#discussion_r358425347) [](ancestors = 358425347)

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -638,13 +652,64 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
...
+                    collectionConstants.Add(new EdmStringConstant(edmTypeMap[type].FullTypeName()));
+                }
+                else
+                {
+                    throw Error.InvalidOperation("Cannot find EdmType corresponding to {0}", type);
+                }
+            }
+
+            var collectionExpression = new EdmCollectionExpression(collectionConstants);
+            EdmVocabularyAnnotation annotation = new EdmVocabularyAnnotation(target, term, collectionExpression);
```
xuzhg: >EdmVocabularyAnnotation annotation = new EdmVocabularyAnnotation(target, term, collectionExpression); [](start = 12, length = 101)

Discussion: Maybe need configuration to set the annotation location (Inline or outofLine)

src/Microsoft.AspNet.OData.Shared/Builder/PropertyConfiguration.cs
```diff
@@ -157,6 +164,29 @@ public bool Unsortable
         /// </summary>
         public QueryConfiguration QueryConfiguration { get; set; }
 
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
```
xuzhg: >TSubClass [](start = 17, length = 9)

resentence the comments

src/Microsoft.AspNet.OData.Shared/Builder/PropertyConfiguration.cs
```diff
@@ -157,6 +164,29 @@ public bool Unsortable
...
+        /// </summary>
+        internal PropertyConfiguration AddDerivedTypeConstraintImpl(Type t)
+        {
+            if (RelatedClrType == null)
```
xuzhg: >RelatedClrType [](start = 16, length = 14)

why do we test this? It should be tested on the constructor, right?

src/Microsoft.AspNet.OData.Shared/Builder/PropertyConfiguration.cs
```diff
@@ -157,6 +164,29 @@ public bool Unsortable
...
+            {
+                throw Error.InvalidOperation("No CLR Type specified for this configuration.");
```
xuzhg: >No CLR Type specified for this configuration. [](start = 46, length = 45)

add the errors into resource manage

xuzhg: same for other

---
In reply to: [362588158](https://github.com/OData/WebApi/pull/1977#discussion_r362588158) [](ancestors = 362588158)

src/Microsoft.AspNet.OData.Shared/Builder/PropertyConfiguration.cs
```diff
@@ -157,6 +164,29 @@ public bool Unsortable
         /// </summary>
         public QueryConfiguration QueryConfiguration { get; set; }
 
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        internal PropertyConfiguration AddDerivedTypeConstraintImpl(Type t)
```
xuzhg: >t [](start = 73, length = 1)

rename the parameter name

src/Microsoft.AspNet.OData.Shared/Builder/SingletonConfiguration.cs
```diff
@@ -41,5 +41,14 @@ public SingletonConfiguration(ODataModelBuilder modelBuilder, EntityTypeConfigur
...
+        {
+            this.AddDerivedTypeConstraintImpl(typeof(TSubClass));
```
xuzhg: >this. [](start = 12, length = 5)

don't need "this."

src/Microsoft.AspNet.OData.Shared/Builder/NavigationSourceConfiguration.cs
```diff
@@ -457,6 +464,28 @@ public virtual IEnumerable<NavigationPropertyBindingConfiguration> FindBindings(
             return Enumerable.Empty<NavigationPropertyBindingConfiguration>();
         }
 
+        /// <summary>
+        /// Adds types to the list of derived type constraints.
+        /// </summary>
+        internal void AddDerivedTypeConstraintImpl(Type t)
+        {
+            if (ClrType == null)
```
xuzhg: >if (ClrType == null) [](start = 12, length = 20)

we don't need to test this, right?
It's verify at the constuctor

src/Microsoft.AspNet.OData.Shared/Builder/NavigationPropertyConfiguration.cs
```diff
@@ -105,6 +105,15 @@ public IEnumerable<PropertyInfo> PrincipalProperties
             get { return _referentialConstraint.Values; }
         }
 
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public NavigationPropertyConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
+        {
+            this.AddDerivedTypeConstraintImpl(typeof(TSubClass));
```
xuzhg: remove "this."

src/Microsoft.AspNet.OData.Shared/Builder/EnumPropertyConfiguration.cs
```diff
@@ -72,5 +72,14 @@ public EnumPropertyConfiguration IsConcurrencyToken()
             ConcurrencyToken = true;
             return this;
         }
+
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public EnumPropertyConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
```
xuzhg: >EnumPropertyConfiguration [](start = 15, length = 25)

do we have the enum property derived type scenario?

src/Microsoft.AspNet.OData.Shared/Builder/EntitySetConfiguration.cs
```diff
@@ -66,5 +66,14 @@ public virtual NavigationSourceConfiguration HasFeedSelfLink(Func<ResourceSetCon
         {
             return _feedSelfLinkFactory;
         }
+
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public EntitySetConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
+        {
+            this.AddDerivedTypeConstraintImpl(typeof(TSubClass));
```
xuzhg: >this. [](start = 12, length = 5)

remove this.

src/Microsoft.AspNet.OData.Shared/Builder/ComplexPropertyConfiguration.cs
```diff
@@ -52,5 +52,14 @@ public ComplexPropertyConfiguration IsRequired()
             OptionalProperty = false;
             return this;
         }
+
+        /// <summary>
+        /// Adds TSubClass to the list of derived type constraints.
+        /// </summary>
+        public ComplexPropertyConfiguration AddDerivedTypeConstraint<TSubClass>() where TSubClass : class
+        {
+            this.AddDerivedTypeConstraintImpl(typeof(TSubClass));
```
xuzhg: >this. [](start = 12, length = 5)

remove this.

src/Microsoft.AspNet.OData.Shared/Builder/EdmTypeBuilder.cs
KanishManuja-MS: We need PropertyConfiguration. _properties does not have any configuration. 

---
In reply to: [358418941](https://github.com/OData/WebApi/pull/1977#discussion_r358418941) [](ancestors = 358418941)

src/Microsoft.AspNet.OData.Shared/Builder/SingletonConfiguration.cs
xuzhg: somthing likes `public SingletonConfiguration AddDerivedTypeConstraint(params Type type)
{}`

---
In reply to: [358416861](https://github.com/OData/WebApi/pull/1977#discussion_r358416861) [](ancestors = 358416861,358416790)

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
KanishManuja-MS: We do that through the ISet. At this point in time, there should be no duplicates. 


---
In reply to: [358420621](https://github.com/OData/WebApi/pull/1977#discussion_r358420621) [](ancestors = 358420621)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DerivedTypeConstraintAttribute"/> class
+        /// with the allowed types. 
+        /// </summary>
+        /// <param name="types">CLR types </param>
+        public DerivedTypeConstraintAttribute(params Type[] types)
+        {
+            if (DerivedTypeConstraints == null)
```
mikepizzo: >if (DerivedTypeConstraints == null) [](start = 12, length = 35)

This is a constructor; wouldn't DerivedTypeConstraints always be null?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                if (!DerivedTypeConstraints.Contains(type))
+                {
+                    DerivedTypeConstraints.Add(type);
+                }
+                else
+                {
+                    throw Error.InvalidOperation(SRResources.ConstraintAlreadyExists, type.Name);
+                }
+            }
```
mikepizzo: HashSet.Add returns false if the item already existed, so could just do:

                if (!DerivedTypeConstraints.Add(type))
                {
                    throw Error.InvalidOperation(SRResources.ConstraintAlreadyExists, type.Name);
                }

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    internal static class DerivedTypeConstraintHelper
+    {
+        internal static void ValidateAndAddConstraint(this ISet<Type> Constraints, Type ClrType, Type typeToAdd)
+        {
+            if (ClrType == null)
+            {
+                throw Error.InvalidOperation(SRResources.NoClrTypeSpecified);
+            }
+            else if (!typeToAdd.IsSubclassOf(ClrType) && typeToAdd != ClrType)
```
mikepizzo: >IsSubclassOf [](start = 32, length = 12)

Consider using IsAssignableFrom rather than IsSubclassOf.

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -191,28 +192,34 @@ private static string ConvertBindingPath(EdmTypeMap edmMap, NavigationPropertyBi
...
                     }
                 }
                 else
                 {
-                    IEdmOperationParameter operationParameter = new EdmOperationParameter(operation, parameter.Name, parameterTypeReference);
+                    operationParameter = new EdmOperationParameter(operation, parameter.Name, parameterTypeReference);
                     operation.AddParameter(operationParameter);
                 }
+
+                if (parameter.DerivedTypeConstraints.Any())
```
mikepizzo: >parameter.DerivedTypeConstraints.Any() [](start = 20, length = 38)

maybe handle this in AddDerivedTypeConstraintAnnotation?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -638,13 +652,64 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
...
+            {
+                model.AddDerivedTypeConstraintAnnotation(target,
+                    edmTypeMap.EdmTypes, navigationSourceConfiguration.DerivedTypeConstraints.ToList());
+            }
+        }
+
+        internal static void AddDerivedTypeConstraintAnnotation(this EdmModel model, IEdmVocabularyAnnotatable target, 
+            Dictionary<Type, IEdmType> edmTypeMap, List<Type> types)
+        {
+            if (types.Count == 0)
```
mikepizzo: >types.Count == 0 [](start = 16, length = 16)

Better to use types.Any()?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -321,7 +328,12 @@ private static void ValidateOperationEntitySetPath(IEdmModel model, IEdmOperatio
                     model.SetAnnotationValue(operation, new ReturnedEntitySetAnnotation(operationConfiguration.NavigationSource.Name));
                 }
 
-                AddOperationParameters(operation, operationConfiguration, edmTypeMap);
+                if (operationConfiguration.DerivedTypeConstraint != null && operationConfiguration.DerivedTypeConstraint.Any())
```
mikepizzo: >&& operationConfiguration.DerivedTypeConstraint.Any() [](start = 73, length = 53)

this is already handled in AddDerivedTypeConstraintAnnotation; don't need to check twice.

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
```diff
@@ -638,13 +652,64 @@ private static void AddCoreVocabularyAnnotations(this EdmModel model, IEnumerabl
...
+                }
+                else
+                {
+                    throw Error.InvalidOperation(SRResources.CantFindEdmType, type);
+                }
+            }
+
+            var collectionExpression = new EdmCollectionExpression(collectionConstants);
+            EdmVocabularyAnnotation annotation = new EdmVocabularyAnnotation(target, term, collectionExpression);
+            model.SetVocabularyAnnotation(annotation);
```
mikepizzo: >SetVocabularyAnnotation [](start = 18, length = 23)

I would generally think (at least by default) we should add the derived type constraint to the model element itself, not the model.

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
```diff
@@ -240,6 +246,20 @@ internal void SetBindingParameterImplementation(string name, IEdmTypeConfigurati
             _bindingParameter = new BindingParameterConfiguration(name, bindingParameterType);
         }
 
+        internal void AddDerivedTypeConstraintToReturnTypeImpl(params Type[] constraints)
+        {
+            foreach (Type t in constraints)
+            {
+                Type returnTypeConstraint = t;
+                if (returnTypeConstraint == null && ReturnType != null)
```
mikepizzo: >&& ReturnType != null [](start = 49, length = 21)

what if ReturnType == null?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.AspNet.OData.Common;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNet.OData.Builder
+{
+    internal static class DerivedTypeConstraintHelper
+    {
+        internal static void ValidateAndAddConstraint(this ISet<Type> Constraints, Type ClrType, Type typeToAdd)
```
mikepizzo: >ValidateAndAddConstraint [](start = 29, length = 24)

Do you really want to add this to any ISet?  Wouldn't it be better to create a DerivedTypeConstraintSet<T> : HashSet<T> that implements this method, and then use that to hold derived type constraints?

mikepizzo: >Type ClrType [](start = 83, length = 12)

Isn't this always going to be the Type of the ISet<Type>?

src/Microsoft.AspNet.OData.Shared/Builder/ParameterConfiguration.cs
```diff
@@ -31,6 +32,8 @@ protected ParameterConfiguration(string name, IEdmTypeConfiguration parameterTyp
             Name = name;
             TypeConfiguration = parameterType;
 
+            DerivedTypeConstraints = new HashSet<Type>();
```
mikepizzo: >Type> [](start = 49, length = 5)

is this not paramterType.ClrType?

src/Microsoft.AspNet.OData.Shared/Builder/PropertyConfiguration.cs
```diff
@@ -157,6 +164,17 @@ public bool Unsortable
...
         public QueryConfiguration QueryConfiguration { get; set; }
 
+        /// <summary>
+        /// Adds subtypes to the set of derived type constraints.
+        /// </summary>
+        internal void AddDerivedTypeConstraintImpl(params Type[] subtypes)
+        {
+            foreach (Type t in subtypes)
+            {
+                DerivedTypeConstraints.ValidateAndAddConstraint(RelatedClrType, t);
```
mikepizzo: >ValidateAndAddConstraint [](start = 39, length = 24)

Maybe define a ValidateAndAddConstraints(Type RelatedClrType, params Type[] subtypes) method to DerivedTypeConstraints class?

src/Microsoft.AspNet.OData.Shared/Builder/ComplexPropertyConfiguration.cs
```diff
@@ -52,5 +52,20 @@ public ComplexPropertyConfiguration IsRequired()
...
+        /// Adds subtypes to the list of derived type constraints.
+        /// </summary>
+        /// <param name="subtypes">The subtypes for which the constraint needs to be added.</param>
+        /// <returns>Updated configuration object.</returns>
+        public ComplexPropertyConfiguration AddDerivedTypeConstraint(params Type[] subtypes)
+        {
+            foreach (Type subType in subtypes)
+            {
+                AddDerivedTypeConstraintImpl(subType);
+            }
```
mikepizzo: define an AddDerivedTypeConstraintsImpl(params Type[] subtypes)?

src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs
KanishManuja-MS: The model element is specified by the target in the annotation. As per my discussion with Sam, it seemed like the right thing to do. We can sync up if I misunderstood something.

mikepizzo: That's fine; I just want to make sure that this constraint, in particular, is written inline by default.

---
In reply to: [383581324](https://github.com/OData/WebApi/pull/1977#discussion_r383581324) [](ancestors = 383581324)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            if (ClrBaseType == null)
+            {
+                Error.InvalidOperation(SRResources.BaseTypeNotSpecified);
+            }
+
+            ValidateAndAddConstraints(ClrBaseType, types);
+        }
+
+        public void ValidateAndAddConstraints(Type baseType, IEnumerable<Type> types)
```
mikepizzo: >baseType [](start = 51, length = 8)

This dual-capability of having a ClrBaseType declared on the DerivedTypeConstraintSet *and* having overloads that allow passing the base type is a bit strange.  It took me a bit to figure out that what (I think) you're doing is accounting for the fact that the base type may not be known when the DerivedTypeConstraintSet is created in the element constructor.  But it's still strange, as the baseType passed in to ValidateAndAddConstraints could have nothing to do with the type of the element on which the DerivedTypeConstraint is defined.

What if, instead, the element lazily created the DerivedTypeConstraintSet the first time it was accessed, and returned an error if the type had not yet been set of the element.  Then you wouldn't ever need to pass in the baseType, or worry if the passed in baseType was at all related to the ClrBaseType (if specified).  You could then replace ValidateAndAddSingleConstraint with an override of the Add method that did the validation before calling base.Add. This would make it easier to expose as a public property on the element.

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,72 @@
...
+        public DerivedTypeConstraintSet()
+        {
+        }
+
+        public DerivedTypeConstraintSet(Type baseType)
+        {
+            ClrBaseType = baseType;
+        }
+
+        public Type ClrBaseType { get; set; }
```
mikepizzo: >ClrBaseType [](start = 20, length = 11)

is there any value to exposing this property? Hopefully it wouldn't ever be set/retrieved after the DerivedTypeConstraintSet was constructed (and certainly couldn't be changed once set?)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            ValidateAndAddConstraints(ClrBaseType, types);
+        }
+
+        public void ValidateAndAddConstraints(Type baseType, IEnumerable<Type> types)
+        {
+            if (!types.Any())
+            {
+                ValidateAndAddSingleConstraint(baseType, baseType);
```
mikepizzo: >ValidateAndAddSingleConstraint [](start = 16, length = 30)

What if other derived types have already been added to the constraint? What's the behavior of *also* adding the base type? What if other derived types are added *after* adding the base type?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                throw Error.InvalidOperation(SRResources.NotADerivedType, typeToAdd.Name, baseType.Name);
+            }
+
+            if (this.Contains(typeToAdd))
+            {
+                throw Error.InvalidOperation(SRResources.ConstraintAlreadyExists, typeToAdd.Name);
+            }
+
+            this.Add(typeToAdd);
```
mikepizzo: Rather than first checking to see if the hashset contains the item, and then adding it (which will cause the hashset to again check to see if it already contains the item, and return false if it's already there), can you just replace 64-67 with:

            if (!this.Add(typeToAdd))
            {
                throw Error.InvalidOperation(SRResources.ConstraintAlreadyExists, typeToAdd.Name);
            }

mikepizzo:
See my comments on DerivedTypeConstraintSet -- I like it better than what we had before, but I don't like sometimes having a ClrBaseType that is sometimes ignored, and think we can simplify this area.  Otherwise looks good!

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
```diff
@@ -240,6 +246,16 @@ internal void SetBindingParameterImplementation(string name, IEdmTypeConfigurati
             _bindingParameter = new BindingParameterConfiguration(name, bindingParameterType);
         }
 
+        internal void AddDerivedTypeConstraintToReturnTypeImpl(params Type[] constraints)
+        {
+            if (ReturnType == null)
+            {
+                Error.InvalidOperation(SRResources.ReturnTypeOfOperationNotSpecified);
```
xuzhg: >Error.InvalidOperation(SRResources.ReturnTypeOfOperationNotSpecified); [](start = 16, length = 70)

throw?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
KanishManuja-MS: I tried to address many things. I still have the validation in place though because I felt that invalidating the model later will make it harder for developers to find the issue. We can discuss if the current design meets the need. 

---
In reply to: [384819539](https://github.com/OData/WebApi/pull/1977#discussion_r384819539) [](ancestors = 384819539)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
KanishManuja-MS: I have to expose it mainly because of the operation configuration. We allow setting the return type after constructing the operation configuration object. 

---
In reply to: [384820274](https://github.com/OData/WebApi/pull/1977#discussion_r384820274) [](ancestors = 384820274)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information
+
```
xuzhg: remove duplicated

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
xuzhg: +1, If a property value constructs from Ctor, don't provide the public set for that property.
Maybe consider to change your operation configuration.

---
In reply to: [384820274](https://github.com/OData/WebApi/pull/1977#discussion_r384820274) [](ancestors = 384820274)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a DerivedTypeContraintSet instance with a base type specified.
+        /// </summary>
+        /// <param name="baseType">CLR type of the base class.</param>
+        public DerivedTypeConstraintSet(Type baseType)
+            : base()
+        {
+            ClrBaseType = baseType;
```
xuzhg: >ClrBaseType = baseType; [](start = 12, length = 23)

no config for the Location?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    /// <summary>
+    /// Set of derived type constraints. 
+    /// </summary>
+    public class DerivedTypeConstraintSet : HashSet<Type>
+    {
+        /// <summary>
+        /// Initializes a DerivedTypeConstraintSet instance without setting a base type.
+        /// </summary>
+        public DerivedTypeConstraintSet()
```
xuzhg: >DerivedTypeConstraintSet [](start = 15, length = 24)

maybe call other ctor with arguments

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
xuzhg: >ClrBaseType = baseType; [](start = 12, length = 23)

check the baseType nullable?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Add the derived type constraints to the set. 
+        /// </summary>
+        /// <param name="derivedTypes">Derived types to be added to the set.</param>
+        public void AddConstraints(IEnumerable<Type> derivedTypes)
+        {
+            if (!derivedTypes.Any())
+            {
+                ValidateAndAddSingleConstraint(ClrBaseType);
+                return;
+            }
```
xuzhg: if the input is empty, add the ClrBaseType ? Why don't we have this strange logic?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+                ValidateAndAddSingleConstraint(type);
+            }
+        }
+
+        private void ValidateAndAddSingleConstraint(Type typeToAdd)
+        {
+            if (ClrBaseType == null)
+            {
+                throw Error.InvalidOperation(SRResources.NoClrTypeSpecified);
+            }
```
xuzhg: A DerivedTypeConstraintSet should have a ClrBaseType configuired already, why should we check this?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                throw Error.InvalidOperation(SRResources.NotADerivedType, typeToAdd.Name, ClrBaseType.Name);
+            }
+
+            if (!this.Add(typeToAdd))
+            {
+                throw Error.InvalidOperation(SRResources.ConstraintAlreadyExists, typeToAdd.Name);
+            }
+        }
+    }
```
xuzhg: I'd like to copy/paste my assumption:
I don't want to use my name, but, for the "Add" logic, we can make it simple as blow

    /// <summary>
    /// 
    /// </summary>
    public class DerivedTypeConstraintConfiguration
    {
        private ISet<Type> _derivedTypes;

        /// <summary>
        /// 
        /// </summary>
        /// <param name="baseType"></param>
        public DerivedTypeConstraintConfiguration(Type baseType)
        {
            ClrBaseType = baseType;
            _derivedTypes = new HashSet<Type>();
            Location = EdmVocabularyAnnotationSerializationLocation.OutOfLine;
        }

        /// <summary>
        /// 
        /// </summary>
        public ISet<Type> DerivedTypes { get { return _derivedTypes; } }

        /// <summary>
        /// 
        /// </summary>
        public Type ClrBaseType { get; private set; }

        /// <summary>
        /// 
        /// </summary>
        /// <typeparam name="TDerived"></typeparam>
        public void Add<TDerived>()
        {
            Add(typeof(TDerived));
        }

        /// <summary>
        /// 
        /// </summary>
        public EdmVocabularyAnnotationSerializationLocation Location { get; set; }

        /// <summary>
        /// 
        /// </summary>
        /// <param name="derivedType"></param>
        public void Add(Type derivedType)
        {
            if (derivedType == ClrBaseType)
            {
                return;
            }

            if (_derivedTypes.Contains(derivedType))
            {
                return;
            }

            if (!ClrBaseType.IsAssignableFrom(derivedType))
            {
                throw new Exception("blabla");
            }

            _derivedTypes.Add(derivedType);
        }
    }

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Linq;
+using Microsoft.AspNet.OData.Common;
+using Microsoft.OData.Edm.Csdl;
+
+namespace Microsoft.AspNet.OData.Builder
+{
+    /// <summary>
+    /// Set of derived type constraints. 
+    /// </summary>
+    public class DerivedTypeConstraintSet : HashSet<Type>
```
xuzhg: >: HashSet<Type> [](start = 42, length = 15)

I am not fan for either "derived" or "Composition" 

However, i'd like to mention:
1) Derived "HasSet<Type>" , we can't change later for this type owing breaking changes.
2) Composition and make it private means we can change it any time without any breaking changes.

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
```diff
@@ -91,7 +93,18 @@ public string Namespace
...
-        public IEdmTypeConfiguration ReturnType { get; set; }
+        public IEdmTypeConfiguration ReturnType
+        {
+            get { return _returnType; }
+            set
+            {
+                _returnType = value;
+                if (value != null)
+                {
+                    DerivedTypeConstraints.ClrBaseType = value.ClrType;
```
xuzhg: >DerivedTypeConstraints.ClrBaseType = value.ClrType; [](start = 20, length = 51)

That's weird and change/add the configuration even no derived type configuration happens in customer scenarios

src/Microsoft.AspNet.OData.Shared/Builder/ActionConfiguration.cs
```diff
@@ -233,6 +233,37 @@ public ActionConfiguration SetBindingParameter(string name, IEdmTypeConfiguratio
             return this;
         }
 
+        /// <summary>
+        /// Adds subtypes to the list of derived type constraints for the return type.
+        /// </summary>
+        /// <param name="subtypes">The subtypes for which the constraint needs to be added.</param>
+        /// <returns>Updated configuration object.</returns>
+        public ActionConfiguration HasDerivedTypeConstraints(params Type[] subtypes)
```
xuzhg: >HasDerivedTypeConstraints [](start = 35, length = 25)

I can't understand which derived constraint happens if customer call this method.

It's for the Binding type constraint, or return type constraint.

Be caution, return type and binding type could be the same type.

KanishManuja-MS: It is for the return type. Binding type is more of a parameter configuration than function configuration.

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.AspNet.OData.Common;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.OData.Builder
+{
+    /// <summary>
+    /// Represents an <see cref="Attribute"/> that can be placed on a property or placed on a class to specify the derived type constraints. 
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property)]
```
xuzhg: >[AttributeUsage(AttributeTargets.Property)] [](start = 4, length = 43)

It's only for the property, we can't put it on the class? 

My assumption is like this:

```
[DerivedTypeConstraint(typeof(Human)]
[DerivedTypeConstraint(typeof(Dog)]
....
public class Creature
{
    ....
}
```

It support to multiple config, each for one. (I am also ok to use param[])

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+    public sealed class DerivedTypeConstraintAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DerivedTypeConstraintAttribute"/> class.
+        /// </summary>
+        public DerivedTypeConstraintAttribute()
+        {
+            DerivedTypeConstraints = new HashSet<Type>();
+        }
```
xuzhg: why do we have a empty here?

src/Microsoft.AspNet.OData.Shared/Builder/ActionConfiguration.cs
xuzhg: See my changes: [https://github.com/OData/WebApi/blob/9c9257fd05783ae5b8ed427b9eef1848f0fa48c5/src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs](https://github.com/OData/WebApi/blob/9c9257fd05783ae5b8ed427b9eef1848f0fa48c5/src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs)

---
In reply to: [394514464](https://github.com/OData/WebApi/pull/1977#discussion_r394514464) [](ancestors = 394514464)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
KanishManuja-MS: An operation can be created without a ClrBaseType.

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
KanishManuja-MS: Base type configures the location.

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+
+        private void ValidateAndAddSingleConstraint(Type typeToAdd)
+        {
+            if (ClrBaseType == null)
+            {
+                throw Error.InvalidOperation(SRResources.NoClrTypeSpecified);
+            }
+            else if (!ClrBaseType.IsAssignableFrom(typeToAdd) && typeToAdd != ClrBaseType)
```
mikepizzo: >&& typeToAdd != ClrBaseType [](start = 62, length = 27)

I don't think you need this second check; if typeToAdd == ClrBaseType, then ClrBaseType.IssAssignableFrom(typeToAdd) would be true.

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
KanishManuja-MS: If you start putting it on Classes, user will expect to put constraints on entity types which we don't want to do. And Params seems cleaner. 

---
In reply to: [394517334](https://github.com/OData/WebApi/pull/1977#discussion_r394517334) [](ancestors = 394517334)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
KanishManuja-MS: The main difference that I see is that you don't support adding enumerables. I think it makes the implementation a lot cleaner if you can add enumerables to the set. 

---
In reply to: [394509903](https://github.com/OData/WebApi/pull/1977#discussion_r394509903) [](ancestors = 394509903)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintSet.cs
KanishManuja-MS: I had two options; One was to throw an error and the other was to consider that as an indication to allow only base types. Consider that this is called from HasDerivedTypeConstraints(). What should be the behavior?

---
In reply to: [394506704](https://github.com/OData/WebApi/pull/1977#discussion_r394506704) [](ancestors = 394506704)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
KanishManuja-MS: Required by the compiler. Need to have at least one constructor with CLS compliant parameters for these classes. 

---
In reply to: [394517978](https://github.com/OData/WebApi/pull/1977#discussion_r394517978) [](ancestors = 394517978)

src/Microsoft.AspNet.OData.Shared/Builder/OperationConfiguration.cs
KanishManuja-MS: The option was to lazy initialize and save extra allocations or initialize with the constructor. Model builder not being the most performance critical piece, I decided to initialize with the constructor. Since the return type does not have to be specified in the operation configuration initialization,I need to update it for the derivedTypeConstraint whenever it gets set. 

---
In reply to: [394512858](https://github.com/OData/WebApi/pull/1977#discussion_r394512858) [](ancestors = 394512858)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintConfiguration.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    /// <summary>
+    /// Configuration for derived type constraints. 
+    /// </summary>
+    public class DerivedTypeConstraintConfiguration
+    {
+        /// <summary>
+        /// Initializes a DerivedTypeConstraintConfiguration instance without setting a base type.
+        /// </summary>
+        internal DerivedTypeConstraintConfiguration()
```
xuzhg, line 20: >internal [](start = 8, length = 8)

why it's internal?

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintAttribute.cs
xuzhg: SO, we don't have the C# attribute for entity set? 
can we do like this:

And a string argument into the Constructor of "DerivedTypeConstraintAttribute", use it to set the entity set or singleton or navigation property built from this type?

[DerivedTypeConstraint(typeof(Human), "Creatures"]
[DerivedTypeConstraint(typeof(Dog)]
....
public class Creature
{
    ....
}

---
In reply to: [399119699](https://github.com/OData/WebApi/pull/1977#discussion_r399119699) [](ancestors = 399119699,394517334)

src/Microsoft.AspNet.OData.Shared/Builder/FunctionConfiguration.cs
```diff
@@ -202,6 +202,47 @@ public FunctionConfiguration SetBindingParameter(string name, IEdmTypeConfigurat
...
+            {
+                throw Error.InvalidOperation(SRResources.ReturnTypeOfOperationNotSpecified);
+            }
+
+            if (ReturnTypeConstraints == null)
+            {
+                ReturnTypeConstraints = new DerivedTypeConstraintConfiguration(ReturnType.ClrType);
+            }
+
+            ReturnTypeConstraints.AddConstraint<TDerivedType>();
```
xuzhg, line 242: why don't you move all these codes into base class OperationConfiguration?

src/Microsoft.AspNet.OData.Shared/Builder/ActionConfiguration.cs
```diff
@@ -233,6 +233,47 @@ public ActionConfiguration SetBindingParameter(string name, IEdmTypeConfiguratio
...
+            {
+                throw Error.InvalidOperation(SRResources.ReturnTypeOfOperationNotSpecified);
+            }
+
+            if (ReturnTypeConstraints == null)
+            {
+                ReturnTypeConstraints = new DerivedTypeConstraintConfiguration(ReturnType.ClrType);
+            }
+
+            ReturnTypeConstraints.AddConstraints(subtypes);
```
xuzhg, line 253: why don't you move all these codes into base class OperationConfiguration?

KanishManuja-MS: To avoid the cast while using the fluent APIs. 

---
In reply to: [401813541](https://github.com/OData/WebApi/pull/1977#discussion_r401813541) [](ancestors = 401813541)

src/Microsoft.AspNet.OData.Shared/Builder/FunctionConfiguration.cs
KanishManuja-MS: To avoid the cast while using the fluent APIs. I can have an internal implementation on operation configuration but it is only 3 lines of code. 

---
In reply to: [401813407](https://github.com/OData/WebApi/pull/1977#discussion_r401813407) [](ancestors = 401813407)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintConfiguration.cs
KanishManuja-MS: We don't want people to create a DerivedTypeConstraintConfiguration without a base type. 

---
In reply to: [401810329](https://github.com/OData/WebApi/pull/1977#discussion_r401810329) [](ancestors = 401810329)

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintConfiguration.cs
```diff
@@ -0,0 +1,104 @@
...
+            : this(null, EdmVocabularyAnnotationSerializationLocation.Inline)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a DerivedTypeConstraintConfiguration instance without setting a base type.
```
mikepizzo, line 26: >without setting a base type [](start = 70, length = 27)

copy/paste - this overload doe sspecify a base type

mikepizzo:
I think we'll ultimately want to put the validation as a rule in EdmModel.Validate -- there may be scenarios where we (or the user) want to turn off this validation. But okay with it here for now, since it's internal, unless we have to make additional changes for other reasons.

src/Microsoft.AspNet.OData.Shared/Builder/DerivedTypeConstraintConfiguration.cs
```diff
@@ -0,0 +1,93 @@
...
+            foreach (Type type in derivedTypes)
+            {
+                if (!this.ConstraintSet.Add(type))
+                {
+                    throw Error.InvalidOperation(SRResources.ConstraintAlreadyExists, type.Name);
+                }
+            }
+        }
+
+        internal ISet<Type> ConstraintSet { get; }
```
xuzhg, line 75: >internal ISet<Type> ConstraintSet { get; } [](start = 8, length = 42)

cannot be private?

Issue #1979, opened 2019-12-10
Not able to expand a null navigation property that is based on a foreign key with shadow property


obohaciak:
I have an ASP.NET Core app that uses OData to query DB using entities mapped with EntityFramework Core. The DB contains 2 tables joined with FK, both having composite PKs. Both tables are exposed as entities where one of the columns in PK is a shadow property. When expanding a navigation property based on this FK, OData fails to serialize the result when navigation property is null.

### Assemblies affected

Microsoft.AspNetCore.OData (7.2.2), targetting ASP.NET Core 2.2 (contains Microsoft.EntityFrameworkCore 2.2.6)

### Reproduce steps
These are the tables
```
CREATE TABLE [dbo].[Person](
	[ShadowId] [int] NOT NULL DEFAULT 1,
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[Name] [nvarchar](100) NULL,
    CONSTRAINT [PK_Owner] PRIMARY KEY CLUSTERED ([ShadowId] ASC, [Id] ASC)
)
GO

CREATE TABLE [dbo].[Pet](
	[ShadowId] [int] NOT NULL DEFAULT 1,
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[Name] [nvarchar](100) NULL,
	[OwnerId] [int] NULL,
    CONSTRAINT [PK_Pets] PRIMARY KEY CLUSTERED ([ShadowId] ASC, [Id] ASC)
)
GO

ALTER TABLE [dbo].[Pet]  WITH CHECK ADD  CONSTRAINT [FK_Pet_Owner] FOREIGN KEY([ShadowId], [OwnerId]) REFERENCES [dbo].[Person] ([ShadowId], [Id])
GO
```
Let's add 2 rows to show cases when the principal entity is null and non-null
```
SET IDENTITY_INSERT [dbo].[Person] ON 
INSERT [dbo].[Person] ([Id], [Name]) VALUES (4, N'Annie')
SET IDENTITY_INSERT [dbo].[Person] OFF
GO

SET IDENTITY_INSERT [dbo].[Pet] ON 
INSERT [dbo].[Pet] ([Id], [Name], [OwnerId]) VALUES (4, N'Mittens', 4)
INSERT [dbo].[Pet] ([Id], [Name], [OwnerId]) VALUES (5, N'Patches', NULL)
SET IDENTITY_INSERT [dbo].[Pet] OFF
GO 
```

This is my DbContext's model
```
[...]

PR #1983, opened 2019-12-12
Fix incorrect evaluation of IsNull expression

obohaciak:
### Issues

This pull request fixes issue #1979

### Description

The problem: when multiple keys with shadow property is used, NullCheck expression is incorrectly evaluated. This causes null value to be incorrectly recognized as actual object and referencing parameters of this object fails.

The solution: we have introduced new property RawValue which holds actual object referenced by navigation property we use it to double check if the property is null.

- [x] Test cases added
- [x] Build and test with one-click build and test script passed

### Additional work necessary
None

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/NamedPropertyExpression.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/NamedPropertyExpression.cs
@@ -34,6 +34,8 @@ namespace Microsoft.AspNet.OData.Query.Expressions
         // and use Value only if IsNull is false.
         public Expression NullCheck { get; set; }
 
+        public Expression RawValue { get; set; }
+
         public int? PageSize { get; set; }
 
         public bool AutoSelected { get; set; }
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
@@ -137,6 +137,11 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                 memberBindings.Add(Expression.Bind(namedPropertyType.GetProperty("IsNull"), property.NullCheck));
             }
 
+            if (property.RawValue != null)
+            {
+                memberBindings.Add(Expression.Bind(namedPropertyType.GetProperty("RawValue"), property.RawValue));
+            }
+
             return Expression.MemberInit(Expression.New(namedPropertyType), memberBindings);
         }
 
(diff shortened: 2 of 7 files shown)


odero:
@obohaciak Would be great if you could resolve the conflicts

obohaciak:
@odero I'm on it.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
```diff
@@ -212,7 +219,7 @@ private class SingleExpandedProperty<T> : NamedProperty<T>
 
             public override object GetValue()
             {
-                return IsNull ? (object)null : Value;
+                return IsNull ? (object)null : (RawValue == null ? (object)null : Value);
```
odero: Easier to read as 
```suggestion
                return IsNull || RawValue == null ? (object)null : Value);
```

src/Microsoft.AspNet.OData.Shared/Query/Expressions/NamedPropertyExpression.cs
```diff
@@ -34,6 +34,8 @@ public NamedPropertyExpression(Expression name, Expression value)
         // and use Value only if IsNull is false.
         public Expression NullCheck { get; set; }
 
+        public Expression RawValue { get; set; }
```
xuzhg: >RawValue [](start = 26, length = 8)

what's the difference between "Value" and "RawValue"?

xsmerek: Value might contain not the actual instance but rather wrapper created in SelectExpandBinder method BuildExpandedProperty (line 712).

src/Microsoft.AspNet.OData.Shared/Query/Expressions/PropertyContainer.cs
xsmerek: Thank you, sounds good. I'll make the change when I'm finished with the failing tests.

odero:
@xuzhg Any reason not to approve this and have it merged?

odero:
@xsmerek Could you force push this again for a rebuild? Can't figure out what the failing test was.

xsmerek:
I'm still trying to debug the tests to see why projectionLambda in SelectExpandBinder.Bind for some tests throws `System.Security.VerificationException: 'Operation could destabilize the runtime.'`. Interesting is that Microsoft.Test.E2E.AspNetCore3x.OData version of the tests does not fail. This makes me think if the original error has something to do with Core version. I'll try to reproduce the bug with Core 3 version.

xsmerek:
I've tried to reproduce the issue on version **7.3.0** with **.NET Core 3.1** and the problem is not reproducible. The call returns expected result.

@odero Could you please close the pull request as the issue is resolved for us and we do not want to add unnecessary code to the library? Thank you very much.

odero:
Awesome. I just saw the comments in the PR

PR #1985, opened 2019-12-13
Fix the byte[] and char[] property in select

xuzhg:
### Issues

*This pull request fixes issue #1985.*

### Description

*For property defined by byte[] and char[] is mapping to "Edm.Binary" and "Edm.String".
So, for these properties, we don't need to further process the nested query options.

- [x] *Test cases added*

### Additional work necessary


xuzhg:
Merged. Close it.

Issue #1973, opened 2019-12-05
EntitySet and Entity GET requests are routed to a single Get() method regardless of parameters


habbes:
When you have a controller with only one `Get()` method when using convention-based routing, both requests from `GET /odata/EntitySet` and `GET /odata/EntitySet(id)` will be routed to that `Get` method regardless of whether it is defined with a key parameter (i.e. `Get(int key)`) or without one (i.e. `Get()`)

### Assemblies affected

- Microsoft.AspNetCore.OData 7.2.2

### Reproduce steps

- Create an `Entity` class with a key property
- Create an `EdmModel` with an entity set called `Entities` using the `ODataConventionModelBuilder`
- Create an `EntitiesController`
- Define a `Get()` method in the controller, and add a breakpoint in that method
- Try making the following requests to your service: `GET /odata/Entities` and `GET /odata/Entities(10)`
- Replace the `Get()` method with a `Get(int key)` method and try to make the requests again

### Expected result

The following table summarizes the expected result

| Controller action defined | `GET /odata/Entities` is routed to | `GET /odata/Entities(10)` is routed to
-----------------------------|--------------------------------------|----------------------------------------
only `Get()`       | `Get()`  |  returns 404 error |
only `Get(int key)` | returns 404 error | `Get(int key)` with `key` set to 10

### Actual result

The following table summarizes the actual result:

| Controller action defined | `GET /odata/Entities` is routed to | `GET /odata/Entities(10)` is routed to
-----------------------------|--------------------------------------|----------------------------------------
only `Get()`       | `Get()`  | `Get()` |
only `Get(int key)` | `Get(int key)` with `key` set to 0 | `Get(int key)` with `key` set to 10

### Additional detail

This issue does not occur when you have only `GetEntity(int key)` or only `GetEntities()` defined, i.e. the service returns 404 as expected when the appropriate action is not defined.

I suspect the issue comes from the following statement on line 120 of `Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs` in the method `SelectBestCandidates`:
```cs
return (matchedCandidates.Any())
  ? candidates.Where(c => c.Id == matchedCandidates.FirstOrDefault().Id).FirstOrDefault()
  : candidates.FirstOrDefault();
```
When the action selector does not find an exact candidate action, it just picks the first candidate available, which would have the correct method name but might not match the parameters in the URL.

PR #1988, opened 2019-12-17
Fix false positives in action selector

habbes:
### Issues

*This pull request fixes issue #1973*

### Description

I updated the `ODataActionSelector` in both AspNetCore (2.x and 3.0) and AspNet classic and made them more strict. It rejects candidate actions which have parameters which cannot be matched to the route parameters, body, query options, optional parameters and `ODataPath`.

- [x] *Test cases added*

(PS: the build command seems to fail due to a recent library update that added support for .net 3.0, not due to my PR).

### Additional work necessary

If this PR is accepted, it would be a good idea to update the docs mentioning that a 404 would be returned if the parameters of the action don't match those of the request, even if the name matches.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -13,7 +13,6 @@ using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
-using Microsoft.AspNetCore.Mvc.ModelBinding;
 
 namespace Microsoft.AspNet.OData.Routing
 {
@@ -82,15 +81,8 @@ namespace Microsoft.AspNet.OData.Routing
             RouteData routeData = context.RouteData;
             ODataPath odataPath = context.HttpContext.ODataFeature().Path;
 
-            var routers = routeData.Routers;
-            var odataRoute = routeData.Routers.FirstOrDefault((router) =>
-            {
-                var _odataRoute = router as ODataRoute;
-                return _odataRoute == null ?
-                    false : true;
-            }) as ODataRoute;
-
-            var routePrefix = odataRoute == null ? null : odataRoute.RoutePrefix;
+            var odataRoute = routeData.Routers.OfType<ODataRoute>().FirstOrDefault();
+            var routePrefix = odataRoute?.RoutePrefix;
             
             if (odataPath != null && routeData.Values.ContainsKey(ODataRouteConstants.Action))
             {
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -96,15 +96,18 @@ namespace Microsoft.AspNet.OData.Routing
         {
             RouteData routeData = context.RouteData;
             ODataPath odataPath = context.HttpContext.ODataFeature().Path;
-
-            var odataRoute = routeData.Routers.OfType<ODataRoute>().FirstOrDefault();
-            var routePrefix = odataRoute?.RoutePrefix;
             
             if (odataPath != null && routeData.Values.ContainsKey(ODataRouteConstants.Action))
             {
+                var odataRoute = routeData.Routers.OfType<ODataRoute>().FirstOrDefault();
+                var routePrefix = odataRoute?.RoutePrefix;
                 // Get the available parameter names from the route data. Ignore case of key names.
-                IList<string> availableKeys = routeData.Values.Keys.Select(k => k.ToLowerInvariant())
-                    .Where((key) => routePrefix != "{" + key + "}")
+                // Remove route prefix and other non-parameter values from availableKeys
+                IList<string> availableKeys = routeData.Values.Keys
+                    .Where((key) => routePrefix != "{" + key + "}"
+                        && key != ODataRouteConstants.Action
+                        && key != ODataRouteConstants.ODataPath)
+                    .Select(k => k.ToLowerInvariant())
                     .ToList();
 
                 // Filter out types we know how to bind out of the parameter lists. These values
@@ -122,7 +125,8 @@ namespace Microsoft.AspNet.OData.Routing
                 ODataOptionalParameter optionalWrapper = wrapper as ODataOptionalParameter;
 
                 // Find the action with the all matched parameters from available keys including
-                // matches with no parameters. Ordered first by the total number of matched
+                // matches with no parameters and matches with parameters bound to the body.
+                // Ordered first by the total number of matched
                 // parameters followed by the total number of parameters.  Ignore case of
                 // parameter names. The first one is the best match.
                 //
@@ -135,10 +139,9 @@ namespace Microsoft.AspNet.OData.Routing
                 var matchedCandidates = considerCandidates
                     .Where(c => TryMatch(context, c.FilteredParameters, availableKeys, optionalWrapper, c.TotalParameterCount))
                     .OrderByDescending(c => c.FilteredParameters.Count)
-                    .ThenByDescending(c => c.TotalParameterCount);
+                    .ThenByDescending(c => c.TotalParameterCount)
+                    .ToList();
 
-                // Return either the best matched candidate or the first
-                // candidate if none matched.
                 return (matchedCandidates.Any())
                     ? candidates.Where(c => c.Id == matchedCandidates.FirstOrDefault().Id).FirstOrDefault()
                     : null;
@@ -149,16 +152,22 @@ namespace Microsoft.AspNet.OData.Routing
 
         private bool TryMatch(RouteContext context, IList<ParameterDescriptor> parameters, IList<string> availableKeys, ODataOptionalParameter optionalWrapper, int totalParameterCount)
         {
-            if (totalParameterCount == 0 && availableKeys.Count > 2)
+            if (totalParameterCount == 0 && availableKeys.Count > 0)
             {
                 return false;
             }
             bool matchedBody = false;
+            var conventionsStore = context.HttpContext.ODataFeature().RoutingConventionsStore;
             // use the parameter name to match.
             foreach(var p in parameters)
             {
                 string parameterName = p.Name.ToLowerInvariant();
                 if (availableKeys.Contains(parameterName))
+                {
+                    continue;
+                }
+
+                if (conventionsStore != null && conventionsStore.ContainsKey(p.Name))
                 {
                     continue;
                 }
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -175,7 +175,7 @@ namespace Microsoft.AspNet.OData.Routing
                 // if parameter is not bound to a key in the path,
                 // assume that it's bound to the request body
                 // only one parameter should be considered bound to the body
-                if (!matchedBody && context.HttpContext.Request.Body.Length > 0)
+                if (!matchedBody && RequestHasBody(context))
                 {
                     matchedBody = true;
                     continue;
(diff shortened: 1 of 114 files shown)


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingTest.cs
```diff
@@ -893,4 +911,40 @@ public string GetName(int keyID)
             return String.Format(CultureInfo.InvariantCulture, "GetName({0}) with keyID", keyID);
         }
     }
+
+    public class NotFoundWithIdCustomersController: TestODataController
+    {
+        public string Get(int key)
+        {
+            return String.Format(CultureInfo.InvariantCulture, "Get({0})", key);
```
odero:
```suggestion
            return $"Get({key})");
```

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -80,10 +81,23 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
             RouteData routeData = context.RouteData;
             ODataPath odataPath = context.HttpContext.ODataFeature().Path;
+
+            var routers = routeData.Routers;
+            var odataRoute = routeData.Routers.FirstOrDefault((router) =>
+            {
+                var _odataRoute = router as ODataRoute;
+                return _odataRoute == null ?
+                    false : true;
+            }) as ODataRoute;
```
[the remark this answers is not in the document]
habbes: `routeData.Routers` happens to have objects of different types that implement `IRouter`. Only one of them was an `ODataRoute` type, and that's the one that has the `RoutePrefix` property.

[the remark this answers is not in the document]
odero: Untested but maybe something along these lines then?
```c#
ODataRoute odataRoute = routeData.Routers.Where(router => router is ODataRoute).FirstOrDefault();
```

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -80,10 +81,23 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
         {
             RouteData routeData = context.RouteData;
             ODataPath odataPath = context.HttpContext.ODataFeature().Path;
+
+            var routers = routeData.Routers;
+            var odataRoute = routeData.Routers.FirstOrDefault((router) =>
```
xuzhg: >routeData.Routers [](start = 29, length = 17)

it's basic scenario to register multiple ODataRoute in the application. It's supported in Web API OData.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
xuzhg: You can accomplish it:
1) routeData.Routers.OfType<ODataRoute>().FirstOrDefault()
2) routeData.Routers.FirstOrDefault(r => r is ODataRoute) as ODataRoute;

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ControllerDescriptorFactory.cs
```diff
@@ -8,6 +8,8 @@
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc.Abstractions;
```
xuzhg: sorting the using

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.OData;
+
+namespace Microsoft.AspNet.OData.Test.Abstraction
+{
+    class ODataActionSelectorTestHelper
```
xuzhg: >class [](start = 4, length = 5)

explicitly decorate it as internal

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -80,10 +81,23 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+            }) as ODataRoute;
+
+            var routePrefix = odataRoute == null ? null : odataRoute.RoutePrefix;
+            
             if (odataPath != null && routeData.Values.ContainsKey(ODataRouteConstants.Action))
             {
                 // Get the available parameter names from the route data. Ignore case of key names.
-                IList<string> availableKeys = routeData.Values.Keys.Select(k => k.ToLowerInvariant()).AsList();
+                IList<string> availableKeys = routeData.Values.Keys.Select(k => k.ToLowerInvariant())
+                    .Where((key) => routePrefix != "{" + key + "}")
```
xuzhg: >(key) => routePrefix != "{" + key + "}") [](start = 27, length = 40)

what's the meaning to compare the {key} with the RoutePrefix ?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: Thanks for the clarification.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: I have found that when the Web API OData is configured with a parameterized prefix, that prefix is included in the available keys. So in that statement I was removing the prefix from the available keys in order to easily detect when a candidate action has no parameters to match the actual keys in the path. But this does feel like a hack and there's probably a better or clearer way to achieve the goal. Or maybe I just misunderstand the use case of a parameterized route prefix. Why is the route prefix added to the route data, is it meant to be bound to action parameters?

habbes:
Hi @xuzhg @odero, here are some updates on the task:

I've updated my branch with the latest updates from master, including support for .net core 3.0. My ODataActionsSelector (AspNet Core) tests were failing to build because the ODataActionSelector constructor I was using is not supported in 3.0. I'll work on fixing these tests. Alternatively I could just get rid of these tests since I also added other more general tests in ODataRoutingTest (shared by both core and classic) that test against the expected routing behaviour.

When running the tests under `netcoreapp3` environment, some tests fail due to how I was testing whether the request contains a body. I was using `context.HttpContext.Request.Body.Length`. This worked in `netcoreapp2`. But under `netcoreapp3` tests, the request body is not seekable and therefore getting the length throws an exception.

To provide some context, the reason I was checking whether the request has a body, is to see whether an action parameter that is not bound to a key in the path is bound to the body. This helped match actions like `Post(Person body)` to routes like `POST /Persons` where the parameter does not match any segment in the route, but will receive its value from the request body. On the other hand,
if the action has parameters that are neither in the route segments or in the body, then a 404 should be returned. For example, if the only controller action is `Get(int key)` and the route is `GET /Persons` then a 404 is returned.

Using the request body length allowed me to achieve that behaviour. But probably that's a hack. Kindly advise if you have a better approach.

I'm still working on an implementaton for AspNet classic.

habbes:
I've made slight adjustments to my AspNetCore `ODataActionSelector` changes and now they build and pass tests in both net core 3.0 and 2.0. I used conditional compiler directives in some tests for this to work.

I've also ported the changes I've made over to the AspNet classic project. However, there's one pre-existing unit test that fails in AspNet classic as a result of my changes, I have not fixed this yet because I wanted to get feedback on my approach first, because this test seems to break the assumptions I had about how routing is supposed to behave.

What I did was to take the existing `ODataActionSelector.SelectBestCandidate` and `TryMatch` methods and made them stricter in the following ways:
All the parameters of the selected action have to either:
- match a key in the route data values
- be an optional parameter
- match query options
- be of type `ODataPath`
- be bound to the request body

If the method has extra parameters that do not match any of the above conditions, then that method is rejected (i.e. `Get(key)` will not get selected for`/odata/Entities`). If the method has no parameters but the route data has values (e.g. `key` or `relatedkey`), then the method is rejected (this means that `Get()` won't get selected when the route is `/odata/Entities(key)`). With these adjustments, the
selector will return `null` (i.e. 404 error) when no action matches the conditions.

The `ODataActionSelector` assumes that a parameter will be bound to the body if the parameter does not match any of the other conditions and if the request has a request body (as determined by the `Content-Length` header). Since binding happens later, this was the best estimate I could come up with.

In cases where the odata route has a parameterized prefix, this route prefix is added to the route data values. Is this something I should account for? Is there a case where this would be bound to parameters? Currently all tests pass when I don't include it in the conditions I listed above.

These additional checks and conditions ensured that the I eliminated the behavior described in the issue, and also kept existing tests* passing.

The test that fails is enabled in AspNet classic but not in Core, namely `CanModelBindNonStringData()`  in `Microsoft.AspNet.OData.Test.Shared\Routing\ODataValueProviderFactoryTest.cs`.
It calls the route `http://localhost/Test`, manually adds some data to the request's `RoutingConventionStore` and expects it to call `Get(TestClass test)` with the data bound to the `test` parameter.

This does not work with my update because `Get(test)` has a parameter that's neither in the route nor in the body. My update assumes that what the test is trying to do is invalid. I could probably make the test pass simply by adding the parameter name to its route data. But I first would like to confirm that my assumption is correct and that in the real application This argument would either come from the route or the body.

cc @xuzhg @mikepizzo @odero

habbes:
I've fixed the failing test in AspNet classic by updating the `ODataActionSelector` to look for the parameter in the `RoutingConventionsStore` as well.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNet.OData.Query;
+using Microsoft.AspNet.OData.Test.Common;
+using System.Collections.Generic;
+using Xunit;
+using System;
```
xuzhg: sorting
System.* at head

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,213 @@
...
+namespace Microsoft.AspNet.OData.Test.Routing
+{
+    public class ODataActionSelectorTest
+    {
+
+        public static TheoryDataSet<Dictionary<string, object>, string, System.Type, string, System.Type[]> ScenariosWithCorrectAction
+        {
+            get
+            {
+                return new TheoryDataSet<Dictionary<string, object>, string, System.Type, string, System.Type[]>
```
xuzhg: >System. [](start = 98, length = 7)

can get rid of System.

xuzhg:
Now, you can run the build.cmd netfx|netcore on command line. please let me know any problems. 

---
In reply to: [568395516](https://github.com/OData/WebApi/pull/1988#issuecomment-568395516) [](ancestors = 568395516)

habbes:
@xuzhg I've updated my code to work with .Net Core 3.1 following the recent updates to the WebApi library. My branch now builds correctly when I run `build.cmd quick`. However, it fails on Azure pipelines because: `The current .NET SDK does not support targeting .NET Core 3.1`.

habbes:
This PR fails to build on CI because it's using an old build agent. To resolve this, I have open a separate PR checkout from this branch.

Let's the conversation continue over on this PR: https://github.com/OData/WebApi/pull/2007

odero:
@habbes does that mean this PR can be closed?

habbes:
@odero yeah, I guess it makes sense to close the PR.

PR #1997, opened 2020-01-02
Fix the int16 overflow during build

ap3rus:
### Description

This change fixes the overflow during generation of VersionDateCode

`F:\repos\OData-WebApi\src\System.Web.Http.OData\System.Web.Http.OData.csproj : error  : The expression "[System.Convert]::ToInt16(40102)" cannot be evaluated. Value was either too large or too small for an Int16.  F:\repos\OData-WebApi\tools\WebStack.versions.settings.targets`


tools/WebStack.versions.settings.targets
```diff
@@ -48,7 +48,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionStartYear Condition="'$(VersionStartYear)' == ''">2017</VersionStartYear>
+    <VersionStartYear Condition="'$(VersionStartYear)' == ''">2020</VersionStartYear>
```
xuzhg: >2020 [](start = 62, length = 4)

what's the value to change the base ?

PR #2007, opened 2020-01-08
Fix routing false positives in action selector

habbes:
**Note**: This is a "fork" of [this PR](https://github.com/OData/WebApi/pull/1988). I create a new PR because the previous had issues building on the CI environment.

This PR has been replaced by #2168 

### Issues

*This pull request fixes issue #1973*

### Description

I updated the `ODataActionSelector` in both AspNetCore (2.x and 3.1) and AspNet classic and made them more strict. It rejects candidate actions which have parameters which cannot be matched to the route parameters, body, query options, optional parameters and `ODataPath`.

- [x] *Test cases added*

### Additional work necessary

If this PR is accepted, it would be a good idea to update the docs mentioning that a 404 would be returned if the parameters of the action don't match those of the request, even if the name matches.

### DIFF
--- a/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
@@ -122,7 +122,7 @@ namespace Microsoft.AspNet.OData.Routing
             var route = routeData.Route as ODataRoute;
             var routePrefix = route?.RoutePrefix;
             var availableKeys = routeData.Values.Keys
-                .Where(k => routePrefix != "{" + k + "}" 
+                .Where(k => routePrefix != "{" + k + "}"
                     && k != ODataRouteConstants.Action
                     && k != ODataRouteConstants.Controller
                     && k != ODataRouteConstants.ODataPath)
@@ -145,6 +145,11 @@ namespace Microsoft.AspNet.OData.Routing
                 {
                     continue;
                 }
+                if (conventionsStore.Keys.FirstOrDefault(key =>
+                    key.Split(new char[] { '.' })[0] == p.ParameterName) != null)
+                {
+                    continue;
+                }
                 if (!matchedBody && RequestHasBody(context))
                 {
                     matchedBody = true;
@@ -162,6 +167,10 @@ namespace Microsoft.AspNet.OData.Routing
                 {
                     continue;
                 }
+                if (context.Request.RequestUri.Query.Contains(p.ParameterName))
+                {
+                    continue;
+                }
                 return false;
             }
             return true;
@@ -170,7 +179,7 @@ namespace Microsoft.AspNet.OData.Routing
         private static bool RequestHasBody(HttpControllerContext context)
         {
             var content = context.Request.Content;
-            return content?.Headers.ContentLength > 0;
+            return content?.Headers.ContentLength > 0 || content?.Headers.ContentType != null;
         }
 
         private static bool IsODataQueryOptions(Type parameterType)
--- a/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
@@ -11,6 +12,7 @@ using System.Web.Http.Controllers;
 using System.Web.Http.Routing;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Query;
 using Microsoft.AspNet.OData.Routing.Conventions;
 
 namespace Microsoft.AspNet.OData.Routing
@@ -87,7 +89,11 @@ namespace Microsoft.AspNet.OData.Routing
                 if (actionName != null)
                 {
                     routeData.Values[ODataRouteConstants.Action] = actionName;
-                    return _innerSelector.SelectAction(controllerContext);
+                    var action = _innerSelector.SelectAction(controllerContext);
+                    if (ActionParametersMatchRequest(action, controllerContext))
+                    {
+                        return action;
+                    }
                 }
             }
 
@@ -105,6 +111,69 @@ namespace Microsoft.AspNet.OData.Routing
             }
 
             return request.CreateErrorResponse(statusCode, error);
+        }
+
+        private static bool ActionParametersMatchRequest(HttpActionDescriptor action, HttpControllerContext context)
+        {
+            var parameters = action.GetParameters();
+            var routeData = context.RouteData;
+            var matchedBody = false;
+            var route = routeData.Route as ODataRoute;
+            var routePrefix = route?.RoutePrefix;
+            var availableKeys = routeData.Values.Keys
+                .Where(k => routePrefix != "{" + k + "}")
+                .Select(k => k.ToUpperInvariant())
+                .ToList();
+
+            if (parameters.Count == 0 && availableKeys.Count > 3)
+            {
+                return false;
+            }
+
+            foreach (var p in parameters)
+            {
+                string parameterName = p.ParameterName.ToUpperInvariant();
+                if (availableKeys.Contains(parameterName))
+                {
+                    continue;
+                }
+                if (!matchedBody && RequestHasBody(context))
+                {
+                    matchedBody = true;
+                    continue;
+                }
+                if (p.ParameterType == typeof(ODataPath))
+                {
+                    continue;
+                }
+                if (IsODataQueryOptions(p.ParameterType))
+                {
+                    continue;
+                }
+                if (p.IsOptional)
+                {
+                    continue;
+                }
+                return false;
+            }
+            return true;
+        }
+
+        private static bool RequestHasBody(HttpControllerContext context)
+        {
+            var content = context.Request.Content;
+            return content?.Headers.ContentLength > 0;
+        }
+
+        private static bool IsODataQueryOptions(Type parameterType)
+        {
+            if (parameterType == null)
+            {
+                return false;
+            }
+            return ((parameterType == typeof(ODataQueryOptions)) ||
+                    (parameterType.IsGenericType &&
+                     parameterType.GetGenericTypeDefinition() == typeof(ODataQueryOptions<>)));
         }
     }
 }
--- a/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
@@ -141,12 +141,11 @@ namespace Microsoft.AspNet.OData.Routing
                 {
                     continue;
                 }
-                if (conventionsStore.ContainsKey(p.ParameterName))
+                if (conventionsStore.Keys.Any(key => key.Contains(p.ParameterName)))
                 {
                     continue;
                 }
-                if (conventionsStore.Keys.FirstOrDefault(key =>
-                    key.Split(new char[] { '.' })[0] == p.ParameterName) != null)
+                if (context.Request.GetQueryNameValuePairs().Any(kvp => kvp.Key == p.ParameterName))
                 {
                     continue;
                 }
@@ -164,10 +163,6 @@ namespace Microsoft.AspNet.OData.Routing
                     continue;
                 }
                 if (p.IsOptional)
-                {
-                    continue;
-                }
-                if (context.Request.RequestUri.Query.Contains(p.ParameterName))
                 {
                     continue;
                 }
--- a/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
@@ -117,15 +117,19 @@ namespace Microsoft.AspNet.OData.Routing
         {
             var parameters = action.GetParameters();
             var routeData = context.RouteData;
+            var conventionsStore = context.Request.ODataProperties().RoutingConventionsStore;
             var matchedBody = false;
             var route = routeData.Route as ODataRoute;
             var routePrefix = route?.RoutePrefix;
             var availableKeys = routeData.Values.Keys
-                .Where(k => routePrefix != "{" + k + "}")
+                .Where(k => routePrefix != "{" + k + "}" 
+                    && k != ODataRouteConstants.Action
+                    && k != ODataRouteConstants.Controller
+                    && k != ODataRouteConstants.ODataPath)
                 .Select(k => k.ToUpperInvariant())
                 .ToList();
 
-            if (parameters.Count == 0 && availableKeys.Count > 3)
+            if (parameters.Count == 0 && availableKeys.Count > 0)
             {
                 return false;
             }
@@ -134,6 +138,10 @@ namespace Microsoft.AspNet.OData.Routing
             {
                 string parameterName = p.ParameterName.ToUpperInvariant();
                 if (availableKeys.Contains(parameterName))
+                {
+                    continue;
+                }
+                if (conventionsStore.ContainsKey(p.ParameterName))
                 {
                     continue;
                 }
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -148,9 +148,6 @@ namespace Microsoft.AspNet.OData.Batch
                 request.Body = bufferedStream;
             }
 
-            // remove the content length and content type headers in case the batchRequest item doesn't have a body
-            request.ContentLength = null;
-            request.ContentType = null;
             foreach (var header in batchRequest.Headers)
             {
                 string headerName = header.Key;
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -148,6 +148,9 @@ namespace Microsoft.AspNet.OData.Batch
                 request.Body = bufferedStream;
             }
 
+            // remove the content length and content type headers in case the batchRequest item doesn't have a body
+            request.ContentLength = null;
+            request.ContentType = null;
             foreach (var header in batchRequest.Headers)
             {
                 string headerName = header.Key;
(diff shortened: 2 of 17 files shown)


habbes:
I've fixed the failing E2E tests, and now finally all the tests are passing and the build is successful on ADO.

So this is ready for review again @xuzhg  @KanishManuja-MS @odero

habbes:
I made the following changes to my `ODataActionSelector` update to pass the tests:
- it now uses either the `Content-Length` or `Content-Type` header to determine whether the  request has body. This is used for matching controller actions that expect a parameter bound from the body, since these parameters are neither included in the request context `RouteData` or convention store.
- it relaxed the checks for the action parameters in the convention store and query string.

I also made the following changes in other parts of the codebase:
- the `SpecialCharactersLinkGenerationTests` and `SpecialCharactersLinkGenerationWebTests` tests had what I think is a bug. The tests make single-entity get requests that are routed to the `Get()` action that returns all entities. This action always returns a successful response, and the tests did not verify the response, so the bug probably went unnoticed. Because my PR  makes the `ODataActionSelector` stricter, they requests not routed to the `Get()` method, which led to me to discover this issue and correct the tests. In the process I found that one of the test cases actually fails, in particular if you make a request to fetch an entity by key, where the key is a single whitespace, the key will be set to `null` (this causes the test to fail). If this whitespace-only string should be supported as a key parameter, then I would propose creating a separate issue to handle that. For now I have omitted the whitespace from the test cases so that the tests can pass.

I should also mention that I have some concerns regarding the cost vs benefit of this PR: It has modified a lot of files and made the `ODataActionSelector` stricter in order to fix a bug that no seemingly no one had ever complained about, and it could potentially introduce a breaking change if people ever relied on that bug as a feature or if there it breaks valid scenarios that are not captured by the test suite. What do you think?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DateAndTimeOfDayTest.cs
```diff
@@ -177,6 +177,7 @@ public async Task PostEntity_WithDateAndTimeOfDayProperties()
             HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, Uri);
 
             request.Content = new StringContent(Payload);
+            request.Content.Headers.ContentLength = Payload.Length;
```
xuzhg: >ContentLength = Payload.Length; [](start = 36, length = 31)

Is Length setting required?  Why?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBuilder/SpecialCharactersLinkGenerationTests.cs
```diff
@@ -51,12 +51,12 @@ public IQueryable<SpecialCharactersLinkGenerationTestsModel> Get()
             return Todoes.AsQueryable();
         }
 
-        protected SpecialCharactersLinkGenerationTestsModel GetEntityByKey(string key)
```
xuzhg: >GetEntityByKey [](start = 60, length = 14)

GetEntityByKey can't work now? It maybe a regression?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,103 @@
...
+    using Microsoft.OData;
+    using Microsoft.AspNetCore.Mvc.Internal;
+#else
+    using Moq;
+#endif
+
+
+namespace Microsoft.AspNet.OData.Test.Abstraction
+{
+    internal class ODataActionSelectorTestHelper
```
xuzhg: >internal class ODataActionSelectorTestHelper [](start = 4, length = 44)

would you please add a simple comments about this Helper class?

xuzhg: >class ODataActionSelectorTestHelper [](start = 13, length = 35)

It seems this class can be static?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -95,10 +95,19 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
         {
             RouteData routeData = context.RouteData;
             ODataPath odataPath = context.HttpContext.ODataFeature().Path;
+
             if (odataPath != null && routeData.Values.ContainsKey(ODataRouteConstants.Action))
             {
+                var odataRoute = routeData.Routers.OfType<ODataRoute>().FirstOrDefault();
```
xuzhg: >routeData.Routers.OfType<ODataRoute>().FirstOrDefault(); [](start = 33, length = 56)

I'd like to use "IODataFeature" to save/retrieve the route name and route prefix. "ODataRoute" will be removed later.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -95,10 +95,19 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+                var routePrefix = odataRoute?.RoutePrefix;
                 // Get the available parameter names from the route data. Ignore case of key names.
-                IList<string> availableKeys = routeData.Values.Keys.Select(k => k.ToLowerInvariant()).AsList();
+                // Remove route prefix and other non-parameter values from availableKeys
+                IList<string> availableKeys = routeData.Values.Keys
+                    .Where((key) => routePrefix != "{" + key + "}"
```
xuzhg: >routePrefix != "{" + key + "} [](start = 36, length = 29)

what about if prefix has the same key name template?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -160,12 +190,27 @@ private bool TryMatch(IList<ParameterDescriptor> parameters, IList<string> avail
...
+
                 return false;
             }
 
             return true;
         }
 
+        private bool RequestHasBody(RouteContext context)
+        {
+            return context.HttpContext.Request.ContentLength > 0 ||
```
xuzhg: >context.HttpContext.Request.ContentLength > 0 [](start = 19, length = 45)

this change seems to require the setting of content length ahead? It maybe a regression? what can we find a workaround for it?

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -138,6 +138,9 @@ public static class ODataBatchReaderExtensions
                 request.Body = bufferedStream;
             }
 
+            // remove the content length and content type headers in case the batchRequest item doesn't have a body
+            request.ContentLength = null;
+            request.ContentType = null;
```
xuzhg: is there any risk?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DateAndTimeOfDayTest.cs
habbes: I use the content length to determine whether the request has a body. So a controller action like 
```c#
Post(Person person)
{
  // ...
}
```
will only be matched if the request has a body, otherwise a 404 error is thrown. Because these parameters are not stored in the `RouteData`, this was the only way I could reasonably get such routes to be matched correctly without affecting other routes. A more detailed explanation of the criteria I added to the ODataActionSelector can be found in the second comment of this PR, as well as in [the comment on this link](https://github.com/OData/WebApi/pull/1988#issuecomment-567586566).

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBuilder/SpecialCharactersLinkGenerationTests.cs
habbes: The test had a bug and this method (`GetEntityByKey`) was not getting called, either because it was protected or because of the name, I don't remember. And the test case was not verifying the response of the request. So the test was getting an incorrect result but it was reporting a success. Specifically OData does not support single whitespace (`' '`) as key but the test marked it as supported. This was not related to my PR, it was a bug that existed in the test, but my PR helped expose the bug because of the stricter routing selection that I implemented that caused the request to return a 404 error instead of routing to the wrong method.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: My reasoning was that all real-world requests coming from a compliant http client would include the content-length header for requests with a body since it's required by the [HTTP spec](https://www.w3.org/Protocols/HTTP/1.0/draft-ietf-http-spec.html#Content-Length) and so I assume it would be safe to rely on it. Initially I had tried to get the length directly from the stream but this was not reliable cause the stream is not guaranteed to be seekable and didn't work consistently between .net framework and .net core.

I also updated tests that create requests with bodies to include the content-length header. I don't know of a better workaround.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
habbes: @xuzhg Initially, the headers from the batch request were being copied to the individual requests inside the batch. This led to some of the requests getting headers that didn't make sense, for example some GET requests inside the batch would end up having a Content-Length and Content-Type even if they did not have any content/body. But if the individual requests (e.g. POST and PUT) have a body, then they would overwrite the content headers with their own. So I don't believe there is a risk.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: What do you mean by that?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: @xuzhg Can the route prefix be currently retrieved from the `IODataFeature`? Or will it be added to it in a future PR?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
habbes: Done

habbes: Done

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DateAndTimeOfDayTest.cs
xuzhg: From the Http Spec, content length is optional for Http request. and the Length must be zero in Get and Delete request. In the codes, if it's not required in previous version, now, it's required. That's a breaking change.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
mikepizzo: I buy that applying Content-Length across all of the requests is unhelpful, and that copying Content-Type to GET/DELETE requests is similarly not helpful. I do think there may be some use in copying Content-Type from the overall batch request to individual requests within the batch that have content (PUT/POST/PATCH) but don't have Content-Type header specified, and *in those cases* not setting Content-Type could be a breaking change.

habbes: I created a separate issue to discuss/address this: https://github.com/OData/WebApi/issues/2055

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DateAndTimeOfDayTest.cs
habbes: When you say the Length must be zero in Get and Delete, does that mean that the Length header is not optional in Get and Delete?

Anyway, the reason I check the content length was just to check whether the request has a body or not. If there's a reliable way to find out whether the request body is not empty without using the content length or content type, then I can use that method. Initially I tried to check the length of the stream, but this is not reliable because the stream is not always seekable. The options is to read the stream and copy it, but I don't think this is worth it, it seems too wasteful.

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
```diff
@@ -106,5 +112,80 @@ private static HttpResponseMessage CreateErrorResponse(HttpRequestMessage reques
...
+            var routePrefix = route?.RoutePrefix;
+            var availableKeys = routeData.Values.Keys
+                .Where(k => routePrefix != "{" + k + "}"
+                    && k != ODataRouteConstants.Action
+                    && k != ODataRouteConstants.Controller
+                    && k != ODataRouteConstants.ODataPath)
+                .Select(k => k.ToUpperInvariant())
+                .ToList();
+
+            if (parameters.Count == 0 && availableKeys.Count > 0)
```
mikepizzo: >availableKeys.Count > 0 [](start = 41, length = 23)

more efficient to do availableKeys.Any(); you don't care about the actual count, so no need to count them. you just want to know if there are any, and any can return after it finds the first one.

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
```diff
@@ -106,5 +112,80 @@ private static HttpResponseMessage CreateErrorResponse(HttpRequestMessage reques
...
+                }
+                if (context.Request.GetQueryNameValuePairs().Any(kvp => kvp.Key == p.ParameterName))
+                {
+                    continue;
+                }
+                if (!matchedBody && RequestHasBody(context))
+                {
+                    matchedBody = true;
+                    continue;
+                }
```
mikepizzo: Why is this done inside of the loop? couldn't you do this check before (or after) checking parameters?

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
```diff
@@ -106,5 +112,80 @@ private static HttpResponseMessage CreateErrorResponse(HttpRequestMessage reques
...
+            var routeData = context.RouteData;
+            var conventionsStore = context.Request.ODataProperties().RoutingConventionsStore;
+            var matchedBody = false;
+            var route = routeData.Route as ODataRoute;
+            var routePrefix = route?.RoutePrefix;
+            var availableKeys = routeData.Values.Keys
+                .Where(k => routePrefix != "{" + k + "}"
+                    && k != ODataRouteConstants.Action
+                    && k != ODataRouteConstants.Controller
+                    && k != ODataRouteConstants.ODataPath)
```
mikepizzo: This logic seems a bit brittle, given that you do the check on line 132 to make sure that you have eliminated any "special" route values if the action has no parameters.  If someone were to add another special-purpose route value in the future, then this would have to be updated to omit it.  Is this line, plus line 132, just an optimization? If all of the parameters for the action are accounted for, should we care if there are other route values?

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
mikepizzo: >&& [](start = 38, length = 2)

Why are you special casing ==0? if you are enforcing that all of the keys in the route value are mapped, wouldn't you want to do availableKeys.Count > parameters.Count?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
mikepizzo: >.FirstOrDefault() [](start = 71, length = 17)

Is it safe to assume that the route data we care about will be the first one? @saxu@microsoft.com; aren't there cases where the user might register multiple odata routes?

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
```diff
@@ -106,5 +112,80 @@ private static HttpResponseMessage CreateErrorResponse(HttpRequestMessage reques
...
+                .Select(k => k.ToUpperInvariant())
```
mikepizzo: >ToUpperInvariant [](start = 31, length = 16)

In AspNetCore we use ToLowerInvariant.  Might be better to be consistent in case we write common routines/merge code later?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -107,15 +116,16 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
                         .Where(p =>
                         {
                             return p.ParameterType != typeof(ODataPath) &&
-                                !ODataQueryParameterBindingAttribute.ODataQueryParameterBinding.IsODataQueryOptions(p.ParameterType);
```
mikepizzo: >                                 [](start = 0, length = 32)

did you change the spacing intentionally?  All things being equal, I prefer it indented.

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
```diff
@@ -106,5 +112,80 @@ private static HttpResponseMessage CreateErrorResponse(HttpRequestMessage reques
...
+                .ToList();
+
+            if (parameters.Count == 0 && availableKeys.Count > 0)
+            {
+                return false;
+            }
+
+            foreach (var p in parameters)
```
mikepizzo: Please add comments below, to explain where the logic is not obvious, as you do in the equivalent .NETCore code.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -126,31 +136,51 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
             return _innerSelector.SelectBestCandidate(context, candidates);
         }
 
-        private bool TryMatch(IList<ParameterDescriptor> parameters, IList<string> availableKeys, ODataOptionalParameter optionalWrapper)
+        private bool TryMatch(RouteContext context, IList<ParameterDescriptor> parameters, IList<string> availableKeys, ODataOptionalParameter optionalWrapper, int totalParameterCount)
         {
+            if (totalParameterCount == 0 && availableKeys.Count > 0)
+            {
+                return false;
+            }
```
mikepizzo: similar to asp.net case -- this seems to special case ==0, and be brittle (assuming we've removed any inapplicable keys).

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
mikepizzo: Rather than set these in CreateHttpContext and then remove them here, if we don't think certain headers should be copied (and I certainly agree for content-length) then we should put the right logic in CreateHttpContext for copying the relevant headers and omitting others.  [http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_HeaderFields](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_HeaderFields) spells out how well-known headers on the overall batch should be applied to individual statements within the batch. For headers in this list, if it doesn't specifically say that they are inherited, then we shouldn't copy. We should continue to copy headers not in this list.

habbes:
I'm moving this PR to #2168 because this one was no longer triggering the build pipeline

Issue #1610, opened 2018-09-07
Unable to have query parameters on functions bound to types with auto-expanding properties


D4N14L:
All functions which are bound to types with auto-expanding properties fail to return correctly, instead failing with:
```"message": "Could not find a property named 'fields' on type 'oneDrive.itemActivityStat'.",
            "type": "Microsoft.OData.ODataException",
            "stacktrace": "   at Microsoft.OData.UriParser.SelectExpandBinder.GenerateExpandItem(ExpandTermToken tokenIn)\r\n   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()\r\n   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()\r\n   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)\r\n   at Microsoft.OData.UriParser.SelectExpandBinder.Bind(ExpandToken tokenIn)\r\n   at Microsoft.OData.UriParser.SelectExpandSemanticBinder.Bind(ODataPathInfo odataPathInfo, ExpandToken expandToken, SelectToken selectToken, ODataUriParserConfiguration configuration)\r\n   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseSelectAndExpandImplementation(String select, String expand, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)\r\n   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseSelectAndExpand()\r\n   at System.Web.OData.Query.SelectExpandQueryOption.ProcessLevels()\r\n   at System.Web.OData.Query.ODataQueryOptions.ApplySelectExpand[T](T entity, ODataQuerySettings querySettings)\r\n   at System.Web.OData.Query.ODataQueryOptions.ApplyTo(IQueryable query, ODataQuerySettings querySettings)\r\n   at System.Web.OData.EnableQueryAttribute.ExecuteQuery(Object response, HttpRequestMessage request, HttpActionDescriptor actionDescriptor, ODataQueryContext queryContext)\r\n   at System.Web.OData.EnableQueryAttribute.OnActionExecuted(HttpActionExecutedContext actionExecutedContext)"
```

### Assemblies affected

OData WebApi lib 6.0

### Reproduce steps

Create an entity type with auto-expanding properties, and create a function that binds to this type. Make a request to the function endpoint, with a query string (can be unrelated to OData, eg. '?test=1'

### Expected result

[...]

PR #2012, opened 2020-01-09
#1610 - Fix OData AddLink (CreateRef) error when performing batch operations

gathogojr:
### Issues

*This pull request fixes issue [#1610](https://github.com/OData/odata.net/issues/1610).*

### Description

An error is observed when an entity is created by an insert request and then referenced in a subsequent add link (create ref) request _in the same batch_. This behaviour is observed on an OData service created using **Microsoft.AspNetCore.OData 7.3.0** or lower. The behaviour is not observed against an equivalent service created using **Microsoft.AspNet.OData**.

The bug seems to have been introduced by the addition of an extension method `GetODataContentIdMapping()` in the [**Microsoft.AspNetCore.OData.Extensions**](https://github.com/OData/WebApi/blob/d669038523e8dbb8e1709b1044ac0d51aed40a6a/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs#L307) namespace that returns `null` instead of the expected dictionary comprising of Content-ID to Location mapping. An extension method with a similar signature is defined in the [**Microsoft.AspNetCore.OData.Batch**](https://github.com/OData/WebApi/blob/d669038523e8dbb8e1709b1044ac0d51aed40a6a/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs#L307) namespace where it returns the expected value.

The inability to fetch the Content-ID to Location mapping results into the create ref action method in the controller being sent the $-prefixed request identifier as the Uri instead of the resource location.

By examining the logic in **Microsoft.AspNet.OData** codebase, and based on the fact that the Content-ID to Location mapping is (currently) applicable to batch requests, the `GetODataContentIdMapping()` in **Microsoft.AspNetCore.OData.Extensions** namespace is dropped to allow **Microsoft.AspNetCore.OData.Batch** namespace provide the implementation of `GetODataContentIdMapping()` required for use within the [**ODataInputFormatter**](https://github.com/OData/WebApi/blob/d669038523e8dbb8e1709b1044ac0d51aed40a6a/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs#L159) class.

This change fixes the error.

A related fix was made to a property `ODataContentIdMapping` in the [**WebApiRequestMessage**](https://github.com/OData/WebApi/blob/d669038523e8dbb8e1709b1044ac0d51aed40a6a/src/Microsoft.AspNetCore.OData/Adapters/WebApiRequestMessage.cs#L186) adapter class that also return `null` instead of the Content-ID to Location mapping.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
@@ -8,6 +8,7 @@ using System.Runtime.Serialization;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Adapters;
+using Microsoft.AspNet.OData.Batch;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter.Deserialization;
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
-#if !NETCORE // TODO #939: Enable these test on AspNetCore.
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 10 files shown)


habbes:
I would suggest adding tests that check whether AspNet(Core) OData correctly handles batch requests where one request references another request in the same change set, since it's part of the protocol.

gathogojr:
> I would suggest adding tests that check whether AspNet(Core) OData correctly handles batch requests where one request references another request in the same change set, since it's part of the protocol.

Thanks @habbes Working on the tests at the moment

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
```diff
@@ -418,6 +451,149 @@ public async Task ValidateRequest_Throws_IfRequestContentTypeDoesNotHaveBoundary
...
                 (await errorResponse.Response.Content.ReadAsAsync<HttpError>()).Message);
         }
+#endif
+        [Fact]
+        public async Task SendAsync_Works_ForBatchRequestWithInsertedEntityReferencedInAnotherRequest()
+        {
+            var endpoint = "http://localhost";
+
+            // Create entity request
+            var createOrderPayload = "{\"@odata.type\":\"Microsoft.AspNet.OData.Test.Batch.BatchTestOrder\",\"Id\":2,\"Amount\":50}";
```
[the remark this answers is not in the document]
gathogojr: @odero Fixed

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
```diff
@@ -8,6 +8,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Adapters;
+using Microsoft.AspNet.OData.Batch;
```
xuzhg: No codes changed in this file, but adding a using?

gathogojr: @xuzhg There was an extension method `GetODataContentIdMapping()` in the `Microsoft.AspNetCore.OData.Extensions` namespace that was returning null instead of the expected dictionary comprising of Content-ID to Location mapping. An extension method with a similar signature is defined in the `Microsoft.AspNetCore.OData.Batch` namespace where it returns the expected value. After deleting the extension method responsible from the bug from [here](https://github.com/OData/WebApi/pull/2012/files#diff-062764cf9e600d884f48c90b5e3c6100L302), I needed to add this `using` to import the extension method that works as expected

xuzhg:
Merged at https://github.com/OData/WebApi/commit/82c54cb70af7b932d65da3bcd9bbe16388b842ea

Issue #2001, opened 2020-01-03
Performance kill when using $expand and the model has many entities and actions.


CarlosMontagut:
[TestExpand.zip](https://github.com/OData/WebApi/files/4019441/TestExpand.zip)

When requesting an entity collection with a simple $expand of a many to one relation, performance is killed when the model has many entities and actions defined (not related with the entity requested, about 200 entities with 20 actions each).

If the model has only two entities, requesting the collection with the $expand takes around 150ms. If you add to the model the 200 entities with 20 actions each, the request takes 20/30 times more. 

The bigger the model, the bigger the performance is killed.

If I remove the $expand in the request, performance is good.

### Assemblies affected

Microsoft.AspNet.OData 7.3

### Reproduce steps

1. Define two entities, ObjectDummy and ObjectDummy2. ObjectDummy has a reference to ObjectDummy2 called Campo8
2. Request http://localhost:59703/api/odata/ObjetoDummy?$expand=Campo8
3. Result is ready in about 150ms
4. Define 200 more entities (not related to ObjectDummy), and define 20 actions on each entity.
5. Request same http://localhost:59703/api/odata/ObjetoDummy?$expand=Campo8
6. Result is ready in about 3s / 5s

The more entities and action define, the more the performance is killed.

### Expected result

Performance shouldn't be affected in that grade when there are more entities in the model

### Actual result

Performance is killed, take around 20 / 30 times more (and even more if the model grows).

### Additional detail

I attach a simple proyect that demonstrate it. In the WebApiConfig.cs (line 95) there is a code to uncomment and check the performance with or without the additional entities).

KanishManuja-MS:
@CarlosMontagut Thanks for submitting the issue. We will look into it and get back.

CarlosMontagut:
Thanks! @KanishManuja-MS let me know if you need any information or sample !

CarlosMontagut:
@KanishManuja-MS  / @odero 

I think I found the performance problem. 

The method GetEdmType in SelectExpandWrapper is calling model.GetEdmTypeReference(elementType). The bigger the model, the longer it takes for GetEdmTypeReference to return the EdmType.

 It only happend when you use $expand in the ODATA query string.

In my case, I was quering a full entity set wtih $expand, and the result returned N entities, calling GetEdmType N times.

I did a test using a simple cache and calling GetEdmTypeReference only if I have a cache miss, and the performance rocket.

[...]

PR #2014, opened 2020-01-10
Fix Performance drop when using a big model.

CarlosMontagut:
### Issues

This pull request fixes issue  #2001

### Description

When requesting an entity collection with a simple $expand of a many to one relation, performance is killed when the model has many entities and actions defined

- [ ] *Test cases added*

The ticket has a exported project to reproduce the issue.

### Additional work necessary

The pull request has a static cache, with has as a key the model. The cache will be holding a reference to the model, wich might be a memory problem if there is an escenario where models are create and discarded often. An option could be to use a hash of the model or any other unique identity that the model could have.

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/ETagMessageHandler.cs
+++ b/src/Microsoft.AspNet.OData.Shared/ETagMessageHandler.cs
@@ -83,7 +83,7 @@ namespace Microsoft.AspNet.OData
                 return edmTypeReference.AsEntity();
             }
 
-            IEdmTypeReference reference = EdmLibHelpers.GetEdmTypeReference(model, value.GetType());
+            IEdmTypeReference reference = model.GetTypeMappingCache().GetEdmType(value.GetType(), model);
             if (reference != null && reference.Definition.IsOrInheritsFrom(edmType))
             {
                 return (IEdmEntityTypeReference)reference;
--- a/src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
@@ -751,8 +751,8 @@ namespace Microsoft.AspNet.OData
                 throw Error.InvalidOperation(SRResources.QueryGetModelMustNotReturnNull);
             }
 
-            IEdmEntityType baseEntityType = model.GetTypeMappingCache().GetEdmType(elementClrType, model).Definition as IEdmEntityType;
-            IEdmStructuredType structuredType = model.GetTypeMappingCache().GetEdmType(elementClrType, model).Definition as IEdmStructuredType;
+            IEdmEntityType baseEntityType     = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition as IEdmEntityType;
+            IEdmStructuredType structuredType = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition as IEdmStructuredType;
             IEdmProperty property = null;
             if (path != null)
             {
--- a/src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
@@ -751,8 +751,9 @@ namespace Microsoft.AspNet.OData
                 throw Error.InvalidOperation(SRResources.QueryGetModelMustNotReturnNull);
             }
 
-            IEdmEntityType baseEntityType     = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition as IEdmEntityType;
-            IEdmStructuredType structuredType = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition as IEdmStructuredType;
+            IEdmType edmType = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition;
+            IEdmEntityType baseEntityType = edmType as IEdmEntityType;
+            IEdmStructuredType structuredType = edmType as IEdmStructuredType;
             IEdmProperty property = null;
             if (path != null)
             {
--- a/src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
@@ -751,8 +751,8 @@ namespace Microsoft.AspNet.OData
                 throw Error.InvalidOperation(SRResources.QueryGetModelMustNotReturnNull);
             }
 
-            IEdmEntityType baseEntityType = model.GetEdmType(elementClrType) as IEdmEntityType;
-            IEdmStructuredType structuredType = model.GetEdmType(elementClrType) as IEdmStructuredType;
+            IEdmEntityType baseEntityType = model.GetTypeMappingCache().GetEdmType(elementClrType, model).Definition as IEdmEntityType;
+            IEdmStructuredType structuredType = model.GetTypeMappingCache().GetEdmType(elementClrType, model).Definition as IEdmStructuredType;
             IEdmProperty property = null;
             if (path != null)
             {
--- a/src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
@@ -418,7 +418,7 @@ namespace Microsoft.AspNet.OData.Query
             }
 
             Type clrType = value.GetType();
-            return EdmLibHelpers.GetEdmType(model, clrType);
+            return model.GetTypeMappingCache().GetEdmType(clrType, model).Definition;
         }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
@@ -418,7 +418,7 @@ namespace Microsoft.AspNet.OData.Query
             }
 
             Type clrType = value.GetType();
-            return model.GetTypeMappingCache().GetEdmType(clrType, model).Definition;
+            return model.GetTypeMappingCache().GetEdmType(clrType, model)?.Definition;
         }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
         /// <inheritdoc />
         public IEdmTypeReference GetEdmType()
-        {
+        { 
             IEdmModel model = GetModel();
 
             if (InstanceType != null)
@@ -62,7 +62,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             }
 
             Type elementType = GetElementType();
-            return model.GetEdmTypeReference(elementType);
+            return model.GetTypeMappingCache().GetEdmType(elementType, model);
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
         /// <inheritdoc />
         public IEdmTypeReference GetEdmType()
-        { 
+        {
             IEdmModel model = GetModel();
 
             if (InstanceType != null)
--- a/src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
@@ -42,7 +42,7 @@ namespace Microsoft.AspNet.OData
         /// <param name="actionExecutedContext">The context related to this action, including the response message,
         /// request message and HttpConfiguration etc.</param>
         [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling",
-                Justification = "The majority of types referenced by this method result from HttpActionExecutedContext")]
+            Justification = "The majority of types referenced by this method result from HttpActionExecutedContext")]
         public override void OnActionExecuted(HttpActionExecutedContext actionExecutedContext)
         {
             if (actionExecutedContext == null)
--- a/src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
@@ -27,6 +27,7 @@ namespace Microsoft.AspNet.OData
     /// <see cref="EnableQueryAttribute"/> to validate incoming queries. For more information, visit
     /// http://go.microsoft.com/fwlink/?LinkId=279712.
     /// </summary>
+    [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling", Justification = "The majority of types referenced by this method result from HttpActionExecutedContext")]
     [SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes",
         Justification = "We want to be able to subclass this type.")]
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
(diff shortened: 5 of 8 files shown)


habbes:
Thanks for the PR @CarlosMontagut 
I have realized that there's a `ClrTypeCache` class implemented in https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
and a `model.GetTypeMappingCache()` method implemented as an extension method in https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/EdmModelExtensions.cs

If this could be reused in your fix, then use it instead of creating another ad-hoc cache.

CarlosMontagut:
@habbes great, I will check with the already existing caches and let you know!

CarlosMontagut:
@habbes, just update the pull request, delete the ad-hoc cache, and use the ClrTypeCache in SelectExpandWrapper.

It worked perfect, performance kept pretty much as the ad-cache solution.

Let me know if you see any other possible change !

habbes:
@CarlosMontagut That looks great. I think it would also be worthwhile to use the cache inside the `GetEdmType()` method where it make recursive calls on lines 145 and 175 of `EdmLibHelpers.cs`, otherwise there might still be accesses that bypass the cache.

habbes:
@CarlosMontagut following the additional discussion on the issue regarding the complexity of using the cache as-is inside `EdmLibHelpers.GetEdmType()`, I think what you've done is good enough to address the issue in question.

habbes:
@KanishManuja-MS @xuzhg could you double-check this, and if it's okay please merge.

Here's a summary of the discussions:
- @CarlosMontagut use the `ClrTypeCache` in `SelectExpandWrapper.GetEdmType()` and fixed the performance issue
- The internal `GetEdmType()` in `EdmLibHelpers` is recursive, and the recursive calls are not cached. Unless the result of the calls are cached by external methods (like in the current PR), then this might lead to performance loss. But we don't know if there are such cases. Making the recursive call use the type cache would require making changes to the `EdmLibHelpers`'s public API as well as the `ClrTypeCache`. Do you think this is something we should handle in this issue as well?

CarlosMontagut:
@habbes, I will:

- Replace all the calls (6 places) to GetEdmType by the call to ClrTypeCache.
- Leave the recursive call to GetEdmType as it is, I don't think this will be an issue in performance, since will only happend when a new entry is added to the cache.

Will update the PR soon.

CarlosMontagut:
@habbes I got an error in the building process due a desing rule:

> CA1506	'EnableQueryAttribute' is coupled with 81 different types from 24 different namespaces. Rewrite or refactor this class's methods to decrease its class coupling, or consider moving some of the class's methods to some of the other types it is tightly coupled with. A class coupling above 95 indicates poor maintainability, a class coupling between 95 and 80 indicates moderate maintainability, and a class coupling below 80 indicates good maintainability.	Microsoft.AspNet.OData	D:\NewTech\Proyectos\WebApi2\src\Microsoft.AspNet.OData\EnableQueryAttribute.cs <

I don't have a deep knowlege of this library design, any suggestion to fix this ?

habbes:
@CarlosMontagut I see that class already has an attribute to suppress that rule on some of its methods. Maybe it would be fair to suppress the rule on the entire class. What do you think @KanishManuja-MS  @xuzhg?

habbes:
@CarlosMontagut there are also 2 failing tests in the `ODataQueryContext` class in the AspNet classic solution (`WebApiOData.AspNet.sln`).

CarlosMontagut:
@habbes I can't access the continuous-integration/vsts details due Authorization problems. Do I need to get access from someone to view the details?

I'm trying to run the test in my environment but I'm getting many many errores, probably because something wrong in the setup of my environment...

habbes:
@CarlosMontagut the failed tests are related to null pointer exceptions as a result of doing `GetEdmType(...).Definition)` when `GetEdmType()` returns null. If I add the suppression of the CA rule and use the null-conditional operator to access the `Definition` property, the tests pass on my machine.
Regarding access to vsts build, it's something we'd have to look into.

CarlosMontagut:
@habbes  just added the null-conditional operator. 

Please confirm if I should add the suppression of the CA rule.

Thanks!

habbes:
@CarlosMontagut  the build on CI is now successful.

Regarding the CA rule, since the build is failing locally due to that rule, and since that rule has already been suppressed in that class on some of the methods, I would advocate just adding the suppression of the rule to the class to complete this PR.

If @KanishManuja-MS or @xuzhg have a different view when they add their review(s), then we can remove the suppression.

CarlosMontagut:
@habbes, added the suppression at class level 

The original suppression at method level (OnActionExectured) was left to avoid another rule fail, seems that the same rule is applied at class and at method level.

habbes:
Thanks @CarlosMontagut

src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
```diff
@@ -418,7 +418,7 @@ private static IEdmType GetTypeFromObject(object value, IEdmModel model)
             }
 
             Type clrType = value.GetType();
-            return EdmLibHelpers.GetEdmType(model, clrType);
+            return model.GetTypeMappingCache().GetEdmType(clrType, model).Definition;
```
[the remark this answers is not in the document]
CarlosMontagut: @habbes Yes, I agree, want me to do the chance ?

src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
```diff
@@ -27,6 +27,7 @@ namespace Microsoft.AspNet.OData
     /// <see cref="EnableQueryAttribute"/> to validate incoming queries. For more information, visit
     /// http://go.microsoft.com/fwlink/?LinkId=279712.
     /// </summary>
+    [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling", Justification = "The majority of types referenced by this method result from HttpActionExecutedContext")]
```
odero: Is it really necessary to do this at class level? Or can it only be applied to the specific method(s) that are failing?

CarlosMontagut: @odero, I think the metric this rules check can be calculated at method and class level. Is failing at method level (in the ones that has the supress warning), but also is failing at class level (so even if we put the supress warning in other methods, fails).

CarlosMontagut: @odero, regarding the test, we don't have an automated test to show this issue, just a sample project attached to the issue. I'm leaving office today and won't be able to produce a ticket until 2 February.

src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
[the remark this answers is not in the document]
habbes: @CarlosMontagut yes I think that would be great.

habbes:
@odero here are new results based the latest build of this PR.

the tests were made using a model with 20,000 functions

Before PR:
- Query without expand: 579ms
- Query with expand: 25.07s

After PR:
- Query without expand: 561ms
- Query with expand: 556ms

habbes:
@odero here's the project I used to test performance: https://github.com/habbes/ODataWebApiExpandPerformanceBug

src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
```diff
@@ -41,7 +42,7 @@ public partial class EnableQueryAttribute : ActionFilterAttribute
         /// <param name="actionExecutedContext">The context related to this action, including the response message,
         /// request message and HttpConfiguration etc.</param>
         [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling",
-            Justification = "The majority of types referenced by this method result from HttpActionExecutedContext")]
+                Justification = "The majority of types referenced by this method result from HttpActionExecutedContext")]
         public override void OnActionExecuted(HttpActionExecutedContext actionExecutedContext)
```
KanishManuja-MS: nit: revert spacing.

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
```diff
@@ -751,8 +751,8 @@ internal static void ValidateSelectExpandOnly(ODataQueryOptions queryOptions)
                 throw Error.InvalidOperation(SRResources.QueryGetModelMustNotReturnNull);
             }
 
-            IEdmEntityType baseEntityType = model.GetEdmType(elementClrType) as IEdmEntityType;
-            IEdmStructuredType structuredType = model.GetEdmType(elementClrType) as IEdmStructuredType;
+            IEdmEntityType baseEntityType     = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition as IEdmEntityType;
```
KanishManuja-MS: >      [](start = 41, length = 5)

Fix spacing.

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
```diff
@@ -751,8 +751,8 @@ internal static void ValidateSelectExpandOnly(ODataQueryOptions queryOptions)
...
+            IEdmStructuredType structuredType = model.GetTypeMappingCache().GetEdmType(elementClrType, model)?.Definition as IEdmStructuredType;
```
KanishManuja-MS: >?. [](start = 109, length = 2)

I think the build used to fail for one of the older solutions when you used "?.". I guess that is no longer the case.

src/Microsoft.AspNet.OData.Shared/ETagMessageHandler.cs
```diff
@@ -83,7 +83,7 @@ private static IEdmEntityTypeReference GetTypeReference(IEdmModel model, IEdmEnt
                 return edmTypeReference.AsEntity();
             }
 
-            IEdmTypeReference reference = EdmLibHelpers.GetEdmTypeReference(model, value.GetType());
+            IEdmTypeReference reference = model.GetTypeMappingCache().GetEdmType(value.GetType(), model);
```
KanishManuja-MS: >model.GetTypeMappingCache(). [](start = 42, length = 28)

Does is it make sense to use GetTypeMappingCache everytime when we try to get the EdmType? If the answer is yes, should we move getting the TypeMapping cache to GetEdmType/GetEdmTypeReference. If no, when does it make sense to go through the cache.

src/Microsoft.AspNet.OData.Shared/Query/DefaultSkipTokenHandler.cs
```diff
@@ -418,7 +418,7 @@ private static IEdmType GetTypeFromObject(object value, IEdmModel model)
             }
 
             Type clrType = value.GetType();
-            return EdmLibHelpers.GetEdmType(model, clrType);
+            return model.GetTypeMappingCache().GetEdmType(clrType, model)?.Definition;
```
KanishManuja-MS: +1;

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
```diff
@@ -45,7 +45,7 @@ internal abstract class SelectExpandWrapper : IEdmEntityObject, ISelectExpandWra
 
         /// <inheritdoc />
         public IEdmTypeReference GetEdmType()
-        {
+        {
```
KanishManuja-MS: Revert this line.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
```diff
@@ -62,7 +62,7 @@ public IEdmTypeReference GetEdmType()
             }
 
             Type elementType = GetElementType();
-            return model.GetEdmTypeReference(elementType);
+            return model.GetTypeMappingCache().GetEdmType(elementType, model);
```
KanishManuja-MS: >GetTypeMappingCache [](start = 25, length = 19)

Inside GetTypeMappingCache, we call this every time - typeMappingCache = model.GetAnnotationValue<ClrTypeCache>(model); is this a cheap enough call? Should we cache the ClrTypeCache somewhere and may be pass it as an argument?

src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
xuzhg: +1

---
In reply to: [368624365](https://github.com/OData/WebApi/pull/2014#discussion_r368624365) [](ancestors = 368624365)

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
xuzhg: >      [](start = 41, length = 5)

align the code

xuzhg: >model.GetTypeMappingCache().GetEdmType(elementClrType, model) [](start = 48, length = 61)

can we save the middle value, it seems here duplicate the codes twice?

src/Microsoft.AspNet.OData.Shared/ETagMessageHandler.cs
habbes: @KanishManuja-MS I had initially proposed that the cache be used also inside `GetEdmType()`, but there was some complexity in achieving that, partly because of a recursive call in `GetEdmType` and the method signatures. I don't know if it would be possible to refactor it to use the cache internally without changing public method signatures all over the place (see [this comment](https://github.com/OData/WebApi/pull/2014#issuecomment-573722066), and [this](https://github.com/OData/WebApi/issues/2001#issuecomment-573634346) and [this too](https://github.com/OData/WebApi/issues/2001#issuecomment-573675387)).

However, I could give this another look to see whether we can find a clean way to add the cache in `GetEdmType` without affect too much of the existing methods.

CarlosMontagut:
Hi @habbes , I'm back. Sorry, this is my first pull request and I'm not familiar with the process. 

Should I apply the changes suggested by @KanishManuja-MS (spacing, saving middle value, etc) and commit that on this pull request? 

I was in doubt because saw the pull request approved...

Thanks

src/Microsoft.AspNet.OData.Shared/ETagMessageHandler.cs
habbes: @CarlosMontagut
Maybe you could modify `EdmHelpers.GetEdmTypeReference` to call
```
return edmModel.GetTypeMappingCache().GetEdmType(..., ...);
```
and then create a new public method like `EdmHelpers.GetEdmTypeReferenceWithoutCache()` 
method that does exactly what `GetEdmTypeReference` is currently doing. Then `ClrTypeChance.GetEdmType()` should be modified to call `GetEdmTypeReferenceWithoutCache()`. Then the other methods can just call `EdmHelpers.GetEdmTypeReference()` directly and will be using the cache. But `GetEdmModel()` should not be called directly.

@KanishManuja-MS what do you think about that?

habbes:
@CarlosMontagut, yes please apply the changes. Usually more than person reviews and approves PR's.

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
CarlosMontagut: Done!

CarlosMontagut: Done!

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
CarlosMontagut: Done!

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
CarlosMontagut: Done!

src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
CarlosMontagut: Done!

src/Microsoft.AspNet.OData.Shared/ETagMessageHandler.cs
CarlosMontagut: @habbes, I think we can do that, and hide the use of the Cache inside GetEdmTypeReference, also keeping there a reference to the cache for all the future calls (avoiding the cost of getting the cache each time).

@habbes / @KanishManuja-MS  let me know your thoughts and if you agree I can proceed doing that change.

xuzhg:
Merged at https://github.com/OData/WebApi/commits/master

PR #2035, opened 2020-01-29
[Endpoint routing review]First draft to enable endpoint routing using dynamic controller

xuzhg:
### Issues

### Description

* Enable Endpoint routing in ASP.NET Core OData

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            DefaultQuerySettings querySettings = serviceProvider.GetRequiredService<DefaultQuerySettings>();
+            if (querySettings == null)
+            {
+                throw Error.InvalidOperation(SRResources.MissingODataServices, nameof(DefaultQuerySettings));
+            }
+
+            querySettings = defaultQuerySettings;
+        }
```
KanishManuja-MS: How does this even work?

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
```diff
@@ -39,23 +38,7 @@ public static IRouteBuilder SetDefaultQuerySettings(this IRouteBuilder builder,
...
-
-            DefaultQuerySettings querySettings = builder.ServiceProvider.GetRequiredService<DefaultQuerySettings>();
-            if (querySettings == null)
-            {
-                throw Error.InvalidOperation(SRResources.MissingODataServices, nameof(DefaultQuerySettings));
-            }
-
-            querySettings = defaultQuerySettings;
+            builder.ServiceProvider.SetDefaultQuerySettings(defaultQuerySettings);
             return builder;
```
KanishManuja-MS: What is the reason behind this change? Extension method for service provider is not intuitive. May be you want to create a private static methods rather than extension methods.

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -10,6 +11,10 @@
 using Microsoft.AspNet.OData.Routing;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
+using Microsoft.AspNetCore.Routing;
+#if NETCOREAPP3_1
+using Microsoft.AspNetCore.Routing.Matching;
```
KanishManuja-MS: Once we upgrade to higher versions of net core, this would need to be changed right? Is there a conditional compile that specifies greater than?

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
```diff
@@ -39,23 +38,7 @@ public static IRouteBuilder SetDefaultQuerySettings(this IRouteBuilder builder,
                 throw Error.ArgumentNull("builder");
             }
```
KanishManuja-MS: Now I understand the reason for this change. I still feel extension method may not be right. Just a shared static method with a normal parameter may be better.

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointBuilderExtensions.cs
```diff
@@ -0,0 +1,698 @@
...
+        /// Maps the specified OData route and the OData route attributes. When the <paramref name="batchHandler"/> is
+        /// non-<c>null</c>, it will create a '$batch' endpoint to handle the batch requests.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="routeName">The name of the route to map.</param>
+        /// <param name="routePrefix">The prefix to add to the OData route's path template.</param>
+        /// <param name="model">The EDM model to use for parsing OData paths.</param>
+        /// <param name="batchHandler">The <see cref="ODataBatchHandler"/>.</param>
+        /// <returns>The <see cref="IEndpointRouteBuilder"/>.</returns>
+        public static IEndpointRouteBuilder MapODataRoute(this IEndpointRouteBuilder builder,
```
KanishManuja-MS: >MapODataRoute [](start = 44, length = 13)

MapODataServiceRoute to maintain parity?

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointBuilderExtensions.cs
```diff
@@ -0,0 +1,698 @@
...
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="routeName">The name of the route to map.</param>
+        /// <param name="routePrefix">The prefix to add to the OData route's path template.</param>
+        /// <param name="model">The EDM model to use for parsing OData paths.</param>
+        /// <param name="pathHandler">The <see cref="IODataPathHandler"/> to use for parsing the OData path.</param>
+        /// <param name="routingConventions">
+        /// The OData routing conventions to use for controller and action selection.
+        /// </param>
+        /// <returns>The <see cref="IEndpointRouteBuilder"/>.</returns>
+        public static IEndpointRouteBuilder MapODataRoute(this IEndpointRouteBuilder builder,
```
KanishManuja-MS: >MapODataRoute [](start = 44, length = 13)

Same for other. Change the name.

tools/WebStack.versions.settings.targets
```diff
@@ -4,7 +4,7 @@
   <!-- Set the version number: major, minor, build and release (i.e. alpha, beta or blank for RTM)-->
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">7</VersionMajor>
-    <VersionMinor Condition="'$(VersionMinor)' == ''">3</VersionMinor>
+    <VersionMinor Condition="'$(VersionMinor)' == ''">4</VersionMinor>
     <VersionBuild Condition="'$(VersionBuild)' == ''">0</VersionBuild>
     <VersionRelease Condition="'$(VersionRelease)' == ''"></VersionRelease>
   </PropertyGroup>
```
KanishManuja-MS: Add beta here.

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+            IList<string> templates = new List<string>();
+            int startIndex = 0;
+            while (true)
+            {
+                int start = prefix.IndexOf('{', startIndex);
+                int end = prefix.IndexOf('}', start + 1); // start == -1 is ok
+                if (start == -1 || end == -1)
+                {
+                    break;
+                }
```
KanishManuja-MS: I am missing something may be. It seems that the if prefix has '{' and '}', this will loop indefinitely as prefix never changes.

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+                startIndex = end + 1;
+
+                string subStr = prefix.Substring(start, end - start + 1);
+                templates.Add(subStr);
+            }
+
+            foreach (var item in templates)
+            {
+                string variable = item.Substring(1, item.Length - 2); // remove { and }
+
```
KanishManuja-MS: Why substring again? We should add to templates after striping the braces already.

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+            else
+            {
+                return String.IsNullOrEmpty(odataPath) ? routePrefix : routePrefix + '/' + odataPath;
+            }
+        }
+
+        private static string UriEncode(string str)
+        {
+            string escape = Uri.EscapeUriString(str);
+            escape = escape.Replace("#", _escapedHashMark);
```
KanishManuja-MS: >"#" [](start = 36, length = 3)

Declare these strings as constants? It may be better to use the char overload instead of the string overload for better performance.

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointSelector.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        /// <param name="selector">The inner Endpoint selector</param>
+        public ODataEndpointSelector(EndpointSelector selector)
+        {
+            _innerSelector = selector;
+        }
+
+        /// <inheritdoc/>
+        public override Task SelectAsync(HttpContext httpContext, CandidateSet candidateSet)
+        {
```
KanishManuja-MS: What are the other things that can happen in the pipeline while this runs asynchronously?

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointPattern.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+#if !NETSTANDARD2_0
```
KanishManuja-MS: >!NETSTANDARD2_0 [](start = 4, length = 15)

This conditional is different from the other conditional. Can you explain the reason?

KanishManuja-MS:
I see the sample but there are no test cases. Can you please add test cases for endpoint routing? Specially for the ones where controller methods are overloaded?

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointSelector.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Microsoft.AspNet.OData.Extensions
+{
+    /// <summary>
+    /// A service that is responsible for the final OData Endpoint selection decision.
+    /// This selector is used to resolve the duplicated Endpoint selection.
+    /// </summary>
+    internal class ODataEndpointSelector : EndpointSelector
+    {
+        private EndpointSelector _innerSelector;
```
KanishManuja-MS: >_innerSelector [](start = 33, length = 14)

I am unclear on what does the coding style guidelines say about using _ for private members. It was the old naming convention and I see that not all classes use this. We should discuss what is the recommendation, since this convention is different in ODL as well.

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        private LinkGenerator _innerGenerator;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataEndpointLinkGenerator"/> class.
+        /// </summary>
+        /// <param name="generator">The inner Link generator</param>
+        public ODataEndpointLinkGenerator(LinkGenerator generator)
+        {
+            _innerGenerator = generator;
```
gathogojr: @xuzhg Is there a chance of this constructor being called with a null value for the `LinkGenerator` parameter? That would most certain precipitate chaos when the overridden methods that reference `_innerGenerator` are called

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointPattern.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// The route pattern is in this format: "routePrefix/{*ODataEndpointPath_routeName}"
+        /// </summary>
+        /// <param name="routeName">The route name. It can not be null and verify upper layer.</param>
+        /// <param name="routePrefix">The route prefix. It could be null or empty</param>
+        /// <returns>The OData route endpoint pattern.</returns>
+        public static string CreateODataEndpointPattern(string routeName, string routePrefix)
+        {
+            if (routeName == null)
+            {
+                throw Error.ArgumentNull("routeName");
```
gathogojr: @xuzhg Better to use `nameof` here

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
gathogojr: This loop can easily be eliminated by stripping the braces before adding them to the collection

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    internal class ODataDynamicControllerEndpointMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
+    {
+        private readonly EndpointMetadataComparer _comparer;
+        public ODataDynamicControllerEndpointMatcherPolicy(EndpointMetadataComparer comparer)
+        {
+            if (comparer == null)
+            {
+                throw new ArgumentNullException(nameof(comparer));
```
gathogojr: @xuzhg Maybe use `throw Error.ArgumentNull(...)` for consistency

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        public static DefaultQuerySettings GetDefaultQuerySettings(this IServiceProvider serviceProvider)
+        {
+            if (serviceProvider == null)
+            {
+                throw Error.ArgumentNull("serviceProvider");
```
gathogojr: @xuzhg Better use `nameof` here

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            return querySettings;
+        }
+
+        public static void MaxTop(this IServiceProvider serviceProvider, int? maxTopValue)
+        {
+            if (serviceProvider == null)
+            {
+                throw Error.ArgumentNull("serviceProvider");
```
gathogojr: @xuzhg Better use `nameof` here

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
```diff
@@ -251,8 +214,7 @@ public static IRouteBuilder Count(this IRouteBuilder builder)
...
 
-            DefaultQuerySettings defaultQuerySettings = builder.GetDefaultQuerySettings();
-            defaultQuerySettings.EnableCount = true;
+            builder.ServiceProvider.Count();
```
gathogojr: @xuzhg Do we need the variant of Count that accepts a `QueryOptionSetting` parameter? Given this overload [here](https://github.com/OData/WebApi/pull/2035/files#diff-1ac400cb4d772495bb6ba2aa4f28b4b5R159)

src/Microsoft.AspNetCore.OData/Routing/Conventions/ODataRoutingConventions.cs
```diff
@@ -38,6 +39,27 @@ public static class ODataRoutingConventions
...
+        public static IList<IODataRoutingConvention> CreateDefaultWithAttributeRouting(
+            string routeName,
+            IServiceProvider serviceProvider)
+        {
+            if (routeName == null)
+            {
+                throw Error.ArgumentNull("routeName");
+            }
+
+            IList<IODataRoutingConvention> routingConventions = CreateDefault();
```
gathogojr: @xuzhg The 4 lines here are a replica of these others [here](https://github.com/OData/WebApi/pull/2035/files#diff-eb87a7b003f2e163e4936620def60725R36). Maybe the existing `CreateDefaultWithAttributeRouting` overload could call this new one like this `return CreateDefaultWithAttributeRouting(routeName, builder.ServiceProvider`) - in which case the check for null `routeName` wouldn't be required in the existing overload, or consider adding a helper private method that both overloads call

mikepizzo:
>     public interface IODataFeature

Given that we are adding to a public interface, I assume we don't expect anyone to have their own custom implementations of this interface that would now break?

---
Refers to: src/Microsoft.AspNetCore.OData/Interfaces/IODataFeature.cs:20 in 59b98be. [](commit_id = 59b98beaecb4ce4ad19dd8564e79f33d04c39f8a, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/IPerRouteContainer.cs
```diff
@@ -11,6 +11,20 @@ namespace Microsoft.AspNet.OData
     /// </summary>
     public interface IPerRouteContainer
```
mikepizzo: >public [](start = 4, length = 6)

I assume there are no custom implementations that would break by our adding members to this interface?

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+
+    //public class ODataDynamicControllerEndpointSelector : IDisposable
+    //{
+    //    public DynamicControllerEndpointSelector(EndpointDataSource dataSource)
+    //        : this((EndpointDataSource)dataSource)
+    //    {
+    //    }
+    //}
```
mikepizzo: clean this up?

src/Microsoft.AspNetCore.OData/Routing/Conventions/ODataRoutingConventions.cs
mikepizzo: :thumbsup:

---
In reply to: [384475452](https://github.com/OData/WebApi/pull/2035#discussion_r384475452) [](ancestors = 384475452)

xuzhg:
Added the test cases

---
In reply to: [583046981](https://github.com/OData/WebApi/pull/2035#issuecomment-583046981) [](ancestors = 583046981)

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
xuzhg: Revert.

---
In reply to: [375983113](https://github.com/OData/WebApi/pull/2035#discussion_r375983113) [](ancestors = 375983113)

xuzhg:
>     public interface IODataFeature

We have been discussing about to add the public interface for a long time. It seems it's not a breaking change to add new items to a public interface. However, you are right. We don't expect any customer to create their own custom "IODataFeature" implementation.

---
In reply to: [591585939](https://github.com/OData/WebApi/pull/2035#issuecomment-591585939) [](ancestors = 591585939)

---
Refers to: src/Microsoft.AspNetCore.OData/Interfaces/IODataFeature.cs:20 in 59b98be. [](commit_id = 59b98beaecb4ce4ad19dd8564e79f33d04c39f8a, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/IPerRouteContainer.cs
xuzhg: Same as the comment in "IODataFeature".

---
In reply to: [384697141](https://github.com/OData/WebApi/pull/2035#discussion_r384697141) [](ancestors = 384697141)

src/Microsoft.AspNetCore.OData/Routing/Conventions/ODataRoutingConventions.cs
xuzhg: Changed.

Originally, i'd like to change the exiting public to the new one. However, it's a breaking change. So, i add a new one.
 

---
In reply to: [384715327](https://github.com/OData/WebApi/pull/2035#discussion_r384715327) [](ancestors = 384715327,384475452)

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
xuzhg: Thanks. I forgot to change that Count(...) method. Fixed.

---
In reply to: [384452658](https://github.com/OData/WebApi/pull/2035#discussion_r384452658) [](ancestors = 384452658)

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs
xuzhg: This class is removed and replaced using "ODataEndpointSelectorPolicy"

---
In reply to: [384446213](https://github.com/OData/WebApi/pull/2035#discussion_r384446213) [](ancestors = 384446213)

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
xuzhg: Theoretically, no. But, let me add the verification code.

---
In reply to: [384365505](https://github.com/OData/WebApi/pull/2035#discussion_r384365505) [](ancestors = 384365505)

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointPattern.cs
xuzhg: Resolved.

---
In reply to: [376008587](https://github.com/OData/WebApi/pull/2035#discussion_r376008587) [](ancestors = 376008587)

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
xuzhg: We only replace the string within "{" and "}". and keep the other as it. for example:

prefix:  "data{data}" and we have {"data": 1} in route value, the return of this method should be "data1", not "11". See more from my added test cases.

---
In reply to: [376002864](https://github.com/OData/WebApi/pull/2035#discussion_r376002864) [](ancestors = 376002864)

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
xuzhg: No, see my test cases for more information.

---
In reply to: [376002351](https://github.com/OData/WebApi/pull/2035#discussion_r376002351) [](ancestors = 376002351)

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
xuzhg: Extension method, we can call the method using Fluent way.

builder.Fun1().Func2().Func3()....

with static method. we can't do like this.

---
In reply to: [375990865](https://github.com/OData/WebApi/pull/2035#discussion_r375990865) [](ancestors = 375990865)

src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointLinkGenerator.cs
xuzhg: Replace doesn't have a overload to replace a char in a string. and for better readability, I'd like to use "#" directly here. The LinkGenerator mayeb change later. 

---
In reply to: [376006516](https://github.com/OData/WebApi/pull/2035#discussion_r376006516) [](ancestors = 376006516)

src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs
xuzhg: Synced offline, right?

---
In reply to: [375976380](https://github.com/OData/WebApi/pull/2035#discussion_r375976380) [](ancestors = 375976380)

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
xuzhg: Synced offline, right.

---
In reply to: [375972141](https://github.com/OData/WebApi/pull/2035#discussion_r375972141) [](ancestors = 375972141)

xuzhg:
The latest CI passed at: https://identitydivision.visualstudio.com/OData/_build/results?buildId=481353&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=fd490c07-0b22-5182-fac9-6d67fe1e939b

Issue #2043, opened 2020-02-05
Error when SkipToken contains a non-nullable string


Grauenwolf:
When sorting by two columns, the first being a string, it throws an exception when a `$skiptoken` is provided.

```
https://localhost:44322/odata/Transaction?$orderby=ContactCity&$skiptoken=ContactCity-'Atlanta',TxnId-13258
```

### Assemblies affected

```
    <PackageReference Include="Microsoft.AspNetCore.OData" Version="7.3.0" />
    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="3.1.1" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="3.1.1" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="3.1.1">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="3.1.0" />
    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="3.1.0" />
```

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*

### Expected result

*What would happen if there wasn't a bug.*

### Actual result

```
InvalidOperationException: The LINQ expression 'DbSet<Transaction>
.OrderBy(t => t.ContactCity)
.Where(t => string.Compare(
strA: t.ContactCity, 
strB: __TypedProperty_0, 
comparisonType: Ordinal) > 0 || t.ContactCity == __TypedProperty_0 && t.TxnId > __TypedProperty_1)' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to either AsEnumerable(), AsAsyncEnumerable(), ToList(), or ToListAsync(). See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
Microsoft.EntityFrameworkCore.Query.QueryableMethodTranslatingExpressionVisitor.<VisitMethodCall>g__CheckTranslated|8_0(ShapedQueryExpression translated, ref <>c__DisplayClass8_0 )
TargetInvocationException: Exception has been thrown by the target of an invocation.
[...]

PR #2053, opened 2020-02-16
Fix client-side evaluation for string comparison queries

SiberaIndustries:
### Issues

*This pull request fixes issue #2043 and fixes issue #2078.*

### Description

Entity Framework Core 3.0 [throws an exception](https://docs.microsoft.com/de-de/ef/core/what-is-new/ef-core-3.0/breaking-changes#linq-queries-are-no-longer-evaluated-on-the-client) when the framework can't translate a linq expression to SQL (called: client-side evaluation). 

This PR fixes the special case of [string comparisons](https://github.com/dotnet/efcore/issues/18020). EF Core can translate `string.Compare(string, string)` to SQL but it can't translate any [overload](https://elanderson.net/2018/12/entity-framework-core-string-filter-tips) of `string.Compare`. This is why the `StringComparison` argument gets removed.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
     /// </summary>
     public abstract class ExpressionBinderBase
     {
-        internal static readonly MethodInfo StringCompareMethodInfo = typeof(string).GetMethod("Compare", new[] { typeof(string), typeof(string), typeof(StringComparison) });
+        internal static readonly MethodInfo StringCompareMethodInfo = typeof(string).GetMethod("Compare", new[] { typeof(string), typeof(string) });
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -192,7 +191,7 @@ internal Expression CreateBinaryExpression(BinaryOperatorKind binaryOperator, Ex
                     case BinaryOperatorKind.GreaterThanOrEqual:
                     case BinaryOperatorKind.LessThan:
                     case BinaryOperatorKind.LessThanOrEqual:
-                        left = Expression.Call(StringCompareMethodInfo, left, right, OrdinalStringComparisonConstant);
+                        left = Expression.Call(StringCompareMethodInfo, left, right);
```
xuzhg: could it be a breaking change for non-ef core scenario?

SiberaIndustries: Sadly I can't negate this question, so there is a chance that a provider expects the ordinal case.

xuzhg:
Merged at: https://github.com/OData/WebApi/commit/63f1c664ee975a844fe4fbe0141919433e8c5cd3

PR #2060, opened 2020-02-19
Fix #2056: opertation returns path for the return entity set or singl…

### LINKED ISSUE #2056: Action returning a collection of entities with containment silently fails
If an entity type has a containment `EntityType<ProjectTask>().ContainsMany(pt => pt.ProjectTaskAssignments);` then action returning an entity collection of this type would fail. It just stop transmitting data to client somewhere on serialization.

### Assemblies affected
AspNetCore.OData 7.3.0
AspNET Core 3.1

### Reproduce steps
Have entities
```CSharp
public class Project
{
  public Guid Id {get;set;}
  public virtual ICollection<ProjectTasks> ProjectTasks() {get;set;}
}

public class ProjectTask
{
  public Guid Id {get;set;}

  public Guid ProjectId {get;set;}
  
  public virtual ICollection<ProjectTaskAssignment> ProjectTaskAssignments {get;set;}
}

public class ProjectTaskAssignment
{
  public Guid Id {get;set;}

  public Guid ProjectTaskId{get;set;}
}
```
Some ODataController with action
```CSharp
public class ProjectsController : ODataController<Project>
{
  public async Task<IActionResult> UpdateProjectTasks([FromODataUri] Guid key,
    [FromBody] ODataActionParameters parameters)
  {
    var projectTasks = ((IEnumerable<ProjectTask>)parameters["projectTasks"]).ToList();
    var updatedProjectTasks = await _someService.UpdateProjectTasksAsync(key, projectTasks);
    return Ok(updatedProjectTasks);
  }
}
```
config smth like
```CSharp
    var projectsConfig = builder.EntityType<Project>();
    var taskUpdateAction = projectsConfig.Action(nameof(ProjectsController.UpdateProjectTasks));
    taskUpdateAction.CollectionEntityParameter<ProjectTask>("projectTasks");
    taskUpdateAction.ReturnsCollectionFromEntitySet<ProjectTask>(
      "ProjectTasks"));
    builder.EntityType<ProjectTask>().ContainsMany(pt => pt.ProjectTaskAssignments);

```
and some repository behind.
### Expected result

Calling action returns actual tasks after update.
### Actual result
Client logs:
```
* Received 1031 B chunk
* Received 190 B chunk
* transfer closed with outstanding read data remaining
* Closing connection 20
```
Server has an unhandled exception (outside of aspnet.core error handling in pipeline) that can be seen in events hub of studio:
```
Microsoft.OData.ODataException: The Path property in ODataMessageWriterSettings.ODataUri must be set when writing contained elements.
   at Microsoft.OData.ODataWriterCore.EnterScope(WriterState newState, ODataItem item)
   at Microsoft.OData.ODataWriterCore.WriteStartNestedResourceInfoImplementation(ODataNestedResourceInfo nestedResourceInfo)
   at Microsoft.OData.ODataWriterCore.WriteStart(ODataNestedResourceInfo nestedResourceInfo)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.WriteNavigationLinks(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.WriteResource(Object graph, ODataWriter writer, ODataSerializerContext writeContext, IEdmTypeReference expectedType)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.WriteObjectInline(Object graph, IEdmTypeReference expectedType, ODataWriter writer, ODataSerializerContext writeContext)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSetSerializer.WriteResourceSet(IEnumerable enumerable, IEdmTypeReference resourceSetType, ODataWriter writer, ODataSerializerContext writeContext)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSetSerializer.WriteObjectInline(Object graph, IEdmTypeReference expectedType, ODataWriter writer, ODataSerializerContext writeContext)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSetSerializer.WriteObject(Object graph, Type type, ODataMessageWriter messageWriter, ODataSerializerContext writeContext)
   at Microsoft.AspNet.OData.Formatter.ODataOutputFormatterHelper.WriteToStream(Type type, Object value, IEdmModel model, ODataVersion version, Uri baseAddress, MediaTypeHeaderValue contentType, IWebApiUrlHelper internaUrlHelper, IWebApiRequestMessage internalRequest, IWebApiHeaders internalRequestHeaders, Func`2 getODataMessageWrapper, Func`2 getEdmTypeSerializer, Func`2 getODataPayloadSerializer, Func`1 getODataSerializerContext)
   at Microsoft.AspNet.OData.Formatter.ODataOutputFormatter.WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding selectedEncoding)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeResultAsync>g__Logged|21_0(ResourceInvoker invoker, IActionResult result)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResultFilterAsync>g__Awaited|29_0[TFilter,TFilterAsync](ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResultExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext[TFilter,TFilterAsync](State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultFilters()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|24_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeFilterPipelineAsync>g__Awaited|19_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Logged|17_1(ResourceInvoker invoker)
   at Microsoft.AspNetCore.Builder.RouterMiddleware.Invoke(HttpContext httpContext)
```

### Additional detail

Can supply info and check things if needed, but after I sleep )))

xuzhg:
…eton

### Issues

*This pull request fixes issue ##2056.*

### Description

* If Operation returns "entity set". we should set the "Path" on "WriterSetting" for that "entity set".

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Net.Http.Headers;
 using System.Runtime.Serialization;
@@ -379,6 +380,8 @@ namespace Microsoft.AspNet.OData.Formatter
 
         private static Microsoft.OData.UriParser.ODataPath GeneratePath(IEdmNavigationSource navigationSource)
         {
+            Contract.Assert(navigationSource != null);
+
             switch (navigationSource.NavigationSourceKind())
             {
                 case EdmNavigationSourceKind.EntitySet:
@@ -389,14 +392,11 @@ namespace Microsoft.AspNet.OData.Formatter
 
                 case EdmNavigationSourceKind.ContainedEntitySet:
                     IEdmContainedEntitySet containedEntitySet = (IEdmContainedEntitySet)navigationSource;
-                    var path = GeneratePath(containedEntitySet.ParentNavigationSource);
+                    Microsoft.OData.UriParser.ODataPath path = GeneratePath(containedEntitySet.ParentNavigationSource);
                     IList<ODataPathSegment> segments = new List<ODataPathSegment>();
-                    if (path != null)
+                    foreach (var item in path)
                     {
-                        foreach (var item in path)
-                        {
-                            segments.Add(item);
-                        }
+                        segments.Add(item);
                     }
 
                     segments.Add(new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource));
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
@@ -174,10 +174,9 @@ namespace Microsoft.AspNet.OData.Formatter
             {
                 ServiceRoot = baseAddress,
 
-                // TODO: 1604 Convert webapi.odata's ODataPath to ODL's ODataPath, or use ODL's ODataPath.
                 SelectAndExpand = internalRequest.Context.ProcessedSelectExpandClause,
                 Apply = internalRequest.Context.ApplyClause,
-                Path = (path == null || IsOperationPath(path)) ? null : path.Path,
+                Path = ConvertPath(path),
             };
 
             ODataMetadataLevel metadataLevel = ODataMetadataLevel.MinimalMetadata;
@@ -348,6 +347,66 @@ namespace Microsoft.AspNet.OData.Formatter
             }
 
             return false;
+        }
+
+        private static Microsoft.OData.UriParser.ODataPath ConvertPath(ODataPath path)
+        {
+            if (path == null)
+            {
+                return null;
+            }
+
+            if (IsOperationPath(path))
+            {
+                var lastSegment = path.Segments.Last();
+                OperationSegment operation = lastSegment as OperationSegment;
+                if (operation != null && operation.EntitySet != null)
+                {
+                    return GeneratePath(operation.EntitySet);
+                }
+
+                OperationImportSegment operationImport = lastSegment as OperationImportSegment;
+                if (operationImport != null && operationImport.EntitySet != null)
+                {
+                    return GeneratePath(operationImport.EntitySet);
+                }
+
+                return null;
+            }
+
+            return path.Path;
+        }
+
+        private static Microsoft.OData.UriParser.ODataPath GeneratePath(IEdmNavigationSource navigationSource)
+        {
+            switch (navigationSource.NavigationSourceKind())
+            {
+                case EdmNavigationSourceKind.EntitySet:
+                    return new Microsoft.OData.UriParser.ODataPath(new EntitySetSegment((IEdmEntitySet)navigationSource));
+
+                case EdmNavigationSourceKind.Singleton:
+                    return new Microsoft.OData.UriParser.ODataPath(new SingletonSegment((IEdmSingleton)navigationSource));
+
+                case EdmNavigationSourceKind.ContainedEntitySet:
+                    IEdmContainedEntitySet containedEntitySet = (IEdmContainedEntitySet)navigationSource;
+                    var path = GeneratePath(containedEntitySet.ParentNavigationSource);
+                    IList<ODataPathSegment> segments = new List<ODataPathSegment>();
+                    if (path != null)
+                    {
+                        foreach (var item in path)
+                        {
+                            segments.Add(item);
+                        }
+                    }
+
+                    segments.Add(new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource));
+                    return new Microsoft.OData.UriParser.ODataPath(segments);
+
+                case EdmNavigationSourceKind.None:
+                case EdmNavigationSourceKind.UnknownEntitySet:
+                default:
+                    return null;
+            }
         }
     }
 }
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BoundOperation/BoundOperationContextUriTest.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BoundOperation/BoundOperationContextUriTest.cs
@@ -1,0 +1,224 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNet.OData;
+using Microsoft.AspNet.OData.Builder;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.OData.Edm;
+using Microsoft.Test.E2E.AspNet.OData.Common.Controllers;
+using Microsoft.Test.E2E.AspNet.OData.Common.Execution;
+using Xunit;
+
+namespace Microsoft.Test.E2E.AspNet.OData.BoundOperation
+{
+    public class BoundOperationContextUriTest : WebHostTestBase
+    {
+        public BoundOperationContextUriTest(WebHostTestFixture fixture)
+            : base(fixture)
+        {
+        }
+
+        protected override void UpdateConfiguration(WebRouteConfiguration configuration)
+        {
+            var controllers = new[] { typeof(ProjectsController) };
+            configuration.AddControllers(controllers);
+            configuration.Routes.Clear();
+
+            // for Operation context Uri testing
+            configuration.MapODataServiceRoute("odata", "odata", GetEdmModel());
+        }
+
+        [Fact]
+        public async Task BoundFunctionWithNonContainmentReturnTypeWorksWithCorrectContextUri()
+        {
+            // Arrange
+            var requestUri = string.Format("{0}/odata/Projects/Default.GetProjectTask", BaseAddress);
+            var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+
+            // Act
+            HttpResponseMessage response = await Client.SendAsync(request);
+            string responseString = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            response.EnsureSuccessStatusCode();
+            Assert.Contains("odata/$metadata#ProjectTasks/$entity", responseString);
+        }
+
+        [Fact]
+        public async Task BoundFunctionWithContainmentReturnTypeWorksWithCorrectContextUri()
+        {
+            // Arrange
+            var requestUri = string.Format("{0}/odata/Projects/Default.GetAssigments", BaseAddress);
+            var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+
+            // Act
+            HttpResponseMessage response = await Client.SendAsync(request);
+            string responseString = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            response.EnsureSuccessStatusCode();
+            Assert.Contains("odata/$metadata#ProjectTasks/Assignments", responseString);
+        }
+
+        [Fact]
+        public async Task BoundActionWorksWithNonContainmentReturnTypeWorksWithCorrectContextUri()
+        {
+            // Arrange
+            var requestUri = string.Format("{0}/odata/Projects/Default.UpdateProjectTasks", BaseAddress);
+            var request = new HttpRequestMessage(HttpMethod.Post, requestUri);
+
+            // Act
+            HttpResponseMessage response = await Client.SendAsync(request);
+            string responseString = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            response.EnsureSuccessStatusCode();
+
+            Assert.Contains("odata/$metadata#ProjectTasks", responseString);
+        }
+
+        [Fact]
+        public async Task BoundActionWithContainmentReturnTypeWorksWithCorrectContextUri()
+        {
+            // Arrange
+            var requestUri = string.Format("{0}/odata/Projects/Default.UpdateAssignment", BaseAddress);
+            var request = new HttpRequestMessage(HttpMethod.Post, requestUri);
+
+            // Act
+            HttpResponseMessage response = await Client.SendAsync(request);
+            string responseString = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            response.EnsureSuccessStatusCode();
+
+            Assert.Contains("odata/$metadata#ProjectTasks/Assignment/$entity", responseString);
+        }
+
+        private static IEdmModel GetEdmModel()
+        {
+            ODataConventionModelBuilder builder = new ODataConventionModelBuilder();
+            builder.EntitySet<ProjectTask>("ProjectTasks");
+            builder.EntityType<ProjectTaskAssignment>();
+            var projectsConfig = builder.EntitySet<Project>("Projects").EntityType;
+
+            // Action returns "non-Containment"
+            var updateAction = projectsConfig.Collection.Action("UpdateProjectTasks");
+            updateAction.CollectionEntityParameter<ProjectTask>("projectTasks");
+            updateAction.ReturnsCollectionFromEntitySet<ProjectTask>("ProjectTasks");
+
+            // Function returns "non-Containment"
+            var getFunction = projectsConfig.Collection.Function("GetProjectTask");
+            getFunction.ReturnsFromEntitySet<ProjectTask>("ProjectTasks");
+
+            // Action returns "Containment"
+            var action = projectsConfig.Collection.Action("UpdateAssignment");
+            action.ReturnsEntityViaEntitySetPath<ProjectTaskAssignment>("bindingParameter/Task/Assignment");
+
+            // Function returns "Containment"
+            var function = projectsConfig.Collection.Function("GetAssigments");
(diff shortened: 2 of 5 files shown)


habbes:
@xuzhg could you add test cases for this feature? Could you also add a brief description to this PR explaining your approach and why it fixes the issue?

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -349,5 +348,52 @@ private static bool IsOperationPath(ODataPath path)
...
+                    return GeneratePath(operationImport.EntitySet);
+                }
+
+                return null;
+            }
+
+            return path.Path;
+        }
+
+        private static Microsoft.OData.UriParser.ODataPath GeneratePath(IEdmEntitySetBase entitySetBase)
```
KanishManuja-MS: Why not use using?

xuzhg: Both ODL and Web API defines `ODataPath` class. Here, we use "ODL ODataPath" class. 
If use using, it'll be ambiguous . #Resolved.

Sreejithpin: Minor: Can move it to a Factory implementation.

samples/AspNetCore3xODataSample.Web/Controllers/CustomersController.cs
```diff
@@ -10,6 +10,49 @@
...
+         * */
+        [HttpPost]
+        public IActionResult UpdateProjectTasks([FromODataUri] string key,
+            [FromBody] ODataActionParameters parameters)
+        {
+            var projectTasks = ((IEnumerable<ProjectTask>)parameters["projectTasks"]).ToList();
+
+            var tasks = new ProjectTask[]
+            {
+            new ProjectTask { Id = "1", ProjectId = "11",
```
habbes: Maybe indent the elements of the array.

samples/AspNetCore3xODataSample.Web/Controllers/CustomersController.cs
```diff
@@ -10,6 +10,49 @@
...
+                ProjectTaskAssignments = new ProjectTaskAssignment[]
+                {
+                    new ProjectTaskAssignment { Id = "22", ProjectTaskId = "1" }
```
mikepizzo: > Id = "22", ProjectTaskId = "1"  [](start = 47, length = 32)

shouldn't this (and line 47) be using projectTasks from the payload? Currently the payload is read but not used.

samples/AspNetCore3xODataSample.Web/Controllers/CustomersController.cs
```diff
@@ -10,6 +10,49 @@
...
+                }
+            },
+
+            new ProjectTask { Id = "2", ProjectId = "22",
+                ProjectTaskAssignments = new ProjectTaskAssignment[]
+                {
+                    new ProjectTaskAssignment { Id = "22", ProjectTaskId = "2" }
+                }
+            },
```
mikepizzo: shouldn't this use the projectTasks passed in the payload?

mikepizzo:
+1 -- please add a test case

---
In reply to: [589150185](https://github.com/OData/WebApi/pull/2060#issuecomment-589150185) [](ancestors = 589150185)

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -349,5 +348,52 @@ private static bool IsOperationPath(ODataPath path)
...
+            switch (entitySetBase.NavigationSourceKind())
+            {
+                case EdmNavigationSourceKind.EntitySet:
+                    return new Microsoft.OData.UriParser.ODataPath(new EntitySetSegment((IEdmEntitySet)entitySetBase));
+
+                case EdmNavigationSourceKind.Singleton:
+                    return new Microsoft.OData.UriParser.ODataPath(new SingletonSegment((IEdmSingleton)entitySetBase));
+
+                case EdmNavigationSourceKind.ContainedEntitySet:
+                    // TODO: for containment, do we have the scenario to return a containment in an operation?
```
mikepizzo: Yes; this should be supported.  This may seem like an edge case, but for services like Microsoft Graph that organize different sources of data through a singleton navigation property it is actually the common scenario.

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
xuzhg: This is only used in this class once. So, it's better to keep it private in this helper class.

---
In reply to: [384850926](https://github.com/OData/WebApi/pull/2060#discussion_r384850926) [](ancestors = 384850926)

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
xuzhg: I add codes to process this case. Please take a look

---
In reply to: [385351278](https://github.com/OData/WebApi/pull/2060#discussion_r385351278) [](ancestors = 385351278)

xuzhg:
Added E2E test cases. Please take a look.

---
In reply to: [592155477](https://github.com/OData/WebApi/pull/2060#issuecomment-592155477) [](ancestors = 592155477,589150185)

xuzhg:
@mikepizzo The CI at https://identitydivision.visualstudio.com/OData/_build/results?buildId=481545&view=results is green. Thanks.

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -349,5 +348,65 @@ private static bool IsOperationPath(ODataPath path)
...
+            {
+                case EdmNavigationSourceKind.EntitySet:
+                    return new Microsoft.OData.UriParser.ODataPath(new EntitySetSegment((IEdmEntitySet)navigationSource));
+
+                case EdmNavigationSourceKind.Singleton:
+                    return new Microsoft.OData.UriParser.ODataPath(new SingletonSegment((IEdmSingleton)navigationSource));
+
+                case EdmNavigationSourceKind.ContainedEntitySet:
+                    IEdmContainedEntitySet containedEntitySet = (IEdmContainedEntitySet)navigationSource;
+                    var path = GeneratePath(containedEntitySet.ParentNavigationSource);
```
mikepizzo: >var [](start = 20, length = 3)

better not to use var

mikepizzo: >ParentNavigationSource [](start = 63, length = 22)

any chance that this could be null?

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -349,5 +348,65 @@ private static bool IsOperationPath(ODataPath path)
...
+                    IList<ODataPathSegment> segments = new List<ODataPathSegment>();
+                    if (path != null)
+                    {
+                        foreach (var item in path)
+                        {
+                            segments.Add(item);
+                        }
+                    }
+
+                    segments.Add(new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource));
```
mikepizzo: is this the right behavior for path == null?  if path==null, shouldn't you just return null?

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
xuzhg: Yes. I know. I originally thought it's easy to use "var" not "Microsoft.OData.UriParser.ODataPath". But I changed.

---
In reply to: [386706993](https://github.com/OData/WebApi/pull/2060#discussion_r386706993) [](ancestors = 386706993)

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
xuzhg: Emm... probably.  in ODL, IEdmContainedEntitySet interface is public, but there's only one implementation "EdmContainedEntitySet" is internal. That means user can't create a "EdmContainedEntitySet" instance. If you can get an instance of "IEdmContainedEntitySet", the ParentNavigationSource should be valid. So, i'd like to remove the conditional verification. 

---
In reply to: [386708095](https://github.com/OData/WebApi/pull/2060#discussion_r386708095) [](ancestors = 386708095)

mikepizzo: This method looks similar to the ComputePath() method in EdmContainedEntitySet, which is unfortunately internal.  However, that method deals with nav props defined on derived types. Do we need to do anything to handle that case here?

Also a bit concerned that EdmContainedEntitySet allows passing in a navigationPath, which is used in ComputePath. What are the situations in which that would be something other than the navigationProperty? Would this be used for nav props on complex types? Do we need to handle that case here?

---
In reply to: [386750791](https://github.com/OData/WebApi/pull/2060#discussion_r386750791) [](ancestors = 386750791,386708095)

mikepizzo: Consider adding an assert that containedEntitySet.ParentNavigationSource != null.

---
In reply to: [386750791](https://github.com/OData/WebApi/pull/2060#discussion_r386750791) [](ancestors = 386750791,386708095)

xuzhg: EdmContainedEntitySet constructor makes the input check. That can make sure "ParentNavigationSource" never be null if the "EntityContainedEntitySet" is live.

public EdmContainedEntitySet(IEdmNavigationSource parentNavigationSource, IEdmNavigationProperty navigationProperty, IEdmPathExpression navigationPath)
            : base(navigationProperty.Name, navigationProperty.ToEntityType())
        {
            EdmUtil.CheckArgumentNull(parentNavigationSource, "parentNavigationSource");
            ...

            this.parentNavigationSource = parentNavigationSource;
            this.navigationProperty = navigationProperty;
            this.navigationPath = navigationPath;
        }


---
In reply to: [391139997](https://github.com/OData/WebApi/pull/2060#discussion_r391139997) [](ancestors = 391139997,386750791,386708095)

xuzhg: Besides, for the  ComputePath() method in EdmContainedEntitySet, we can call " containedEntitySet.Path" to get the Path. However, "Path" is IEdmPathExpression, it's a string segment.
We want the ODataPath, it's a path segments. So, we have to use "OData Uri parse" to parse the "IEdmPathExpression". That's not feasible here. 
I'd like to create a new issue to track on this feature request.  And I do think that's a feature in the ODL . 

@mikep

---
In reply to: [394725483](https://github.com/OData/WebApi/pull/2060#discussion_r394725483) [](ancestors = 394725483,391139997,386750791,386708095)

mikepizzo: @saxu@microsoft.com -- did you create an issue to track this?  Please specifically call out what is not currently supported (i.e., how does this code handle navigation properties on complex types or navigation properties defined on derived types? does it work with navigation properties on entities returned from contained navigation properties?)

---
In reply to: [394726333](https://github.com/OData/WebApi/pull/2060#discussion_r394726333) [](ancestors = 394726333,394725483,391139997,386750791,386708095)

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -349,5 +348,62 @@ private static bool IsOperationPath(ODataPath path)
...
+
+                return null;
+            }
+
+            return path.Path;
+        }
+
+        private static Microsoft.OData.UriParser.ODataPath GeneratePath(IEdmNavigationSource navigationSource)
+        {
+            switch (navigationSource.NavigationSourceKind())
```
mikepizzo, line 382: Maybe add assert that navigationSource != null

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -349,5 +348,62 @@ private static bool IsOperationPath(ODataPath path)
...
         }
+
+        private static Microsoft.OData.UriParser.ODataPath ConvertPath(ODataPath path)
+        {
+            if (path == null)
+            {
+                return null;
+            }
+
+            if (IsOperationPath(path))
```
mikepizzo, line 359: What is the value in doing this check? It's going to walk the path looking for an operationSegment or operationImportSegment, and if it returns true, we're going to go to the last segment and check to see if it's an operationSegment or operationImportSegment.  Can we remove the IsOperationPath(path) check and just rely on the existing checks on line 363 and 369?  The only difference I can see in doing the check first is that the current logic will return a null if there is an operation (i.e., a composible function) in the path but not as the last segment -- is that desired behavior?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BoundOperation/BoundOperationContextUriTest.cs
```diff
@@ -0,0 +1,224 @@
...
+            // Assert
+            response.EnsureSuccessStatusCode();
+            Assert.Contains("odata/$metadata#ProjectTasks/$entity", responseString);
+        }
+
+        [Fact]
+        public async Task BoundFunctionWithContainmentReturnTypeWorksWithCorrectContextUri()
+        {
+            // Arrange
+            var requestUri = string.Format("{0}/odata/Projects/Default.GetAssigments", BaseAddress);
```
mikepizzo, line 54: how about adding a segment after the bound function call, i.e.: 
-Projects/Default.GetAssignments/111
-Projects/Default.GetAssignments/Default.ProjectTaskAssignment

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BoundOperation/BoundOperationContextUriTest.cs
```diff
@@ -0,0 +1,224 @@
...
+
+            // for Operation context Uri testing
+            configuration.MapODataServiceRoute("odata", "odata", GetEdmModel());
+        }
+
+        [Fact]
+        public async Task BoundFunctionWithNonContainmentReturnTypeWorksWithCorrectContextUri()
+        {
+            // Arrange
+            var requestUri = string.Format("{0}/odata/Projects/Default.GetProjectTask", BaseAddress);
```
mikepizzo, line 38: >            var requestUri = string.Format("{0}/odata/Projects/Default.GetProjectTask" [](start = 0, length = 86)

how about adding a segment after the bound function call, i.e.: 
-Projects/Default.GetProjectTask/1
-Projects/Default.GetDefaultProjectTask/Default.ProjectTask

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
xuzhg: I'd like to keep the behavior unchanged. if we consider the compossible function, the logic here is not enough to cover it.  

---
In reply to: [401841900](https://github.com/OData/WebApi/pull/2060#discussion_r401841900) [](ancestors = 401841900)

mikepizzo:
Still have a couple of small comments, but nothing to block merging.

mikepizzo (approved):
:shipit:

xuzhg:
Merged at https://github.com/OData/WebApi/commit/c79634a93d2efe07528a4c11a03bf6ebc2118782

Issue #2055, opened 2020-02-17
Batch request copies content headers to embedded requests without content


habbes:
When a batch request is received, GET and DELETE requests inside the batch get the content-type and content-length headers of the overall batch request even though they don't have content.

In general, the batch request headers are copied to the individual requests, but if an individual request inside the batch defines the same headers, then they will replace the inherited ones. Since batch requests are POST requests with a body, they always have a content-length and content-type, and since GET and DELETE requests don't have content, and therefore no content-length or content-type headers, they'll keep the inherited ones.

### Assemblies affected

OData WebAPI 7.3.0

### Reproduce steps

Create an OData WebAPI app with a basic model and convention routing. Add a controller with a `Delete(int key)` method.

Inside the method add break points so you can inspect the headers, or add the following print statements:
```c#
Console.WriteLine(ControllerContext.HttpContext.Request.Headers.ContentLength);
Console.WriteLine(ControllerContext.HttpContext.ContentType);
```

### Expected result

ContentLength and ContentType should be null.

### Actual result

ContentLength is set to the size of the entire batch request body.

ContentType is set to the content type of the batch request (e.g. `multipart/mixed; boundary=batch_e22ea856-3770-4803-b579-3cfccc999775`)

### Additional detail

[The ReadOperationInternalAsync method of the ODataBatchReaderExtensions class](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs#L141-L147) creates copies headers from an individual request in the batch to replace the headers inherited from the overall batch request, but this doesn't remove undesired headers inherited from the batch.

mikepizzo:
CreateHttpContext in ODataBatchReaderExtensions should not copy over headers that are listed in http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_HeaderFields that don't specifically say that they are inherited. We should continue to copy headers not in this list.  Specifically, we should NOT copy over:
Content-Length
Content-Type (we *could* copy this over as the default if it is not multipart/mixed (i.e., json) but it SHOULD be specified on the individual request)
continue-on-error preference
respond-async

habbes:
Thanks for the clarifications @mikepizzo.
[...]

PR #2064, opened 2020-02-24
Batch request headers copied to individual requests

habbes:
### Issues

*This pull request fixes issue #2055. *

The batch request handler in AspNetCore WebApi copies headers from the overall batch request to each individual request inside the batch, unless the individual request specify the same headers, in which case they replace the inherited ones.
However, not all headers should be copied, in particular: `Content-Length`, `Content-Type`, some preferences from the `Prefer` header such as `respond-async`, `continue-on-error` and `odata.continue-on-error`.

### Description

I've added some restrictions on how headers are copied from a batch request to individual requests inside the batch:
- `Content-Length` and `Content-Type` headers are no longer copied to individual requests
- I've implemented the following rules for the `Prefer` header:
  - Preferences from the `Prefer` header are copied to individual requests except for `respond-async`, `continue-on-error` and `odata.continue-on-error` preferences.
  - If the individual request already had a `Prefer` header, then the individual request will end up with its own preferences merged with those inherited from the batch request 
  - The individual request will not receive a preference from the batch that is already defined in the individual request

Here's a concrete example to demonstrate how preference headers from the batch and individual requests are merged:

If the overal batch request define the following prefer header:
```http
Prefer: allow-entityreferences, respond-async, include-annotations="display.*,foo", continue-on-error; wait="200;\"300"
```
And an individual request inside the batch specifies the following:
```http
Prefer: wait=100, handling=lenient
```
then the final preference header for the individual request will be:
```http
Prefer: wait=100, handling=lenient, allow-entityreferences; include-annotations="display.*,foo"
```

- [x] *Test cases added*

### Additional work necessary

This implementation has the following issues:
- In the case that the batch had defined multiple `Prefer` headers, only the first one is considered. How should this case be handled?

This PR implements the rules only for AspNetCore. AspNet does not copy headers from the batch to the individual requests.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Linq;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Common;
@@ -140,9 +142,21 @@ namespace Microsoft.AspNet.OData.Batch
 
             foreach (var header in batchRequest.Headers)
             {
-                // Copy headers from batch, overwriting any existing headers.
                 string headerName = header.Key;
                 string headerValue = header.Value;
+
+                if (headerName.Trim().ToLowerInvariant() == "prefer")
+                {
+                    // in the case of Prefer header, we don't want to overwrite,
+                    // instead we merge preferences defined in the individual request with those inherited from the batch
+                    request.Headers.TryGetValue(headerName, out StringValues batchPreferencesList);
+                    string batchPreferences = batchPreferencesList.FirstOrDefault();
+                    // TODO: what if multiple Prefer headers are defined on the request?
+                    request.Headers[headerName] = MergeIndividualAndBatchPreferences(headerValue, batchPreferences);
+                    continue;
+                }
+
+                // Copy headers from batch, overwriting any existing headers.
                 request.Headers[headerName] = headerValue;
             }
 
@@ -234,6 +248,23 @@ namespace Microsoft.AspNet.OData.Batch
             context.Request.Cookies = originalContext.Request.Cookies;
             foreach (KeyValuePair<string, StringValues> header in originalContext.Request.Headers)
             {
+                var headerKey = header.Key.ToLowerInvariant();
+                // do not copy over headers that should not be inherited from batch to individual requests
+                if (headerKey == "content-length" ||
+                    headerKey == "content-type")
+                {
+                    continue;
+                }
+                // some preferences may be inherited, others discarded
+                if (headerKey == "prefer")
+                {
+                    var preferencesToInherit = GetPreferencesToInheritFromBatch(header.Value);
+                    if (!string.IsNullOrEmpty(preferencesToInherit))
+                    {
+                        context.Request.Headers.Add(header.Key, preferencesToInherit);
+                    }
+                    continue;
+                }
                 context.Request.Headers.Add(header);
             }
 
@@ -243,6 +274,56 @@ namespace Microsoft.AspNet.OData.Batch
             context.Response.Body = new ODataBatchStream();
 
             return context;
+        }
+
+        /// <summary>
+        /// Extract preferences that can be inherited from the overall batch request to
+        /// an individual request.
+        /// </summary>
+        /// <param name="batchPreferences">The value of the Prefer header from the batch request</param>
+        /// <returns>comma-separated preferences that can be passed down to an individual request</returns>
+        private static string GetPreferencesToInheritFromBatch(string batchPreferences)
+        {
+            // do not inherit respond-async and continue-on-error (odata.continue-on-error in OData 4.0)
+            var preferencesToIgnore = new string[] { "respond-async", "continue-on-error", "odata.continue-on-error" };
+            var preferencesToInherit = batchPreferences.Split(',')
+                .Where(value => 
+                !preferencesToIgnore.Any(
+                    prefToIgnore => value.Trim().ToLowerInvariant().StartsWith(prefToIgnore)))
+                .Select(value => value.Trim());
+            return string.Join(", ", preferencesToInherit);
+        }
+
+        /// <summary>
+        /// Merges the preferences from the batch request and an individual request inside the batch into one value.
+        /// If a given preference is defined in both the batch and individual request, the one from the individual
+        /// request is retained and the one from the batch is discarded.
+        /// </summary>
+        /// <param name="individualPreferences">The value of the Prefer header from the individual request inside the batch</param>
+        /// <param name="batchPreferences">The value of the Prefer header from the overall batch request</param>
+        /// <returns>Value containing the combined preferences</returns>
+        private static string MergeIndividualAndBatchPreferences(string individualPreferences, string batchPreferences)
+        {
+            if (string.IsNullOrEmpty(individualPreferences))
+            {
+                return batchPreferences;
+            }
+            if (string.IsNullOrEmpty(batchPreferences))
+            {
+                return individualPreferences;
+            }
+            var initialPreferences = new StringBuilder(individualPreferences);
+            // get the name of each preference to avoid adding duplicates from batch
+            var individualList = individualPreferences.Split(',').Select(pref => pref.Trim());
+            var individualPreferenceNames = new HashSet<string>(individualList.Select(pref => pref.Split('=').FirstOrDefault()));
+            
+            
+            var filteredBatchList = batchPreferences.Split(',').Select(pref => pref.Trim())
+                // do not add duplicate preferences from batch
+                .Where(pref => !individualPreferenceNames.Contains(pref.Split('=').FirstOrDefault()));
+            var filteredBatchPreferences = string.Join(", ", filteredBatchList);
+
+            return string.Join(", ", individualPreferences, filteredBatchPreferences);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -307,6 +307,7 @@ namespace Microsoft.AspNet.OData.Batch
             {
                 return batchPreferences;
             }
+
             if (string.IsNullOrEmpty(batchPreferences))
             {
                 return individualPreferences;
@@ -314,7 +315,7 @@ namespace Microsoft.AspNet.OData.Batch
             // get the name of each preference to avoid adding duplicates from batch
             IEnumerable<string> individualList = individualPreferences.SplitPreferences().Select(pref => pref.Trim());
             HashSet<string> individualPreferenceNames = new HashSet<string>(individualList.Select(pref => pref.Split('=').FirstOrDefault()));
-            
+
             
             IEnumerable<string> filteredBatchList = batchPreferences.SplitPreferences().Select(pref => pref.Trim())
                 // do not add duplicate preferences from batch
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -8,6 +8,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Common;
@@ -150,8 +151,8 @@ namespace Microsoft.AspNet.OData.Batch
                     // in the case of Prefer header, we don't want to overwrite,
                     // instead we merge preferences defined in the individual request with those inherited from the batch
                     request.Headers.TryGetValue(headerName, out StringValues batchPreferencesList);
-                    string batchPreferences = batchPreferencesList.FirstOrDefault();
                     // TODO: what if multiple Prefer headers are defined on the request?
+                    string batchPreferences = batchPreferencesList.FirstOrDefault();
                     request.Headers[headerName] = MergeIndividualAndBatchPreferences(headerValue, batchPreferences);
                     continue;
                 }
@@ -286,7 +287,7 @@ namespace Microsoft.AspNet.OData.Batch
         {
             // do not inherit respond-async and continue-on-error (odata.continue-on-error in OData 4.0)
             var preferencesToIgnore = new string[] { "respond-async", "continue-on-error", "odata.continue-on-error" };
-            var preferencesToInherit = batchPreferences.Split(',')
+            var preferencesToInherit = batchPreferences.SplitPreferences()
                 .Where(value => 
                 !preferencesToIgnore.Any(
                     prefToIgnore => value.Trim().ToLowerInvariant().StartsWith(prefToIgnore)))
@@ -312,18 +313,66 @@ namespace Microsoft.AspNet.OData.Batch
             {
                 return individualPreferences;
             }
-            var initialPreferences = new StringBuilder(individualPreferences);
             // get the name of each preference to avoid adding duplicates from batch
-            var individualList = individualPreferences.Split(',').Select(pref => pref.Trim());
+            var individualList = individualPreferences.SplitPreferences().Select(pref => pref.Trim());
             var individualPreferenceNames = new HashSet<string>(individualList.Select(pref => pref.Split('=').FirstOrDefault()));
             
             
-            var filteredBatchList = batchPreferences.Split(',').Select(pref => pref.Trim())
+            var filteredBatchList = batchPreferences.SplitPreferences().Select(pref => pref.Trim())
                 // do not add duplicate preferences from batch
                 .Where(pref => !individualPreferenceNames.Contains(pref.Split('=').FirstOrDefault()));
             var filteredBatchPreferences = string.Join(", ", filteredBatchList);
 
             return string.Join(", ", individualPreferences, filteredBatchPreferences);
+        }
+
+        /// <summary>
+        /// Splits the value of a Prefer header into separate preferences
+        /// e.g. a value like 'a, b=c, foo="bar,baz"' will return an IEnumerable with
+        /// - a
+        /// - b=c
+        /// - foo="bar,baz"
+        /// </summary>
+        /// <param name="preferences"></param>
+        /// <returns></returns>
+        private static IEnumerable<string> SplitPreferences(this string preferences)
+        {
+            StringBuilder currentPreference = new StringBuilder();
+            var insideQuotedValue = false;
+            char prevChar = '\0';
+            foreach (var c in preferences)
+            {
+                if (c == '"')
+                {
+                    if (!insideQuotedValue)
+                    {
+                        // we are starting a double-quoted value
+                        insideQuotedValue = true;
+                    }
+                    else
+                    {
+                        // this could be the end of a quoted value, or it could be an escaped quote
+                        insideQuotedValue = prevChar == '\\';
+                    }
+                }
+
+                if ((c == ',' || c == ';') && !insideQuotedValue)
+                {
+                    yield return currentPreference.ToString();
+                    currentPreference.Clear();
+                }
+                else
+                {
+                    currentPreference.Append(c);
+                }
+
+                prevChar = c;
+            }
+
+            if (currentPreference.Length > 0)
+            {
+                yield return currentPreference.ToString();
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -293,13 +293,12 @@ namespace Microsoft.AspNet.OData.Batch
         private static string GetPreferencesToInheritFromBatch(string batchPreferences)
         {
             IEnumerable<string> preferencesToInherit = SplitPreferences(batchPreferences)
-                .Where(value => 
+                .Where(value =>
                     !nonInheritablePreferences.Any(
-                        prefToIgnore => 
-                        value.Trim().ToLowerInvariant().StartsWith(prefToIgnore)
+                        prefToIgnore =>
+                        value.ToLowerInvariant().StartsWith(prefToIgnore)
                     )
-                )
-                .Select(value => value.Trim());
+                );
             return string.Join(", ", preferencesToInherit);
         }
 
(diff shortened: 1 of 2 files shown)


habbes:
I've updated the preference splitting algorithms to account for quoted values that may include a comma. I also added support for semicolon as a separator. But the output preferences will be combined using a comma regardless of the separator used in the initial header.

habbes:
I had tried using a regular expression to split preferences, but the regex became unwieldly, hard to read and tricky to get right, so I resorted to using a simple loop. But I think it would be better if I'd use an existing parser for this kind of syntax.

I considered using [TextFieldParser from Microsoft.VisualBasic](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualbasic.fileio.textfieldparser?view=netcore-3.1), but it does not support .netstandard 2.0

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -140,9 +143,21 @@ public static class ODataBatchReaderExtensions
 
             foreach (var header in batchRequest.Headers)
             {
-                // Copy headers from batch, overwriting any existing headers.
                 string headerName = header.Key;
                 string headerValue = header.Value;
+
+                if (headerName.Trim().ToLowerInvariant() == "prefer")
```
g2mula: Could invert the ```if``` to take care of the ```request.Headers[headerName] = headerValue;``` line then leave the rest of the code un-indented...

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -227,6 +242,23 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
             context.Request.Cookies = originalContext.Request.Cookies;
             foreach (KeyValuePair<string, StringValues> header in originalContext.Request.Headers)
             {
+                var headerKey = header.Key.ToLowerInvariant();
```
g2mula: The style in this file/project seems to prefer ```type``` over ```var```, need to be consistent. If it were a new class then the ```var``` could be internally consistent in that file but here will need to adhere to existing standards.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -227,6 +242,23 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+                // do not copy over headers that should not be inherited from batch to individual requests
+                if (headerKey == "content-length" ||
+                    headerKey == "content-type")
+                {
+                    continue;
+                }
+                // some preferences may be inherited, others discarded
+                if (headerKey == "prefer")
```
g2mula: Could invert the ```if``` to take care of the one liner and leave the other code indented less...

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,103 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        /// </summary>
+        /// <param name="batchPreferences">The value of the Prefer header from the batch request</param>
+        /// <returns>comma-separated preferences that can be passed down to an individual request</returns>
+        private static string GetPreferencesToInheritFromBatch(string batchPreferences)
+        {
+            // do not inherit respond-async and continue-on-error (odata.continue-on-error in OData 4.0)
+            var preferencesToIgnore = new string[] { "respond-async", "continue-on-error", "odata.continue-on-error" };
+            var preferencesToInherit = batchPreferences.SplitPreferences()
+                .Where(value => 
+                !preferencesToIgnore.Any(
```
g2mula: May want to add indentation by one level to make clearer that it is part of the where clause...

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,103 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        /// </summary>
+        /// <param name="individualPreferences">The value of the Prefer header from the individual request inside the batch</param>
+        /// <param name="batchPreferences">The value of the Prefer header from the overall batch request</param>
+        /// <returns>Value containing the combined preferences</returns>
+        private static string MergeIndividualAndBatchPreferences(string individualPreferences, string batchPreferences)
+        {
+            if (string.IsNullOrEmpty(individualPreferences))
+            {
+                return batchPreferences;
+            }
```
g2mula:
```suggestion
            }
            
``` Newline after closing bracket...

File: src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs [hunk 6]
```diff
@@ -236,5 +268,103 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+            if (string.IsNullOrEmpty(batchPreferences))
+            {
+                return individualPreferences;
+            }
+            // get the name of each preference to avoid adding duplicates from batch
+            var individualList = individualPreferences.SplitPreferences().Select(pref => pref.Trim());
+            var individualPreferenceNames = new HashSet<string>(individualList.Select(pref => pref.Split('=').FirstOrDefault()));
+            
+
```
g2mula:
```suggestion
```
Extra newline.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
```diff
@@ -506,6 +542,182 @@ public async Task SendAsync_Works_ForBatchRequestWithInsertedEntityReferencedInA
 
             Assert.NotNull(customer.Orders?.SingleOrDefault(d => d.Id.Equals(orderId)));
         }
+
+        public static TheoryDataSet<Dictionary<string, string>, (string, string, string)> BatchHeadersTestData
```
g2mula: Is this only accessed once? It should probably be set as a (static) readonly field returned from this getter... as it is currently, youu be rebuilding this on each access...

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -140,9 +144,21 @@ public static class ODataBatchReaderExtensions
...
             {
-                // Copy headers from batch, overwriting any existing headers.
                 string headerName = header.Key;
                 string headerValue = header.Value;
+
+                if (headerName.Trim().ToLowerInvariant() == "prefer")
+                {
+                    // in the case of Prefer header, we don't want to overwrite,
+                    // instead we merge preferences defined in the individual request with those inherited from the batch
+                    request.Headers.TryGetValue(headerName, out StringValues batchPreferencesList);
```
mikepizzo: Why do you need to do a TryGetValue, can't you just use the headerValue that you already have?

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -140,9 +144,21 @@ public static class ODataBatchReaderExtensions
...
+                    // TODO: what if multiple Prefer headers are defined on the request?
```
mikepizzo: >what if multiple Prefer headers are defined on the request [](start = 29, length = 58)

proper behavior would be to merge them, giving preference to the first one specified

From [https://tools.ietf.org/html/rfc7240](https://tools.ietf.org/html/rfc7240) :
" To avoid any possible ambiguity, individual preference tokens SHOULD
   NOT appear multiple times within a single request.  If any preference
   is specified more than once, only the first instance is to be
   considered.  All subsequent occurrences SHOULD be ignored without
   signaling an error or otherwise altering the processing of the
   request.  This is the only case in which the ordering of preferences
   within a request is considered to be significant."

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
mikepizzo: Coding guidelines are to use the actual type, not var.  Existing WebAPI doesn't adhere to this very well, but it's still good practice.

---
In reply to: [383750782](https://github.com/OData/WebApi/pull/2064#discussion_r383750782) [](ancestors = 383750782)

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,103 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+            var individualList = individualPreferences.SplitPreferences().Select(pref => pref.Trim());
+            var individualPreferenceNames = new HashSet<string>(individualList.Select(pref => pref.Split('=').FirstOrDefault()));
+            
+            
+            var filteredBatchList = batchPreferences.SplitPreferences().Select(pref => pref.Trim())
+                // do not add duplicate preferences from batch
+                .Where(pref => !individualPreferenceNames.Contains(pref.Split('=').FirstOrDefault()));
+            var filteredBatchPreferences = string.Join(", ", filteredBatchList);
+
+            return string.Join(", ", individualPreferences, filteredBatchPreferences);
```
mikepizzo: >, [](start = 32, length = 1)

shouldn't this be ";"?

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,103 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+            }
+            // get the name of each preference to avoid adding duplicates from batch
+            var individualList = individualPreferences.SplitPreferences().Select(pref => pref.Trim());
+            var individualPreferenceNames = new HashSet<string>(individualList.Select(pref => pref.Split('=').FirstOrDefault()));
+            
+            
+            var filteredBatchList = batchPreferences.SplitPreferences().Select(pref => pref.Trim())
+                // do not add duplicate preferences from batch
+                .Where(pref => !individualPreferenceNames.Contains(pref.Split('=').FirstOrDefault()));
+            var filteredBatchPreferences = string.Join(", ", filteredBatchList);
```
mikepizzo: >, [](start = 56, length = 1)

shouldn't this be ";"?

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,103 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+                        insideQuotedValue = true;
+                    }
+                    else
+                    {
+                        // this could be the end of a quoted value, or it could be an escaped quote
+                        insideQuotedValue = prevChar == '\\';
+                    }
+                }
+
+                if ((c == ',' || c == ';') && !insideQuotedValue)
```
mikepizzo: >(c == ',' || c == ';' [](start = 20, length = 21)

should just be ;?

habbes: You're right. I misinterpreted the grammar, the semicolon is just used to separate parameters of a given preference, not to separate preferences. I will correct that.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
habbes: In this case, since the prefer header is sort of an edge case, I think it makes sense to handle the edge case, rather than the default case, in the if block.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
mikepizzo: I also had to re-read; I initially thought that semi-colon was the separator and was wondering at the use of comma, until I looked more closely and realized it was the other way around.

---
In reply to: [384954431](https://github.com/OData/WebApi/pull/2064#discussion_r384954431) [](ancestors = 384954431)

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
habbes: I found that the batch reader throws an exception if an individual request includes a duplicate header (e.g. if it assigns the Prefer header twice), so that made things easier.

But I have accounted for the case where the batch requests happens to have multiple occurrences of the Prefer header and multiple occurrences of individual preferences. The `StringValues` returned when you access a header in `request.Headers` can already correctly combine the values from multiple occurrences of the header into one string. So all I did was just make sure that duplicate occurrences of individual preferences are ignored.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -140,9 +146,19 @@ public static class ODataBatchReaderExtensions
...
                 string headerValue = header.Value;
+
+                if (headerName.Trim().ToLowerInvariant() == "prefer")
+                {
+                    // in the case of Prefer header, we don't want to overwrite,
+                    // instead we merge preferences defined in the individual request with those inherited from the batch
+                    request.Headers.TryGetValue(headerName, out StringValues batchReferences);
+                    request.Headers[headerName] = MergeIndividualAndBatchPreferences(headerValue, batchReferences);
+                    continue;
+                }
```
gathogojr: @habbes Out of curiosity, are the headers and preferences finite and known and do we know for sure that only headers: `content-length` and `content-type` and preferences: `respond-async`, `continue-on-error` and `odata.continue-on-error` are not inheritable from parent (or not applicable to nested requests?

habbes: These are what we decided not to inherit following a discussion in the issue that this PR addressed. Whether the possible headers and preferences are finite, I'm not sure what you mean by that. It's possible that users can define custom headers or preferences, or that new headers or preferences may be introduced into the spec over time. I don't know whether this is what you mean by finite.

gathogojr: @habbes If user can define headers and preferences, the question would be whether the nested requests inheriting those from the batch is okay. It'd be hard to know though. The outcome of inheriting or not inheriting the custom headers and preferences can swing one way or the other

habbes: @gathogojr yes it would be hard. But if they are custom headers, then maybe it should be up to the user to decide to place them on the individual requests that they apply to if they don't want them to be copied to all other requests in the batch.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -140,9 +146,19 @@ public static class ODataBatchReaderExtensions
...
+
+                // Copy headers from batch, overwriting any existing headers.
                 request.Headers[headerName] = headerValue;
```
mikepizzo: I think it's a little easier to read if you remove the continue on line 158 and put line 162 in an else.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -227,6 +243,22 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+                if (headerKey == "prefer")
+                {
+                    string preferencesToInherit = GetPreferencesToInheritFromBatch(header.Value);
+                    if (!string.IsNullOrEmpty(preferencesToInherit))
+                    {
+                        context.Request.Headers.Add(header.Key, preferencesToInherit);
+                    }
+                    continue;
+                }
                 context.Request.Headers.Add(header);
```
mikepizzo: I think it's a bit easier to read without the continues:

            if (!nonInheritableHeaders.Contains(headerKey))
            {
                // some preferences may be inherited, others discarded
                if (headerKey == "prefer")
                {
                    string preferencesToInherit = GetPreferencesToInheritFromBatch(header.Value);
                    if (!string.IsNullOrEmpty(preferencesToInherit))
                    {
                        context.Request.Headers.Add(header.Key, preferencesToInherit);
                    }
                }
                else
                {
                    context.Request.Headers.Add(header);
                }
            }

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,110 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        /// <summary>
+        /// Splits the value of a Prefer header into separate preferences
+        /// e.g. a value like 'a, b=c, foo="bar,baz"' will return an IEnumerable with
+        /// - a
+        /// - b=c
+        /// - foo="bar,baz"
+        /// </summary>
+        /// <param name="preferences"></param>
+        /// <returns></returns>
+        private static IEnumerable<string> SplitPreferences(this string preferences)
```
mikepizzo: I wouldn't do this as an extension method (especially since it would then apply to all strings).  Instead, just make it a static method that takes a string parameter:
private static IEnumerable<string> SplitPreferences(string preferences) {...}

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,110 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        {
+            if (string.IsNullOrEmpty(individualPreferences))
+            {
+                return batchPreferences;
+            }
+
+            if (string.IsNullOrEmpty(batchPreferences))
+            {
+                return individualPreferences;
+            }
```
mikepizzo: generally, insert blank line after closing brace (unless next line is another closing brace or else statement)

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,110 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+                        insideQuotedValue = true;
+                    }
+                    else
+                    {
+                        // this could be the end of a quoted value, or it could be an escaped quote
+                        insideQuotedValue = prevChar == '\\';
+                    }
+                }
+
+                if (c == ',' && !insideQuotedValue)
```
mikepizzo: make it an else if... no need to check if we've already determined above that it is a quote.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,110 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        /// e.g. a value like 'a, b=c, foo="bar,baz"' will return an IEnumerable with
+        /// - a
+        /// - b=c
+        /// - foo="bar,baz"
+        /// </summary>
+        /// <param name="preferences"></param>
+        /// <returns></returns>
+        private static IEnumerable<string> SplitPreferences(this string preferences)
+        {
+            StringBuilder currentPreference = new StringBuilder();
```
mikepizzo: StringBuilder is pretty expensive.  If you're just splitting on the comma and not changing/re-arranging the characters, you could just keep track of the index as you read through the preferences text and then use Substring to return each individual preference from within the preferences text string.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +268,110 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+                if (c == ',' && !insideQuotedValue)
+                {
+                    string result = currentPreference.ToString();
+                    string prefName = result.Split('=')[0];
+                    // do not add duplicate preference
+                    if (!addedPreferences.Contains(prefName))
+                    {
+                        yield return result;
+                        addedPreferences.Add(prefName);
+                    }
```
mikepizzo: a little cleaner to add before yielding.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
```diff
@@ -612,4 +857,33 @@ public ITestActionResult Post([FromBody]BatchTestOrder order)
             return Created(order);
         }
     }
+#if NETCORE
```
mikepizzo: why is this only #if NETCORE?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
habbes: VS has recommended that I use an auto property.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
habbes: WebApi for AspNet and AspNetCore have different implementations of the batch handler that are not compatible. For example the batch handler in AspNet did not copy headers to individual requests in the first place, so it seems they worked differently. I added a test case for that. But I did not add the new header copying logic to AspNetCore.

Also, there's no test in that file that runs on both frameworks, they're either inside a `#if NETCORE` or `#if !NETCORE` block. So maybe it doesn't make that much sense to have these test cases in the shared project in the first place.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
mikepizzo: Good question. Unfortunately, there is no easy way to determine whether or not a preference is inherited by individual requests within a batch -- it depends on the definition of the header.  For the preference headers that OData defines (or defines semantics for) we explicitly call out what should be inherited and what should not be, which is what Clement has encoded here.  Unfortunately, adding additional preferences in the future will require that we update this code appropriately.  The safe default for unknown preferences is to copy them over and let the individual statement have the option of handling.

---
In reply to: [386260180](https://github.com/OData/WebApi/pull/2064#discussion_r386260180) [](ancestors = 386260180)

mikepizzo: Exactly. Unfortunately, we don't know the intended behavior of custom headers and preferences, so the default is to copy them over and let the individual statement handle.

---
In reply to: [386295560](https://github.com/OData/WebApi/pull/2064#discussion_r386295560) [](ancestors = 386295560)

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +275,109 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        {
+            IEnumerable<string> preferencesToInherit = SplitPreferences(batchPreferences)
+                .Where(value => 
+                    !nonInheritablePreferences.Any(
+                        prefToIgnore => 
+                        value.Trim().ToLowerInvariant().StartsWith(prefToIgnore)
+                    )
+                )
+                .Select(value => value.Trim());
+            return string.Join(", ", preferencesToInherit);
```
gathogojr: @habbes Here and similar places, I think you should use "," as a separator (as opposed to ", ") to keep the resulting string shorter and compact to minimize request overhead

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -236,5 +275,109 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
+        private static string GetPreferencesToInheritFromBatch(string batchPreferences)
+        {
+            IEnumerable<string> preferencesToInherit = SplitPreferences(batchPreferences)
+                .Where(value => 
+                    !nonInheritablePreferences.Any(
+                        prefToIgnore => 
+                        value.Trim().ToLowerInvariant().StartsWith(prefToIgnore)
+                    )
+                )
+                .Select(value => value.Trim());
```
gathogojr: @habbes Maybe return trimmed values from `SplitPreferences` such that you don't need to call `.Trim()` that many places

PR #2069, opened 2020-02-27
Ensure that $apply=compute() works with EF3.1

kosinsky:
### Description

Moves .NET Core 3.1 tests to EF 3.1 and ensures that $apply=compute() works

- [ ] *Test cases added*

### Additional work necessary


xuzhg:
Merged. Thanks.

Issue #2091, opened 2020-03-18
$count without =true value returns 404.


cdarrigo:
http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360944
states the ODATA api supports a $count operation without specifying a value. 
(e.g.  products/$count vs products/$count=true).

The current implementation requires the =true value to be appended to the query string, violating the Odata specification . 

### Assemblies affected

Microsoft.ASPNetCore.Odata 7.3.0
### Reproduce steps

Configure an asp.net core site for Odata, and issue a call to an exposed odata endpoint, specifying $count with no arguments. 

GET products?$count

### Expected result
```
{
"@odata.context": "...",
"@odata.count": 0,
"value": []
}
```

### Actual result

```
{
"error": {
"code": "",
"message": "The query specified in the URI is not valid. The value for OData query '$count' cannot be empty.",
"details": [],
"innererror": {
"message": "The value for OData query '$count' cannot be empty.",
"type": "Microsoft.OData.ODataException",
[...]

PR #2092, opened 2020-03-18
Added default count support in accordance with http://docs.oasis-open…

cdarrigo:
….org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360944

### Issues

This pull request fixes issue #2091.

### Description

Removed ThrowIfEmptyCheck for count operation in BuildQueryOptions. 
Added a default value of true to provide compliance with the odata specificiation.
Removed inline["$count"] data checks on the unit tests

- [N/A ] *Test cases added*
- [ X] *Build and test with one-click build and test script passed*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
@@ -947,8 +947,7 @@ namespace Microsoft.AspNet.OData.Query
                         RawValues.Select = kvp.Value;
                         break;
                     case "$count":
-                        ThrowIfEmpty(kvp.Value, "$count");
-                        RawValues.Count = kvp.Value;
+                        RawValues.Count = kvp.Value ?? "true";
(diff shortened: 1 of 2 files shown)


mikepizzo:
@cdarrigo -- thanks for the contribution!  The code looks fine, however I think there is confusion around what is/is not supported per protocol.

According to protocol, the $count *query option*, which includes the count along with the results, requires the boolean.  See "inlinecount" in http://docs.oasis-open.org/odata/odata/v4.01/cos01/abnf/odata-abnf-construction-rules.txt:
                  inlinecount = ( "$count" / "count" ) EQ booleanValue

I believe the confusion comes around the fact that $count is *also* used as a path segment, to say that the count of a collection is returned *instead of* the members of the collection. In this case, there is no boolean. See "count" which is used exclusively in path segments:
count = '/$count'

mikepizzo:
Closing as by design, as described in the issue. Please let us know if you need anything else here, and thanks again for your contribution!

PR #2121, opened 2020-04-08
Revert the default null propagation to false for EF Core

xuzhg:
### Issues

*This pull request fixes issue #2080 .*

### Description

The following change makes $select=Collection and $expand=Collection scenarios failing within EF Core 3.x

image

This PR is to revert the default behavior of Null propagation to false for EF Core. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Aggregation/AggregationTests.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Aggregation/AggregationTests.cs
@@ -683,7 +683,7 @@ namespace Microsoft.Test.E2E.AspNet.OData.Aggregation
             string queryUrl =
                 string.Format(
                     AggregationTestBaseUrl + "?$apply=groupby((Name), aggregate(Order/Price with sum as TotalAmount))"
-                    + "/compute(TotalAmount mul 2 as DoubleAmount, length(Name) as NameLen)"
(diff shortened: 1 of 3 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Aggregation/AggregationTests.cs
```diff
@@ -684,7 +684,7 @@ public async Task ComputeAfterGroupByWorks()
                 string.Format(
                     AggregationTestBaseUrl + "?$apply=groupby((Name), aggregate(Order/Price with sum as TotalAmount))"
                     + "/compute(TotalAmount mul 2 as DoubleAmount, length(Name) as NameLen)"
-                    + "&$orderby=TotalAmount",
+                    + "&$orderby=TotalAmount&$filter=Name ne null",
```
kosinsky, line 687: It should be $apply=...../filter(Name null)/compute(...)

Spec says that $apply should be evaluated first and only then $filter. 
As a result, for case with $filter, we will generate LINQ that looks like:
```C#
.Select(x =>new {
        Name = x.Name,
        NameLen=x.Name.Length
})
.Where(x => x.Name != null)
```
I guess .NET or EF optimizes real execution by swapping .Select and .Where. However, that behavior may change.

xuzhg: I changed the query, thanks.

Issue #1755, opened 2019-02-06
Implement support for IAsyncEnumerable


rossbuggins:
Should odata support IAsyncEnumerable?  There is an open work item on asp.net core https://github.com/aspnet/AspNetCore/issues/5357 and the underlying c# support https://github.com/dotnet/corefx/issues/32640

onovotny:
Note that Ix Async will be releasing a preview of the LINQ operators for this shortly.

casperOne:
@KanishManuja-MS I can't speak for @rossbuggins but now that `IAsyncEnumerable` is out in the wild, I think the best ask would be to support allowing controllers to return an `IAsyncEnumerable<T>` implementation (or `Task<IAsyncEnumerable<T>>`).

This would probably require a change in the `TypeHelper` class here:

https://github.com/OData/WebApi/blob/9e447bd77d177cf4cd1e285c318ffc3185a34c12/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs#L269

To allow the `IsCollection` method to recognize `IAsyncEnumerable<T>`.

Additional work would have to be done in the outputformatters/serializers to support `async`/`Task` changes when iterating over said `IAsyncEnumerable<T>`.

The current implementation iterates over the non-generic `IEnumerable`; a conditional branch there would probably have to be checked to see if there is something assignable to `IAsyncEnumerable<object>` or `IEnumerable` and then iterate appropriately.

aladdinhussein:
What is the best approach to deal with MaxIAsyncEnumerableBufferLimit in odata?

darraghjones:
Any plan to add support for IAsyncEnumerable to Odata?

medboz:
Is there any plan to implement this ?

LucasVos:
Any news on this one?

jwevans:
Also interested in this feature

PR #2142, opened 2020-04-29
Remove the null check, fix issue #1755

xuzhg:
### Issues

*This pull request fixes issue https://github.com/OData/odata.net/issues/1755.*

https://github.com/OData/WebApi/issues/2281

### Description

* SelectExpandBinder generates the "Null" expression can't work for "EntityFramework6".

```
IsNull = .If ($$it == null) {
            null
        } .Else {
            $$it.Description
        } == null
```

However, it seems we don't need the "IsNull" check for the complex. So, let's remove this.

- [ ] *Test cases added*

### Additional work necessary


mikepizzo (approved):
Do we know why this was added originally? Is this needed for proper null propagation for other data sources?

Can you re-run the tests to verify no breaking changes? It looks like there was a problem executing the CI.

:shipit:

xuzhg:
I checked with "Smit Patel", and the changes here looks good to him. So, I'd merge it

Issue #2126, opened 2020-04-14
PATCHing a complex type dynamic property does not work


gathogojr:
A bug exists in OData Web API where PATCHing a complex type dynamic property does not work. It however works for a complex type collection dynamic property

### Assemblies affected

Microsoft.AspNetCore.OData 7.3.0

### Reproduce steps

1. Create an OData V4 service with an open entity type, a complex type and a controller

    **Customer** class
    ```c#
    public class Customer
    {
        [Key]
        public int Id { get; set; }
        public string Name { get; set; }
        public IDictionary<string, object> DynamicProperties { get; set; }
    }
    ```

    **Address** class
    ```c#
    public class Address
    {
        public string AddressLine { get; set; }
        public string City { get; set; }
    }
    ```

    **CustomersController** controller
    ```c#
    [ApiController]
    [Route("[controller]")]
    public class CustomersController : ControllerBase
    {
        private static readonly Collection<Customer> _customers = new Collection<Customer>
        {
            new Customer
            {
                Id = 1,
                Name = "Customer 1",
                DynamicProperties = new Dictionary<string, object>
                {
                    { "StreetAddress", new Address { AddressLine = "AL1", City = "C1" } },
                    { "OtherAddresses", new Collection<Address>
                        {
                            new Address { AddressLine = "AL2", City = "C2" },
                            new Address { AddressLine = "AL3", City = "C3" }
                        }
                    }
                }
            }
        };

        [EnableQuery]
        public IQueryable<Customer> Get()
        {
            return _customers.AsQueryable();
        }

        [EnableQuery]
        public SingleResult<Customer> Get([FromODataUri]int key)
        {
            var customer = _customers.SingleOrDefault(d => d.Id.Equals(key));

            return SingleResult.Create(new[] { customer }.AsQueryable());
        }

        public async Task<IActionResult> Patch([FromODataUri] int key, [FromBody]Delta<Customer> delta)
        {
            var customer = _customers.SingleOrDefault(d => d.Id.Equals(key));

            if (customer != null)
                delta.Patch(customer);

            await Task.Run(() =>
            {
                // ...            
            });

            return Ok();
        }
    }
    ```
    Take note of the two dynamic properties `StreetAddress` (complex) and `OtherAddresses` (complex collection)
2. Apply the required service configuration in _Startup.cs_
3. Run the service and then initiate a PATCH request

    Endpoint: http://ServiceRoot/Customers(1)

    Payload:
    ```json
    {
[...]

Issue #1780, opened 2019-03-11
Unable to PATCH a complex property on an Open Type


andyward:
Given an open type `A` and EntitySet `AEntities`, and a Complex Type of `B`, it IS possible to patch an `A` Entity with a Collection of `B`s as follows:

```
PATCH /odata/AEntities(1)
{
    'BCollection@odata.type':'#Collection(B)',
    'BCollection': [{ 'Foo': 'bar' }]
}
```
... however it's NOT possible to PATCH a single Property with a Complex Type of `B`:

```
PATCH /odata/AEntities(1)
{
    'B': {
         '@odata.type':'#B',
         'Foo': 'bar'
   }
}
```
There seems to be an issue in the deserialisation of `Delta<A>` that causes this discrepancy. 

Is this supported?

### Assemblies affected

OData 7.1.0 on .NET Framework.

### Reproduce steps

I've created a E2E test with a more real-world example in the existing `OpenType/TypedTest` class that reproduces the issue using the standard test controllers: https://github.com/andyward/WebApi/commit/9cd34ff6176380cfa89d7b23d57a5535e663fd68

### Expected result

I would expect `A.B` to be deserialised in `Delta<A>` so the property could be persisted.

### Actual result

`Delta<A>.B` is null, while `Delta<A>.BCollection` is instantiated and populated with deserialised array of Bs

### Additional detail

The Controller Action is defined as

``` csharp
[HttpPatch]
public IActionResult Patch(int key, [FromBody]Delta<A> patch, ODataQueryOptions<A> queryOptions)
{
    // Patch implementation
}
```

xuzhg:
@andyward Thanks for reporting it to us. we will take a look. Basically, i think it supports. Thanks.

raheph:
@xuzhg please take a look

andyward:
Did you get a chance to look at this @xuzhg ?

jinhong-:
Any updates? We seem to be experiencing the same issue

PR #2160, opened 2020-05-11
Fix logic for patching structured type with complex type dynamic property

gathogojr:
### Issues

*This pull request fixes issue #2126 and #1780*

### Description

Logic for patching an structured type containing a complex type dynamic property is currently broken. This appears to have happened in a refactor done in [this](https://github.com/OData/WebApi/pull/1445/files#diff-da4fb05d65976a67ce2a8a1eb8ef0c8fL117) PR. Two helper methods were created: `TrySetPropertyValueInternal` and `TrySetNestedResourceInternal`. The first helper method ended up with the logic for handling dynamic properties in the delta. However, a complex type property (including dynamic) is routed to `TrySetNestedResourceInternal` that does not bother with dynamic properties at all. This PR fixes the issue by moving the logic for processing dynamic properties in the delta back to the `TrySetPropertyValue` public method

- [ ] *Test cases added*


xuzhg:
Looks good to me. Let's ship it.

Issue #1611, opened 2018-09-09
CORS preflight doesn't work with OData in ASP.NET Core


brjohnstmsft:
Sending CORS preflight requests to an OData service running on ASP.NET Core fails with 404 when it shouldn't. Note that this is a different symptom than [this similar issue](https://github.com/OData/WebApi/issues/208), and also likely a different cause since the implementation of CORS in ASP.NET Core is substantially different than the one for classic ASP.NET.

### Assemblies affected

Microsoft.AspNetCore.OData 7.0.1

### Reproduce steps

I managed to reproduce this by creating a solution with two projects: One ASP.NET Core Web API project, and one MSTest-based unit test project. For some reason, I couldn't get ASP.NET Core to work at all if I combined all the code in a single project.

The repro is [here](https://github.com/brjohnstmsft/odata-repros/tree/master/ODataAndCors)

### Expected result

All four tests pass

### Actual result

The test that uses CORS with OData fails. The first preflight request returns 404 instead of 200.

### Additional detail

I haven't been able to determine the root cause of this. The failure mode is different than it is in classic ASP.NET.

cornem:
Related to https://github.com/aspnet/CORS/issues/98?

A possible workaround is to let middleware handle the CORS pre-flight requests (but it means less granular control over CORS policies):

	services.AddCors(options =>
	{
		options.AddPolicy(
			"AllowAllOrigins",
			builder => builder
                            .AllowAnyOrigin()
                            .AllowAnyHeader()
                            .AllowAnyMethod()
                            .AllowCredentials());

		options.DefaultPolicyName = "AllowAllOrigins";
	});

And adding `app.UseCors()` before the call to `UseMvc()`.

The [`ODataRequestMethod`](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/RequestMethod.cs) enum doesn't know `Options`, so `Unkown` is returned [here](https://github.com/OData/WebApi/blob/3ec3a69fc557a9f6a0380ee21940c2998c02b394/src/Microsoft.AspNetCore.OData/Adapters/WebApiRequestMessage.cs#L91). As a result [`EntitySetRoutingConvention`](https://github.com/OData/WebApi/blob/3ec3a69fc557a9f6a0380ee21940c2998c02b394/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs#L24) doesn't find a matching action.

[...]

Issue #1616, opened 2018-09-20
CreateRequestContainer fails when called multiple times


brjohnstmsft:
`HttpRequestMessageExtensions.CreateRequestContainer` is not idempotent; if you call it more than once, it throws an exception. `ODataPathRouteConstraint.Match` attempts to mitigate this by calling it inside a lambda:
 https://github.com/OData/WebApi/blob/5c59a95076011f0d9d5ebfa8f96553724af88ca1/src/Microsoft.AspNet.OData/Routing/ODataPathRouteConstraint.cs#L53

However, this doesn't help in the case where `Match` is called more than once. This can happen if `GetRouteData()` is called more than once, which indirectly calls all route constraints. This is necessary to work around [this issue](https://github.com/OData/WebApi/issues/208).

### Assemblies affected

Microsoft.AspNet.OData 7.0.1

### Reproduce steps

Call `GetRouteData` before `CorsMessageHandler` is called. See [this project](https://github.com/brjohnstmsft/odata-repros/tree/cors-workaround-fails/ODataAndCorsClassic/ODataAndCorsClassic) for an example.

### Expected result

All tests in the repro would pass.

### Actual result

Both tests that call OData endpoints fail with this stack trace:

```
Test Name:	MakeNonCorsODataRequestExpectSuccess
Test Outcome:	Failed
Result Message:	Assert.AreEqual failed. Expected:<OK>. Actual:<InternalServerError>. Expected matching statuses for request GET http://localhost:9999/odata/$metadata
Result StandardOutput:	
Running workaround...
{"Message":"An error has occurred.","ExceptionMessage":"A request container already exists on the request.","ExceptionType":"System.InvalidOperationException","StackTrace":"   at Microsoft.AspNet.OData.Extensions.HttpRequestMessageExtensions.CreateRequestContainer(HttpRequestMessage request, String routeName)
   at Microsoft.AspNet.OData.Routing.ODataPathRouteConstraint.<>c__DisplayClass0_0.<Match>b__0()
   at Microsoft.AspNet.OData.Routing.ODataPathRouteConstraint.GetODataPath(String oDataPathString, String uriPathString, String queryString, Func`1 requestContainerFactory)
   at Microsoft.AspNet.OData.Routing.ODataPathRouteConstraint.Match(HttpRequestMessage request, IHttpRoute route, String parameterName, IDictionary`2 values, HttpRouteDirection routeDirection)
[...]

Issue #208, opened 2015-02-03
Cors OData Preflight


LianwMS:
When the browser sends an "OPTIONS" pre-flight request for an OData endpoint, the response may be a 500 error when it should not be. 

A simple "Get" action on an Cors attributed ODataController like the following will succeed:  

```
[EnableCors(origins: "http://localhost:52135", headers: "*", methods: "*")]
public class SampleController : ODataController
{
    [EnableQuery]
        public IQueryable<JobDTO> Get()
        {... }
```

However a call to an action in the same controller with a "delete" verb and an argument like this fails:

```
public async Task<IHttpActionResult> Delete([FromODataUri] Guid key)
        {...}
```

Sample Error Message:

```
{
"error":{
"code":"","message":"An error has occurred.","innererror":{
  "message":"Object reference not set to an instance of an object.","type":
"System.NullReferenceException","stacktrace":" 
at 
System.Web.Http.Controllers.ApiControllerActionSelector.ActionSelectorCacheItem.Create405Response(HttpControllerContext controllerContext, IEnumerable1 allowedCandidates)\r\n 
at System.Web.Http.Controllers.ApiControllerActionSelector.ActionSelectorCacheItem.CreateSelectionError(HttpControllerContext controllerContext)\r\n 
at System.Web.Http.Controllers.ApiControllerActionSelector.ActionSelectorCacheItem.SelectAction(HttpControllerContext controllerContext)\r\n 
at System.Web.Http.Controllers.ApiControllerActionSelector.SelectAction(HttpControllerContext controllerContext)\r\n 
at System.Web.Http.Tracing.Tracers.HttpActionSelectorTracer.<>c__DisplayClass2.<System.Web.Http.Controllers.IHttpActionSelector.SelectAction>b__0()\r\n 
at System.Web.Http.Tracing.ITraceWriterExtensions.TraceBeginEnd(ITraceWriter traceWriter, HttpRequestMessage request, String category, TraceLevel level, String operatorName, String operationName, Action1 beginTrace, Action execute, Action1 endTrace, Action1 errorTrace)\r\n 
at System.Web.Http.Tracing.Tracers.HttpActionSelectorTracer.System.Web.Http.Controllers.IHttpActionSelector.SelectAction(HttpControllerContext controllerContext)\r\n 
at System.Web.OData.Routing.ODataActionSelector.SelectAction(HttpControllerContext controllerContext)\r\n 
[...]

PR #2167, opened 2020-05-20
Add fine grained cors access control Support

marabooy:
### Issues

*This pull request fixes issue #1611, #208  and #1616 *

### Description
Introduces a way where the routing conventions pick the targeted controller method and provides the metadata to the cors middlware enabling cors preflights to be possible when using attributes
https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-3.1#enable-cors-with-attributes

- [x ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNet.OData.Interfaces;
 
@@ -14,43 +15,19 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
         /// <returns></returns>
         internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
         {
-            if (request.Method == ODataRequestMethod.Options)
+            if (request.Method != ODataRequestMethod.Options)
             {
-                IEnumerable<string> values;
-                if (request.Headers.TryGetValues("Access-Control-Request-Method", out values))
-                {
-                    return ConvertToODataRequestMethod(values.FirstOrDefault());
-                }
-                else
-                {
-                    return ODataRequestMethod.Options;
-                }
+                return request.Method;
             }
 
-            return request.Method;
-        }
-        /// <summary>
-        /// Converts string request method to ODataRequestMethod
-        /// </summary>
-        /// <param name="requestMethod"></param>
-        /// <returns></returns>
-        private static ODataRequestMethod ConvertToODataRequestMethod(string requestMethod)
-        {
-            if (requestMethod == null)
+            IEnumerable<string> values;
+            if (!request.Headers.TryGetValues("Access-Control-Request-Method", out values))
             {
                 return ODataRequestMethod.Unknown;
             }
 
-            switch (requestMethod.ToUpperInvariant())
-            {
-                case "GET": return ODataRequestMethod.Get;
-                case "POST": return ODataRequestMethod.Post;
-                case "PUT": return ODataRequestMethod.Put;
-                case "PATCH": return ODataRequestMethod.Patch;
-                case "DELETE": return ODataRequestMethod.Delete;
-                case "MERGE": return ODataRequestMethod.Merge;
-                default: return ODataRequestMethod.Unknown;
-            }
+            ODataRequestMethod preflightMethod;
+            return Enum.TryParse<ODataRequestMethod>(values.FirstOrDefault(), true, out preflightMethod) ? preflightMethod : ODataRequestMethod.Unknown;
         }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
@@ -1,0 +1,41 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNet.OData.Interfaces;
+
+namespace Microsoft.AspNet.OData.Routing.Conventions
+{
+    internal static class RequestMethodExtensions 
+    {
+        internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
+        {
+            if (request.Method == ODataRequestMethod.Options)
+            {
+                IEnumerable<string> values;
+                request.Headers.TryGetValues("Access-Control-Request-Method", out values);
+                var enumerable = values.ToList();
+                if (enumerable.Count()!=0)
+                {
+                    return ToOdataRequestMethod(enumerable.First());
+                }
+
+                return ODataRequestMethod.Unknown;
+            }
+
+            return request.Method;
+        }
+
+        private static ODataRequestMethod ToOdataRequestMethod(string requestMethod)
+        {
+            switch (requestMethod.ToUpperInvariant())
+            {
+                case "GET": return ODataRequestMethod.Get;
+                case "POST": return ODataRequestMethod.Post;
+                case "PUT": return ODataRequestMethod.Put;
+                case "PATCH": return ODataRequestMethod.Patch;
+                case "DELETE": return ODataRequestMethod.Delete;
+                case "MERGE": return ODataRequestMethod.Merge;
+                default: return ODataRequestMethod.Unknown;
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
                 }
                 else
                 {
-                    return ODataRequestMethod.Unknown;
+                    return ODataRequestMethod.Options;
                 }
             }
 
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
@@ -4,28 +4,43 @@ using Microsoft.AspNet.OData.Interfaces;
 
 namespace Microsoft.AspNet.OData.Routing.Conventions
 {
-    internal static class RequestMethodExtensions 
+    internal static class RequestMethodExtensions
     {
+        /// <summary>
+        /// Returns the request method and in the case of Options request it returns the Access-Control-Request-Method present in the
+        /// preflight request for the request method that will be used for the actual request
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
         internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
         {
             if (request.Method == ODataRequestMethod.Options)
             {
                 IEnumerable<string> values;
-                request.Headers.TryGetValues("Access-Control-Request-Method", out values);
-                var enumerable = values.ToList();
-                if (enumerable.Count()!=0)
+                if (request.Headers.TryGetValues("Access-Control-Request-Method", out values))
                 {
-                    return ToOdataRequestMethod(enumerable.First());
+                    return ConvertToODataRequestMethod(values.FirstOrDefault());
                 }
-
-                return ODataRequestMethod.Unknown;
+                else
+                {
+                    return ODataRequestMethod.Unknown;
+                }
             }
 
             return request.Method;
         }
-
-        private static ODataRequestMethod ToOdataRequestMethod(string requestMethod)
+        /// <summary>
+        /// Converts string request method to ODataRequestMethod
+        /// </summary>
+        /// <param name="requestMethod"></param>
+        /// <returns></returns>
+        private static ODataRequestMethod ConvertToODataRequestMethod(string requestMethod)
         {
+            if (requestMethod == null)
+            {
+                return ODataRequestMethod.Unknown;
+            }
+
             switch (requestMethod.ToUpperInvariant())
             {
                 case "GET": return ODataRequestMethod.Get;
--- a/src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs
@@ -111,12 +111,12 @@ namespace Microsoft.AspNet.OData.Extensions
                         // to select the correct Endpoint, we save the ActionDescriptor value into ODataFeature.
                         odataFeature.ActionDescriptor = controllerActionDescriptor;
                         // Add handler to handle options calls. The routing criteria has been patched to allow endpoint discovery using the correct cors headers
-                        if (request.Method.Equals("OPTIONS",StringComparison.InvariantCultureIgnoreCase))
+                        if (request.Method.Equals("OPTIONS", StringComparison.OrdinalIgnoreCase))
                         {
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs
```diff
@@ -109,6 +110,18 @@ public override ValueTask<RouteValueDictionary> TransformAsync(HttpContext httpC
                         // For example, we have two "Get" methods in same controller, in order to help "EndpointSelector" 
                         // to select the correct Endpoint, we save the ActionDescriptor value into ODataFeature.
                         odataFeature.ActionDescriptor = controllerActionDescriptor;
+                        // Add handler to handle options calls. The routing criteria has been patched to allow endpoint discovery using the correct cors headers
+                        if (request.Method.Equals("OPTIONS",StringComparison.InvariantCultureIgnoreCase))
+                        {
+                            var metadata = actionDescriptor.EndpointMetadata;
+                            // For option request can set this as the action will be handled by the cors middleware 
+                            var metadataCollection = metadata!=null && metadata.Any()
+                                ? new EndpointMetadataCollection(actionDescriptor.EndpointMetadata)
```
odero, line 119:
```suggestion
                                ? new EndpointMetadataCollection(metadata)
```

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
```diff
@@ -104,6 +107,96 @@ public void TransformAsyncReturnsCorrectRouteValues()
...
+                ControllerName = "Customers",
+                ActionName = "Get",
+                EndpointMetadata = GenerateMetadata(includeMetadata)
+            };
+
+            IActionSelector actionSelector = new MockActionSelector(actionDescriptor);
+
+            // Act
+            ODataEndpointRouteValueTransformer transformer = new ODataEndpointRouteValueTransformer(actionSelector);
+            ValueTask<RouteValueDictionary> actual = transformer.TransformAsync(httpContext, values);
```
odero, line 141:
```suggestion
            _ = transformer.TransformAsync(httpContext, values);
```
Since `actual` is assigned but never used maybe this might be better?

Also do you need to Wait for `TransformAsync`?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
```diff
@@ -104,6 +107,96 @@ public void TransformAsyncReturnsCorrectRouteValues()
...
+            var endpoint = httpContext.GetEndpoint();
```
odero, line 179: Can all this repeated logic be encapsulated into some build/helper method that the test calls - keep things DRY

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    internal static class RequestMethodExtensions 
+    {
+        internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
+        {
+            if (request.Method == ODataRequestMethod.Options)
+            {
+                IEnumerable<string> values;
+                request.Headers.TryGetValues("Access-Control-Request-Method", out values);
+                var enumerable = values.ToList();
+                if (enumerable.Count()!=0)
```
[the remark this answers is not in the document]
marabooy: will refactor to .Any() since i want to know if there is anything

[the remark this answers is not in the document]
odero: Since `enumerable` is a list then you don't need to use `.Any`,  `.Count` property is preferred in this case

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                    return ToOdataRequestMethod(enumerable.First());
+                }
+
+                return ODataRequestMethod.Unknown;
+            }
+
+            return request.Method;
+        }
+
+        private static ODataRequestMethod ToOdataRequestMethod(string requestMethod)
```
gathogojr, line 27: @marabooy `ToODataRequestMethod(string requestMethod)` is better for the eyes :)

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.AspNet.OData.Routing.Conventions
+{
+    internal static class RequestMethodExtensions 
+    {
+        internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
+        {
+            if (request.Method == ODataRequestMethod.Options)
+            {
+                IEnumerable<string> values;
+                request.Headers.TryGetValues("Access-Control-Request-Method", out values);
```
gathogojr, line 14: @marabooy According to the docs, `TryGetValues` returns true if the specified header and specified values are stored are stored in the `HttpHeaders` collection: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpheaders.trygetvalues?view=netcore-3.1
For that reason, I think you could rewrite this section as:
```c#
IEnumerable<string> values;
if (request.Headers.TryGetValues("Access-Control-Request-Method", out values)
{
    return ToODataRequestMethod(values.FirstOrDefault());
}
else
{
    return ODataRequestMethod.Unknown;
}
```
Note that the `TryGetValues()` method won't take a `List<string>` or even `IList<string>` out parameter - I tried it out.
In addition, you could use `.First()` (on the strength of the docs) but I think it may be safer to use `.FirstOrDefault()` and do a null-check first in the private `ToODataRequestMethod` helper method

gathogojr:
@marabooy FYI, I pulled your branch and tried out your fix and it worked

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
marabooy: To be sure I made the methods use async await

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
marabooy: Updated

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
marabooy: Thanks I chose the last suggestion as it communicates the methods intention clearly

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
marabooy: 👍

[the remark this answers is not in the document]
marabooy: When it returns false I do not check the header values and I resolve the method to the same method in case the user has added their own way to resolve options requests then this would go to that. 
Also according to the cors specification the method should be populated with one of the known request methods.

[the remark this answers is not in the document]
odero: I see john more explicitly addressed my concern here

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <param name="request"></param>
+        /// <returns></returns>
+        internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
+        {
+            if (request.Method == ODataRequestMethod.Options)
+            {
+                IEnumerable<string> values;
+                if (request.Headers.TryGetValues("Access-Control-Request-Method", out values))
+                {
+                    return ConvertToODataRequestMethod(values.FirstOrDefault());
```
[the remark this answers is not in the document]
marabooy: The cors specification specifies that only one value should be there. That is why I am picking the first and discarding everything else.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            IEnumerable<string> values;
+            if (!request.Headers.TryGetValues("Access-Control-Request-Method", out values))
+            {
+                return ODataRequestMethod.Unknown;
+            }
+
+            ODataRequestMethod preflightMethod;
+            return Enum.TryParse<ODataRequestMethod>(values.FirstOrDefault(), true, out preflightMethod) ? preflightMethod : ODataRequestMethod.Unknown;
```
odero, line 30:
```suggestion
            return Enum.TryParse<ODataRequestMethod>(values.SingleOrDefault(), true, out preflightMethod) ? preflightMethod : ODataRequestMethod.Unknown;
```
If cors spec states that only one value should be there then maybe you should use `SingleOrDefault` instead of `FirstOrDefault`?

mikepizzo (approved):
:shipit:

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
```diff
@@ -104,6 +107,73 @@ public void TransformAsyncReturnsCorrectRouteValues()
             Assert.Same(actionDescriptor, httpContext.ODataFeature().ActionDescriptor);
         }
 
+        [Theory]
+        [InlineData(false, true)]
+        [InlineData(false, false)]
```
mikepizzo, line 112: >false [](start = 20, length = 5)

don't really have to pass this in, since it's false in all cases.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
```diff
@@ -104,6 +107,73 @@ public void TransformAsyncReturnsCorrectRouteValues()
...
+        public async void TransformAsyncReturnsCorrectRouteValuesForNonOptionRequests(bool isOptionsRequest, bool includeMetadata)
+        {
+            var endpoint = await GenerateCorsEndPoint(isOptionsRequest, includeMetadata);
+            // Assert the metadata is empty when options requests are sent and the controller is not annotated
+
+            Assert.Null(endpoint);
+        }
+
+        [Theory]
+        [InlineData(true, true)]
```
mikepizzo, line 122: >true [](start = 20, length = 4)

don't really have to pass this in since it's true for all cases in this test.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal static class RequestMethodExtensions
+    {
+        /// <summary>
+        /// Returns the request method and in the case of Options request it returns the Access-Control-Request-Method present in the
+        /// preflight request for the request method that will be used for the actual request
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        internal static ODataRequestMethod GetRequestMethodOrPreflightMethod(this IWebApiRequestMessage request)
```
mikepizzo, line 16: >GetRequestMethodOrPreflightMethod [](start = 43, length = 33)

What if the developer already has explicit logic to implement OPTIONS? Would this override a custom implementation? Would they have to rely on ordering to check for their implementation first?

marabooy: I believe this won't block their current usecase. This fix exposes the endpoint metadata for the cors middleware to use.

marabooy: they can also override the behaviour easily by extending the value transformer or providing a custom one.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
marabooy: resolved

test/UnitTest/Microsoft.AspNetCore.OData.Test/Extensions/Endpoint/ODataEndpointRouteValueTransformerTests.cs
marabooy: resolved

xuzhg:
Merged at https://github.com/OData/WebApi/commit/bd30d43152a951ebed0c7ceb0e784d1d90e48f35

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RequestMethodExtensions.cs
GrayPockets: I had explicit logic to implement OPTIONS and this blocked my use case.

@marabooy could you show me how to follow your suggestion to customize/override the transformer? I could use a hand.

marabooy: @GrayPockets do you mean the routing or response transformers?

marabooy: @GrayPockets You could use the following.

Override the TransformAsync method in [the ODataEndpointRouteValueTransformer  class.](https://github.com/OData/WebApi/blob/2ccbefd65e19e13f015f653f3c446a6a19f08406/src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointRouteValueTransformer.cs#L130) The default method sets an endpoint so that the default cors middleware can read the annotations on the method or controller.  Do let me know if you do experience issues with this approach, I would be happy to create a fix to make it easier for users to override this.

GrayPockets: > Override the TransformAsync method in the ODataEndpointRouteValueTransformer class. 

@marabooy perhaps I am dense, I do not see how I can do this.
Attempting to create a class to inherit from ODataEndpointRouteValueTransformer in the project of my API gives me tons of accessibility/protection errors.

image

I'm glad that the CORS updates occurred, but I have custom [HttpOptions] methods on my controllers, and this change causes them to throw 404 errors. This allows me to have fine-grain control of the Option header responses to each controller. I'm also using Microsoft.AspNetCore.OData.Versioning, which complicates the middleware layer on top of it.

As a work-around for today, I downloaded the source for this project to replace the NuGet package, and commented out lines 130-140.

If there was setting to disable these lines, I would be greatly appreciative. I may not be able to grok how to make that change on my own.

marabooy: Thanks for the update, Let me create an issue that can be followed to allow extensibility of this. Thanks for the update.

GrayPockets: Awesome!

Issue #2075, opened 2020-03-03
Routing selects wrong method in controller


henrikdahl8240:
I have these definitions in my Entity Model:
```
var function__Bike__SellA = this.EntityType<Bike>()
    .Function("Sell").Returns<SellResult_DTO>();
function__Bike__SellA.Parameter<string>("aaa");
function__Bike__SellA.Parameter<int>("ccc");
function__Bike__SellA.Parameter<string>("ddd");
function__Bike__SellA.Parameter<string>("eee");
function__Bike__SellA.Parameter<string>("fff");
function__Bike__SellA.Parameter<bool>("ggg");
function__Bike__SellA.Parameter<bool>("hhh");
function__Bike__SellA.Parameter<bool>("iii");
function__Bike__SellA.Parameter<bool>("jjj");
function__Bike__SellA.Parameter<bool>("kkk");
function__Bike__SellA.Parameter<bool>("lll");
function__Bike__SellA.Parameter<int>("mmm");
function__Bike__SellA.Parameter<int>("nnn");
function__Bike__SellA.Parameter<int>("ooo");

var function__Bike__SellB = this.EntityType<Bike>()
    .Function("Sell").Returns<SellResult_DTO>();
function__Bike__SellB.Parameter<string>("aaa");
function__Bike__SellB.Parameter<int>("ccc");
function__Bike__SellB.Parameter<string>("open");
function__Bike__SellB.Parameter<string>("ddd");
function__Bike__SellB.Parameter<string>("eee");
function__Bike__SellB.Parameter<string>("fff");
function__Bike__SellB.Parameter<bool>("ggg");
function__Bike__SellB.Parameter<bool>("hhh");
function__Bike__SellB.Parameter<bool>("iii");
function__Bike__SellB.Parameter<bool>("jjj");
function__Bike__SellB.Parameter<bool>("kkk");
function__Bike__SellB.Parameter<bool>("lll");
function__Bike__SellB.Parameter<int>("mmm");
function__Bike__SellB.Parameter<int>("nnn");
function__Bike__SellB.Parameter<int>("ooo");

var function__Bike__SellIdC = this.EntityType<Bike>()
    .Function("Sell").Returns<SellResult_DTO>();
function__Bike__SellIdC.Parameter<string>("aaa");
function__Bike__SellIdC.Parameter<int>("ccc");
function__Bike__SellIdC.Parameter<bool>("ppp");
function__Bike__SellIdC.Parameter<string>("qqq");
[...]

PR #2168, opened 2020-05-20
Improve action selector routing

habbes:
### Issues

This pull request fixes issues #1973 #2075

This is a "fork" of PRs #2007 and #1988. I created a new PR because the older ones were not longer triggering the build pipeline.

### Description

The implementation of `ODataActionSelector` was relatively relaxed, resulting in returning the incorrect route in some instances (see related issues). In this PR, I updated the `ODataActionSelector` in both AspNetCore (2.x and 3.1) and AspNet classic and made them more strict. 

1. It rejects candidate actions which have parameters which cannot be matched to the route parameters, body, query options, optional parameters and `ODataPath`.
2. It also rejects a candidate which has fewer parameters than there are segment keys in the route. For example, it would reject `Get()` if the route is like `/People(1)`.
3. If the action has parameters that cannot be matched from the request (see point 1.), then it checks whether there are registered model binders that can bind that parameter type (e.g. `CancellationToken`), if it doesn't find a suitable binder, the action is rejected.
4. In case it ends up with multiple suitable candidates, it prioritizes the ones that have declared support for the current request method through attributes like `AcceptVerbs("POST")` and `HttpPost`
5. In case 4. doesn't work, another tie-breaker used is to prioritize the action which has as many parameters as keys (i.e. no extra parameters). (Note: I added a key count value in the route data to keep track of the number of detected key segments in the route)

In previous versions of this PR, I used headers like content-length and media type to determine whether the request has a body. In this PR, I'm using http methods (PUT, POST, PATCH, MERGE) instead to make that decision. This introduced some issue when dealing with an edm action that has multiple overloads. For example `DoSomething(int key)` and `DoSomething()`. Assuming the route `/DoSomething()` was called and did not have any parameters, naturally, one would expect the second parameter-less overload `DoSomething()` to be called, however it the action selector would call the first one `DoSomething(int key)` because it thinks the `key` parameter could be bound to the body (even if this request does not have a body, so long as it's a POST request, it can't tell). To address that, I implemented step 5. above.

In previous versions of this PR, I also attempted to perform the same checks in the ODataActionSelector in both aspnet core and classic. In this revision if I have removed the checks from net classic's implementation except the one needed to fix the bug. I think this was safer since the actual action selection is handled by aspnet, it doesn't make much sense to perform all those checks *after* the action has already been selected.

#### Fixed bugs in other tests
`DateTimeTests` had a test case that calls the url: `http://localhost/odata/DateTimeModels(2)/Default.CalcBirthday(dto=2012-12-22T01:02:03Z)` and defined controller action:
```c#
[HttpGet]
 public DateTime CalcBirthday(DateTimeOffset dto)
 {
      Assert.Equal(DateTimeOffset.Parse("2012-12-22T01:02:03Z"), dto);
      return new DateTime(1978, 11, 15, 0, 12, 0, DateTimeKind.Utc);
 }
```
The test used to run correctly, but this shouldn't be the case. The url is calling a function bound to the entity with id 2, but the controller action does not declare a key parameter. This test fails with a 404 error with the new PR. I added an `int key` parameter in order for the test to pass.

 `SpecialCharactersLinkGenerationModelTests` had tests which were passing when they shouldn't have, because they were routing to the wrong action and they and they didn't have sufficient assertions to verify the result. With this new action selector, the test requests would return a 404 instead of routing to the wrong action, causing the test to fail

-  The `SpecialCharactersLinkGenerationTestController` actions were protected, causing routing not to work properly. I made them public. Also renamed `GetEntityByKey(int key)`, which does not seem to be supported by the implemented entity set routing conventions, to `Get(int key)`), did the same for `PatchEntity`
- The `SpecialCharactersLinkGenerationWebTestsController` included a test that asserts that a single whitespace should be supported as a key parameter. It turns out that it's not supported and the test was only passing because of the incorrect routing and insufficient assertions to notice this. I added more assertions to the test, I also excluded the whitespace from those tests. It should be added back if we implement support for whitespace as key parameter.

- [x] *Test cases added*

### Additional work necessary

I have some doubt about these changes. I get the impression that it overfits the test cases, and could break on edge cases that are not covered in the tests but may currently exist in our customers' codebases. I think the problem is that I initially approached this task as a bug fix to the original `ODataActionSelector`, but it turned out to be an exercise of fixing one edge-case after another without breaking anything else. In retrospect, I think it should have been approached as a feature that overhauls and revamps the action selector. I propose we plan for a feature to rethink and re-design the action selector and routing logic in order to implement something robust.

Alternatively, if you find that this approach is to brittle to merge, I propose that I implement a quicker patch fix that only adds support for http method attributes (step 4. in the list above). This wouldn't solve all the bugs in the action selector, but I think it will be sufficient to address [this customer-reported bug](#2075). And I believe it's fairly simple and straightforward to implement. Then after that we can plan for a feature to re-design and re-implement the action selector.

If this PR is accepted, we should update the docs mentioning that a 404 would be returned if the parameters of the action don't match those of the request, even if the name matches.


src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -126,31 +162,78 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+                    {
+                        return bestCandidate.ActionDescriptor;
+                    }
+                    
+                    // also priorize actions that have the exact number of parameters as available keys
+                    // this helps disambiguate between overloads of actions that implement actions
+                    // e.g. DoSomething(int key) vs DoSomething(), if there are no availableKeys, the
+                    // selector could still think that the `int key` param will come from the request body
+                    // and end up returning DoSomething(int key) instead of DoSomething()
+                    bestCandidate = matchedCandidates.FirstOrDefault(candidate => candidate.FilteredParameters.Count() == availableKeys.Count());
```
habbes, line 188:
```diff
- availableKyes.Count()
+ availableKeysCount
```

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -108,14 +143,15 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
                         {
                             return p.ParameterType != typeof(ODataPath) &&
                                 !ODataQueryParameterBindingAttribute.ODataQueryParameterBinding.IsODataQueryOptions(p.ParameterType);
-                        })));
+                        }), c));
```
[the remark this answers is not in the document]
habbes: You mean I should remove the curly braces?

henrikdahl8240:
@xuzhg In case you find the approach to be sound, do you see a possibility, that your nightly build may include this, so I may try it?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -35,21 +41,35 @@ public class ODataActionSelector : IActionSelector
...
         /// <summary>
         /// Initializes a new instance of the <see cref="ODataActionSelector" /> class.
         /// </summary>
         /// <param name="innerSelector">The inner action selector.</param>
-        public ODataActionSelector(IActionSelector innerSelector)
+        /// <param name="modelBinderFactory">IModelBinderFactory instance from dependency injection</param>
+        /// <param name="modelMetadataProvider">IModelMetadataProvider instance from dependency injection</param>
+        public ODataActionSelector(
+            IActionSelector innerSelector,
+            IModelBinderFactory modelBinderFactory = null,
```
xuzhg, line 67: > = null [](start = 50, length = 7)

They are from DI, So why makes them as optional parameter?

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RoutingConventionHelpers.cs
```diff
@@ -191,6 +191,8 @@ public static void AddKeyValueToRouteData(this IWebApiControllerContext controll
                         AddKeyValues(anotherKeyName, keyValuePair.Value, keyProperty.Type, controllerContext.RouteData, routingConventionsStore);
                     }
                 }
+
+                IncrementKeyCount(controllerContext.RouteData);
```
xuzhg, line 195: >controllerContext.RouteData [](start = 34, length = 27)

for composite keys (multiple keys), only count 1?

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RefRoutingConvention.cs
```diff
@@ -69,7 +69,7 @@ internal static string SelectActionImpl(ODataPath odataPath, IWebApiControllerCo
                         controllerContext.AddKeyValueToRouteData((KeySegment)odataPath.Segments[1]);
                     }
 
-                    controllerContext.RouteData.Add(ODataRouteConstants.NavigationProperty, navigationLinkSegment.NavigationProperty.Name);
+                    controllerContext.AddNavigationPropertyToRouteData(navigationLinkSegment);
```
xuzhg, line 72: >AddNavigationPropertyToRout [](start = 38, length = 27)

shall we modify NavigationRoutingConvention.cs

src/Microsoft.AspNet.OData.Shared/Routing/ODataRouteConstants.cs
```diff
@@ -67,5 +67,11 @@ public static class ODataRouteConstants
         /// Route data key for the OData optional parameters.
         /// </summary>
         public static readonly string OptionalParameters = typeof(ODataOptionalParameter).FullName;
+
+        /// <summary>
+        /// Route data key that tracks the number of key segments,
+        /// navigation properties and operation parameters in the request URI
+        /// </summary>
+        public static readonly string KeyCount = "{keyCount}";
```
xuzhg, line 75: >{ [](start = 50, length = 1)

{ } typically used for template, It might be confusing here. If it's only used as the "key" in the dictionary, consider to remove braces and add "OData" as prefix?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingTest.cs
```diff
@@ -491,6 +493,22 @@ public async Task BadQueryString_ReturnsBadRequest(string uri, string expectedEr
...
             Assert.Contains(expectedError, responseString);
         }
+
+        [Theory]
+        [InlineData("NotFoundWithIdCustomers", "GET")]
+        [InlineData("NotFoundCustomers(10)", "GET")]
+        [InlineData("NotFoundCustomers(10)", "DELETE")]
+        [InlineData("NotFoundCustomers(10)", "PUT")]
+        [InlineData("NotFoundCustomers(10)", "PATCH")]
+        public async Task ActionsDontMatch_ReturnsNotFound(string uri, string httpMethod)
```
xuzhg, line 503: >Dont [](start = 33, length = 4)

Donot

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// Helper methods for working with ODataActionSelector tests
+    /// </summary>
+    public static class ODataActionSelectorTestHelper
+    {
+        /// <summary>
+        /// Checks whether the specified method is a suitable match
+        /// for the specified action
+        /// </summary>
+        /// <param name="action"></param>
+        /// <param name="method"></param>
```
xuzhg, line 37: If you add the comments, please finish them?
Otherwise, FYI, in the Test project, it's optional to add the comments.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,249 @@
...
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNet.OData.Query;
+using Microsoft.AspNet.OData.Test.Common;
+using Xunit;
+using System.Threading;
```
xuzhg, line 11: Sort

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,249 @@
...
+
+namespace Microsoft.AspNet.OData.Test.Routing
+{
+    public class ODataActionSelectorTest
+    {
+
+        //public static TheoryDataSet<Dictionary<string, object>, string, string, Type, string, Type[]> ScenariosWithCorrectAction
+        public static TheoryDataSet<Dictionary<string, object>, (string Method, string Body), (Type ControllerType, string ActionName, Type[] Signature)> ScenariosWithCorrectAction
```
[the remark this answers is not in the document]
habbes: The tuple was meant to organize related parameters together, but also to avoid having to many type parameters to the `TheoryDataSet` generic class because it only supports a limited number of type parameters.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: I didn't want to make it a breaking change since it's a public class. But I can remove the `= null` if that's not a problem.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RoutingConventionHelpers.cs
```diff
@@ -302,6 +304,33 @@ public static void AddFunctionParameterToRouteData(this IWebApiControllerContext
...
+
+        /// <summary>
+        /// Increments the number of keys, navigation properties and operation parameters
+        /// parameters. This count is tracked by the <see cref="ODataRouteConstants.KeyCount"/> key
+        /// in the controller context route data.
+        /// </summary>
+        /// <param name="routeValues"></param>
+        public static void IncrementKeyCount(IDictionary<string, object> routeValues)
+        {
+            if (routeValues.TryGetValue(ODataRouteConstants.KeyCount, out object count))
```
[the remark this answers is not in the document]
habbes: Noted

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: I've updated it. The parameters are no longer optional.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RefRoutingConvention.cs
habbes: What do you mean by this?

I added the `AddNavigationPropertyToRouteData()` extension method to encapsulate the adding the navigation property from the link segment to the route data manually. And also to ensure that the key count was incremented.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RoutingConventionHelpers.cs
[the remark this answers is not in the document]
habbes: Is the RoutingConventionsStore property also available on WebApi AspNet classic?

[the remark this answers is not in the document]
habbes: Done. Found it on aspnet classic in `request.ODataProperties().RoutingConventionsStore`

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -10,14 +11,18 @@
...
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Binders;
+using Microsoft.AspNetCore.Mvc.Routing;
 #if NETSTANDARD2_0
     using Microsoft.AspNetCore.Mvc.Internal;
     using Microsoft.AspNetCore.Routing;
     using Microsoft.Extensions.Logging;
 #else
-    using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing;
```
marabooy, line 22: Seems to be imported either way. any reason we have this in the else?

habbes: Good point.

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -126,31 +162,90 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
-                    ? candidates.Where(c => c.Id == matchedCandidates.FirstOrDefault().Id).FirstOrDefault()
-                    : candidates.FirstOrDefault();
+                return matchedCandidates.Select(c => c.ActionDescriptor).FirstOrDefault();
             }
 
             return _innerSelector.SelectBestCandidate(context, candidates);
         }
 
-        private bool TryMatch(IList<ParameterDescriptor> parameters, IList<string> availableKeys, ODataOptionalParameter optionalWrapper)
+        private bool TryMatch(
```
marabooy, line 201: Could you add a comment to showcase what is happening in this function as the number of params has increased.

marabooy (changes requested):
LGTM

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -160,26 +255,96 @@ private bool TryMatch(IList<ParameterDescriptor> parameters, IList<string> avail
...
+
                 return false;
             }
 
             return true;
         }
 
+        private bool RequestHasBody(RouteContext context)
+        {
+            string method = context.HttpContext.Request.Method.ToLowerInvariant();
```
gathogojr, line 282: @habbes Is there any need in calling `.ToLowerInvariant()` here if you still have to supply the `StringComparison` comparison parameter in the line that follows?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -126,31 +162,90 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
                 if (availableKeys.Contains(parameterName))
                 {
                     continue;
                 }
 
+                if (conventionsStore != null)
+                {
+                    // the convention store can contain the parameter as key
+                    // with a nested property (e.g. customer.Name) 
+                    if (conventionsStore.Keys.Any(k => k.Contains(p.Name)))
```
gathogojr, line 238: @habbes Do we care about case here?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -126,31 +162,90 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+                    {
+                        continue;
+                    }
+                }
+
+                if (context.HttpContext.Request.Query.ContainsKey(p.Name))
```
gathogojr, line 244: @habbes Do we care about case here?

gathogojr: From the documentation, I can't tell for certain if the comparison is case-sensitive https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.querycollection.containskey?view=aspnetcore-3.1

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -160,26 +255,96 @@ private bool TryMatch(IList<ParameterDescriptor> parameters, IList<string> avail
...
+            string method = context.HttpContext.Request.Method.ToLowerInvariant();
+            return method.Equals("post", StringComparison.OrdinalIgnoreCase) ||
+                method.Equals("put", StringComparison.OrdinalIgnoreCase) ||
+                method.Equals("patch", StringComparison.OrdinalIgnoreCase) ||
+                method.Equals("merge", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private bool ActionAcceptsMethod(ControllerActionDescriptor action, string method)
+        {
+            if (action == null)
```
gathogojr, line 291: @habbes Shouldn't we instead use a `Debug.Assert()` or `Contract.Assert()` here?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using System.IO;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using System.Linq;
+#if NETCOREAPP2_0
+using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Logging;
```
gathogojr, line 17: @habbes Are these indents necessary?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            var parameters = method.GetParameters();
+            if (parameters.Length != action.Parameters.Count)
+            {
+                return false;
+            }
+
+            for (var i = 0; i < parameters.Length; i++)
+            {
+                if (parameters[i].Name != action.Parameters[i].Name ||
```
gathogojr, line 54: @habbes I know this class is for test purposes but I hope that if in the actual logic the parameter types are playing any part we're careful with the comparison so we don't trip at implicit conversion... e.g. from `int` to `long`

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <param name="actionName">Action name of the action descriptors to return</param>
+        /// <param name="routeBuilder"></param>
+        /// <param name="actionSelector"></param>
+        /// <param name="actionDescriptors"></param>
+        public static void SetupActionSelector(System.Type controllerType,
+           string actionName,
+           out IRouteBuilder routeBuilder,
+           out ODataActionSelector actionSelector,
+           out IReadOnlyList<ControllerActionDescriptor> actionDescriptors)
+#if NETCOREAPP2_0
```
gathogojr, line 78: @habbes `routeBuilder` and `actionDescriptors` seems to be getting initialized the same way in both `#if` and `#else` block. Maybe you could move that logic above this line?

gathogojr: Maybe even `serviceProvider` and `modelBinderFactory` - DRY ;)

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// Create route context with the specified with the specified request data
+        /// </summary>
+        /// <param name="routeBuilder"></param>
+        /// <param name="actionName">Name of the action ebging routed to</param>
+        /// <param name="routeDataValues">Key-value pairs to add to the route data</param>
+        /// <param name="method">HTTP request method</param>
+        /// <param name="bodyContent">Request body content</param>
+        /// <returns></returns>
+        public static RouteContext SetupRouteContext(IRouteBuilder routeBuilder, string actionName,
+            Dictionary<string, object> routeDataValues,string method, string bodyContent)
```
gathogojr, line 121:
```suggestion
            Dictionary<string, object> routeDataValues, string method, string bodyContent)
```

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+                keyCount++;
+            }
+
+            routingConventionsStore[ODataRouteConstants.KeyCount] = keyCount;
+
+            request.Method = method;
+
+            if (bodyContent != null)
+            {
+                var stream = new MemoryStream();
```
gathogojr, line 145: @habbes Easier way to achieve this: `var stream = new MemoryStream(Encoding.UTF8.GetBytes(bodyContent));`

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+                var writer = new StreamWriter(stream);
+                writer.Write(bodyContent);
+                request.Body = stream;
```
gathogojr, line 148: @habbes Is assignment the safe option here (in terms of object ownership) or should you be using `Body.CopyTo()` or .Body.CopyToAsync()`?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,248 @@
...
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNet.OData.Test.Common;
+using Xunit;
+
+namespace Microsoft.AspNet.OData.Test.Routing
+{
+    public class ODataActionSelectorTest
+    {
+
+        //public static TheoryDataSet<Dictionary<string, object>, string, string, Type, string, Type[]> ScenariosWithCorrectAction
```
gathogojr, line 17: @habbes Remove instead of commenting?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,248 @@
...
+        public void Get(int key, int relatedKey) { }
+
+        public void Get(int key, int relatedKey, int navigationProperty) { }
+
+        public void Get(int key, ODataPath path) { }
+    }
+
+    public class KeyBodyController : TestODataController
+    {
+        public void Put(int key, Delta<object> dt) { }
```
gathogojr, line 221: @habbes Should this be patch?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,248 @@
...
+    public class NoParamController : TestODataController
+    {
+        public void Get() { }
+    }
+
+    public class ExtraParametersWithOwnModelBindersController
+    {
+        public void Get(int key, System.Threading.CancellationToken cancellationToken) { }
+
+        public void Put(int key, System.Threading.CancellationToken cancellationToken, Delta<object> delta) { }
```
gathogojr, line 238: @habbes Should this be patch?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    public class ExtraParametersWithoutModelBindersController
+    {
+        public void Get(int key, UnknownModel other) { }
+        public void Put(int key, UnknownModel other, Delta<object> delta) { }
```
gathogojr, line 244: @habbes Should this be patch?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    public class UnknownModel { }
+}
```
gathogojr, line 248: @habbes Add newline...

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
```diff
@@ -0,0 +1,248 @@
...
+            {
+                return new TheoryDataSet<Dictionary<string, object>, (string, string), (Type, string, Type[])>
+                {
+                    // Actions with single key parameter
+                    {
+                        new Dictionary<string, object> { { "key", 1 } },
+                        ("GET", null),
+                        (typeof(SingleKeyController), "Get", new [] { typeof(int) })
+                    },
+                    {
```
gathogojr, line 30: @habbes It'd help if each scenario had a comment like the one above

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/RoutingConventionHelpers.cs
```diff
@@ -341,6 +304,33 @@ public static void AddFunctionParameterToRouteData(this IWebApiControllerContext
...
+        public static void AddNavigationPropertyToRouteData(this IWebApiControllerContext controllerContext,
+            NavigationPropertyLinkSegment navigationLinkSegment)
+        {
+            controllerContext.RouteData.Add(ODataRouteConstants.NavigationProperty, navigationLinkSegment.NavigationProperty.Name);
+            IncrementKeyCount(controllerContext.Request.Context.RoutingConventionsStore);
+        }
+
+        /// <summary>
+        /// Increments the number of keys, navigation properties and operation parameters
+        /// parameters. This count is tracked by the <see cref="ODataRouteConstants.KeyCount"/> key
```
gathogojr, line 320: @habbes `parameters` is repeated - `operation parameters parameters`

src/Microsoft.AspNet.OData.Shared/Routing/ODataRouteConstants.cs
```diff
@@ -67,5 +67,11 @@ public static class ODataRouteConstants
...
         /// </summary>
         public static readonly string OptionalParameters = typeof(ODataOptionalParameter).FullName;
+
+        /// <summary>
+        /// Route data key that tracks the number of key segments,
+        /// navigation properties and operation parameters in the request URI
+        /// </summary>
+        public static readonly string KeyCount = "ODataRouteKeyCount";
     }
 }
```
gathogojr, line 77: @habbes Add newline

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
```diff
@@ -106,5 +112,32 @@ private static HttpResponseMessage CreateErrorResponse(HttpRequestMessage reques
...
+            }
+
+            // navigationProperty is optional in some cases, therefore an action
+            // should not be rejected simply because it does not declare a navigationProperty parameter
+            if (routeData.Values.ContainsKey(ODataRouteConstants.NavigationProperty))
+            {
+                keyCount -= 1;
+            }
+
+            if (parameters.Count < keyCount)
```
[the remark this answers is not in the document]
gathogojr: Or is that only a likelyhood when there's a navigation property?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -10,14 +11,16 @@
...
+using Microsoft.AspNetCore.Routing;
 #if NETSTANDARD2_0
     using Microsoft.AspNetCore.Mvc.Internal;
-    using Microsoft.AspNetCore.Routing;
     using Microsoft.Extensions.Logging;
-#else
-    using Microsoft.AspNetCore.Routing;
 #endif
 
+
```
gathogojr, line 23: nit: Probably added this line inadvertently?

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -126,31 +160,102 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+                return matchedCandidates.Select(c => c.ActionDescriptor).FirstOrDefault();
             }
 
             return _innerSelector.SelectBestCandidate(context, candidates);
         }
 
-        private bool TryMatch(IList<ParameterDescriptor> parameters, IList<string> availableKeys, ODataOptionalParameter optionalWrapper)
+        /// <summary>
+        /// Checks whether the a controller action matches the current route by comparing the parameters
+        /// of te action with the data in the route.
```
gathogojr, line 201:
```suggestion
        /// of the action with the data in the route.
```

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
```diff
@@ -3080,7 +3081,7 @@ public class Microsoft.AspNet.OData.Routing.DefaultODataPathValidator : Microsof
 }
 
 public class Microsoft.AspNet.OData.Routing.ODataActionSelector : IActionSelector {
-	public ODataActionSelector (Microsoft.AspNetCore.Mvc.Infrastructure.IActionDescriptorCollectionProvider actionDescriptorCollectionProvider, Microsoft.AspNetCore.Mvc.Internal.ActionConstraintCache actionConstraintProviders, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory)
+	public ODataActionSelector (Microsoft.AspNetCore.Mvc.Infrastructure.IActionDescriptorCollectionProvider actionDescriptorCollectionProvider, Microsoft.AspNetCore.Mvc.Internal.ActionConstraintCache actionConstraintProviders, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.AspNetCore.Mvc.ModelBinding.IModelBinderFactory modelBinderFactory, Microsoft.AspNetCore.Mvc.ModelBinding.IModelMetadataProvider modelMetadataProvider)
```
gathogojr, line 3084: @habbes Could this result into a breaking change? E.g. in the event that someone had overridden the `ODataActionSelector`?

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore3x.OData.PublicApi.bsl
```diff
@@ -3251,7 +3252,7 @@ public class Microsoft.AspNet.OData.Routing.DefaultODataPathValidator : Microsof
 }
 
 public class Microsoft.AspNet.OData.Routing.ODataActionSelector : IActionSelector {
-	public ODataActionSelector (Microsoft.AspNetCore.Mvc.Infrastructure.IActionSelector innerSelector)
+	public ODataActionSelector (Microsoft.AspNetCore.Mvc.Infrastructure.IActionSelector innerSelector, Microsoft.AspNetCore.Mvc.ModelBinding.IModelBinderFactory modelBinderFactory, Microsoft.AspNetCore.Mvc.ModelBinding.IModelMetadataProvider modelMetadataProvider)
```
gathogojr, line 3255: https://github.com/OData/WebApi/pull/2168/files#r482919044

src/Microsoft.AspNet.OData/Routing/ODataActionSelector.cs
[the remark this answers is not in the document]
habbes: In this case it will return `true`  if it reaches this point. If `parameters.Count` is less than the number of keys, it means that there are not enough parameters on the action to match the keys in the uri. But if `parameters.Count` is greater, it could mean that there are optional parameters on the action or parameters that are bound by values other than those in the uri.

Something else to note is that the `ODataActionSelector` in Asp.Net classic works a bit differently. It uses Asp.Net's internal action selector to return the correct match, so the method I added here just tries to ensure that the selected match is not the wrong one, that's why it does not implement rules as strict as the one in AspNetCore

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
habbes: Since the `stream` is created here locally and not used anywhere else, I think it's safe. The stream is created for the sole purpose of becoming the request body. But if what you propose is better practice, I could switch to it. But would the following code be better (it combines your previous suggestion and avoids creating a temporary variable):
```c#
request.Body = new MemoryStream(Encoding.UTF8.GetBytes(bodyContent));
```

test/UnitTest/Microsoft.AspNetCore.OData.Test/Routing/ODataActionSelectorTest.cs
habbes: It was meant to be `Put`. But I assume you're asking that because I used `Delta<object>`. I can change it to `Patch` to match the standard practice, it shouldn't affect the validity of the test.

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
habbes: I had made the extra arguments optional, but @xuzhg didn't think they should be optional since they are resolved by DI: https://github.com/OData/WebApi/pull/2168#discussion_r432594077

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
habbes: I believe that the parameters are not converted to lower case when added to `conventionsStore`

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
habbes: This is not used in the actual logic, it's just used to compare the selected controller action with the expected controller method for the test assertions.

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ODataActionSelectorTestHelper.cs
habbes: I thought it would be easier to read if I completely separated the netcore2.0 and netcore3.1 implementations, that way you only focus on the block that's relevant to the version you're viewing at any one time.

But if you believe that keeping the common part in one place and only separating the parts that are different is easier to read, then I can change it.

gathogojr (approved):
LGTM

marabooy (approved):

houbi56:
This is a BREAKING CHANGE and should be documented as such!

marabooy:
@houbi56 have you encountered any issues relating to this pr?

houbi56:
Our code was built to work around #1973 #2075. Went from 7.4.1 to 7.5.2. This changed the routing to another Action for the GET request, and we got 404 when sending request to the DELETE action.

Our OData service is built for an ERP and we have a single generic controller (serving ~300 entities) which relied on the previous "relaxed" behaviour. Unfortunately, these Actions did not have receive a lot of requests and we didn't have sufficient test coverage so this was discovered after deployment to production.

AdamCaviness:
@marabooy  @Houbi56 We are also hitting this same DELETE action problem for the same reason you describe. The action parameter count comparison in `ODataActionSelector.TryMatch` on line 271 evaluates that our general purpose Delete action in our base controller that takes a single string key is no longer a match.

PR #2169, opened 2020-05-21
Copy query settings correctly

KanishManuja-MS:
### Issues

Setting DefaultQuerySettings using the extensions on the route builder does nothing. 

### Description

Now, we will respect the default query settings. 

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/DefaultQuerySettings.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/DefaultQuerySettings.cs
@@ -64,5 +64,16 @@ namespace Microsoft.AspNet.OData.Query
         /// Gets or sets a value indicating whether the service will use skiptoken or not.
         /// </summary>
         public bool EnableSkipToken { get; set; }
+
+        internal void CopySettings(DefaultQuerySettings settings)
+        {
+            this.EnableCount = settings.EnableCount;
+            this.EnableExpand = settings.EnableExpand;
+            this.EnableFilter = settings.EnableFilter;
+            this.EnableOrderBy = settings.EnableOrderBy;
+            this.EnableSelect = settings.EnableSelect;
+            this.EnableSkipToken = settings.EnableSkipToken;
+            this.MaxTop = settings.MaxTop;
+        }
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
@@ -34,7 +34,7 @@ namespace Microsoft.AspNet.OData.Extensions
                 throw Error.InvalidOperation(SRResources.MissingODataServices, nameof(DefaultQuerySettings));
             }
 
-            querySettings = defaultQuerySettings;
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/DefaultQuerySettings.cs
```diff
@@ -64,5 +64,16 @@ public class DefaultQuerySettings
         /// Gets or sets a value indicating whether the service will use skiptoken or not.
         /// </summary>
         public bool EnableSkipToken { get; set; }
+
+        internal void CopySettings(DefaultQuerySettings settings)
```
xuzhg, line 68: Clone?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
```diff
@@ -34,7 +34,7 @@ public static void SetDefaultQuerySettings(this IServiceProvider serviceProvider
                 throw Error.InvalidOperation(SRResources.MissingODataServices, nameof(DefaultQuerySettings));
             }
 
-            querySettings = defaultQuerySettings;
+            querySettings.CopySettings(defaultQuerySettings);
```
mikepizzo, line 37: So we are overwriting whatever settings were obtained from Dependency Injection with the settings passed in to SetDefaultQuerySettings. Is this the expected behavior?

Note that we are copying MaxTop from the default settings returned from DI *after* we have already applied semantics based on the value passed in (see line 26).  Should we move the logic on line 26-29 to *after* we do copysettings?

src/Microsoft.AspNet.OData.Shared/Query/DefaultQuerySettings.cs
```diff
@@ -64,5 +64,16 @@ public class DefaultQuerySettings
...
+        {
+            this.EnableCount = settings.EnableCount;
+            this.EnableExpand = settings.EnableExpand;
+            this.EnableFilter = settings.EnableFilter;
+            this.EnableOrderBy = settings.EnableOrderBy;
+            this.EnableSelect = settings.EnableSelect;
+            this.EnableSkipToken = settings.EnableSkipToken;
+            this.MaxTop = settings.MaxTop;
```
mikepizzo, line 76: Should we only copy this over if this.MaxTop.HasValue == true, or would we need to copy it over in order to allow clearing a MaxTop through DI?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceProviderConfigExtenions.cs
KanishManuja-MS: Yes. Looking at the name of the extension method it should reset the default query settings with the current object. 

Moved the three lines after copying.

src/Microsoft.AspNet.OData.Shared/Query/DefaultQuerySettings.cs
KanishManuja-MS: I think we need to copy to reset the MaxTop to even null.

xuzhg:
Merged at: https://github.com/OData/WebApi/commit/4d8c2baeaf5bdbe650d5b7d80b0753f7efba6229

Issue #2089, opened 2020-03-17
Memory leak when using OData 7.3.0 and EF Core 3.1.2


vladimir-kanev:
I forked this [sample](https://github.com/hassanhabib/OData3Beta) to add the EF Core support. It is configured to use in-memory DB, but the effect is the same with MySQL for example.

When doing OData requests the memory keeps growing and is never released. 

[Here](https://github.com/vladimir-kanev/OData3Beta) is the forked version with the memory leak problem.

### Assemblies affected

AspNetCore OData 7.3.0 and EF Core 3.1.2

### Reproduce steps

1. Run the [demo](https://github.com/vladimir-kanev/OData3Beta) project.
2. Make several hundred GET requests to: localhost:3509/odata/persons?$filter=Name+in+('some name')&$orderby=Date+asc

### Expected result

Memory to stop increasing and get released at some point.

### Actual result

Memory consumption keeps growing and is never released.

![MemoryConsumption](https://user-images.githubusercontent.com/41116615/76781306-30957a00-67b7-11ea-9c75-7c51eca1876d.png)

habbes:
Hi @vladimir-kanev I am curious to know whether the 3 GC collections that appear on the screenshot were triggered automatically or whether you forced them with `GC.Collect()`. If they were triggered automatically, it's possible that the GC did not perform an exhaustive collection since your system was not under a lot of memory pressure.

I tried your demo app and I was able to reproduce the behaviour. When the automatic GC ran, I did not observe a memory reduction. I then added an endpoint that class `GC.Collect()`, which performs garbage collection across all generations. When I called this endpoint, I did see the memory usage decreasing as you can see in the attached screenshot.

![image](https://user-images.githubusercontent.com/8460169/79202639-93277780-7e42-11ea-9b7a-fefcb14908c1.png)

However, memory seems to reduce up to a point. And looking at the objects in the managed heap after GC, there's still 100's of instances of classes like PersonContext and EntityQueryable<Person>. Maybe this has to do with the caching that you mentioned. But I have not investigated this.

![image](https://user-images.githubusercontent.com/8460169/79203389-a129c800-7e43-11ea-89ce-18793ecb0bad.png)

Just a few more clarifying questions:
- How much free memory did you have on your system
- Did your memory continue to grow as a result of making additional calls to the endpoint, or was there any other activity running on your server? In my case there were memory spikes when I hit the endpoint multiple times then curve remained flat until I flood the endpoint again.

[...]

PR #2172, opened 2020-05-21
Fix EfCore memory leak when query contains collection constant

habbes:
### Issues

This pull request fixes issue #2089 

### Description

EF core caches compiled queries so it doesn't have to re-compile the same query twice. The compiled queries are usually parameterized so that if the same query is made with different input values, the same cached query will be re-used. However, if a constant expression is inside a lambda, it's not parameterized, this means the same query with different values will result in queries being compiled and cache (see https://github.com/dotnet/efcore/issues/10535#issuecomment-375116203). If such queries are made frequently, then the size of the cache quickly grows over time unnecessarily (see https://dzone.com/articles/investigating-a-memory-leak-in-entity-framework-co).

WebApi has a setting `EnableConstantParameterization` which is enabled by default to prevent this issue. It uses `LinqParameterContainer.ParameterizeConstant()` to wrap the constant value in a property of a class instance and replaces the `ConstantExpression` with a `PropertyExpression`. This mimics how the compiler handles closures. This parameterization was applied to Uri Parser's `ConstantNode` but not to `CollectionConstantNode`. This PR simply applies it to `CollectionConstanNode`.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -1360,7 +1360,12 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                 castedList.Add(member);
             }
 
-            return Expression.Constant(castedList);
+            if (QuerySettings.EnableConstantParameterization)
+            {
+                return LinqParameterContainer.Parameterize(listType, castedList);
+            }
+
+            return Expression.Constant(castedList, listType);
         }
 
         private Expression BindCastSingleValue(SingleValueFunctionCallNode node)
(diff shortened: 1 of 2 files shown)


habbes:
I did notice some strange behaviour related to the number of cached entries before my PR. When I flooded the server with 1000 queries, then took a snapshot of the memory (using VS's studio diagnostic tools) I would find that the number of entries in `Microsoft.Extensions.Caching.Memory.MemoryCache` has increased by about 1000 (from 2 to 1002). But when I flooded it with subsequent queries, the entries do not increase, in fact they decreased to about 964. At this point I cannot explain this behaviour, but then again I'm not very well versed with the internal mechanisms of EF Core caching and what the expected behaviour should be. After I implemented the fix, the number of entries only increase by 1 (from 2 to 3).

xuzhg:
:thumbsup: 

---
In reply to: [632511089](https://github.com/OData/WebApi/pull/2172#issuecomment-632511089) [](ancestors = 632511089)

xuzhg:
Looks good to me. Thanks.

---
In reply to: [417038318](https://github.com/OData/WebApi/pull/2172#issuecomment-417038318) [](ancestors = 417038318)

xuzhg:
Merged at: 8de67a7e4e40973f3a03227863069a9ec14cbdab

PR #2189, opened 2020-06-04
Enable convention routing for operation import

xuzhg:
### Issues

*This pull request fixes issue ##2188.*

### Description

*Enable convention routing for operation import

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/OperationImportRoutingConvention.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/OperationImportRoutingConvention.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.OData.Edm;
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNet.OData.Routing.Conventions
+{
+    /// <summary>
+    /// An implementation of <see cref="IODataRoutingConvention"/> that handles function invocations.
+    /// </summary>
+    public partial class OperationImportRoutingConvention : NavigationSourceRoutingConvention
+    {
+        /// <inheritdoc/>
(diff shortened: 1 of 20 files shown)


src/Microsoft.AspNet.OData.Shared/Routing/Conventions/OperationImportRoutingConvention.cs
```diff
@@ -0,0 +1,63 @@
...
+                ODataRequestMethod.Post == controllerContext.Request.Method)
+            {
+                // The same action name may be used multiple times within a schema provided there is at most one unbound overload
+                operationImportSegment = (OperationImportSegment)odataPath.Segments[0];
+            }
+
+            string actionName = SelectAction(operationImportSegment, actionMap);
+            if (actionName != null)
+            {
+                controllerContext.AddFunctionParameterToRouteData(operationImportSegment);
```
gathogojr, line 40: @xuzhg Curious, does this handle only function imports?

xuzhg: Yes, it's for operation import convention routing.

gathogojr: @xuzhg What I meant was, does this line handle both action and function imports (i.e. operations imports in general) or just function imports? And the reason for my question is because the function the manner in which the function is implemented suggests it only handles functions... for example [here](https://github.com/OData/WebApi/pull/2189/files#diff-81f57b7f872fe2d221327178aae70fd4R236)

xuzhg: actually action/action import parameters are passed in payload body, we don't need to add parameter for action/action import. 
Here, i don't want to test it before the function call. But, maybe it's better to move the test out of the method.

gathogojr: Got it. I had forgotten we use Post for actions

xuzhg:
merged at: https://github.com/OData/WebApi/commit/8e2c3a7c48e9c32519408c87ff8ea13b0e0fb961

Issue #2190, opened 2020-06-07
Feature: Support Long URL Pattern in WebAPI OData


gathogojr:
While Http imposes no limits on the length of a URL, many hosting environments (including IIS) pose limitations which may restrict the client's ability to request the exact set of data that they are interested in. This is especially true as we push clients to use explicit select statements.

OData v4.01 introduced a [pattern](http://docs.oasis-open.org/odata/odata/v4.01/csprd05/part2-url-conventions/odata-v4.01-csprd05-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody) for executing long URLs by placing the query options in the body of a POST request to the resource. This makes it easy to reuse existing routing and resource/workload resolution logic while passing URLs with arbitrarily long query strings to the workload in the body.

**Example**
POST http://ServiceRoot/Users/$query
Content-Type: text/plain
```
$filter=id in ('<guid1>','<guid2>','<guid3>')
```

### Assemblies affected

Microsoft.AspNet.OData/Microsoft.AspNetCore.OData 7.4.1

PR #2191, opened 2020-06-07
Implement support for long URL pattern

gathogojr:
### Issues

*This pull request fixes issue #2190 .*

### Description

OData v4.01 introduced a [pattern](http://docs.oasis-open.org/odata/odata/v4.01/csprd05/part2-url-conventions/odata-v4.01-csprd05-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody) for executing long URLs by placing the query options in the body of a POST request to the resource. This makes it easy to reuse existing routing and resource/workload resolution logic while passing URLs with arbitrarily long query strings to the workload in the body.

**Example**
POST http://ServiceRoot/Users/$query
Content-Type: text/plain
```
$filter=id in ('<guid1>','<guid2>','<guid3>')
```
In this PR, this objective is achieved by intercepting the request and checking if it is a POST request targeted at a resource path ending in `/$query`. Where the outcome of that check is positive, a search is conducted in the request DI container for a parser than can parse the request body and return the query options part of the OData URL.

After the parsing the request body for the query string, the request is transformed from a POST to a GET and the URL also transformed to accommodate the query string. The request is then handled the same way any GET request for a resource would be handled. 

It is anticipated that in the future OData WebApi may support different parsers for different media types (e.g. `application/json`). Only support for `text/plain` media type is included in this PR but the implementation is extensible to make it easy to add other parsers in the future. Each parser is tagged with the media type its designed to handle. The parsers are added to the request DI container when the service is being bootstrapped.

Note that this change affects Microsoft.AspNet.OData and Microsoft.AspNetCore.OData project

- [x] *Test cases added*


src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,66 @@
...
+    public partial class TextPlainODataQueryOptionsParser
+    {
+        /// <inheritdoc/>
+        public string Parse(Stream requestStream)
+        {
+            string queryString = string.Empty;
+
+            MemoryStream memoryStream = new MemoryStream();
+            TextReader reader;
+
```
gathogojr, line 20: Just to preempt possible questions around this seemingly "unattractive" approach, the following "attractive" approaches are flagged by the analyzer in **Microsoft.AspNet.OData** project. The analyzer in **Microsoft.AspNetCore.OData** however didn't appear to make a fuss.

1. [CA2202: Do not dispose object multiple times](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2202-do-not-dispose-objects-multiple-times?view=vs-2015) against the following two approaches:

    a. Approach # 1
    ```c#
    using (var memoryStream = new MemoryStream())
    {
        requestStream.CopyTo(memoryStream);
        memoryStream.Position = 0;
    
        using (var reader = new StreamReader(memoryStream))
        {
            string result = reader.ReadToEnd();
        }
    }
    ```

    b. Approach # 2
    ```c#
    var memoryStream = new MemoryStream();

    try
    {
        requestStream.CopyTo(memoryStream);
        memoryStream.Position = 0;
    
        using (var reader = new StreamReader(memoryStream))
        {
            string result = reader.ReadToEnd();
        }
    }
    catch
    {
        // CopyTo() may throw ObjectDisposedException
        memoryStream.Dispose();
    }
    ```

2. [CA2000: Dispose objects before losing scope](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2000-dispose-objects-before-losing-scope?view=vs-2015) against the following approach:
    ```c#
    var memoryStream = new MemoryStream();

    using (var reader = new StreamReader(memoryStream))
    {
        requestStream.CopyTo(memoryStream);
        memoryStream.Position = 0;
    
        string result = reader.ReadToEnd();
    }
    ```

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore3x/Microsoft.Test.E2E.AspNetCore3x.OData.csproj
```diff
@@ -28,6 +28,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.5" />
     <PackageReference Include="Microsoft.OData.Client" Version="7.6.1" />
+    <PackageReference Include="System.Diagnostics.EventLog" Version="4.6.0" />
```
gathogojr, line 31: This fixes an issue being reported on ADO even for other PRs:
```
Error CS1069: The type name 'EventLog' could not be found in the namespace 'System.Diagnostics'
```
More details https://stackoverflow.com/questions/57688012/reading-from-eventlog-error-the-type-name-eventlog-could-not-be-found-in-the

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
...
+                reader = new StreamReader(memoryStream);
+            }
+            catch
+            {
+                memoryStream.Dispose();
+
+                throw new ODataException(SRResources.CannotParseQueryOptionsPayload);
+            }
+
+            using (reader)
```
odero, line 41: How does this work if an exception was thrown and `reader` was never set?

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
odero: What about:
```c#
var memoryStream = new MemoryStream();
try {
    using(var reader = new StreamReader(memoryStream))
    {
    // do your parsing ...
    }
}
catch
{
    throw new ODataException(SRResources.CannotParseQueryOptionsPayload);
}
finally
{
    if (memoryStream != null) memoryStream.Dispose();
}

File: src/Microsoft.AspNet.OData/Routing/ODataPathRouteConstraint.cs [hunk 4]
```diff
@@ -47,10 +47,31 @@ public virtual bool Match(HttpRequestMessage request, IHttpRoute route, string p
                 object oDataPathValue;
                 if (values.TryGetValue(ODataRouteConstants.ODataPath, out oDataPathValue))
                 {
+                    string odataPath = oDataPathValue as string;
```
odero, line 50:
```suggestion
                    string oDataPath = oDataPathValue as string;
```

src/Microsoft.AspNet.OData/Routing/ODataPathRouteConstraint.cs
```diff
@@ -47,10 +47,31 @@ public virtual bool Match(HttpRequestMessage request, IHttpRoute route, string p
...
+                    Func<IServiceProvider> requestContainerFactory = () =>
+                    {
+                        // Delegate is reused. We need to ensure only one request container is created
+                        if (request.Properties.ContainsKey(Extensions.HttpRequestMessageExtensions.RequestContainerKey))
```
odero, line 54: Might want to consider using a `TryGetValue` instead

src/Microsoft.AspNetCore.OData/Routing/ODataPathRouteConstraint.cs
```diff
@@ -48,13 +48,36 @@ public virtual bool Match(HttpContext httpContext, IRouter route, string routeKe
...
+                        return request.CreateRequestContainer(RouteName);
+                    };
+
+                    // Check whether the request is a POST targeted at a resource path ending in /$query
+                    if (request.IsQueryRequest(odataPath))
+                    {
+                        request.TransformQueryRequest(requestContainerFactory);
+
+                        odataPath = odataPath.Substring(0, odataPath.LastIndexOf('/' + ODataRouteConstants.QuerySegment, StringComparison.OrdinalIgnoreCase));
+                        values[ODataRouteConstants.ODataPath] = odataPath;
```
odero, line 71: What does this do? Or why is it necessary?

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
...
+                    requestStream.Position = 0;
+                }
+
+                requestStream.CopyTo(memoryStream);
+                memoryStream.Position = 0;
+                reader = new StreamReader(memoryStream);
+            }
+            catch
+            {
+                memoryStream.Dispose();
```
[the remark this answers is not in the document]
gathogojr: @odero We actually don't want to dispose the stream when an exception is not thrown. The stream reader in the section that follows depends on it. We should only dispose it if an exception occurred and we terminate execution of the method prematurely

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
gathogojr: @odero The only scenario where the `reader` is not set is when an exception is thrown in the first `try...catch` and this code will never be executed in that scenario. In the `try...catch` above, we'll dispose the memory stream and exit from the function prematurely if either `CopyTo` or stream reader initialization fails. We shouldn't get to this line if we don't have an initialized reader and a memory stream

src/Microsoft.AspNetCore.OData/Routing/ODataPathRouteConstraint.cs
gathogojr: @odero When `/$query` is appended to the resource path, the `oDataPath` key in the route values dictionary will be mapped to the value `/Users/$query`. Say after parsing the request body we end up with query options string like this: `$filter=Role eq 'Administrator'&$orderby=Id`.

The reconstructed OData URL will look something like this: `/Users?$filter=Role eq 'Administrator'&$orderby=Id`

Note that we drop the `/$query` segment from the URL completely at this point and this is what necessitates updating the route values as well, replacing `/Users/$query` with `/Users`

gathogojr:
https://github.com/OData/WebApi/pull/2191#discussion_r436558480
@odero Will try this out

src/Microsoft.AspNetCore.OData/Routing/ODataPathRouteConstraint.cs
odero: I see so could you just set it as
```c#
values[ODataRouteConstants.ODataPath] = request.Path.Value;
// or set oDataPath as
oDataPath = request.Path.Value;
```

gathogojr:
@odero In relation to your suggested approach here https://github.com/OData/WebApi/pull/2191#discussion_r436558480
Its also categorized under **CA2202: Do not dispose object multiple times** because the check for nullity in this statement `if (memoryStream != null) memoryStream.Dispose();` is not sufficient for purposes of checking that the stream has not been disposed. I did try quite a number of similar permutations

src/Microsoft.AspNetCore.OData/Routing/ODataPathRouteConstraint.cs
gathogojr: @odero The problem here is the `oDataPath` value from the route dictionary is not an exact match of `request.Path` value. For instance, based on the example above, `oDataPath` would initially have the value `Users/$query` while `request.Path` would have the value `/Users/$query` (starts with a /).

Whichever route you take, some string manipulation will be involved. I just found it easier/safer to drop `/$query` segment

gathogojr: Here's a screenshot

image

gathogojr: Notice that the Request Path includes even the route prefix

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
...
+
+using System;
+using System.IO;
+using Microsoft.OData;
+using Microsoft.AspNet.OData.Common;
+using System.Diagnostics;
+
+namespace Microsoft.AspNet.OData.Query
+{
+    public partial class TextPlainODataQueryOptionsParser
```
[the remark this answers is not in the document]
gathogojr: @odero David and I spent some time debating the naming. In the end we settled on `TextPlainODataQueryOptionsParser`. Looking at it now I think `PlainTextODataQueryOptionsParser` is a better fit

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
g2mula: @gathogojr, it is possible to suppress analyzers when you have a valid justification...

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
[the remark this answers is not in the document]
gathogojr: @odero FYI, Microsoft.AspNet.OData project analyzer had this to say about my attempt to use `PlainTextODataQueryOptionsParser`
```
The compound word 'PlainText' in type name 'PlainTextODataQueryOptionsParser' exists 
as a discrete term. 
If your usage is intended to be single word, case it as 'Plaintext'
```
Unless you have any strong feelings in support of `PlainText...` I'll adopt `PlaintextODataQueryOptionsParser`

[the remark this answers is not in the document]
gathogojr: ☝️ After seeking opinions from 2 other people...

src/Microsoft.AspNet.OData/Routing/ODataPathRouteConstraint.cs
gathogojr: Fixed

src/Microsoft.AspNet.OData/Routing/ODataPathRouteConstraint.cs
gathogojr: Fixed

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.OData;
+using Microsoft.AspNet.OData.Common;
+using System.Diagnostics;
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -410,6 +411,64 @@ public static IEnumerable<IODataRoutingConvention> GetRoutingConventions(this Ht
...
+
+            string queryString = queryOptionsParser.Parse(request.Content.ReadAsStreamAsync().Result);
+
+            string requestPath = request.RequestUri.LocalPath;
+            requestPath = requestPath.Substring(0, requestPath.LastIndexOf('/' + ODataRouteConstants.QuerySegment, StringComparison.OrdinalIgnoreCase));
+
+            Uri requestUri = request.RequestUri;
+            request.RequestUri = new Uri(string.Format(
+                CultureInfo.InvariantCulture,
+                "{0}://{1}{2}{3}{4}",
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -410,6 +411,64 @@ public static IEnumerable<IODataRoutingConvention> GetRoutingConventions(this Ht
...
+            }
+
+            IServiceProvider requestContainer = requestContainerFactory();
+
+            // Fetch parsers available in the request container for parsing the query options in the request body
+            IEnumerable<IODataQueryOptionsParser> queryOptionsParsers = requestContainer.GetRequiredService<IEnumerable<IODataQueryOptionsParser>>();
+            IODataQueryOptionsParser queryOptionsParser = queryOptionsParsers.FirstOrDefault(
+                d => d.MediaTypeMapping.TryMatchMediaType(request) > 0);
+
+            string mediaType = request.Content.Headers.ContentType != null ? request.Content.Headers.ContentType.MediaType : string.Empty;
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.AspNet.OData/Query/TextPlainODataQueryOptionsParser.cs
gathogojr: @g2mula In this case I'm not 100% sure that I have a valid justification to suppress the warning so it might be better to err on the side of caution. I have used the nested `using`s liberally before but as I scrutinized it this time, I realized the argument by the analyzer would be valid for certain scenarios.

I think the reason why we're able to get away with it is best captured in [this answer](https://stackoverflow.com/a/4274769/752739) to a StackOverflow post I came across - `MemoryStream` does not have any unmanaged resources to dispose. Plus unlike other streams, I'd expect `MemoryStream` to be less prone to typical failures

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.OData;
+using Microsoft.AspNet.OData.Common;
+
+namespace Microsoft.AspNet.OData.Query
+{
+    [SuppressMessage("Microsoft.Naming", "CA1702:CompoundWordsShouldBeCasedCorrectly", Justification="Spelling correct in this context")]
+    public partial class PlainTextODataQueryOptionsParser
+    {
+        /// <inheritdoc/>
+        public string Parse(Stream requestStream)
```
habbes, line 15: Out of curiosity, why isn't this method async?

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    // Based on OData OASIS Standard, the request body is expected to contain the query options part of OData URL 
+                    // and MUST use the same percent-encoding as in URLs (especially: no spaces, tabs, or line breaks allowed) 
+                    // and MUST follow the expected syntax rules
+
+                    // For consideration: support multiline
+                    // - Would it be a single query option per line?
+                    // - Would the parser be responsible for adding the & separator?
+                    // - Would the parser try to detect the & separator and add where necessary?
+                    string result = reader.ReadToEnd();
```
habbes, line 53: Why isn't reading the stream async?

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptionsParserFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+            IList<IODataQueryOptionsParser> parsers = new List<IODataQueryOptionsParser>();
+
+            // Create an instance of each implementing type
+            foreach(var type in implementingTypes)
+            {
+                parsers.Add((IODataQueryOptionsParser)Activator.CreateInstance(type));
+            }
+
+            return parsers;
+        }
```
habbes, line 39: Why is this method getting all parsers implementation in the assembly? If this is initializing the default, built-in parsers, then in my opinion the list should be created manually (like how the routing conventions does it). And if the user wants to add custom parser, then they should do so explicitly in the service provider, that way there's better control on which parsers are registered, and in what order. I believe the order matters, right?

src/Microsoft.AspNet.OData.Shared/Query/IODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Reads and parses the content of a <see cref="T:System.IO.Stream" /> 
+        /// into a query options part of an OData URL.
+        /// </summary>
+        /// <param name="requestStream">A <see cref="T:System.IO.Stream" /> containing the query options.</param>
+        /// <returns>A string representing the query options part of an OData URL.</returns>
+        string Parse(Stream requestStream);
+    }
+}
```
habbes, line 33: In my opinion, instead of making the interface expose a single `MediaTypeMapping` that it supports, I think it should expose a method that encapsulates the logic of figuring out whether the parser can handle the request, e.g. `CanParse(context)` that returns a boolean. If the goal is to be extensible, I think this would be more flexible.

Furthermore, I think its consistent with the design of other AspNetCore concepts like input formatters:
```c#
interface IInputFormatter
{
   bool CanRead(InputFormatterContext);
   void ReadAsync(InputFormatterContext);
}
```

habbes: After reading more of your code, I've realized that `MediaTypeMapping` does have a `TryMatchMediaType` which technically does what I was saying, so I don't know if it's still worth going with the `CanParse(context)` approach, but would like to hear your thoughts on this.

habbes:
Could you also add unit tests that demonstrate making $query requests and seeing them getting routed to the correct controller methods? I think this would be beneficial for future PRs to notice that they've broken a feature without having to wait for the e2e tests to complete. I've you've already added such tests and I missed them, then just ignore this.

habbes:
Also, could you show an example of how a user would register a new custom query options parser?

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,70 @@
...
+                // Reset request stream position if possible
+                if (requestStream.CanSeek)
+                {
+                    requestStream.Position = 0;
+                }
+
+                requestStream.CopyTo(memoryStream);
+                memoryStream.Position = 0;
+                reader = new StreamReader(memoryStream);
+            }
```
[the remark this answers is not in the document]
gathogojr: @habbes The reason for copying the content to a memory stream is because to read the content I'm using a reader and how they behave is that they dispose the underlying stream after they're done. I don't want that to happen here. The responsibility of disposing that request stream lies somewhere else.
The reason for attempting to rewind the stream if it supports seeking to make sure it's at position 0 before copying commences. If the stream position was (say) at position 13 and you called `CopyTo` (or `CopyAsync()`), the copying starts at that position.
It's also worth mentioning that not all streams support seeking.
Luckily, we intercept the request very early in the lifecycle such that we're almost always guaranteed that it's at position 0.

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -451,6 +452,57 @@ public static void DeleteRequestContainer(this HttpRequest request, bool dispose
...
+
+            string queryString = queryOptionsParser.Parse(request.Body);
+
+            // Request path starts with a /
+            string requestPath = request.Path.Value;
+            requestPath = requestPath.Substring(0, requestPath.LastIndexOf('/' + ODataRouteConstants.QuerySegment, StringComparison.OrdinalIgnoreCase));
+
+            request.Path = new PathString(requestPath);
+            request.QueryString = new QueryString(queryString);
+            request.Method = HttpMethods.Get;
```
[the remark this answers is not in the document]
gathogojr: @habbes I'm not exact sure I understand what you're asking here. However, just to note that this feature rides on the content-type, based on OASIS OData standard [here](http://docs.oasis-open.org/odata/odata/v4.01/csprd05/part2-url-conventions/odata-v4.01-csprd05-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody). In addition, the request from the client will be a POST

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptionsParserFactory.cs
gathogojr: I considered that approach as well. See **Alternative Approaches and Enhancements** section in  the design document I shared internally.

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
gathogojr: In Microsoft.AspNet.OData project, asynchronous processing is not 

> Also, could you show an example of how a user would register a new custom query options parser?

@habbes Here's how:
```c#
IList<IODataQueryOptionsParser> queryOptionsParsers = ODataQueryOptionsParserFactory.Create();
queryOptionsParsers.Insert(0, new CustomODataQueryOptionsParser());

app.UseMvc(routeBuilder =>
{                
	// ...
	routeBuilder.MapODataServiceRoute(
		routeName: "odata",
		routePrefix: "odata",
		configureAction: containerBuilder => containerBuilder
			// ...
			.AddService(Microsoft.OData.ServiceLifetime.Singleton, typeof(IEnumerable<IODataQueryOptionsParser>), _ => queryOptionsParsers), 
			// ...
	);
	// ...
});
```

gathogojr:
> Could you also add unit tests that demonstrate making $query requests and seeing them getting routed to the correct controller methods? I think this would be beneficial for future PRs to notice that they've broken a feature without having to wait for the e2e tests to complete. I've you've already added such tests and I missed them, then just ignore this.

@habbes [Here](https://github.com/OData/WebApi/pull/2191/files#diff-aff302550e6cee06e873fd659ea65487R86) are the tests for /$query request

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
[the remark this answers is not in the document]
habbes: I'm still wondering whether it's a good idea to rewind to position 0. We don't expect the request stream to have been read at this point. If someone did it, either they did it in error or they knew what they were doing. If they did it in error, I don't think we should attempt to fix that by rewinding it, I think it should just give incomplete results and the user would trace the issue back to their mistake, otherwise it will be harder to detect that mistake. If they did it intentionally for some obscure reason, then I don't think we should interfere with that decision. Maybe they have this hack where the payload has some random string at the beginning followed by the actual query options, so they decide to read that initial part of the stream in advance so that the odata parser can pick it up from where the query options actually start. I know that's very contrived and arbitrary scenario, but I think any scenario that intentionally leads the stream not to start at 0 would be. If it should be rewound, I think that should be the responsibility of the one who read it prematurely to rewind it.

gathogojr:
> Also, could you show an example of how a user would register a new custom query options parser?

@habbes Here's how:
```c#
IList<IODataQueryOptionsParser> queryOptionsParsers = ODataQueryOptionsParserFactory.Create();
queryOptionsParsers.Insert(0, new CustomODataQueryOptionsParser());

app.UseMvc(routeBuilder =>
{                
	// ...
	routeBuilder.MapODataServiceRoute(
		routeName: "odata",
		routePrefix: "odata",
		configureAction: containerBuilder => containerBuilder
			// ...
			.AddService(Microsoft.OData.ServiceLifetime.Singleton, typeof(IEnumerable<IODataQueryOptionsParser>), _ => queryOptionsParsers), 
			// ...
	);
	// ...
});
```

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
gathogojr: @habbes Because async processing wasn't/isn't a design consideration for Microsoft.AspNet.OData project

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
gathogojr: @habbes I also considered making it async. However, based on how this entire feature is used, I didn't find any tangible benefit that would be accrued from that. We receive a request and the determination how we should proceed to process the request needs to happen pronto. Note that the execution starts from `Match` method that is also not async. I'd be open to reconsider the approach though if making it async is the better choice

src/Microsoft.AspNet.OData.Shared/Query/IODataQueryOptionsParser.cs
gathogojr: @habbes Input and output formatters also use the `MediaTypeMapping` approach to determine how a resource should be materialized/serialized

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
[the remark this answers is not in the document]
gathogojr: @habbes I don't mind removing it. The stream in `HttpRequest` doesn't support seeking either way. But just as an FYI, we actually do manipulate stream position quite significantly across OData repos:
- https://github.com/search?q=%22Stream.Position+%3D+%22+repo%3AOData%2FWebApi+path%3Asrc&type=Code&ref=advsearch&l=&l=
- https://github.com/search?q=%22Stream.Position+%3D+%22+repo%3AOData%2Fodata.net+path%3Asrc&type=Code&ref=advsearch&l=&l=
However, the execution flow in this case would support your argument

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
gathogojr: @habbes I eventually with async processing to address an issue where a particular test was failing intermittently

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
gathogojr: @habbes I eventually with async processing to address an issue where a particular test was failing intermittently

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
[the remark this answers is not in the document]
gathogojr: @habbes Fixed

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DollarQuery/DollarQueryTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+#if NETCORE
+            // KestrelServerLimits.MaxRequestLineSize equals 8192 so setting query string length to that value
+            // should be enough to make us breach the threshold
+            // NOTE: Change of limit could cause test to start failing
+            var queryStringLength = 8192;
+#else
+            // Microsoft Owin Hosting is used
+            // 414 Request-URI too long failure code expected
+            var queryStringLength = 16384;
```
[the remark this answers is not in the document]
gathogojr: @habbes They are defaults. We don't override them in the tests.
The default for Kestrel Server was easy to find: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits.maxrequestlinesize?view=aspnetcore-3.1

The one for Microsoft Owin Hosting took a lot of digging through internet resources to find. The docs are actually quite scanty\imprecise. The figure of 16384 applied by the Owin framework for request filtering was the result of scouring lots of materials for "collaborating evidence" and trying it out practically. The MaxRequestLength limit in this case applies to both the request line and the headers:
- https://support.microsoft.com/en-us/help/820129/http-sys-registry-settings-for-windows
- http://www.pointshared.com/2012/05/23/bad-request-request-too-long/
- https://stackoverflow.com/questions/1097651/is-there-a-practical-http-header-length-limit
- https://www.grouppolicy.biz/2013/06/how-to-configure-iis-to-support-large-ad-token-with-group-policy/
- https://social.microsoft.com/Forums/Windows/de-DE/e22a3ac9-930a-4705-8766-bf84c08ccf5d/sending-more-than-16384-bytes-from-wcf?forum=wcf
- https://flylib.com/books/en/1.271.1.70/1/

src/Microsoft.AspNet.OData.Shared/Query/IODataQueryOptionsParser.cs
gathogojr: > In my opinion, instead of making the interface expose a single `MediaTypeMapping` that it supports, I think it should expose a method that encapsulates the logic of figuring out whether the parser can handle the request, e.g. `CanParse(context)` that returns a boolean. If the goal is to be extensible, I think this would be more flexible.
> 
> Furthermore, I think its consistent with the design of other AspNetCore concepts like input formatters:
> 
> ```cs
> interface IInputFormatter
> {
>    bool CanRead(InputFormatterContext);
>    void ReadAsync(InputFormatterContext);
> }
> ```

@habbes I implemented this suggestion

marabooy (approved):
LGTM

habbes (approved):

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptionsParserFactory.cs
mikepizzo: I agree that I'd rather have the parsers explicitly added rather than look across all loaded types to see if they implement IQueryOptionsParser.

---
In reply to: [437212714](https://github.com/OData/WebApi/pull/2191#discussion_r437212714) [](ancestors = 437212714)

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -410,6 +412,57 @@ public static IEnumerable<IODataRoutingConvention> GetRoutingConventions(this Ht
...
+        {
+            if (requestContainerFactory == null)
+            {
+                throw Error.ArgumentNull("requestContainerFactory");
+            }
+
+            IServiceProvider requestContainer = requestContainerFactory();
+
+            // Fetch parsers available in the request container for parsing the query options in the request body
+            IEnumerable<IODataQueryOptionsParser> queryOptionsParsers = requestContainer.GetRequiredService<IEnumerable<IODataQueryOptionsParser>>();
```
mikepizzo, line 443: >requestContainer.GetRequiredService<IEnumerable<IODataQueryOptionsParser>> [](start = 72, length = 74)

I would rather see ODataQueryOptionsParserFactory as an instantiable class that has a GetQueryOptionParser(request) method that returns the query option parser. I could pass my custom query option parser, or the default query option parser, to the constructor or add my custom query option parser to that class. Lines 444-454 could be encapsulated within that method.

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -410,6 +412,57 @@ public static IEnumerable<IODataRoutingConvention> GetRoutingConventions(this Ht
...
+                oDataPath?.TrimEnd('/').EndsWith('/' + ODataRouteConstants.QuerySegment, StringComparison.OrdinalIgnoreCase) == true;
+        }
+
+        /// <summary>
+        /// Transforms a POST request targeted at a resource path ending in $query into a GET request. 
+        /// The query options are parsed from the request body and appended to the request URL.
+        /// </summary>
+        /// <param name="request">The request.</param>
+        /// <param name="requestContainerFactory">Request container factory.</param>
+        internal static void TransformQueryRequest(this HttpRequestMessage request, Func<IServiceProvider> requestContainerFactory)
```
mikepizzo, line 433: >, Func<IServiceProvider> requestContainerFactory [](start = 82, length = 48)

why pass in requestContainerFactory rather than using request.GetRequestContainer() as is done in the other extension methods?

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                await requestStream.CopyToAsync(memoryStream);
+                memoryStream.Position = 0;
+                reader = new StreamReader(memoryStream);
+            }
+            catch
+            {
+                memoryStream.Dispose();
+
+                throw new ODataException(SRResources.CannotParseQueryOptionsPayload);
+            }
```
mikepizzo, line 58: What is the value of copying to a memory stream, versus reading directly from the requestStream?

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -410,6 +412,57 @@ public static IEnumerable<IODataRoutingConvention> GetRoutingConventions(this Ht
...
+            }
+
+            Stream requestStream = request.Content.ReadAsStreamAsync().Result;
+            string queryString = queryOptionsParser.ParseAsync(requestStream).Result;
+
+            string requestPath = request.RequestUri.LocalPath;
+            requestPath = requestPath.Substring(0, requestPath.LastIndexOf('/' + ODataRouteConstants.QuerySegment, StringComparison.OrdinalIgnoreCase));
+
+            Uri requestUri = request.RequestUri;
+            request.RequestUri = new UriBuilder(requestUri.Scheme, requestUri.Host, requestUri.Port, requestPath, queryString).Uri;
```
mikepizzo, line 463: >queryString [](start = 114, length = 11)

Note that, according to spec (see [http://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody](http://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody))
   "Query options specified in the request body and query options specified in the request URL are processed together."
i.e., if query options exist in the original request, they should be merged with query options specified in the body.

Note that the spec also says (see [http://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_SystemQueryOptions](http://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_SystemQueryOptions)):
   "The same system query option, irrespective of casing or whether or not it is prefixed with a $, MUST NOT be specified more than once for any resource."
so we should return an error if the same query option is specified both in the body and in the request.

src/Microsoft.AspNet.OData/Routing/ODataPathRouteConstraint.cs
```diff
@@ -47,10 +47,32 @@ public virtual bool Match(HttpRequestMessage request, IHttpRoute route, string p
...
+                    Func<IServiceProvider> requestContainerFactory = () =>
+                    {
+                        // Delegate is reused. We need to ensure only one request container is created
+                        object requestContainer;
+                        if (request.Properties.TryGetValue(Extensions.HttpRequestMessageExtensions.RequestContainerKey, out requestContainer))
+                        {
+                            return (IServiceProvider)requestContainer;
+                        }
+
+                        return request.CreateRequestContainer(RouteName);
```
mikepizzo, line 60: >request.CreateRequestContainer(RouteName) [](start = 31, length = 41)

If there is a per-route request container, should it take precedence over one associated with the request? Couldn't CreateRequestContainer(RouteName) default to reqeust.Properties.TryGetValue(...) if there was no request container explicitly associated with the route?

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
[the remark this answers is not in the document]
mikepizzo: I think he's asking if we should remove headers describing the request body (i.e., content-type and content-length) from the transformed query request, since they don't apply to the transformed GET request. Might be nice to clean them up, but not sure it's really a lot of value, and we would have to enumerate the full list of such headers to strip out.  As long as nothing downstream complains about these non-relevant headers, I think I'd just leave them.

---
In reply to: [437210686](https://github.com/OData/WebApi/pull/2191#discussion_r437210686) [](ancestors = 437210686)

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -451,6 +454,56 @@ public static void DeleteRequestContainer(this HttpRequest request, bool dispose
...
+            }
+
+            string queryString = queryOptionsParser.ParseAsync(request.Body).Result;
+
+            // Request path starts with a /
+            string requestPath = request.Path.Value;
+            requestPath = requestPath.Substring(0, requestPath.LastIndexOf('/' + ODataRouteConstants.QuerySegment, StringComparison.OrdinalIgnoreCase));
+
+            request.Path = new PathString(requestPath);
+            request.QueryString = new QueryString(queryString);
```
mikepizzo, line 503: >QueryString(queryString) [](start = 38, length = 24)

as per above, the spec says that query options in the body can be combined with query options in the request.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DollarQuery/DollarQueryTests.cs
```diff
@@ -0,0 +1,205 @@
...
+                builder.EntitySet<DollarQueryCustomer>("DollarQueryCustomers");
+
+            EntitySetConfiguration<DollarQueryOrder> dollarQueryOrders =
+                builder.EntitySet<DollarQueryOrder>("DollarQueryOrders");
+
+            return builder.GetEdmModel();
+        }
+
+        [Theory]
+        [MemberData(nameof(ODataQueryOptionsData))]
```
mikepizzo, line 87: >[MemberData(nameof(ODataQueryOptionsData))] [](start = 8, length = 43)

add a variant where there are some query options specified in the body and some specified in the request URL, both where they don't overlap (in which case, they should be merged) and where they do overlap (in which case, an error should be thrown)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DollarQuery/DollarQueryTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+            var response = await this.Client.SendAsync(request);
+            Assert.True(response.IsSuccessStatusCode);
+
+            var contentAsString = response.Content.ReadAsStringAsync().Result;
+            Assert.Contains("\"value\":[{\"Id\":1,\"Name\":\"Customer Name 1\"}]", contentAsString);
+        }
+
+        [Fact]
+        public async Task ODataQueryOptionsInRequestBody_ForUnsupportedMediaType()
```
mikepizzo, line 125: >    public async Task ODataQueryOptionsInRequestBody_ForUnsupportedMediaType() [](start = 4, length = 78)

any similar tests for a media type supported through a custom queryoptionparser?

mikepizzo (changes requested):
:clock1:

mikepizzo (changes requested):
:clock1:

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
gathogojr: @mikepizzo If I call `request.GetRequestContainer()` at this point, the method will in turn call `request.CreateRequestContainer(request.ODataFeature().RuoteName)` and the problem will be that `RouteName` will be `null` at that point. We initialize the `Path` and `RouteName` properties of `request.ODataFeature()` [much further below](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNetCore.OData/Routing/ODataPathRouteConstraint.cs#L60) after the call to `GetODataPath()`.
Things are made worse by the fact that there's a call to `request.CreateRequestContainer(RouteName)` on the `GetODataPath()` [method call](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNetCore.OData/Routing/ODataPathRouteConstraint.cs#L57). The `CreateRequestContainer()` method is designed in such a way that it throws an exception if a request container already exists.
I therefore opted to add [this delegate](https://github.com/OData/WebApi/pull/2191/files#diff-a08977e326b20ea2c1a30644a20643beR51) that both `TransformQueryRequest()` and `GetODataPath()` methods invoke. The delegate won't blow up if the request container already exists.

src/Microsoft.AspNet.OData/Query/PlainTextODataQueryOptionsParser.cs
gathogojr: @mikepizzo The idea of copying to a memory stream was to avoid tinkering with the request stream - at the time I thought that `CopyTo`/`CopyToAsync` do not progress the stream position. I just did some tests that confirmed that thinking to be wrong so I'll change the logic to just read directly from the request stream

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
gathogojr: @mikepizzo Fixed.
Just to note, currently, when a system query option is specified more than once, e.g., `/People?$orderby=Name&$orderby=Name`, we respond with a 404 - NotFound. It made sense to keep the logic here lean by just merging the options in the request body to those in the request url and have any repeated query options caught downstream the same place it happens currently.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DollarQuery/DollarQueryTests.cs
gathogojr: @mikepizzo Add tests to cover those two scenarios:
https://github.com/OData/WebApi/pull/2191/files#diff-9e480fb8b0555db1128de67981ff5e0aR123
https://github.com/OData/WebApi/pull/2191/files#diff-9e480fb8b0555db1128de67981ff5e0aR142

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
gathogojr: @mikepizzo Fixed

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
gathogojr: Fixed

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DollarQuery/DollarQueryTests.cs
gathogojr: Added a test to cover that scenario
https://github.com/OData/WebApi/pull/2191/files#diff-9e480fb8b0555db1128de67981ff5e0aR321
Here is the custom query options cursor for media type _text/xml_
https://github.com/OData/WebApi/pull/2191/files#diff-9e480fb8b0555db1128de67981ff5e0aR244

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
gathogojr: @mikepizzo Here's how I anticipated plugging in of a custom query options parser to work
https://github.com/OData/WebApi/pull/2191/files#diff-9e480fb8b0555db1128de67981ff5e0aR305
I'd like to know if your idea about making the class instantiatable was primarily to support plugging in of parsers or whether my approach suffices.
I added the `GetQueryOptionsParser(request)` method to the factory method

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
mikepizzo: Slightly cleaner (so you don't have to check IsNullOrWhitespace twice):

            if (!string.IsNullOrWhiteSpace(queryOptions))
            {
                   if(string.IsNullOrWhiteSpace(queryString)
                   {
                        queryString = '?' + queryOptions;
                   }
                   else
                  {
                        queryString += ('&' + queryOptions);
                  }
            }


---
In reply to: [488699478](https://github.com/OData/WebApi/pull/2191#discussion_r488699478) [](ancestors = 488699478)

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
mikepizzo: :thumbs up:

---
In reply to: [488693700](https://github.com/OData/WebApi/pull/2191#discussion_r488693700) [](ancestors = 488693700)

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DollarQuery/DollarQueryTests.cs
mikepizzo: nice!

---
In reply to: [488802104](https://github.com/OData/WebApi/pull/2191#discussion_r488802104) [](ancestors = 488802104)

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
mikepizzo: Slightly different that I had envisioned it, but I think this is fine.  I liked having a strongly typed class/interface passed in to the service container, but IEnumerable<IODataQueryOptionsParser> works fine.

---
In reply to: [488835406](https://github.com/OData/WebApi/pull/2191#discussion_r488835406) [](ancestors = 488835406)

mikepizzo (approved):
:shipit:

mikepizzo:
Add a very minor comment, but overall looks good -- thanks!

mikepizzo (approved):
:shipit:

odero (approved):

PR #2196, opened 2020-06-13
Fix NullReferenceException using $select on null value type array

bdebaere:
### Issues

When calling $select on arrays with a value type (or string), and its value is null, an error is thrown. This is due to the fact that null.Select(it => it).ToList() obviously throws a NullReferenceException. Null values for arrays of these types are common when dealing with navigation properties which are null (e.g. $expand=parent where parent is null).
This pull request avoids Expression.Call for Select and Expression.Call for ToList on arrays of these types, as they are not necessary anyway.

### Description

A check is added whether the collection currently being projected is an array with a value type (or string). If that is the case, then Expression.Call for Select and Expression.Call for ToList are avoided.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -1050,6 +1050,23 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                 }
             }
 
+            // Avoid calling source.Select($it => $it).ToList() on array types.
+            if (source.Type.IsArray)
+            {
+                if (_settings.HandleNullPropagation == HandleNullPropagationOption.True)
+                {
+                    // source == null ? null : projectedCollection
+                    return Expression.Condition(
+                           test: Expression.Equal(source, Expression.Constant(null)),
+                           ifTrue: Expression.Constant(null, source.Type),
+                           ifFalse: source);
+                }
+                else
+                {
+                    return source;
+                }
+            }
+
             // expression
             //      source.Select((ElementType element) => new Wrapper { })
             var selectMethod = GetSelectMethod(elementType, projection.Type);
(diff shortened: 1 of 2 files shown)


bdebaere:
@xuzhg Looks like the build is failing outside of the scope of this pull request, as this test succeeds locally. If I am wrong about this, do let me know!

xuzhg:
@bdebaere Can you see the "re-run" button? [I clicked the re-run button for you. :) ]
Click the "detail" from the "WebApi-master-pipeline-Rolling", under azure-pipeline, you can find a "re-run" button to rerun the whole PR, if You think the failing test case is not related to your changes, or it can pass locally. If it's still failing, need more investigation.

bdebaere:
@xuzhg Thanks for your help! I can't see the re-run button sadly, but I am happy you clicked it. Is the button available on github or is it on Azure itself? Either way, since the re-run, the build succeeds. Just one test fails in the Azure pipeline, but I can't figure out why as nothing I changed influences that test.

This is the query generated locally (and locally the test succeeds):
```sql
SELECT 
    [Project1].[C2] AS [C1], 
    [Project1].[C3] AS [C2], 
    [Project1].[Name] AS [Name], 
    [Project1].[C4] AS [C3], 
    [Project1].[C1] AS [C4]
    FROM ( SELECT 
        [GroupBy1].[A1] AS [C1], 
        [GroupBy1].[K1] AS [Name], 
        [GroupBy1].[K2] AS [C2], 
        [GroupBy1].[K3] AS [C3], 
        N'TotalPrice' AS [C4]
        FROM ( SELECT 
            [Join1].[K1] AS [K1], 
            [Join1].[K2] AS [K2], 
            [Join1].[K3] AS [K3], 
            SUM([Join1].[A1_0]) AS [A1]
            FROM ( SELECT 
                [Extent1].[Name] AS [K1], 
                1 AS [K2], 
                N'Name' AS [K3], 
                [Extent2].[Price] AS [A1_0]
                FROM  [dbo].[Customers] AS [Extent1]
                LEFT OUTER JOIN [dbo].[Orders] AS [Extent2] ON [Extent1].[Order_Id] = [Extent2].[Id]
            )  AS [Join1]
            GROUP BY [K1], [K2], [K3]
        )  AS [GroupBy1]
    )  AS [Project1]
    ORDER BY [Project1].[C1] ASC
```

Issue #2177, opened 2020-05-28
Problem when generating OData nextlink skiptoken with date containing offset


YoannaAtanasova:
There is a problem when using Odata Web API with dates that are not in UTC time zone. When we get the first page of data there is a nextlink with skip token generated, like so:

"@odata.nextLink":"https://localhost:44331/odata/weatherforecast?$skiptoken=StartDate-2020-05-29T08:48:29.2464819+01:00"

This is problematic. because of the special characters like '+', that cannot be passed as a query string parameter. So when we use this link we get the following exception:

The query specified in the URI is not valid. The DateTimeOffset text '2020-05-29T08:48:29.2464819' should be in format 'yyyy-mm-ddThh:mm:ss('.'s+)?(zzzzzz)?' and each field value is within valid range.

For the generated nextlink to work the skip token must be encoded, like so:

"@odata.nextLink":"https://localhost:44331/odata/weatherforecast?$skiptoken=StartDate-2020-05-29T08%3A48%3A29.2464819%2B01%3A00"

For the purpose of reproducing the issue I have created a simple .Net Core 3.1 project, with the latest version of the library:
https://github.com/YoannaAtanasova/ODataOffsetProblemPOC

xuzhg:
@YoannaAtanasova I remember it supports to customize the SkipToken through DI. See https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Extensions/ContainerBuilderExtensions.cs#L54
Can you try it?

YoannaAtanasova:
Hi @xuzhg ,

I don't need to customize the SkipToken.
I need the SkipToken values to be encoded as you do with the Filters values.

Otherwise the default SkipToken is useless when special characters are included.

I think it's a bug that need to be fixed.

Do you suggest me to try to find a workaround for the issue?

YoannaAtanasova:
The issue is resolved with version 7.5

PR #2197, opened 2020-06-15
Issue 2177 - Fix skip token encoding

YoannaAtanasova:
### Issues

This pull request fixes issue #2177 

### Description

With the changes in this PR we now encode all values that are put in the generated skiptoken.

- [x] *Test cases added*


YoannaAtanasova:
> Rather than pass in a boolean for whether or not to encode, which is going to be hard for the service to decide, can we just add logic to DefaultSkipTokenHandler to encode any date, string, or other values that may have suspect characters?

I am not sure I understand you. I am not passing a boolean and the encoding is not conditional.

mikepizzo:
Sorry; added my comment to the wrong PR.  This one looks fine.

---
In reply to: [650314297](https://github.com/OData/WebApi/pull/2197#issuecomment-650314297) [](ancestors = 650314297)

PR #2199, opened 2020-06-17
Draft PR - nextlink encode

Sreejithpin:
### Issues
Draft PR for nextlink

### Description

- [ ] *Test cases added*

### Additional work necessary


mikepizzo (changes requested):
Rather than pass in a boolean for whether or not to encode, which is going to be hard for the service to decide, can we just add logic to DefaultSkipTokenHandler to encode any date, string, or other values that may have suspect characters?  See similar logic in PR #2197.

PR #2201, opened 2020-06-17
Modify ToInt16 to ToUInt16

bdebaere:
I understand that there is some work underway to modify the project structure so that this becomes irrelevant. Nonetheless I'd like to add this change of ToUInt16, instead of the current ToInt16, just in case it takes longer than 3 years to accomplish.


xuzhg:
Merged at: https://github.com/OData/WebApi/commit/235b62601000e44cd6480809e391ce898ac70f31

Issue #1645, opened 2018-10-16
Support: Add interface to create ODataQueryOptions via a mock framework


ynauls:
Add interface/support to unit test ODataQueryOptions with a mock framework.  I implemented logic using the following sample from here [Server-Driven Paging](https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/supporting-odata-query-options).  The code works, however, i am unable to unit test due to complexity of creating ODataQueryOptions

```c#
public PageResult<Product> Get(ODataQueryOptions<Product> options)
{
    ODataQuerySettings settings = new ODataQuerySettings()
    {
        PageSize = 5
    };

    IQueryable results = options.ApplyTo(_products.AsQueryable(), settings);

    return new PageResult<Product>(
        results as IEnumerable<Product>, 
        Request.GetNextPageLink(), 
        Request.GetInlineCount());
}

```

### Assemblies affected

Microsoft.AspNet.OData.Query

### Reproduce steps

Create unit test sample outside "Microsoft.AspNet.OData" package.  

### Expected result

Able to easily create unit test for logic passing in a ODataQueryOptions<T>

### Actual result

unable to create unit tests

freeranger:
@ynauls https://github.com/OData/WebApi/issues/1352#issuecomment-397385479 may help you with your unit testing - not nice, but it works :)

ynauls:
@freeranger, thank you, i saw that earlier and it works.  I would def like some support out of the box :)

PR #2214, opened 2020-06-30
#1645 pull request

kccarter76:
### Issues

This pull request fixes issue #1645

### Description

implements the feature that allows the use of IODataQueryOptions as a parameter in a controller method

- [x] *Test cases added*

### Additional work necessary

the new feature implemented needs to be documented.


kccarter76:
I have zero experience with azure pipelines, do I need to duplicate the interface somewhere or have I missed something which needed to be setup in my VS 2019?

kccarter76:
it looks like I forgot about stuff that was not .net core

PR #2231, opened 2020-07-14
Enhance batch accept header decision

iwate:
### Issues

Enhance batch content type decision.  

### Description

 `CreateODataBatchResponseAsync`'s content type decision is uncertain.

Currently

| Request Accept | Request Content-Type | Response Content-Type |
|:------------------|:-------------------------|:---------------------------|
|                           |  application/json          |application/json             |
|                           |  *                                  | multipart/mixed             |
|  multipart/mixed   |  *                                  | multipart/mixed             |
|   application/json  |  *                                  |  application/json            |
|  multipart/mixed, application/json   |  *                                  | application/json              |
|  application/json, multipart/mixed   |  *                                  | application/json              |
|  multipart/mixed; q=0.8, application/json; q=0.8   |  *                                  | application/json              |
|  application/json; q=0.8, multipart/mixed; q=0.8   |  *                                  | application/json              |
|  multipart/mixed; q=0.9, application/json; q=0.5   |  *                                  | application/json              |
|  application/json; q=0.5, multipart/mixed; q=0.9   |  *                                  | application/json              |
|  multipart/mixed; q=0.5, application/json; q=0.9   |  *                                  | application/json              |
|  application/json; q=0.9, multipart/mixed; q=0.5   |  *                                  | application/json              |

Expected

| Request Accept | Request Content-Type | Response Content-Type |
|:------------------|:-------------------------|:---------------------------|
|                           |  application/json          |application/json             |
|                           |  *                                  | multipart/mixed             |
|  multipart/mixed   |  *                                  | multipart/mixed             |
|   application/json  |  *                                  |  application/json            |
|  multipart/mixed, application/json   |  *                                  | multipart/mixed              |
|  application/json, multipart/mixed   |  *                                  | application/json              |
|  multipart/mixed; q=0.8, application/json; q=0.8   |  *                                  | multipart/mixed             |
|  application/json; q=0.8, multipart/mixed; q=0.8   |  *                                  | application/json              |
|  multipart/mixed; q=0.9, application/json; q=0.5   |  *                                  | multipart/mixed              |
|  application/json; q=0.5, multipart/mixed; q=0.9   |  *                                  | multipart/mixed             |
|  multipart/mixed; q=0.5, application/json; q=0.9   |  *                                  | application/json              |
|  application/json; q=0.9, multipart/mixed; q=0.5   |  *                                  | application/json              |

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/UnitTest/Microsoft.AspNetCore.OData.Test/Batch/ODataBatchHttpRequestMessageExtensionsTest.cs
+++ b/test/UnitTest/Microsoft.AspNetCore.OData.Test/Batch/ODataBatchHttpRequestMessageExtensionsTest.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNet.OData.Batch;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNetCore.Http;
+using Microsoft.OData;
+using Xunit;
+
+namespace Microsoft.AspNet.OData.Test.Batch
+{
+    public class ODataBatchHttpRequestMessageExtensionsTest
+    {
+        private static ODataMessageQuotas _odataMessageQuotas = new ODataMessageQuotas { MaxReceivedMessageSize = Int64.MaxValue };
+
+        [Theory]
+        // if no accept header, return multipart/mixed
+        [InlineData(null, "multipart/mixed")]
+
+        // if accept is multipart/mixed, return multipart/mixed
+        [InlineData(new[] { "multipart/mixed" }, "multipart/mixed")]
+        
+        // if accept is application/json, return application/json
+        [InlineData(new[] { "application/json" }, "application/json")]
+        
+        // if accept is application/json with charset, return application/json
+        [InlineData(new[] { "application/json; charset=utf-8" }, "application/json")]
+        
+        // if multipart/mixed is high proprity, return multipart/mixed
+        [InlineData(new[] { "multipart/mixed;q=0.9", "application/json;q=0.5" }, "multipart/mixed")]
+        [InlineData(new[] { "application/json;q=0.5", "multipart/mixed;q=0.9" }, "multipart/mixed")]
+        
+        // if application/json is high proprity, return application/json
+        [InlineData(new[] { "application/json;q=0.9", "multipart/mixed;q=0.5" }, "application/json")]
+        [InlineData(new[] { "multipart/mixed;q=0.5", "application/json;q=0.9" }, "application/json")]
+        
+        // if proprities are same, return first
+        [InlineData(new[] { "multipart/mixed;q=0.9", "application/json;q=0.9" }, "multipart/mixed")]
+        [InlineData(new[] { "multipart/mixed", "application/json" }, "multipart/mixed")]
+        
+        // if proprities are same, return first
+        [InlineData(new[] { "application/json;q=0.9", "multipart/mixed;q=0.9" }, "application/json")]
+        [InlineData(new[] { "application/json", "multipart/mixed" }, "application/json")]
+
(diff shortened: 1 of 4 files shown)


test/UnitTest/Microsoft.AspNetCore.OData.Test/Batch/ODataBatchHttpRequestMessageExtensionsTest.cs
```diff
@@ -0,0 +1,88 @@
...
+        
+        // if application/json is high proprity, return application/json
+        [InlineData(new[] { "application/json;q=0.9", "multipart/mixed;q=0.5" }, "application/json")]
+        [InlineData(new[] { "multipart/mixed;q=0.5", "application/json;q=0.9" }, "application/json")]
+        
+        // if proprities are same, return first
+        [InlineData(new[] { "multipart/mixed;q=0.9", "application/json;q=0.9" }, "multipart/mixed")]
+        [InlineData(new[] { "multipart/mixed", "application/json" }, "multipart/mixed")]
+        
+        // if proprities are same, return first
```
KenitoInc, line 44: - Rename `proprities` to `priorities` or `relative quality factor` or `q-value`.
- I am interested to know what the protocol says in situations where media types have the same `relative quality factor`. I can't get anything here https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

iwate: Thank you for your review.

I fix the typo.

> I am interested to know what the protocol says in situations where media types have the same relative quality factor.

The protocol docs says

> Each media-range MAY be followed by one or more accept-params, beginning with the "q" parameter for indicating a relative quality factor. https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

I consider that accept-params need `q` parameters if media-range has over two items.

If all accept-params each have not `q` prameter, it is invalid as multiple accept-param.

If accept-params are as following, it is overridden  as`text/html`, because `text/html` is more specific than `text/*`.

    Accept: text/*, text/html

But, the test case `Accept: mulpart/mixed, application/json` has same level.

Thus,  it is invalid as multiple accept-param from this perspective.

I consider that the response should apply first accept-param as one accept-param because it is invalid as multiple accept-param.

Or, I consider for other way that the response use request content-type because the accept header cannot decide.
(like following case
https://github.com/OData/WebApi/blob/5ea42200f805ea0fda8ee2ce7f23f704c305486a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchHttpRequestMessageExtensions.cs#L214-L224

KenitoInc: @mikepizzo @xuzhg Any thoughts on this?

marabooy (approved):
LGTM

test/UnitTest/Microsoft.AspNetCore.OData.Test/Batch/ODataBatchHttpRequestMessageExtensionsTest.cs
mikepizzo: I do like this logic much better than what we have today; it comes closer to aligning with the intention of the accept header.

Note that no q-value should be equivalent to a q-value of 1.  It would be good to add tests cases such as:
 [InlineData(new[] { "application/json", "multipart/mixed;q=0.9" }, "application/json")]
 [InlineData(new[] { "application/json;q=0.9", "multipart/mixed" }, "multipart/mixed")]

Also, technically, if the client specifies one or more accept headers, and none are supported, it should return with a 406 Not Acceptable, but that does not appear to be what the code does today.

Finally, I don't see anything in the standard that would prohibit having two media types with the same (or no) q-value, so I'm glad this PR doesn't enforce that.

mikepizzo (approved):
I believe that all comments have been addressed.  Thanks for the contribution @iwate!

Issue #2071, opened 2019-08-07
Add related object to a newly created entity with ODataController and DataServiceContext


tfabian:
I would like to save a newly created entity with a newly created child in a single changeset (batch), with Microsoft.OData.Client 7.6.0 and Microsoft.AspNet.OData 7.1.0. The relationship between them is required: [dbo].[BoxInstance].[IdTemplateInstance] (FK, int, not null)
I've tried to call this:
```
var templateInstance = new TemplateInstance();
var boxInstance = new BoxInstance();

dataServiceContext.AddObject("TemplateInstances", templateInstance);
dataServiceContext.AddObject("BoxInstances", boxInstance);
dataServiceContext.AddLink(templateInstance, "BoxInstances", boxInstance);
```

This doesn't work because BoxInstance.IdTemplateInstance is 0 during the post, (ForeignKey violation).
My second attempt was to call AddRelatedObject:

```
dataServiceContext.AddObject("TemplateInstances", templateInstance);
dataServiceContext.AddRelatedObject(templateInstance, "BoxInstances", boxInstance);
dataServiceContext.SaveChangesAsync(SaveChangesOptions.BatchWithSingleChangeset);
```

In this case I receive the following error message:
```_{"error":{"code":"","message":"No HTTP resource was found that matches the request URI 'http://localhost:41256/TemplateInstances(372)/BoxInstances'.","innererror":{"message":"No routing convention was found to select an action for the OData path with template '~/entityset/key/navigation'.","type":"","stacktrace":""}}}_```

I have the following pocos and a ODataController:
```
  public class TemplateInstance : DomainEntity
  {
    public List<BoxInstance> BoxInstances { get; set; }
  }

  public class BoxInstance : DomainEntity
  {
    public TemplateInstance TemplateInstance { get; set; }
    public int IdTemplateInstance { get; set; }
  }

  //server side
  public class TemplateInstancesController : ODataController
  {
    [AcceptVerbs("POST", "PUT")]
    public IHttpActionResult CreateLink([FromODataUri]int key, string navigationProperty, [FromBody]Uri link)
    {
    }

    [AcceptVerbs("POST", "PUT")]
    public IHttpActionResult CreateRef(string navigationProperty, [FromBody] Uri link, ODataQueryOptions<TEntity> queryOptions)
    {
    }
  }
```

This is not a valid solution, because the first step can't be rolled back in case of failure of the second save changes:
```
dataServiceContext.Add(templateInstance);

[...]

PR #2232, opened 2020-07-17
Fix AspNetCore Resolving of Batch ContentId for AddRelatedObject (#2071)

AdamCaviness:
AspNetCore.OData ODataBatchRequestItem does not properly call ContentIdHelpers.ResolveContentId() with a url but instead passes querystring.

### Issues

*This pull request fixes issue #2071 .*

### Description

*During a migration of a project from .NET Framework to .NET Core 3.1 I found that Microsoft.AspNetCore.OData's handling of batch ContentId paths behave differently than the framework version of the library. The result being a server response of 404 on an inner request when a client calls AddRelatedObject(). I have created complete samples of this behavior for [aspnet](https://github.com/AdamCaviness/ODataAddRelatedObject) and [aspnetcore](https://github.com/AdamCaviness/ODataCoreAddRelatedObject). You can observe that only aspnetcore is broken.*

*I found that [ODataBatchRequestItem](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs#L25) was improperly calling [ContentIdHelpers.ResolveContentId()](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs#L39) with a querystring instead of a url.*

*The [existing unit tests for ODataBatchRequestItem](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/ODataBatchRequestItemTest.cs#L4) were commented out for aspnetcore so I copied the aspnet version and updated them for aspnetcore while adding a new test for my changes as well.*

### Checklist

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
@@ -7,6 +7,7 @@ using System.Diagnostics.Contracts;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
 
 namespace Microsoft.AspNet.OData.Batch
 {
@@ -35,19 +36,12 @@ namespace Microsoft.AspNet.OData.Batch
 
             if (contentIdToLocationMapping != null)
             {
-                string queryString = context.Request.QueryString.HasValue ? context.Request.QueryString.Value : String.Empty;
-                string resolvedRequestUrl = ContentIdHelpers.ResolveContentId(queryString, contentIdToLocationMapping);
+                string displayUrl = context.Request.GetDisplayUrl();
+                string resolvedRequestUrl = ContentIdHelpers.ResolveContentId(displayUrl, contentIdToLocationMapping);
                 if (!string.IsNullOrEmpty(resolvedRequestUrl))
                 {
-                    Uri resolvedUri = new Uri(resolvedRequestUrl, UriKind.RelativeOrAbsolute);
-                    if (resolvedUri.IsAbsoluteUri)
-                    {
-                        context.Request.CopyAbsoluteUrl(resolvedUri);
-                    }
-                    else
-                    {
-                        context.Request.QueryString = new QueryString(resolvedRequestUrl);
-                    }
+                    Uri resolvedUri = new Uri(resolvedRequestUrl, UriKind.Absolute);
+                    context.Request.CopyAbsoluteUrl(resolvedUri);
                 }
 
                 context.Request.SetODataContentIdMapping(contentIdToLocationMapping);
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
@@ -36,11 +36,12 @@ namespace Microsoft.AspNet.OData.Batch
 
             if (contentIdToLocationMapping != null)
             {
-                string displayUrl = context.Request.GetDisplayUrl();
-                string resolvedRequestUrl = ContentIdHelpers.ResolveContentId(displayUrl, contentIdToLocationMapping);
-                if (!string.IsNullOrEmpty(resolvedRequestUrl))
-                {
-                    Uri resolvedUri = new Uri(resolvedRequestUrl, UriKind.Absolute);
+                string encodedUrl = context.Request.GetEncodedUrl();
+                string resolvedRequestUrl = ContentIdHelpers.ResolveContentId(encodedUrl, contentIdToLocationMapping);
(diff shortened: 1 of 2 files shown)


AdamCaviness:
The _WebApi-master-pipeline-Rolling_ check reports a test failure about the bsl baseline files not matching. I'm not familiar with the bsl system you have in place but the test failure reports that the .out file doesn't match the .bsl file but on _my machine_ they are identical. The test failure instructs me to `copy /y` my out file to the bsl file but when I do that there are no changes since they are identical. Additionally, I made no changes to any public api of any method as you can see in the the [PR files changed](https://github.com/OData/WebApi/pull/2232/files) page. I'll be happy to correct anything necessary! Thank you.

AdamCaviness:
@mikepizzo Would you be willing to review my PR or point me in the right direction? I don't want this to get lost in the mix. Any advice or guidance is very much appreciated. Thank you.

marabooy:
@AdamCaviness just tried it out and it seems to be an issue with the test case which is sensitive to line endings from the looks of it. 
Therefore I will be adding the ready to review label.

KenitoInc:
@AdamCaviness Are all your tests passing locally?
@marabooy Can you have a look what's up with the build pipeline.

The fix looks good for me though!

AdamCaviness:
@KenitoInc Correct, the tests pass locally. Thanks.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
```diff
@@ -35,19 +36,12 @@ public static async Task SendRequestAsync(RequestDelegate handler, HttpContext c
...
-                    Uri resolvedUri = new Uri(resolvedRequestUrl, UriKind.RelativeOrAbsolute);
-                    if (resolvedUri.IsAbsoluteUri)
-                    {
-                        context.Request.CopyAbsoluteUrl(resolvedUri);
-                    }
-                    else
-                    {
-                        context.Request.QueryString = new QueryString(resolvedRequestUrl);
-                    }
+                    Uri resolvedUri = new Uri(resolvedRequestUrl, UriKind.Absolute);
```
mikepizzo, line 43: >Absolute [](start = 74, length = 8)

I don't think that we can be sure the Uri returned from ResolveContentId is absolute. In other places that we call ResolveContentId, we create the Uri as UriKind.RelativeOrAbsolute and then, if !resolvedUri.IsAbsoluteUri we create a new uri using the base address.  See, for example, Ln57-61 in ODataEntityReferenceLinkDeserializer.

mikepizzo (changes requested):
:clock1:

src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
```diff
@@ -35,19 +36,12 @@ public static async Task SendRequestAsync(RequestDelegate handler, HttpContext c
 
             if (contentIdToLocationMapping != null)
             {
-                string queryString = context.Request.QueryString.HasValue ? context.Request.QueryString.Value : String.Empty;
-                string resolvedRequestUrl = ContentIdHelpers.ResolveContentId(queryString, contentIdToLocationMapping);
+                string displayUrl = context.Request.GetDisplayUrl();
```
mikepizzo, line 39: >GetDisplayUrl [](start = 52, length = 13)

Is GetDisplayUrl() the right method to use here?  Would GetEncodedUrl() be better?

src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
mikepizzo: Since this is retrieved from the HttpRequest it should generally be absolute, but added a TryCreate just in case.

---
In reply to: [487336879](https://github.com/OData/WebApi/pull/2232#discussion_r487336879) [](ancestors = 487336879)

src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs
mikepizzo: Updated to use GetEncodedUrl()

---
In reply to: [487340257](https://github.com/OData/WebApi/pull/2232#discussion_r487340257) [](ancestors = 487340257)

mikepizzo (approved):

xuzhg:
Mergefd at: 0fbd644fecb513f2cc1433e65cbb6f751598ce4c

xuzhg:
Merged at: 0fbd644fecb513f2cc1433e65cbb6f751598ce4c

PR #2233, opened 2020-07-17
Fix/request binding

KenitoInc:
### Issues

*This pull request fixes issue https://github.com/OData/odata.net/issues/1261*

### Description

The exception is thrown when we try extracting the edmProperty for the StructuredType by passing a property name that doesn’t match what was in the model.
This issue can be fixed in 2 ways:
1.	In ODL, update IEdmStructuredTypeReference.FindProperty() to handle case insensitive property names
2.	Update the WebApi in a way that we can relate the passed property name with the exact property name declared in the model.
i.e we should have a way of matching naMe with Name.

Also we can a configuration to make this feature optional.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
@@ -9,8 +9,13 @@ using System.Globalization;
 using System.Reflection;
 using System.Runtime.Serialization;
 using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Formatter;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
+using Microsoft.OData.UriParser;
 
 namespace Microsoft.AspNet.OData.Formatter.Deserialization
 {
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,7 +19,15 @@ internal static class DeserializationHelpers
         internal static void ApplyProperty(ODataProperty property, IEdmStructuredTypeReference resourceType, object resource,
             ODataDeserializerProvider deserializerProvider, ODataDeserializerContext readContext)
         {
-            IEdmProperty edmProperty = resourceType.FindProperty(property.Name);
+            IDictionary<string, string> propertyNamesMapping = new Dictionary<string, string>();
+            var sProperties = resourceType.StructuralProperties();
```
xuzhg, line 23: There's a scenario that we don't support property name case insensitive is that:
Do we consider 'NaMe" and 'Name' two different properties. 'NaMe' could be a dynamic property, how can we distinguish it?

xuzhg: @mikepizzo 

---
In reply to: [456543378](https://github.com/OData/WebApi/pull/2233#discussion_r456543378) [](ancestors = 456543378)

KenitoInc:
New PR #2249

PR #2234, opened 2020-07-20
support complex typed parameters using FromODataUri attribute

ElizabethOkerio:
### Issues

### Description

This PR add support for the use of complex-typed parameters using the FromODataUri attribute:
Like below:
<pre>
   //Type
    public class Product
    { 
        public int Id { get; set; } 
        public double Price { get; set; } 
    } 
</pre>
If you have the following action methods in your controller, you can use the <code> FromODataUri </code> attribute
to bind parameter values for a complex-typed parameter from a query string 
<pre>
 public class ProductsController : ControllerBase
    {
        [EnableQuery]
        public IActionResult Get([FromODataUri]Product prod)
        {
            return Ok(prod);
        }

        public IActionResult Post([FromODataUri]Product prod)
        {
            if (!ModelState.IsValid)
            {
                return BadRequest();
            }
            return Ok(prod);
        }
    }

</pre>

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataComplexTypeModelBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataComplexTypeModelBinder.cs
@@ -14,8 +14,6 @@ namespace Microsoft.AspNet.OData.Formatter
 {
     internal class ODataComplexTypeModelBinder : IModelBinder
     {
-        private readonly IDictionary<ModelMetadata, IModelBinder> _propertyBinders = null;
-
         private Func<object> _modelCreator;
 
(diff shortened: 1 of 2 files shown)


xuzhg:
@ElizabethOkerio What issue is this PR for? Would you please share more detail in the description? And I remember you have a spec? right? but i can't find it where it is.

ElizabethOkerio:
> @ElizabethOkerio What issue is this PR for? Would you please share more detail in the description? And I remember you have a spec? right? but i can't find it where it is.

@xuzhg Ok. Will add the description and share the spec.

Issue #2240, opened 2020-08-03
Wrong escaped Header (Location) for Objects created with an unicode char within a string-ID causes an Exception


ckuetbach:
Errors with escaping of unicode Characters as Id of type string.

Given there is a model with a Key of type string. The value of this key is 'ä'. The generated URL will contains chars, that are not valid in Header-Fields. The error may only occur for HTTP 1.1 requests

### Assemblies affected

Microsoft.AspNetCore.OData up to 7.7.0

### Reproduce steps

- Create a model with a string key of type string. 
- POST an element with key "ä"

### Expected result
Object is stored,
HTTP Created is returned,
The Location Header points to the newly created entity: http://localhost:5000/odata/test/umlaute(%27%C3%A4%27)/

### Actual result
Object is stored,
Internal ServerError is returned
An Exception occured, because of an unescaped non-ASCII char http://localhost:5000/odata/test/umlaute(%27ä%27)/ This is invalid for Header Values

### Additional detail

The LinkGenerationHelp creates an correct URI 
https://github.com/OData/WebApi/blob/80999f0dca10cc90ef48906c4323a0724a134ca5/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs#L57, but in https://github.com/OData/WebApi/blob/250e955ff68cedfb0e0553747da11bc1f687a63d/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs#L57 the ToString() Method is called, this will unescape the Non-ASCII chars back to unicode.

```		
Uri u = new Uri("http://localhost:5000/odata/test/umlaute(%27ö%27)/");
Console.WriteLine(u.AbsoluteUri);
Console.WriteLine(u.ToString());
```
will return
``` 
http://localhost:5000/odata/test/umlaute(%27%C3%B6%27)/
http://localhost:5000/odata/test/umlaute(%27ö%27)/
```
This causes an exception while writing the value into the location Header.

``` 
[23:53:52 ERR] An unhandled exception has occurred while executing the request.
System.InvalidOperationException: Invalid non-ASCII or control character in header: 0x00E4
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.ThrowInvalidHeaderCharacter(Char ch)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.ValidateHeaderCharacters(String headerCharacters)
[...]

PR #2248, opened 2020-08-12
Fix invalid location header generated when key property for a model contains unicode chars

gathogojr:
### Issues

*This pull request fixes issue #2240 .*

### Description

When the key property of a model is of type string and unicode characters are used (e.g. umlaut), the generated URL will contain chars that are not valid in headers.

For instance, for key equals **Ärne Bjørn**, the generated location is
```
/EntitySet('Ã�rne BjÃ¸rn')
```
instead of
```
/EntitySet(%C3%84rne%20Bj%C3%B8rn)
```

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed
### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData/Results/ResultHelpers.cs
+++ b/src/Microsoft.AspNet.OData/Results/ResultHelpers.cs
@@ -60,7 +60,8 @@ namespace Microsoft.AspNet.OData.Results
         {
             if (response.StatusCode == HttpStatusCode.NoContent)
             {
-                response.Headers.TryAddWithoutValidation(EntityIdHeaderName, entityId().ToString());
+                Uri location = entityId();
+                response.Headers.TryAddWithoutValidation(EntityIdHeaderName, location.AbsoluteUri);
             }
         }
 
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/UnicodeCharactersTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/UnicodeCharactersTest.cs
@@ -1,31 +1,21 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
-#if NETCORE
 using System;
 using System.ComponentModel.DataAnnotations;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
-using Microsoft.AspNet.OData.Builder;
-using Microsoft.AspNet.OData.Extensions;
-using Microsoft.AspNet.OData.Test.Abstraction;
-using Microsoft.AspNetCore.Mvc;
-using Xunit;
-#else
-using System;
-using System.ComponentModel.DataAnnotations;
-using System.Net;
-using System.Net.Http;
-using System.Net.Http.Headers;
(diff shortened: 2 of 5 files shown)


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/UnicodeCharactersTest.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+#if NETCORE
+using System;
```
marabooy, line 5: Most of these look the same any reason for including them in different blocks

gathogojr: @marabooy Fixed

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/UnicodeCharactersTest.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.AspNetCore.Mvc;
+#endif
+using Microsoft.AspNet.OData.Builder;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Xunit;
+
+namespace Microsoft.AspNet.OData.Test
+{
+    public class UnicodeCharactersTest
```
xuzhg, line 24: the file alignment is not correct?

src/Microsoft.AspNet.OData/Results/ResultHelpers.cs
```diff
@@ -60,7 +60,8 @@ public static void AddEntityId(HttpResponseMessage response, Func<Uri> entityId)
         {
             if (response.StatusCode == HttpStatusCode.NoContent)
             {
-                response.Headers.TryAddWithoutValidation(EntityIdHeaderName, entityId().ToString());
+                Uri location = entityId();
```
xuzhg, line 63: why use an intermediate variable can resolve this problem?

gathogojr: @xuzhg The intermediate problem doesn't resolve the problem. If you look at the line that follows, we proceed to call `location.AbsoluteUri` and that works, as opposed to `entityId().ToString()`. I could however ditch the intermediate variable and use `entityId().AbsoluteUri` in the line that follows if this is causing confusion

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/UnicodeCharactersTest.cs
gathogojr: @xuzhg Fixed

src/Microsoft.AspNet.OData/Results/ResultHelpers.cs
paulodero: @gathogojr Your first sentence is a bit confusing but yes, I would also prefer to use `entityId().AbsoluteUri` instead of introducing another variable.

paulodero (approved):
@gathogojr I would be curious to know if the same issue is happening for `action `and `function` params. Otherwise looks good to 🚢

Issue #2247, opened 2018-09-25
How to make Odata request binding case insensitive


brijeshb:
*How do i make odata bind request json in a case insensitive manner? I really appreciated this feature in web api and now as i try toward better odata compatibility this feels like an unnecessary hindrance on my clients. (I'm currently using the latest release on .net core. If there does exist documentation about this, please direct me to it.)*

GreenEyedAndy:
Any progress here?
We have the same problem. All is working well GET, POST, DELETE but PATCH isn't working because of the different case sensitive handling.

KenitoInc:
@GreenEyedAndy We have a PR under review. We hope to have the fix in the next release.

jsgoupil:
I appreciate the work being done on this. This worked fine prior the version 7, but when moving to .NET Core, then case sensitivity started to bite us.
I am happy we are going towards a fix for this, but is there a way we can workaround before this gets in? Because I have doubts it will happen quickly.

I am interested in this `Delta` that is now case sensitive in .NET Core.

I spent about 4h digging in plenty of code trying to figure out the right way to go around this.

jsgoupil:
Alright, another 2h spent. I have noticed a couple of things:

- The `ODataConventionModelBuilder` has something like `EnableLowerCamelCase()`, and when enabled, it just works!
- However, I don't use EdmModel like this in .NET Core. I actually just used `Delta<T>`, and I have noticed that it's using the base of the default Json Serializer installed with .NET Core. I have Newtonsoft, configured to use the naming strategy of CamelCase. However, the `Delta<T>` does not like receiving a `camelCase` for property.

After spending so much time on this, I only needed the .Patch method to work. I will simply drop the OData and come up with my own Pathcher. Something more manual like this:

https://stackoverflow.com/a/19662467/734742

KenitoInc:
PR Merged @GreenEyedAndy @brijeshb

PR #2249, opened 2020-08-12
Fix/2247 Allow case insensitive request binding

KenitoInc:
### Issues

*This pull request fixes issue #2247 .*

### Description

The exception is thrown when we try extracting the `edmProperty` for the `StructuredType` by passing a property name that doesn’t match what was in the model.
```csharp
```
This issue can be fixed in 2 ways:

- In ODL, update IEdmStructuredTypeReference.FindProperty() to handle case insensitive property names. This will be a breaking change.
- In WebApi, add a way that we can relate the passed property name with the exact property name declared in the model.
i.e we should have a way of matching naMe with Name. We should have a configuration to make this feature optional.

We resolve this by injecting a dependency in the service collection as follows:

**AspNetCore WebApi**
```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddOData().EnableODataCaseInsensitiveModelBinding(true);
    ....
    ....
}
```
OR
```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddOData().EnableODataCaseInsensitiveModelBinding();
    ....
    ....
}
```
**AspNet WebApi**
```csharp
public static void Register(HttpConfiguration config)
{
    var builder = new ODataConventionModelBuilder();
    ....
    ....
    config.EnableODataCaseInsensitiveModelBinding();
}
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
@@ -6,12 +6,17 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Net.Http;
 using System.Reflection;
 using System.Runtime.Serialization;
 using Microsoft.AspNet.OData.Builder;
+using Microsoft.AspNet.OData.Adapters;
 using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
+using System.Linq;
 
 namespace Microsoft.AspNet.OData.Formatter.Deserialization
 {
@@ -19,9 +24,60 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
     {
         internal static void ApplyProperty(ODataProperty property, IEdmStructuredTypeReference resourceType, object resource,
             ODataDeserializerProvider deserializerProvider, ODataDeserializerContext readContext)
-        { 
-            IEdmProperty edmProperty = resourceType.FindProperty(property.Name);
-                       
+        {
+            WebApiRequestMessage request = null;
+            if(readContext != null)
+            {
+                request = (WebApiRequestMessage)readContext.InternalRequest;
+            }
+
+            bool enableCaseInsensitiveModelBinding = false;
+
+            if(request != null)
+            {
+#if NETCORE
+                // In AspNetCore, we add our setting by adding a service to the ServiceCollection as below:
+                // services.AddSingleton(new ODataOptions() { EnableCaseInsensitiveModelBinding = true });
+                // We need to resolve the service so that we can get our setting.
+                var settings = request.innerRequest.HttpContext.RequestServices.GetService<ODataOptions>();
+                if(settings != null)
+                {
+                    enableCaseInsensitiveModelBinding = settings.EnableCaseInsensitiveModelBinding;
+                }
+#else
+                // In AspNet classic, we add our setting to HttpConfiguration.Properties as below:
+                // configuration.Properties[EnableCaseInsensitiveModelBindingKey] = true;
+                var container = request.innerRequest.GetConfiguration();
+                if(container != null)
+                {
+                    enableCaseInsensitiveModelBinding = container.HasEnabledCaseInsensitiveModelBinding();
+                }
+#endif
+            }
+
+            IEdmProperty edmProperty;
+
+            // Extract edmProperty using a Case-Sensitive match
+            edmProperty = resourceType.FindProperty(property.Name);
+
+            if (enableCaseInsensitiveModelBinding && edmProperty == null)
+            {
+                // if edmProperty is null, we try a Case-Insensitive match
+                var sProperties = resourceType.StructuralProperties().Where(k => k.Name.Equals(property.Name, StringComparison.OrdinalIgnoreCase));
+
+                // We throw an exception when we have more than 1 case-insensitive matches and no case sensitive matches.
+                if (sProperties.Count() > 1)
+                {
+                    throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
+                }
+
+                if (sProperties.Count() == 1)
+                {
+                    var sProperty = sProperties.Single();
+                    edmProperty = resourceType.FindProperty(sProperty.Name);
+                }
+            }
+
             bool isDynamicProperty = false;
             string propertyName = property.Name;
             if (edmProperty != null)
--- a/src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
+++ b/src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
@@ -26,7 +26,7 @@ namespace Microsoft.AspNet.OData.Adapters
         /// <summary>
         /// The inner request wrapped by this instance.
         /// </summary>
-        private HttpRequestMessage innerRequest;
+        internal HttpRequestMessage innerRequest;
 
         /// <summary>
         /// Initializes a new instance of the WebApiRequestMessage class.
--- a/src/Microsoft.AspNet.OData/Extensions/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.AspNet.OData/Extensions/HttpConfigurationExtensions.cs
@@ -53,6 +53,8 @@ namespace Microsoft.AspNet.OData.Extensions
 
         private const string CompatibilityOptionsKey = "Microsoft.AspNet.OData.CompatibilityOptionsKey";
 
+        private const string EnableCaseInsensitiveModelBindingKey = "Microsoft.AspNet.OData.EnableCaseInsensitiveModelBindingKey";
+
         /// <summary>
         /// Enables query support for actions with an <see cref="IQueryable" /> or <see cref="IQueryable{T}" /> return
         /// type. To avoid processing unexpected or malicious queries, use the validation settings on
@@ -264,6 +266,39 @@ namespace Microsoft.AspNet.OData.Extensions
             }
 
             configuration.Services.Add(typeof(IFilterProvider), new QueryFilterProvider(queryFilter));
+        }
+
+        /// <summary>
+        /// Enables support for case insensitive model binding.
+        /// </summary>
+        /// <param name="configuration">The server configuration.</param>
+        public static void EnableODataCaseInsensitiveModelBinding(this HttpConfiguration configuration)
+        {
+            if (configuration == null)
+            {
+                throw Error.ArgumentNull("configuration");
+            }
+
+            configuration.Properties[EnableCaseInsensitiveModelBindingKey] = true;
+        }
+
+        /// <summary>
+        /// Check the CaseInsensitiveModelBinding is enabled or not.
+        /// </summary>
+        /// <returns>True if CaseInsensitiveModelBinding is enabled; false otherwise</returns>
+        internal static bool HasEnabledCaseInsensitiveModelBinding(this HttpConfiguration configuration)
+        {
+            if (configuration == null)
+            {
+                throw Error.ArgumentNull("configuration");
+            }
+
+            object value;
+            if (configuration.Properties.TryGetValue(EnableCaseInsensitiveModelBindingKey, out value))
+            {
+                return (bool)value;
+            }
+            return false;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
@@ -114,6 +114,33 @@ namespace Microsoft.AspNet.OData.Extensions
         }
 
         /// <summary>
+        /// Enables case-insensitive model binding.
+        /// </summary>
+        /// <param name="builder">Odata Builder.</param>
+        public static IODataBuilder EnableODataCaseInsensitiveModelBinding(this IODataBuilder builder)
+        {
+            return EnableODataCaseInsensitiveModelBinding(builder, true);
+        }
+
+        /// <summary>
+        /// Enables case-insensitive model binding.
+        /// </summary>
+        /// <param name="builder">Odata Builder.</param>
+        /// <param name="enable">Whether to enable case-insensitive model binding.</param>
+        public static IODataBuilder EnableODataCaseInsensitiveModelBinding(this IODataBuilder builder, bool enable)
+        {
+            if (builder == null)
+            {
+                throw Error.ArgumentNull("builder");
+            }
+
+            var odataOptions = builder.Services.First(s => s.ServiceType == typeof(ODataOptions) && s.ImplementationType == typeof(ODataOptions));
+            builder.Services.Remove(odataOptions);
+            builder.Services.AddSingleton(new ODataOptions() { EnableCaseInsensitiveModelBinding = enable });
+            return builder;
+        }
+
+        /// <summary>
         /// Enables query support for actions with an <see cref="IQueryable" /> or <see cref="IQueryable{T}" /> return
         /// type. To avoid processing unexpected or malicious queries, use the validation settings on
         /// <see cref="EnableQueryAttribute"/> to validate incoming queries. For more information, visit
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
@@ -8,12 +8,16 @@ using System.Linq;
 using System.Runtime.Serialization;
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter;
 using Microsoft.AspNet.OData.Formatter.Deserialization;
 using Microsoft.AspNet.OData.Test.Abstraction;
 using Microsoft.AspNet.OData.Test.Builder.Conventions;
 using Microsoft.AspNet.OData.Test.Common;
 using Microsoft.AspNet.OData.Test.Common.Types;
+#if NETCORE
+using Microsoft.AspNetCore.Routing;
+#endif
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Moq;
@@ -367,6 +371,93 @@ namespace Microsoft.AspNet.OData.Test.Formatter.Deserialization
                     resource: null,
                     deserializerProvider: null,
                     readContext: null));
+
+            // Assert
+            Assert.Equal(HelpfulErrorMessage, exception.Message);
+        }
+        
+        [Fact]
+        public void ApplyProperty_PassesWithCaseInsensitivePropertyName()
+        {
+            // Arrange
+            ODataProperty property = new ODataProperty { Name = "keY1", Value = "Value1" };
+            EdmEntityType entityType = new EdmEntityType("namespace", "name");
+            entityType.AddKeys(entityType.AddStructuralProperty("Key1",
+                EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string))));
+
+            EdmEntityTypeReference entityTypeReference = new EdmEntityTypeReference(entityType, isNullable: false);
+            ODataDeserializerProvider provider = ODataDeserializerProviderFactory.Create();
+
+            var resource = new Mock<IDelta>(MockBehavior.Strict);
+            Type propertyType = typeof(string);
+            resource.Setup(r => r.TryGetPropertyType("Key1", out propertyType)).Returns(true).Verifiable();
+            resource.Setup(r => r.TrySetPropertyValue("Key1", "Value1")).Returns(true).Verifiable();
+
+#if NETCORE
+
+            IRouteBuilder builder = RoutingConfigurationFactory.CreateWithCaseInsensitiveModelBinding();
+
+            var request = RequestFactory.Create(builder);
+#else
+            var configuration = RoutingConfigurationFactory.CreateWithRootContainer("OData");
+            configuration.EnableODataCaseInsensitiveModelBinding();
+            var request = RequestFactory.Create(configuration);
+#endif
+
+            ODataDeserializerContext context = new ODataDeserializerContext
+            {
+                Model = new EdmModel(),
+                Request = request
+            };
+
+            // Act
+            DeserializationHelpers.ApplyProperty(property, entityTypeReference, resource.Object, provider,
+                context);
+
+            // Assert
+            resource.Verify();
+        }
+
+        [Fact]
+        public void ApplyProperty_FailWithTwoCaseInsensitiveMatchesAndCaseSensitiveMatch()
+        {
+            const string HelpfulErrorMessage = "The property 'proPerty1' does not exist on type 'namespace.name'. Make sure to only use property names that are defined by the type.";
+            // Arrange
+            ODataProperty property = new ODataProperty { Name = "proPerty1", Value = "Value1" };
+            EdmEntityType entityType = new EdmEntityType("namespace", "name");
+            entityType.AddStructuralProperty("Property1", EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string)));
+            entityType.AddStructuralProperty("property1", EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string)));
+            entityType.AddKeys(entityType.AddStructuralProperty("Key1",
+                EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string))));
+
+            EdmEntityTypeReference entityTypeReference = new EdmEntityTypeReference(entityType, isNullable: false);
+            ODataDeserializerProvider provider = ODataDeserializerProviderFactory.Create();
+
+#if NETCORE
+
+            IRouteBuilder builder = RoutingConfigurationFactory.CreateWithCaseInsensitiveModelBinding();
+
+            var request = RequestFactory.Create(builder);
+#else
+            var configuration = RoutingConfigurationFactory.CreateWithRootContainer("OData");
+            configuration.EnableODataCaseInsensitiveModelBinding();
+            var request = RequestFactory.Create(configuration);
+#endif
+
+            ODataDeserializerContext context = new ODataDeserializerContext
+            {
+                Model = new EdmModel(),
+                Request = request
+            };
+
+            // Act
+            var exception = Assert.Throws<ODataException>(() =>
+                DeserializationHelpers.ApplyProperty(
+                    property,
+                    entityTypeReference,
+                    resource: null,
+                    provider,
+                    context));
 
             // Assert
             Assert.Equal(HelpfulErrorMessage, exception.Message);
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/HttpConfigurationExtensionsTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/HttpConfigurationExtensionsTest.cs
@@ -193,6 +193,21 @@ namespace Microsoft.AspNet.OData.Test
         }
 
         [Fact]
+        public void EnableODataCaseInsensitiveModelBinding_Sets_CaseInsensitiveModelBindingKeyFlag()
+        {
+            // Arrange
+            HttpConfiguration config = new HttpConfiguration();
+            bool initialConfig = config.HasEnabledCaseInsensitiveModelBinding();
+
+            // Act
+            config.EnableODataCaseInsensitiveModelBinding();
+
+            // Assert
+            Assert.False(initialConfig);
+            Assert.True(config.HasEnabledCaseInsensitiveModelBinding());
+        }
+
+        [Fact]
         public void SetUrlKeyDelimiter_Sets_UrlKeyDelimiter()
         {
             // Arrange
--- a/test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
+++ b/test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
@@ -14,8 +14,8 @@ using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.ApplicationParts;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 #if NETCOREAPP2_0
-    using Microsoft.AspNetCore.Builder.Internal;
-    using Microsoft.AspNetCore.Mvc.Internal;
+using Microsoft.AspNetCore.Builder.Internal;
+using Microsoft.AspNetCore.Mvc.Internal;
 #endif
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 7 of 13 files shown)


src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -139,5 +139,38 @@ public static IServiceCollection AddODataQueryFilter(this IServiceCollection ser
...
+        /// </summary>
+        /// <param name="services">The services collection.</param>
+        public static IServiceCollection EnableODataCaseInsensitiveModelBinding(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw Error.ArgumentNull("services");
+            }
+
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IODataModelBinderSettings>(new ODataModelBinderSettings { EnableCaseInsensitiveModelBinding = true }));
```
marabooy, line 154: Probably much better for brevity to call the method below with the model

marabooy:
Lgtm let's fix the failing tests.

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// Settings to use during model binding.
+    /// </summary>
+    public interface IODataModelBinderSettings
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether request body binding should be case insensitive.
+        /// </summary>
+        /// <value>The default value is <c>false</c>.</value>
+        bool EnableCaseInsensitiveModelBinding { get; set; }
```
xuzhg, line 15: why don't use "ODataOptions" to enable it configuration?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -139,5 +139,38 @@ public static IServiceCollection AddODataQueryFilter(this IServiceCollection ser
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IFilterProvider>(new QueryFilterProvider(queryFilter)));
             return services;
         }
+
+        /// <summary>
+        /// Enables support for case insensitive model binding.
+        /// </summary>
+        /// <param name="services">The services collection.</param>
+        public static IServiceCollection EnableODataCaseInsensitiveModelBinding(this IServiceCollection services)
```
xuzhg, line 147: how can developer use this interface in ASP.NET Classic?

KenitoInc: I have added the capability

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            get
+            {
+                return _enableCaseInsensitiveModelBinding;
+            }
+            set
+            {
+                _enableCaseInsensitiveModelBinding = value;
+            }
+        }
```
paulodero, line 35: Any reason why you have both private field and a public property for this instead of just using automatic property?

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// and initializes the default settings.
+        /// </summary>
+        public ODataModelBinderSettings()
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether request body binding should be case insensitive.
+        /// </summary>
+        /// <value>The default value is <c>false</c>.</value>
```
paulodero, line 24: I'm not sure if this needs to be explicitly stated in the comment, since it is almost obvious that `bool ` variable type defaults to `false `in most programming languages including C#.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -6,9 +6,13 @@
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Net.Http;
 using System.Reflection;
 using System.Runtime.Serialization;
+using Microsoft.AspNet.OData.Adapters;
 using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.Extensions.DependencyInjection;
```
paulodero, line 15: Looks like this import is not necessary.

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        private bool _enableCaseInsensitiveModelBinding;
+
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="ODataModelBinderSettings"/> class
+        /// and initializes the default settings.
+        /// </summary>
+        public ODataModelBinderSettings()
+        {
+        }
```
paulodero, line 19: Any reason to explicitly declare this and not to rely on compiler generated default constructor?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: They are both necessary. One works for AspNet classic, the other for AspNet Core

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderSettings.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderSettings.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderSettings.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
KenitoInc: We are already injecting `ODataOptions` in `AddOData` method of `ODataServiceCollectionExtensions`.
`services.AddSingleton<ODataOptions>();`
I didn't find a clean way of  injecting `ODataOptions` with `EnableCaseInsensitiveModelBinding` configuration.

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
KenitoInc: #Resolved

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -139,5 +139,36 @@ public static IServiceCollection AddODataQueryFilter(this IServiceCollection ser
...
+        {
+            if (services == null)
```
marabooy, line 149: You could remove this as it will be checked again for null in the other call.

marabooy (changes requested):

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
xuzhg: Emm, in this way, can we add a overload to "AddOData" to accept the bool value.

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -139,5 +139,31 @@ public static IServiceCollection AddODataQueryFilter(this IServiceCollection ser
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IFilterProvider>(new QueryFilterProvider(queryFilter)));
             return services;
         }
+
+        /// <summary>
+        /// Enables support for case insensitive model binding.
+        /// </summary>
+        /// <param name="services">The services collection.</param>
+        public static IServiceCollection EnableODataCaseInsensitiveModelBinding(this IServiceCollection services)
```
xuzhg, line 147: In the asp.net core version, please add the extensions to "IODataBuilder".

So, customer should call "AddOData().UseCaseInsensitiveBinding()"

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
xuzhg: To inject ODataOptions is to register ODataOptions again after call "AddOData". you can try it

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -139,5 +139,31 @@ public static IServiceCollection AddODataQueryFilter(this IServiceCollection ser
...
+        {
+            return EnableODataCaseInsensitiveModelBinding(services, new ODataModelBinderSettings { EnableCaseInsensitiveModelBinding = true });
+        }
+
+        /// <summary>
+        /// Enables support for case insensitive model binding.
+        /// </summary>
+        /// <param name="services">The services collection.</param>
+        /// <param name="settings">Instance of <see cref="ODataModelBinderSettings" /></param>
+        public static IServiceCollection EnableODataCaseInsensitiveModelBinding(this IServiceCollection services, ODataModelBinderSettings settings)
```
xuzhg, line 157: add extension method for "IODataBuilder"

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
KenitoInc: #Resolved

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,7 +23,54 @@ internal static class DeserializationHelpers
...
+                    enableCaseInsensitiveModelBinding = container.HasEnabledCaseInsensitiveModelBinding();
+                }
+#endif
+            }
+
+            IEdmProperty edmProperty;
+
+            if (enableCaseInsensitiveModelBinding)
+            {
+                IDictionary<string, string> propertyNamesMapping = new Dictionary<string, string>();
```
odero, line 60:
```suggestion
               var propertyNamesMapping = new Dictionary<string, string>();
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,7 +23,54 @@ internal static class DeserializationHelpers
...
+                var sProperties = resourceType.StructuralProperties();
+                foreach (var prop in sProperties)
+                {
+                    propertyNamesMapping.Add(prop.Name.ToUpperInvariant(), prop.Name);
+                }
+                string name;
+                var findPropertyName = propertyNamesMapping.TryGetValue(property.Name.ToUpperInvariant(), out name) ? name : property.Name;
```
odero, line 67: Alternatively, you could use a `HashSet` initialized with `StringComparer.OrdinalIgnoreCase` instead of a full-blown `Dictionary`

mikepizzo: I like the use of the hashset here.

---
In reply to: [487683925](https://github.com/OData/WebApi/pull/2249#discussion_r487683925) [](ancestors = 487683925)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
mikepizzo, line 60: >IDictionary<string, string> propertyNamesMapping = new Dictionary<string, string>(); [](start = 16, length = 84)

Isn't it a bit expensive to create this each time? Can we create the mapping table on first use and then cache it?

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
mikepizzo: I agree; i'd rather not invent something new here for case insensitivity, but make it part of ODataOptions (even if you have to go through slightly more work to set it).

---
In reply to: [470869995](https://github.com/OData/WebApi/pull/2249#discussion_r470869995) [](ancestors = 470869995)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,7 +23,54 @@ internal static class DeserializationHelpers
...
+            IEdmProperty edmProperty;
+
+            if (enableCaseInsensitiveModelBinding)
+            {
+                IDictionary<string, string> propertyNamesMapping = new Dictionary<string, string>();
+                var sProperties = resourceType.StructuralProperties();
+                foreach (var prop in sProperties)
+                {
+                    propertyNamesMapping.Add(prop.Name.ToUpperInvariant(), prop.Name);
+                }
```
mikepizzo, line 65: What if the model has two different properties that vary only by case?  Can we first try a case-sensitive match and, if that fails, fall back to case-insensitive (note that we typically fail if there were two case-insensitive matches and no case sensitive matches).

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,7 +23,54 @@ internal static class DeserializationHelpers
         internal static void ApplyProperty(ODataProperty property, IEdmStructuredTypeReference resourceType, object resource,
```
mikepizzo, line 23: >ApplyProperty [](start = 29, length = 13)

This provides case insensitivity only for payloads.  Isn't it a bit inconsistent to support case-insensitive properties in payloads, but not in URLs?

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
xuzhg: It can inject in the Config(...) before the MapODataService?

---
In reply to: [484275440](https://github.com/OData/WebApi/pull/2249#discussion_r484275440) [](ancestors = 484275440)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: Unfortunately `Hashset<string>.TryGetValue` is not available in `netstandard 2.0` https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.trygetvalue?view=netcore-3.1

odero: Technically you could still use `hashset.Contains(property.Name)` to achieve the same result - which I think is still better than using a dict

KenitoInc: `hashset.Contains(property.Name)` returns a `bool`. However we need to extract the value from the hashset. This can only be achieved by iteration O(n). While it is a solution to the problem, it kind of defeats the purpose of using a hashset in the first place.

odero: Oh yeah you're right. Had forgotten that part

Looking at this again though, what's the benefit of creating the Dict here? Since you are already looping anyway, you might as well get rid of the Dict altogether and just do the check inside the loop instead of populating a Dict in the loop THEN doing the check. That or consider Mike's comment on creating the mapping earlier and reusing it if there are multiple props to check for

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
KenitoInc: @xuzhg @mikepizzo  I have updated the PR to use `ODataOptions`

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: I think case-insensitivity is already supported in WebAPI https://docs.microsoft.com/en-us/odata/webapi/custom-url-parsing

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/HttpConfigurationExtensionsTest.cs
```diff
@@ -192,6 +192,19 @@ public void EnableContinueOnError_Sets_ContinueOnErrorKeyFlag()
...
+        public void EnableODataCaseInsensitiveModelBinding_Sets_CaseInsensitiveModelBindingKeyFlag()
+        {
+            // Arrange
+            HttpConfiguration config = new HttpConfiguration();
+
+            // Act
+            config.EnableODataCaseInsensitiveModelBinding();
+
+            // Assert
+            Assert.True(config.HasEnabledCaseInsensitiveModelBinding());
```
paulodero, line 205: Assuming the default is `false`, you could check if case insensitive model binding flag is `false `before `config.EnableODataCaseInsensitiveModelBinding();` is executed.

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
```diff
@@ -26,7 +26,7 @@ internal class WebApiRequestMessage : IWebApiRequestMessage
         /// <summary>
         /// The inner request wrapped by this instance.
         /// </summary>
-        private HttpRequestMessage innerRequest;
+        internal HttpRequestMessage innerRequest;
```
paulodero, line 29: Looking at the general design of the class, we do not need to change this field from `private` to `internal`. Seems the original intention was to disallow changing the `HttpRequestMessage` object once created. I would prefer to add another method/property in the class `WebApiRequestMessage` as follows and use that to get configurations.

```
        public HttpConfiguration Configuration
        {
            get { return this.innerRequest.GetConfiguration(); }
        }
```

KenitoInc: Great suggestion. However the NetCore `WebApiRequestMessage` class is marked as `internal`. I thought it would be consistent to make the NetClassic version `internal` too. https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNetCore.OData/Adapters/WebApiRequestMessage.cs

paulodero (approved):
:shipit:

paulodero (changes requested):
Tests failing.

paulodero (approved):

src/Microsoft.AspNet.OData.Shared/Formatter/IODataModelBinderSettings.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: I removed the dictionary.
#Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: #Resolved

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/HttpConfigurationExtensionsTest.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
KenitoInc: #Resolved

paulodero (approved):
:shipit:

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -63,24 +63,16 @@ internal static class DeserializationHelpers
...
-                        // Exception when we have two case-insensitive matches and no case sensitive matches.
-                        throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
-                    }
+                    throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
                 }
 
-                string name;
-                var findPropertyName = structuralPropertyNamesMapping.TryGetValue(property.Name.ToUpperInvariant(), out name) ? name : property.Name;
-                edmProperty = resourceType.FindProperty(findPropertyName);
+                var sProperty = sProperties.First();
```
odero, line 74: Minor nit: I'd go with .`Single` instead of `.First` since we know it's only supposed to be 1 item at this point
Q: is there any likelihood that `sProperties` might be empty at this point in which case we might want to consider `.[First|Single]OrDefault`? Or do we want it to fail instead of returning null?

KenitoInc: I have added a condition to ensure `sProperties` is not empty. When `sProperties` is empty, `edmProperty` will remain as `null` and exception will be thrown later in subsequent code.
#Resolved

odero (approved):

src/Microsoft.AspNet.OData/Extensions/HttpConfigurationExtensions.cs
```diff
@@ -266,6 +268,39 @@ public static void AddODataQueryFilter(this HttpConfiguration configuration, IAc
             configuration.Services.Add(typeof(IFilterProvider), new QueryFilterProvider(queryFilter));
         }
 
+        /// <summary>
+        /// Enables support for case insensitive model binding.
+        /// </summary>
+        /// <param name="configuration">The server configuration.</param>
+        public static void EnableODataCaseInsensitiveModelBinding(this HttpConfiguration configuration)
```
gathogojr, line 275:
```suggestion
        public static void EnableCaseInsensitiveModelBinding(this HttpConfiguration configuration)
```

src/Microsoft.AspNet.OData/Extensions/HttpConfigurationExtensions.cs
```diff
@@ -266,6 +268,39 @@ public static void AddODataQueryFilter(this HttpConfiguration configuration, IAc
             configuration.Services.Add(typeof(IFilterProvider), new QueryFilterProvider(queryFilter));
         }
 
+        /// <summary>
+        /// Enables support for case insensitive model binding.
```
xuzhg, line 272: >Enables [](start = 12, length = 7)

typo?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -113,6 +113,33 @@ public static IODataBuilder AddOData(this IServiceCollection services)
             return new ODataBuilder(services);
         }
 
+        /// <summary>
+        /// Enables case-insensitive model binding.
+        /// </summary>
+        /// <param name="builder">Odata Builder.</param>
+        public static IODataBuilder EnableODataCaseInsensitiveModelBinding(this IODataBuilder builder)
```
xuzhg, line 120: >IODataBuilder [](start = 80, length = 13)

It should be "IServiceCollection" ,because the setting is global, not for the certain OData route?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
```diff
@@ -157,6 +157,67 @@ public static IRouteBuilder CreateWithRootContainerAndTypes(string routeName = n
...
+
+            // Create a logger, diagnostic source and app builder.
+            var mockLoggerFactory = new Mock<ILoggerFactory>();
+            var diagnosticSource = new DiagnosticListener("Microsoft.AspNetCore");
+            IApplicationBuilder appBuilder = new ApplicationBuilder(serviceCollection.BuildServiceProvider());
+
+            // Create a route build with a default path handler.
+            IRouteBuilder routeBuilder = new RouteBuilder(appBuilder);
+
+#if NETCOREAPP2_0
```
xuzhg, line 207: how about 3.1?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: #Resolved

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
KenitoInc: @xuzhg I initially used `IServiceCollection`. You asked me to use `IODataBuilder` in this comment https://github.com/OData/WebApi/pull/2249#discussion_r484522022

xuzhg: > @xuzhg I initially used `IServiceCollection`. You asked me to use `IODataBuilder` in this comment [#2249 (comment)](https://github.com/OData/WebApi/pull/2249#discussion_r484522022)

Oh, yes.

src/Microsoft.AspNet.OData/Extensions/HttpConfigurationExtensions.cs
KenitoInc: We have agreed to retain `EnableODataCaseInsensitiveModelBinding`
#Resolved

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
KenitoInc: I will retain `IODataBuilder` so that the customers will be able to use the setting as follows.
`services.AddOData().EnableODataCaseInsensitiveModelBinding(true);`

#Resolved

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
xuzhg: I can't remember that's the old codes. Thanks.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
xuzhg: @KenitoInc I think Mike's meaning is to support the value case-insensitive in the URL. @mikepizzo Is it correct? The Web API already supported case-insensitivity is for the key words in the request url.  For example:   $FilTer is same as $filter.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -6,22 +6,78 @@
...
+#else
+                // In AspNet classic, we add our setting to HttpConfiguration.Properties as below:
+                // configuration.Properties[EnableCaseInsensitiveModelBindingKey] = true;
+                var container = request.innerRequest.GetConfiguration();
+                if(container != null)
+                {
+                    enableCaseInsensitiveModelBinding = container.HasEnabledCaseInsensitiveModelBinding();
+                }
+#endif
+            }
```
xuzhg, line 56: why don't you move all of these setting related codes into the ODataDeserializerContext?

We can expose a property named "enableCaseInsensitiveModelBinding" from "ODataDeserializerContext' ?

src/Microsoft.AspNetCore.OData/Extensions/ODataServiceCollectionExtensions.cs
```diff
@@ -113,6 +113,33 @@ public static IODataBuilder AddOData(this IServiceCollection services)
...
+        /// <param name="enable">Whether to enable case-insensitive model binding.</param>
+        public static IODataBuilder EnableODataCaseInsensitiveModelBinding(this IODataBuilder builder, bool enable)
+        {
+            if (builder == null)
+            {
+                throw Error.ArgumentNull("builder");
+            }
+
+            var odataOptions = builder.Services.First(s => s.ServiceType == typeof(ODataOptions) && s.ImplementationType == typeof(ODataOptions));
+            builder.Services.Remove(odataOptions);
```
xuzhg, line 138: We can't simple just remove the old one, what about there's some configuration before calling this method?

So, i think we should copy construct or clone the old ODataOptions and reset the model binding property.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
```diff
@@ -371,6 +375,93 @@ public void ApplyAnnotations_FailsWithUsefulErrorMessageOnUnknownProperty()
...
+
+        [Fact]
+        public void ApplyProperty_FailWithTwoCaseInsensitiveMatchesAndCaseSensitiveMatch()
+        {
+            const string HelpfulErrorMessage = "The property 'proPerty1' does not exist on type 'namespace.name'. Make sure to only use property names that are defined by the type.";
+            // Arrange
+            ODataProperty property = new ODataProperty { Name = "proPerty1", Value = "Value1" };
+            EdmEntityType entityType = new EdmEntityType("namespace", "name");
+            entityType.AddStructuralProperty("Property1", EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string)));
+            entityType.AddStructuralProperty("property1", EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string)));
```
xuzhg, line 429: >entityType.AddStructuralProperty("property1", EdmLibHelpers.GetEdmPrimitiveTypeReferenceOrNull(typeof(string))); [](start = 12, length = 112)

why duplicate the code?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
```diff
@@ -157,6 +157,67 @@ public static IRouteBuilder CreateWithRootContainerAndTypes(string routeName = n
...
+        /// </summary>
+        /// <returns>A new instance of the routing configuration class.</returns>
+        public static IRouteBuilder CreateWithCaseInsensitiveModelBinding()
+        {
+            IServiceCollection serviceCollection = new ServiceCollection();
+            serviceCollection.AddMvc();
+            serviceCollection.AddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
+            serviceCollection.AddSingleton<ILoggerFactory>(NullLoggerFactory.Instance);
+            serviceCollection.AddOData();
+            serviceCollection.AddSingleton(new ODataOptions(){ EnableCaseInsensitiveModelBinding = true});
```
xuzhg, line 172: why don't use call the setting extension method after "AddOData"

marabooy (approved):

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: By default, webapi supports case-insenstive values in the url. I can get results correctly from the 2 requests below.
`/Books?$expand=Authors&$filter=Id eq 1`
`/Books?$expand=authors&$filter=id eq 1`

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
KenitoInc: The test validates that we typically fail if there were two case-insensitive matches and no case sensitive matches.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
```diff
@@ -371,6 +375,93 @@ public void ApplyAnnotations_FailsWithUsefulErrorMessageOnUnknownProperty()
...
+                context);
+
+            // Assert
+            resource.Verify();
+        }
+
+        [Fact]
+        public void ApplyProperty_FailWithTwoCaseInsensitiveMatchesAndCaseSensitiveMatch()
+        {
+            const string HelpfulErrorMessage = "The property 'proPerty1' does not exist on type 'namespace.name'. Make sure to only use property names that are defined by the type.";
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
KenitoInc: #Resolved

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/RoutingConfigurationFactory.cs
KenitoInc: #Resolved

PR #2258, opened 2020-08-20
Alwayspersistinstance

ificator:
### Issues

This PR fixes https://github.com/OData/WebApi/issues/987

### Description

The previous attempt at this fix always populated `Instance` on the `SelectExpandWrapper<>`, however this caused performance issues with Entity Framework when it walked the expression tree in order to determine what SQL query was required. This change instead requires callers to opt in to the behavior, and it is off by default.

- [X ] *Test cases added*
- [X ] *Build and test with one-click build and test script passed*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
@@ -11,7 +11,7 @@ using Microsoft.OData.Edm;
 
 namespace Microsoft.AspNet.OData.Query.Expressions
 {
-    internal abstract class SelectExpandWrapper : IEdmEntityObject, ISelectExpandWrapper
+    internal abstract class SelectExpandWrapper : ISelectExpandWrapper
     {
         private static readonly IPropertyMapper DefaultPropertyMapper = new IdentityPropertyMapper();
         private static readonly Func<IEdmModel, IEdmStructuredType, IPropertyMapper> _mapperProvider =
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
@@ -30,7 +30,14 @@ namespace Microsoft.AspNet.OData.Query.Expressions
         /// </summary>
         public string ModelID { get; set; }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The instance of the object that is being wrapped, if available.
+        /// </summary>
+        // Ideally this would be called "Instance" and SelectExpandWrapper<> would use the "new"
+        // keyword to replace it with a variation that returns the correct type. Unfortunately the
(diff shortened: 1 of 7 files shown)


gathogojr (changes requested):
Error reported on ADO - baseline (*.bsl) needs to be updated

ificator:
Thanks, I've let this languish a bit because the immediate need to get it through as diminished. However, it's something that I'd still like to see land so I'll clean it up and push a new iteration.

One question I have is whether I should swap it to use the new `DataSourceProviderKind` and always populate the instance unless it's `EFClassic` or `EFCore`?

mikepizzo:
The reason we disabled this was because it had a significant perf impact -- I don't know if it was related to the data source provider, but I think the perf impact could be significant regardless the provider, in which case we'd want to let developers enable it only if they needed it.

---
In reply to: [707305268](https://github.com/OData/WebApi/pull/2258#issuecomment-707305268) [](ancestors = 707305268)

ificator:
@mikepizzo I believe the performance issue was specifically for Entity Framework, because it would see the instance property and decide that it needs to essentially do a `SELECT * FROM ..` instead of requesting the specific fields. Theoretically populating the instance should be fine for other scenarios because the object already exists in memory, but I agree that it's better to play it safe and have it opt-in by the developer. We can always reevaluate that in the future if we have a driver for it.

src/Microsoft.AspNet.OData.Shared/Query/ISelectExpandWrapper.cs
```diff
@@ -10,8 +10,14 @@ namespace Microsoft.AspNet.OData.Query
...
     /// Represents the result of a $select and $expand query operation.
     /// </summary>
-    public interface ISelectExpandWrapper
+    public interface ISelectExpandWrapper : IEdmEntityObject
     {
+        /// <summary>
+        /// The instance of the object that is being wrapped, if available.
+        /// </summary>
+        // The public contract should avoid propagating the "UntypedInstance" terminology.
+        object Instance { get; }
```
mikepizzo, line 19: >object Instance { get; } [](start = 8, length = 24)

This will be a breaking change if anyone has a custom implementation of ISelectExpandWrapper.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs
```diff
@@ -30,7 +30,14 @@ internal abstract class SelectExpandWrapper : IEdmEntityObject, ISelectExpandWra
...
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The instance of the object that is being wrapped, if available.
+        /// </summary>
+        // Ideally this would be called "Instance" and SelectExpandWrapper<> would use the "new"
+        // keyword to replace it with a variation that returns the correct type. Unfortunately the
+        // scenarios that consume "Instance" do so using reflection, and such an approach would
+        // introduce ambiguity that would need to be handled in numerous places. Since this is an
+        // internal construct it fine to leave like this.
```
mikepizzo, line 40: Please create an issue to address this properly in 8.x.

Issue #2264, opened 2020-09-01
Support EnumMember Attribute


keatkeat87:
It look like currently OData is not working on EnumMember Attribute, when serialize and deserialize payload and $filter.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData WebApi lib 6.1.0
v7.4.1

### Reproduce steps
git clone https://github.com/keatkeat87/odata-enum-member
F5 run
query http://localhost:56503/odata/WeatherForecast
query with $filter 
$filter status eq 'Sold out'

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*

### Expected result
The value of property status should be "Sold out".

### Actual result
The value of property status was "SoldOut"

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*
I m not sure this is a bug or feature request.
I think this is a useful feature, just like EF core can use value converter to convert Enum to string and string to Enum. So we get a nice display in SQL.
for fronted we still need a nice display Enum.
If this is not consider in OData team, can someone share how to implement by customize way ?

xuzhg:
@keatkeat87 Thanks for reporting this.

Actually, "EnumMember Attribute" is supported in model builder and serializer. If i run your project using "http://localhost:5000/odata/WeatherForecast", i can get the following result, if you put '[DataContract]' on the enum type.
![image](https://user-images.githubusercontent.com/9426627/91881611-99be5c00-ec36-11ea-8611-25e11db2a6ea.png)

```C#
[DataContract/*(Name = "Status")*/]
    public enum Status { 
        [EnumMember(Value = "Sold out")]
        SoldOut
    }
```

And yes, there's some problem when we do the query like" $filter=status eq 'Sold out'"
The problem is from https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs#L1939

We should follow up the code here at: https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataEnumSerializer.cs#L94-L102
to replace the string first then call TryParse.

keatkeat87:
Thank for fast reply. I can get it work now.
but for $filter part, have any work around ?
maybe I know when 7.5 will be released ?

xuzhg:
@keatkeat87 We planned to next week. Hope it can happen.

keatkeat87:
[...]

PR #2267, opened 2020-09-01
Fix issue#2264, enable enum member alias for query options

xuzhg:
### Issues

*This pull request fixes issue #2264.*

### Description

* Enable "$filter=Status eq 'Sold out'" if Status enum type has enum alias.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -1954,6 +1954,10 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                         {
                             value = clrMember;
                         }
+                        else
+                        {
+                            throw new ODataException(Error.Format(SRResources.CannotGetEnumClrMember, enumMember.Name));
+                        }
                     }
                     else
                     {
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -1936,7 +1936,34 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                 Contract.Assert(strValue != null);
 
                 constantType = Nullable.GetUnderlyingType(constantType) ?? constantType;
-                value = Enum.Parse(constantType, strValue);
+
+                IEdmEnumType enumType = edmTypeReference.AsEnum().EnumDefinition();
+                ClrEnumMemberAnnotation memberMapAnnotation = Model.GetClrEnumMemberAnnotation(enumType);
+                if (memberMapAnnotation != null)
+                {
+                    IEdmEnumMember enumMember = enumType.Members.FirstOrDefault(m => m.Name == strValue);
+                    if (enumMember == null)
+                    {
+                        enumMember = enumType.Members.FirstOrDefault(m => m.Value.ToString() == strValue);
+                    }
+
+                    if (enumMember != null)
+                    {
+                        Enum clrMember = memberMapAnnotation.GetClrEnumMember(enumMember);
+                        if (clrMember != null)
+                        {
+                            value = clrMember;
+                        }
+                    }
+                    else
+                    {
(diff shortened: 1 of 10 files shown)


gathogojr:
@xuzhg You might want to look at the failing tests

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Enums/EnumsController.cs
```diff
@@ -245,4 +246,24 @@ public ITestActionResult HasAccessLevel([FromODataUri] int id, [FromODataUri] Ac
             return Ok(result);
         }
     }
+
+    public class WeatherForecastController : TestODataController
+    {
+        private static readonly string[] Summaries = new[]
```
mikepizzo, line 252: >Summaries [](start = 41, length = 9)

Is this used anywhere?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1936,7 +1936,34 @@ private Type RetrieveClrTypeForConstant(IEdmTypeReference edmTypeReference, ref
...
+
+                IEdmEnumType enumType = edmTypeReference.AsEnum().EnumDefinition();
+                ClrEnumMemberAnnotation memberMapAnnotation = Model.GetClrEnumMemberAnnotation(enumType);
+                if (memberMapAnnotation != null)
+                {
+                    IEdmEnumMember enumMember = enumType.Members.FirstOrDefault(m => m.Name == strValue);
+                    if (enumMember == null)
+                    {
+                        enumMember = enumType.Members.FirstOrDefault(m => m.Value.ToString() == strValue);
+                    }
```
mikepizzo, line 1948: is this code path tested?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1936,7 +1936,34 @@ private Type RetrieveClrTypeForConstant(IEdmTypeReference edmTypeReference, ref
...
+
+                    if (enumMember != null)
+                    {
+                        Enum clrMember = memberMapAnnotation.GetClrEnumMember(enumMember);
+                        if (clrMember != null)
```
mikepizzo, line 1953: >if (clrMember != null) [](start = 24, length = 22)

what if clrMember is null here?  it looks like value is never set?

mikepizzo (changes requested):
:clock1:

mikepizzo:
Please see a few minor issues in the comments.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Enums/EnumsController.cs
xuzhg: No, Can we leave it unchanged, later, if we add more test cases, we can re-use it?

---
In reply to: [485246056](https://github.com/OData/WebApi/pull/2267#discussion_r485246056) [](ancestors = 485246056)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
xuzhg: It's set in the Model builder when we add the annotation.  If it's null, It's better to throw exception. Let me add

---
In reply to: [485247896](https://github.com/OData/WebApi/pull/2267#discussion_r485247896) [](ancestors = 485247896)

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
xuzhg: Actually, yes. If I don't add this, there's one test case failing because the ODataEnumValue has the "1" as the value, not the string. But, i can't remember the test (name).
If that matters, i can figure it out later

---
In reply to: [485247028](https://github.com/OData/WebApi/pull/2267#discussion_r485247028) [](ancestors = 485247028)

mikepizzo (approved):
:shipit:

Issue #2284, opened 2020-09-15
Still unable to $select complex types after version 7.2.2


MathiasJor:
Hey again, I seem to be unable to re-open the original issue, so I had to open a new one. This #2281 is still there on nightly Microsoft.AspNet.OData 7.5.0-Nightly202009150040. Both on the sample project I linked in that issue, and in our custom project. The issue can still be reproduced in the same steps provided in that issue. 

### Assemblies affected

- Microsoft.AspNet.OData 7.5.0-Nightly202009150040
- From `Microsoft.AspNet.OData 7.2.3` up to and including `Microsoft.AspNet.OData 7.4.1` (current newest).

### Reproduce steps _(Copied from #2281)_
I found [this](https://github.com/OData/ODataSamples/tree/master/RESTier/Trippin) (slightly outdated) sample project, which has a complex type called "Location", similar to what we have in our project. Using that as a base, to reproduce:

1. Clone repo
2. Upgrade nuget packages (that project is on Microsoft.AspNet.OData 7.2.1, which works fine)
3. Run the solution and try to query Events: 
(1) `http://localhost:18384/Events`
(2) `http://localhost:18384/Events?$select=occursAt`

Here the "OccursAt" field is of the complex type "Location". 

### Expected result _(Copied from #2281)_
This is what you get on version 7.2.2 and below:
(1) `{"@odata.context":"http://localhost:18384/$metadata#Events","value":[{"Id":1,"Description":null,"OccursAt":{"Address":"Address1"}}]}`

(2) `{"@odata.context":"http://localhost:18384/$metadata#Events(OccursAt)","value":[{"OccursAt":{"Address":"Address1"}}]}`

### Actual result _(Copied from #2281)_
This is what you get on versions above 7.2.2:
(1) `{"@odata.context":"http://localhost:18384/$metadata#Events","value":[{"Id":1,"Description":null,"OccursAt":{"Address":"Address1"}}]}`

(2) `{"error":{"code":"","message":"Cannot compare elements of type 'Microsoft.OData.Service.Sample.Trippin.Models.Location'. Only primitive types, enumeration types and entity types are supported."}}`

xuzhg:
@MathiasJor How about this version?

 https://www.myget.org/feed/webapinetcore/package/nuget/Microsoft.AspNet.OData/7.5.0-Nightly202009142118

xuzhg:
@MathiasJor Alright, 

1) 
[...]

PR #2285, opened 2020-09-15
Continue on the Nullcheck remove fix

xuzhg:
### Issues

*This pull request fixes issue #2284.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/HandleNullPropagationOptionHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/HandleNullPropagationOptionHelper.cs
@@ -92,6 +92,11 @@ namespace Microsoft.AspNet.OData.Query
         None,
 
         /// <summary>
+        /// In memory data source provider
+        /// </summary>
+        InMemory,
+
+        /// <summary>
         /// EF classic Data source provider
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/HandleNullPropagationOptionHelper.cs
```diff
@@ -69,11 +69,11 @@ public static DataSourceProviderKind GetDataSourceProviderKind(this IQueryable q
             string queryProviderNamespace = query.Provider.GetType().Namespace;
             switch (queryProviderNamespace)
```
gathogojr, line 70: @xuzhg Given that you added `DataSourceProviderKind.InMemory` should the default case still return `DataSourceProviderKind.None`

xuzhg: Emm, maybe not. InMemory is not for the "IQueryable", this method is for the IQueryable. So, it's better to return "None". @gathogojr

gathogojr (approved):
LGTM

mikepizzo (approved):
:shipit:

Issue #2262, opened 2020-08-28
Invalid, partial json response happens when expanding an entity with more items than the allowed 'PageSize' for an endpoint.


UmairB:
Invalid, partial json response happens when expanding a contained entity with more items than the allowed `PageSize` for an endpoint.

### Assemblies affected

Microsoft.AspNetCore.OData 7.4.1

### Reproduce steps

Entities:
```
    public class Store
    {
        public int Id { get; set; }

        public string Name { get; set; }

        public virtual IEnumerable<Employee> Employees { get; set; }
    }

    public class Employee
    {
        public int Id { get; set; }

        public string Name { get; set; }
    }
```

Configuration:
```
            app.UseEndpoints(endpoints =>
            {
                // odata
                //endpoints.MapControllers();
                endpoints
                    .Select().Filter().OrderBy().Count().Expand().MaxTop(50)
                    .SetUrlKeyDelimiter(ODataUrlKeyDelimiter.Parentheses);

                var defaultBatchHandler = new DefaultODataBatchHandler();
                defaultBatchHandler.MessageQuotas.MaxNestingDepth = 2;
                defaultBatchHandler.MessageQuotas.MaxOperationsPerChangeset = 10;

                endpoints.MapODataRoute("odata", string.Empty, GetEdmModel(), defaultBatchHandler);
            });

        private static IEdmModel GetEdmModel()
        {
            var builder = new ODataConventionModelBuilder();

            builder.EntitySet<Store>("Stores")
                .EntityType
                .HasKey(e => e.Id)
                .ContainsMany(e => e.Employees).IsNavigable()
                .Expand()
                .Select()
                .Filter()
                .OrderBy()
                .Page(null, null);

            builder.EntityType<Employee>()
                .HasKey(e => e.Id);

            var model = builder.GetEdmModel();
            return model;
        }
```

Metadata:
```
<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
    <edmx:DataServices>
        <Schema Namespace="Scratchpad.OData.Web.Entities" xmlns="http://docs.oasis-open.org/odata/ns/edm">
            <EntityType Name="Store">
                <Key>
                    <PropertyRef Name="Id" />
                </Key>
                <Property Name="Id" Type="Edm.Int32" Nullable="false" />
                <Property Name="Name" Type="Edm.String" />
[...]

PR #2287, opened 2020-09-15
Fix/2262 invalid partial json

KenitoInc:
### Issues

*This pull request fixes issue #2262 .*

### Description

In contained navigation, we don't have the navigation property binding. In the current implementation, we are relying on the navigation property binding to get the `NavigationLink` which we use in generating the `NextPageLink`. This causes an error that results in a partial json response.
The PR allows us to generate the `NextPageLink` in Contained navigation.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
@@ -361,7 +361,26 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             NavigationSourceLinkBuilderAnnotation linkBuilder = writeContext.Model.GetNavigationSourceLinkBuilder(sourceNavigationSource);
             Uri navigationLink =
                 linkBuilder.BuildNavigationLink(writeContext.ExpandedResource, writeContext.NavigationProperty);
-            Uri nestedNextLink = GenerateQueryFromExpandedItem(writeContext, navigationLink);
+
+            Uri nestedNextLink = null;
+
+            // In Contained Navigation, we don't have navigation property binding,
+            // Hence we cannot get the NavigationLink from the NavigationLinkBuilder
+            if (navigationLink != null)
+            {
+                // Non-Contained navigation
+                nestedNextLink = GenerateQueryFromExpandedItem(writeContext, navigationLink);
+            }
+            else
+            {
+                // Contained navigation
+                var idLinkBuilder = new NavigationSourceLinkBuilderAnnotation(sourceNavigationSource, writeContext.Model);
+                var idlink = idLinkBuilder.BuildIdLink(writeContext.ExpandedResource);
+
+                var link = idlink.ToString() + "/" + writeContext.ExpandedResource.NavigationSource.Name;
+                nestedNextLink = new Uri(link);
+            }
+
             SkipTokenHandler nextLinkGenerator = null;
             if (writeContext.QueryContext != null)
             {
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
@@ -357,7 +357,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
         private static Uri GetNestedNextPageLink(ODataSerializerContext writeContext, int pageSize, object obj)
         {
             Contract.Assert(writeContext.ExpandedResource != null);
-            Uri nestedNextLink = null;
+            Uri navigationLink = null;
             IEdmNavigationSource sourceNavigationSource = writeContext.ExpandedResource.NavigationSource;
             NavigationSourceLinkBuilderAnnotation linkBuilder = writeContext.Model.GetNavigationSourceLinkBuilder(sourceNavigationSource);
 
@@ -368,17 +368,17 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                 // Contained navigation.
                 Uri idlink = linkBuilder.BuildIdLink(writeContext.ExpandedResource);
 
-                var link = idlink.ToString() + "/" + writeContext.NavigationSource.Name;
-                nestedNextLink = new Uri(link);
+                var link = idlink.ToString() + "/" + writeContext.NavigationProperty.Name;
+                navigationLink = new Uri(link);
             }
             else
             {
                 // Non-Contained navigation.
-                Uri navigationLink =
+                navigationLink =
                     linkBuilder.BuildNavigationLink(writeContext.ExpandedResource, writeContext.NavigationProperty);
-
-                nestedNextLink = GenerateQueryFromExpandedItem(writeContext, navigationLink);
             }
+
+            Uri nestedNextLink = GenerateQueryFromExpandedItem(writeContext, navigationLink);
 
             SkipTokenHandler nextLinkGenerator = null;
             if (writeContext.QueryContext != null)
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
@@ -359,14 +359,13 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             Contract.Assert(writeContext.ExpandedResource != null);
             Uri nestedNextLink = null;
             IEdmNavigationSource sourceNavigationSource = writeContext.ExpandedResource.NavigationSource;
+            NavigationSourceLinkBuilderAnnotation linkBuilder = writeContext.Model.GetNavigationSourceLinkBuilder(sourceNavigationSource);
 
             // In Contained Navigation, we don't have navigation property binding,
             // Hence we cannot get the NavigationLink from the NavigationLinkBuilder
             if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet)
             {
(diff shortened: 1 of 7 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Containment/ContainmentDataSource.cs
```diff
@@ -129,5 +142,82 @@ private void InitAccounts()
...
+                        {
+                            PaymentInstrumentID = 202,
+                            FriendlyName = "202 second PI",
+                            Statement =
+                                new Statement()
+                                {
+                                    StatementID=201,
+                                    TransactionDescription="Physical Goods.",
+                                },
+                        },
```
xuzhg, line 217: >, [](start = 25, length = 1)

don't need the comma

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -357,11 +357,30 @@ private IEnumerable<ODataOperation> CreateODataOperations(IEnumerable<IEdmOperat
...
+
+            // In Contained Navigation, we don't have navigation property binding,
+            // Hence we cannot get the NavigationLink from the NavigationLinkBuilder
+            if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet)
+            {
+                // Contained navigation.
+                var linkBuilder = new NavigationSourceLinkBuilderAnnotation(sourceNavigationSource, writeContext.Model);
+                var idlink = linkBuilder.BuildIdLink(writeContext.ExpandedResource);
+
+                var link = idlink.ToString() + "/" + writeContext.NavigationSource.Name;
```
xuzhg, line 371: > var link = idlink.ToString() + "/" + writeContext.NavigationSource.Name; [](start = 15, length = 73)

I can't understand here.

Id link itself has the navigation source, so, append the navigation source again? 

Can you add the test to cover the output link?

xuzhg (changes requested):
:clock1:

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -357,11 +357,30 @@ private IEnumerable<ODataOperation> CreateODataOperations(IEnumerable<IEdmOperat
...
-            Uri navigationLink =
-                linkBuilder.BuildNavigationLink(writeContext.ExpandedResource, writeContext.NavigationProperty);
-            Uri nestedNextLink = GenerateQueryFromExpandedItem(writeContext, navigationLink);
+
+            // In Contained Navigation, we don't have navigation property binding,
+            // Hence we cannot get the NavigationLink from the NavigationLinkBuilder
+            if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet)
+            {
+                // Contained navigation.
+                var linkBuilder = new NavigationSourceLinkBuilderAnnotation(sourceNavigationSource, writeContext.Model);
```
mikepizzo, line 368: >var [](start = 16, length = 3)

prefer not to use "var". Makes code easier to read if we know what the expected types are that are returned from the different methods.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
mikepizzo: I think that the NavigationSourceLinkBuilderAnnotation.BuildIdLink builds the Uri for a particular contained resource (i.e., writeContext.ExpandedResource).  We then need to append the name of the navigation property on that contained resource.

---
In reply to: [488983874](https://github.com/OData/WebApi/pull/2287#discussion_r488983874) [](ancestors = 488983874)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
mikepizzo, line 368: Why not still call writeContext.Model.GetNavigationSourceLinkBuilder(sourceNavigationSource)? Both end up calling the same constructor, the extension method on IEdmModel caches the NavigationSourceLinkBuilders for performance.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -357,11 +357,30 @@ private IEnumerable<ODataOperation> CreateODataOperations(IEnumerable<IEdmOperat
...
+                var idlink = linkBuilder.BuildIdLink(writeContext.ExpandedResource);
+
+                var link = idlink.ToString() + "/" + writeContext.NavigationSource.Name;
+                nestedNextLink = new Uri(link);
```
mikepizzo, line 372: Should this logic be pushed down to where the link is actually created, rather than you having to fix it up here?  i.e., should this fix really be made as a special case for NavigationSourceKind.ContainedEntitySet in LinkGenerationHelpers.GenerateBaseODataPathSegments?  It seems likely that there are other methods that may trying to generate links that are broken for contained navigation properties?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
KenitoInc: @xuzhg In a scenario where we want to build the `NavigationLink` as `https://baseUri/Stores(1)/Employees`,
BuildIdLink will return the Uri for the contained resource as `https://baseUri/Stores(1)`. Therefore we need to append the navigation property `Employees`.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Containment/ContainmentDataSource.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
KenitoInc: #Resolved

KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
xuzhg: @KenitoInc I mean why append the navigation source name, it should be the "navigation property name", right?

mikepizzo: i.e.; should line 371 be wirteContext.NavigationProperty.Name instead of writeContext.NavigationSource.Name?

---
In reply to: [489627956](https://github.com/OData/WebApi/pull/2287#discussion_r489627956) [](ancestors = 489627956)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -357,11 +358,30 @@ private IEnumerable<ODataOperation> CreateODataOperations(IEnumerable<IEdmOperat
...
```
mikepizzo: I would think we would still need to call GenerateQueryFromExpandedItem on this uri, to include other query options (like a nested filter) in the nextlink.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -357,11 +357,29 @@ private IEnumerable<ODataOperation> CreateODataOperations(IEnumerable<IEdmOperat
...
             Contract.Assert(writeContext.ExpandedResource != null);
+            Uri navigationLink = null;
             IEdmNavigationSource sourceNavigationSource = writeContext.ExpandedResource.NavigationSource;
             NavigationSourceLinkBuilderAnnotation linkBuilder = writeContext.Model.GetNavigationSourceLinkBuilder(sourceNavigationSource);
-            Uri navigationLink =
-                linkBuilder.BuildNavigationLink(writeContext.ExpandedResource, writeContext.NavigationProperty);
+
+            // In Contained Navigation, we don't have navigation property binding,
+            // Hence we cannot get the NavigationLink from the NavigationLinkBuilder
+            if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet)
```
xuzhg, line 366: >if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet) [](start = 12, length = 103)

why don't use the ContainedTarget property on Navigation Property to identify whether it's contained navigation or not?

xuzhg (approved):
:shipit:

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
xuzhg: +1


---
In reply to: [489779315](https://github.com/OData/WebApi/pull/2287#discussion_r489779315) [](ancestors = 489779315)

xuzhg:
Merged at: ebe2a09eafe06711fd903c6bea341bb51be84c82

PR #2288, opened 2020-09-15
ResourceContext.ResourceInstance throws exception if there's a select or expand

xuzhg:
### Issues

*This pull request fixes issue https://github.com/OData/WebApi/issues/2284#issuecomment-692531295*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/ResourceContext.cs
+++ b/src/Microsoft.AspNet.OData.Shared/ResourceContext.cs
@@ -240,6 +240,12 @@ namespace Microsoft.AspNet.OData
                 object value;
                 if (EdmObject.TryGetPropertyValue(property.Name, out value) && value != null)
                 {
+                    if (value is SelectExpandWrapper)
+                    {
+                        // Skip the select expand property
+                        continue;
+                    }
+
                     string propertyName = EdmLibHelpers.GetClrPropertyName(property, EdmModel);
 
                     if (TypeHelper.IsCollection(value.GetType()))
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/ResourceContext.cs
```diff
@@ -240,6 +240,13 @@ private object BuildResourceInstance()
                 object value;
                 if (EdmObject.TryGetPropertyValue(property.Name, out value) && value != null)
                 {
+                    SelectExpandWrapper valueWrapper = value as SelectExpandWrapper;
+                    if (valueWrapper != null)
+                    {
+                        // Skip the select expand property
+                        continue;
+                    }
+
```
mikepizzo, line 249: or just:

if (value is SelectExpandWrapper)
{
    // skip the select expand property
    continue;
}

mikepizzo (approved):
:shipit:

mikepizzo:
would be good to add tests covering this case.  Also, suggested a small code simplification, but otherwise lgtm.

src/Microsoft.AspNet.OData.Shared/ResourceContext.cs
xuzhg: > would be good to add tests covering this case. Also, suggested a small code simplification, but otherwise lgtm.

Added.

gathogojr (approved):

Issue #2255, opened 2020-08-18
FilterBinder test "CastToQuotedEntityOrComplexType_DerivedProductName" seems incorrect.


BlaiseD:
The [CastToQuotedEntityOrComplexType_DerivedProductName test](https://github.com/OData/WebApi/blob/40b8e57fe9f29490cbbe04aadfb2acf3e4fd7638/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/FilterBinderTests.cs#L2320) has the following filter clause `cast(Category,'FilterBuilder.Tests.Model.DerivedCategory')/DerivedCategoryName eq null`.  Category is a member of the Product entity.

The expected result as defined for the test is `$it => (($it As DerivedCategory).DerivedCategoryName == null)` so casting DerivedCategory (subclass of Category) against the $it Product parameter.

### Assemblies affected

I'm using the source code in this repository.

### Reproduce steps

Just build and execute [tCastToQuotedEntityOrComplexType_DerivedProductName](https://github.com/OData/WebApi/blob/40b8e57fe9f29490cbbe04aadfb2acf3e4fd7638/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/FilterBinderTests.cs#L2320).

### Expected result

I would have the thought the defined expected result would be:  `$it => (($it.Category As DerivedCategory).DerivedCategoryName == null)`.

### Actual result

The test succeeds hence the question.  Why is that the correct behavior or is it a bug?

xuzhg:
Merged at: https://github.com/OData/WebApi/commit/800166239db589dfb606bb120f793f4aa75eef31

PR #1986, opened 2019-12-16
Fix cast issue for structured types

Sreejithpin:
Fix cast issue for structured types

### Issues

*This pull request fixes issue #1962 .*

### Description

Currently cast bind function node only allow for casting to same type. Added an option for casting to subtype as well

### Checklist (Uncheck if it is not completed)

- [ ] *Test cases added*
- [ ] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #2299, opened 2020-09-25
Create the correct expression when casting entity/complex type to der…

KenitoInc:
…ived type

### Issues

*This pull request fixes issue #2255.*

### Description

When calling `Expression.TypeAs(source, targetClrType);`, we need to get the source from the first argument in `node.Parameters`.
My thinking is since `Expression[] arguments = BindArguments(node.Parameters);` is already creating `PropertyAccessExpressions`, we don't need to call `BindCastSourceNode(node.Source)`.

Related logic:
https://github.com/OData/WebApi/blob/ade333a0d063c0b18787623c1fd5651d0a497ed1/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs#L1377
https://github.com/OData/WebApi/blob/ade333a0d063c0b18787623c1fd5651d0a497ed1/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs#L1519

This PR just does minor fixes to this merged PR #1986 

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -331,12 +331,12 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
             Expression source = arguments[0];
 
-            if (arguments[0].Type == targetClrType)
+            if (source.Type == targetClrType)
             {
                 // We only support to cast Entity type to the same type now.
                 return source;
             }
-            else if (arguments[0].Type.IsAssignableFrom(targetClrType))
+            else if (source.Type.IsAssignableFrom(targetClrType))
             {
                 // To support to cast Entity/Complex type to the sub type now.
                 return Expression.TypeAs(source, targetClrType);
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -329,16 +329,16 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                 targetClrType = EdmLibHelpers.GetClrType(targetEdmType.ToEdmTypeReference(false), Model);
             }
 
+            Expression source = arguments[0];
+
             if (arguments[0].Type == targetClrType)
             {
                 // We only support to cast Entity type to the same type now.
-                return arguments[0];
+                return source;
             }
             else if (arguments[0].Type.IsAssignableFrom(targetClrType))
             {
                 // To support to cast Entity/Complex type to the sub type now.
-                Expression source = BindCastSourceNode(node.Source);
-
                 return Expression.TypeAs(source, targetClrType);
             }
             else
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -329,16 +329,27 @@ namespace Microsoft.AspNet.OData.Query.Expressions
                 targetClrType = EdmLibHelpers.GetClrType(targetEdmType.ToEdmTypeReference(false), Model);
             }
 
-            Expression source = arguments[0];
-
-            if (source.Type == targetClrType)
+            if (arguments[0].Type == targetClrType)
             {
                 // We only support to cast Entity type to the same type now.
-                return source;
+                return arguments[0];
             }
-            else if (source.Type.IsAssignableFrom(targetClrType))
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -329,16 +329,16 @@ private Expression BindSingleResourceCastFunctionCall(SingleResourceFunctionCall
                 targetClrType = EdmLibHelpers.GetClrType(targetEdmType.ToEdmTypeReference(false), Model);
             }
 
+            Expression source = arguments[0];
+
             if (arguments[0].Type == targetClrType)
```
gathogojr, line 334:
```suggestion
            if (source.Type == targetClrType)
```

src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -329,16 +329,16 @@ private Expression BindSingleResourceCastFunctionCall(SingleResourceFunctionCall
...
             {
                 // We only support to cast Entity type to the same type now.
-                return arguments[0];
+                return source;
             }
             else if (arguments[0].Type.IsAssignableFrom(targetClrType))
```
gathogojr, line 339:
```suggestion
            else if (source.Type.IsAssignableFrom(targetClrType))
```

src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -329,16 +329,16 @@ private Expression BindSingleResourceCastFunctionCall(SingleResourceFunctionCall
...
             {
                 // To support to cast Entity/Complex type to the sub type now.
-                Expression source = BindCastSourceNode(node.Source);
```
gathogojr, line 340: @KenitoInc Would you care to elaborate what the call to `BindCastSourceNode` was doing and how/why removing it fixes the issue? You could add that to the PR discussion...

src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -329,16 +329,16 @@ private Expression BindSingleResourceCastFunctionCall(SingleResourceFunctionCall
...
             {
                 // We only support to cast Entity type to the same type now.
-                return arguments[0];
+                return source;
             }
-            else if (arguments[0].Type.IsAssignableFrom(targetClrType))
+            else if (source.Type.IsAssignableFrom(targetClrType))
             {
                 // To support to cast Entity/Complex type to the sub type now.
-                Expression source = BindCastSourceNode(node.Source);
```
xuzhg, line 340: >Expression source = BindCastSourceNode(node.Source); [](start = 16, length = 52)

why remove this line?

KenitoInc: @xuzhg 
`BindCastSourceNode` binds the `Source` of either the `SingleResourceCastNode` or the `CollectionResourceCastNode`.

If the cast is on the root e.g 
`/Products?$filter=cast('Microsoft.AspNet.OData.Test.Query.Expressions.DerivedProduct')/DerivedCategoryName eq null` , the `Source` will always be null hence `BindCastSourceNode` will always return a parameterExpression with the expression `$it`.
In a scenario where we are casting a navigation property to return an expression that queries against the parent property,
we need to have a `memberAccess` expression. e.g `$it.Category` in the `lambdaExpression`

In the example below:
`/Products?$filter=cast(Category,'Microsoft.AspNet.OData.Test.Query.Expressions.DerivedCategory')/DerivedCategoryName eq null`
should return the lambda expression `$it => (($it.Category As DerivedCategory).DerivedCategoryName == null)`.
However if we call `BindCastSourceNode`, we will get
`$it => (($it As DerivedCategory).DerivedCategoryName == null)`

I have modified the code to only Call `BindCastSourceNode` when not casting on the root.

cc @mikepizzo @gathogojr @Sreejithpin

xuzhg (approved):
:shipit:

xuzhg:
Merged at https://github.com/OData/WebApi/commit/800166239db589dfb606bb120f793f4aa75eef31

Issue #2296, opened 2020-07-08
GroupBy on a Date results in an exception


stuartag:
When using a GroupBy on a DateTime field that is holding a date, the output formatter raises an exception:

```
Microsoft.OData.ODataException: An ODataPrimitiveValue was instantiated with a value of type 'System.DateTime'. ODataPrimitiveValue can only wrap values which can be represented as primitive EDM types.
   at Microsoft.OData.ODataPrimitiveValue..ctor(Object value)
   at Microsoft.OData.ODataValueUtils.ToODataValue(Object objectToConvert)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.CreateODataPropertiesFromDynamicType(EdmEntityType entityType, Object graph, Dictionary`2 dynamicTypeProperties)
````
This is despite the underlying EdmModel defining the field as a Date

This is in Microsoft.AspNet.OData.Shared  v7.4.1

### Expected result

The result of the grouping would be interpreted correctly as a Date and outputted correctly.

### Actual result

The formatter throws an ODataException

### Additional detail

The issue appears to be where fields from the EdmModel are added to a dynamic type. They lose their EdmModel type and are only held as CLR types. When the formatter calls CreateODataPropertiesFromDynamicType, the value is not converted to a Edm.Date type. I have managed to work around this by altering this method to check the original type from the EdmEntityType parameter's properties and perform the necessary conversion to Date, DateTimeOffset or TimeOfDay.

Issue #1945, opened 2019-10-31
Custom Serializer ODataProperty DateTime ODataPrimitiveValue can only wrap values which can be represented as primitive EDM types.'


rossbuggins:
When implementing a custom serializer I'm getting the below exception when a DateTime object is added to ODataProperty. DateTimeOffset is fine. Looking at Microsoft.Odata.Core, there are some ifs around DateTime but not DateTimeOffset, lines 104 and 115 at https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs

Whats the reason behind this?

### Assemblies affected

Odata 7.2.2

### Reproduce steps

```
var dateTime = DateTime.Now;
new ODataProperty() { Name = "UtcStart", Value = dateTime}
```

### Expected result

Accept DateTime as a valid object, like DateTimeOffset is:

```
var dateTime = DateTime.Now;
new ODataProperty() { Name = "UtcStart", Value = new DateTimeOffset(dateTime, TimeSpan.Zero) }
```

### Actual result

Microsoft.OData.ODataException: 'An ODataPrimitiveValue was instantiated with a value of type 'System.DateTime'. ODataPrimitiveValue can only wrap values which can be represented as primitive EDM types.'

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

xuzhg:
@rossbuggins DateTime is removed from OData built-in primitive type. Can you change to use "DateTimeOffset"?

Issue #918, opened 2017-02-16
Aggregate Statement on Collection of IEdmEntityObjects cant find property


vrkrishn:
I was wondering whether aggregation was supported for methods that return EdmEntityObjectCollection (i.e. with type parameter Contact). I see that in all the examples, the controller methods return IQueryable but I'm not sure if that would work if the method expects IEdmEntityObjects rather than  aspecific contact object. Currently I am running into an exception where I cannot find a property defined in the Contact entityType.

### Assemblies affected
Microsoft.AspNet.OData 5.9.1
Microsoft.OData.Core 6.15
Microsoft.OData.Edm 6.15    

### Reproduce steps
GET /data/Contacts?$apply=aggregate(Date with min as Earliest)

### Expected result
{
 @odata.context":"http://hostname/data/$metadata#Contacts,
"value": [
    {
      "@odata.id": null,
      "Earliest": Datetime Object
    }
  ]
}

### Actual result
"error":{
    "code":"","message":"The query specified in the URI is not valid. Could not find a property named 'Date' on type 'System.Web.OData.IEdmEntityObject'.
}

TehWardy:
I have a similar problem when I use groupby with a DateTime property in the same version assemblies ...
`~/Api/MyType?$apply=groupby((DueDate))`
With the DueDate property on the type being a DateTime it returns ...
```
{
  "error": {
    "code": "",
    "message": "An error has occurred."
  }
}
```
doing a normal "get all" query with the url "~/Api/MyType" returns the data fine and displays the datetime properties correctly as ISO strings as I would expect.

TehWardy:
Ah looks like my problem is #912 
Could be related though I guess ???

Issue #912, opened 2017-02-09
GroupBy DateTime properties doesn't work


isahrnddev6:
Grouping by DateTime properties does not work. EnableQuery attribute throws System.ArgumentException.

### Assemblies affected
Microsoft.AspNet.OData 5.9.1

### Reproduce steps
Sent a request that contains a grouping transformation on an entity for a DateTime property:
http://services.odata.org/V4/Northwind/Northwind.svc/Orders?$apply=groupby((OrderDate))  

### Expected result
The service should return grouped data.

### Actual result
System.ArgumentException is thrown.

TehWardy:
"Me too" and I think this may be related to #918

TehWardy:
playing with it a bit more I managed to get this exception information from a simple "~/API/Type?$groupby((DateTimeField))" query ...

```
{
  "error": {
    "code": "",
    "message": "An error has occurred.",
    "innererror": {
      "message": "Argument types do not match",
      "type": "System.ArgumentException",
[...]

PR #2302, opened 2020-09-29
group by on date throws exceptions

ElizabethOkerio:
### Issues

*This pull request fixes issue #2296 , #912, #1945, #918*

### Description

OData doesn't include DateTime as a primitive type. If a user defines a model as
<pre>
public class Customer
{
    public int Id { get; set; }
    public DateTime Birthday { get; set; }
}
</pre>
then tries to do a groupBy using the Birthday Column, an exception is thrown. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -293,6 +293,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                                 };
                             }
                         }
+
                         properties.Add(property);
                     }
                 }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -253,11 +253,11 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             {
                 foreach (var prop in dynamicObject.Values)
                 {
+                    IEdmProperty edmProperty = entityType?.Properties()
+                            .FirstOrDefault(p => p.Name.Equals(prop.Key));
                     if (prop.Value != null
                         && (prop.Value is DynamicTypeWrapper || (prop.Value is IEnumerable<DynamicTypeWrapper>)))
-                    {
-                        IEdmProperty edmProperty = entityType.Properties()
-                            .FirstOrDefault(p => p.Name.Equals(prop.Key));
+                    {                
                         if (edmProperty != null)
                         {
                             dynamicTypeProperties.Add(edmProperty, prop.Value);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -264,17 +264,50 @@ private void WriteDeltaResource(object graph, ODataWriter writer, ODataSerialize
                     }
                     else
```
gathogojr, line 265: @ElizabethOkerio I'd say instead of repeating the line for retrieving the `EdmProperty` in both this `else` block and the `if` block above it, you move it right where the `foreach` block starts. You could also use the null-conditional operator to cater for the unlikely scenario of `entityType` being null. Something like this would be neater
```
                    IEdmProperty edmProperty = entityType?.Properties()
                        .FirstOrDefault(p => p.Name.Equals(prop.Key));

                    if (prop.Value != null
                        && (prop.Value is DynamicTypeWrapper || (prop.Value is IEnumerable<DynamicTypeWrapper>)))
                    {
                        if (edmProperty != null)
                        {
                            dynamicTypeProperties.Add(edmProperty, prop.Value);
                        }
                    }
                    else
                    {
                        object propertyValue;

                        if (prop.Value == null)
                        {
                            propertyValue = new ODataNullValue();
                        }
                        else
                        {
                            if (edmProperty != null)
                            {
                                propertyValue = ODataPrimitiveSerializer.ConvertPrimitiveValue(prop.Value, edmProperty.Type.AsPrimitive());
                            }
                            else
                            {
                                propertyValue = prop.Value;
                            }
                        }

                        var property = new ODataProperty
                        {
                            Name = prop.Key,
                            Value = propertyValue
                        };

                        properties.Add(property);
                    }
                }
```
You should also write tests to cover the scenario described in the issue. In addition, link the related issues (some dating back to 2017) to your PR such that they're closed when you merge your PR.

ElizabethOkerio: Well Noted. Thanks

gathogojr (approved):
LGTM

KenitoInc (approved):

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DateAndTimeOfDay/DateWithEfTest.cs
```diff
@@ -59,6 +59,16 @@ public async Task CanFilterByDatePropertyForDateTimePropertyOnEf(string filter,
...
         }
 
+        [Fact]
+        public async Task CanGroupByDatePropertyForDateTimePropertyOnEf()
+        {
+            string requestUri = string.Format("{0}/odata/EfPeople?$apply=groupby((Birthday), aggregate($count as Cnt))", BaseAddress);
+
+            HttpResponseMessage response = await Client.GetAsync(requestUri);
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
xuzhg, line 69: >Assert.Equal(HttpStatusCode.OK, response.StatusCode); [](start = 12, length = 53)

at least verify the repsonse content?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -253,29 +253,51 @@ private void WriteDeltaResource(object graph, ODataWriter writer, ODataSerialize
...
+                            };
+                        }
+                        else 
+                        {
+                            if (edmProperty != null)
+                            {
+                                property = new ODataProperty
+                                {
+                                    Name = prop.Key,
+                                    Value = ODataPrimitiveSerializer.ConvertPrimitiveValue(prop.Value, edmProperty.Type.AsPrimitive())
```
xuzhg, line 284: >edmProperty.Type.AsPrimitive() [](start = 103, length = 30)

It is always a primitives' type?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -253,29 +253,51 @@ private void WriteDeltaResource(object graph, ODataWriter writer, ODataSerialize
             {
                 foreach (var prop in dynamicObject.Values)
                 {
+                    IEdmProperty edmProperty = entityType?.Properties()
+                            .FirstOrDefault(p => p.Name.Equals(prop.Key));
```
xuzhg, line 257: >p.Name.Equals(prop.Key) [](start = 49, length = 23)

Why simple compare the property name here?
1) does it case sensitive?
2) property alias?

Issue #2277, opened 2020-09-09
Buggy behaviour when addressing derived types in OData Service


gathogojr:
Buggy behaviour when addressing derived types in OData Service built using Microsoft.AspNetCore.OData library

### Assemblies affected

- Microsoft.AspNetCore.OData 7.4.1
- Microsoft.OData.Core 7.6.1

### Reproduce steps

Consider the `Customer` model below together with its derived type `VipCustomer`:
```c#
public class Customer
{
    [Key]
    public int Id { get; set; }
    public string Name { get; set; }
    public List<Order> Orders { get; set; }
}

public class VipCustomer : Customer
{
    public string LoyaltyCardNo { get; set; }
}
```

Consider also the `Customers` entity set where customer with `Id` 1 is of type `Customer` and customer with `Id` 2 is of type `VipCustomer`

Based on OData spec, section [4.11 Addressing Derived Types](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingDerivedTypes), the 1st request from the two listed below should respond with a 404 (since customer at `Id` 1 is not a `VipCustomer`) while the 2nd should respond with the relevant entity
```
http://localhost:32522/odata/Customers/Model.VipCustomer(1) -> 404
http://localhost:32522/odata/Customers/Model.VipCustomer(2) -> 404
```
However, it'd appear that there's a bug such that a 404 is being returned for both requests

In addition, the following request should return a 404 since customer with `Id` 1 is not a `VipCustomer`
```
http://localhost:32522/odata/Customers(1)/NS.Models.VipCustomer
```
However, it'd appear there's a bug such that both requests respond with the respective entity:
```
http://localhost:32522/odata/Customers(1)/NS.Models.VipCustomer -> 200
http://localhost:32522/odata/Customers(2)/NS.Models.VipCustomer -> 200
```

Repro: https://github.com/gathogojr/ODataWebApiIssue1870Repro

### Expected result

Behaviour as outlined in the OData specification

### Actual result

Buggy behaviour as explained in the Repro Steps section

### Additional detail

_Apparently, both of the above scenarios worked okay in Microsoft.AspNet.OData library_

PR #2310, opened 2020-10-02
Fix buggy behaviour when addressing derived types

gathogojr:
### Issues

*This pull request fixes issue #2277.*

### Description

Based on the [OData specification](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingDerivedTypes), the following two requests should be applicable when restricting an entity to a derived type instance

http://host/service/Customers/Model.VipCustomer(1) - key appearing after the cast
http://host/service/Customers(1)/Model.VipCustomer - key appearing before the cast (right after entity set)

In addition, if the customer with key 1 is not a VipCustomer, the result should be a 404 Not Found

This PR addresses that. Also added missing tests for these scenarios.

**Note:**
In the [**EntitySetRoutingConvention**](https://github.com/OData/WebApi/blob/ade333a0d063c0b18787623c1fd5651d0a497ed1/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs#L66), the _~/entityset/cast_ path template is handled by either **Get[EntitySet]From[EntityType]** (e.g. `GetCustomersFromVipCustomer()`) or **GetFrom[EntityType]** (e.g. `GetFromVipCustomer()`). However, in regard to the [**EntityRoutingConventation**](https://github.com/OData/WebApi/blob/ade333a0d063c0b18787623c1fd5651d0a497ed1/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntityRoutingConvention.cs#L20), the _~/entityset/key/cast_ path template is handled by **Get[EntityType]** (e.g. `GetVipCustomer(int key)`) and `Get(int key)`. This is actually what our documentation says on [routing conventions](https://docs.microsoft.com/en-us/odata/webapi/built-in-routing-conventions). For this reason, where the relevant action method is not implemented, a 404 Not Found is returned and may create the impression that there's a bug/feature gap.

Q. Would there be any added benefit in having the _~/entityset/key/cast_ path template be handled _**additionally**_ by `GetCustomersFromVipCustomer(int key)` and `GetFromVipCustomer(int key)` to mirror the equivalent methods for derived type entity set routing?

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DerivedTypes/DerivedTypesTests.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DerivedTypes/DerivedTypesTests.cs
@@ -60,29 +60,13 @@ namespace Microsoft.Test.E2E.AspNet.OData.DerivedTypes
             Assert.Equal(expectedContent, await response.Content.ReadAsStringAsync());
         }
 
-        [Fact]
-        public async Task RestrictEntityToDerivedTypeInstance_ForDerivedTypeNameAfterKey()
+        [Theory]
+        [InlineData("Customers(2)/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer")]
+        [InlineData("Customers/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer(2)")]
+        public async Task RestrictEntityToDerivedTypeInstance(string path)
         {
             // Key preceeds name of the derived type
-            string requestUri = this.BaseAddress + "/odata/Customers(2)/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer";
-
-            var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
-
-            var response = await this.Client.SendAsync(request);
-            Assert.True(response.IsSuccessStatusCode);
-
-            var expectedContent = string.Format("{{\"@odata.context\":\"{0}/odata/$metadata" +
-                "#Customers/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer/$entity\"," +
-                "\"Id\":2,\"Name\":\"Customer 2\",\"LoyaltyCardNo\":\"9876543210\"}}", this.BaseAddress);
-
-            Assert.Equal(expectedContent, await response.Content.ReadAsStringAsync());
-        }
-
-        [Fact]
-        public async Task RestrictEntityToDerivedTypeInstance_ForDerivedTypeNameBeforeKey()
-        {
-            // Name of the derived type preceeds key
-            string requestUri = this.BaseAddress + "/odata/Customers/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer(2)";
+            string requestUri = this.BaseAddress + "/odata/" + path;
 
             var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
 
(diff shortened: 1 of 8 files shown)


marabooy (approved):
LGTM

paulodero:
> ### Issues
> _This pull request fixes issue #2277._
> 
> ### Description
> Based on the [OData specification](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingDerivedTypes), the following two requests should be applicable when restricting an entity to a derived type instance
> 
> http://host/service/Customers/Model.VipCustomer(1) - key appearing after the cast
> http://host/service/Customers(1)/Model.VipCustomer - key appearing before the cast (right after entity set)
> 
> In addition, if the customer with key 1 is not a VipCustomer, the result should be a 404 Not Found
> 
> This PR addresses that. Also added missing tests for these scenarios.
> 
> **Note:**
> In the [**EntitySetRoutingConvention**](https://github.com/OData/WebApi/blob/ade333a0d063c0b18787623c1fd5651d0a497ed1/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs#L66), the _~/entityset/cast_ path template is handled by either **Get[EntitySet]From[EntityType]** (e.g. `GetCustomersFromVipCustomer()`) or **GetFrom[EntityType]** (e.g. `GetFromVipCustomer()`). However, in regard to the [**EntityRoutingConventation**](https://github.com/OData/WebApi/blob/ade333a0d063c0b18787623c1fd5651d0a497ed1/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntityRoutingConvention.cs#L20), the _~/entityset/key/cast_ path template is handled by **Get[EntityType]** (e.g. `GetVipCustomer(int key)`) and `Get(int key)`. This is actually what our documentation says on [routing conventions](https://docs.microsoft.com/en-us/odata/webapi/built-in-routing-conventions). For this reason, where the relevant action method is not implemented, a 404 Not Found is returned and may create the impression that there's a bug/feature gap.
> 
> Q. Would there be any added benefit in having the _~/entityset/key/cast_ path template be handled _**additionally**_ by `GetCustomersFromVipCustomer(int key)` and `GetFromVipCustomer(int key)` to mirror the equivalent methods for derived type entity set routing?
> 
> ### Checklist (Uncheck if it is not completed)
> * [x]  _Test cases added_
> * [ ]  _Build and test with one-click build and test script passed_
> 
> ### Additional work necessary
> _If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue._

@gathogojr from the documentation and your explanation looks like there is inconsistency in how _~/entityset/cast_ path template and _~/entityset/key/cast_ path template are handled. One of them could have adopted the other's style. To your question, I don't see any other added benefit apart from consistency which improves predictability. In my opinion if we add support for `GetCustomerFromVipCustomer(int key)` and `GetFromVipCustomer(int key)` then _~/entityset/cast_ should also be handled by **Get[EntitySet]** and **Get** as well which I guess now complicates the situation. 

For _~/entityset/key/cast_ I think you meant `GetCustomerFromVipCustomer(int key)` without an **s**

paulodero (approved):
:shipit:

odero (approved):

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DerivedTypes/DerivedTypesTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            var response = await this.Client.SendAsync(request);
+            Assert.True(response.IsSuccessStatusCode);
+
+            var expectedContent = string.Format("{{\"@odata.context\":\"{0}/odata/$metadata" +
+                "#Customers/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer/$entity\"," +
+                "\"Id\":2,\"Name\":\"Customer 2\",\"LoyaltyCardNo\":\"9876543210\"}}", this.BaseAddress);
+
+            Assert.Equal(expectedContent, await response.Content.ReadAsStringAsync());
+        }
```
xuzhg, line 97: Merge these two test cases using 

[Theory]
[InlineData("Customers/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer(2)")]
[InlineData("Customers(2)/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer")]

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DerivedTypes/DerivedTypesTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                "\"Orders\":[{{\"Id\":2,\"Amount\":230}},{{\"Id\":3,\"Amount\":150}}]}}]}}", this.BaseAddress);
+
+            Assert.Equal(expectedContent, await response.Content.ReadAsStringAsync());
+        }
+
+        [Fact]
+        public async Task RestrictEntityToDerivedTypeInstance_ThenExpandNavProperty()
+        {
+            // Key preceeds name of the derived type
+            string requestUri = this.BaseAddress + "/odata/Customers(2)/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer?$expand=Orders";
```
xuzhg, line 134: > "/odata/Customers(2)/Microsoft.Test.E2E.AspNet.OData.DerivedTypes.VipCustomer?$expand=Orders" [](start = 50, length = 94)

can it work for use $expand after the "/.../cast/key"? Maybe a test to cover it? 
Expand this case to using [Theory]

PR #2311, opened 2020-10-02
Enable case insensitive request binding

KenitoInc:
### Issues

*This pull request fixes issue #2247 .*

### Description

Original PR merged accidentally and reverted #2249 
The exception is thrown when we try extracting the `edmProperty` for the `StructuredType` by passing a property name that doesn’t match what was in the model.
```csharp
```
Since properties in the Url are case-insensitive by default, we will make properties in the request case-insensitive by default. We can disable this through a compatibility option.

**AspNetCore WebApi**
We resolve this by injecting a dependency as follows:
```csharp
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
{
    ...
    app.UseMvc(routeBuilder =>
    { 
         routeBuilder.Select().Filter().Expand().MaxTop(100).OrderBy().Count().SetCompatibilityOptions(CompatibilityOptions.DisableCaseInsensitiveRequestPropertyBinding);
    });
}
```
**AspNet WebApi**
```csharp
public static void Register(HttpConfiguration config)
{
    var builder = new ODataConventionModelBuilder();
    ....
    ....
    config.SetCompatibilityOptions(CompatibilityOptions.DisableCaseInsensitiveRequestPropertyBinding);
}
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
@@ -18,6 +18,11 @@ namespace Microsoft.AspNet.OData
         /// <summary>
         ///  Generate nextlink even if the top value specified in request is less than page size when the request extension method is directly called.
         /// </summary>
-        AllowNextLinkWithNonPositiveTopValue = 0x1
+        AllowNextLinkWithNonPositiveTopValue = 0x1,
+
+        /// <summary>
+        /// Disable case-insensitive request property binding.
+        /// </summary>
+        DisableCaseInsensitiveRequestPropertyBinding = 0x2
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
@@ -19,9 +19,33 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
     {
         internal static void ApplyProperty(ODataProperty property, IEdmStructuredTypeReference resourceType, object resource,
             ODataDeserializerProvider deserializerProvider, ODataDeserializerContext readContext)
-        { 
-            IEdmProperty edmProperty = resourceType.FindProperty(property.Name);
-                       
+        {
+            IEdmProperty edmProperty;
+
+            // Extract edmProperty using a Case-Sensitive match
+            edmProperty = resourceType.FindProperty(property.Name);
+
+            if (readContext != null && !readContext.DisableCaseInsensitiveRequestPropertyBinding && edmProperty == null)
+            {
+                // if edmProperty is null, we try a Case-Insensitive match.
+                bool propertyMatched = false;
+
+                IEnumerable<IEdmStructuralProperty> structuralProperties = resourceType.StructuralProperties();
+                foreach(IEdmStructuralProperty structuralProperty in structuralProperties)
+                {
+                    if(structuralProperty.Name.Equals(property.Name, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // We throw an exception when we have more than 1 case-insensitive matches and no case sensitive matches.
+                        if (propertyMatched)
+                        {
+                            throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
+                        }
+                        propertyMatched = true;
+                        edmProperty = resourceType.FindProperty(structuralProperty.Name);
+                    }
+                }
+            }
+
             bool isDynamicProperty = false;
             string propertyName = property.Name;
             if (edmProperty != null)
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataDeserializerContext.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataDeserializerContext.cs
@@ -47,6 +47,11 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         /// </summary>
         internal IWebApiUrlHelper InternalUrlHelper { get; private set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to disable case-insensitive request property binding.
+        /// </summary>
+        internal bool DisableCaseInsensitiveRequestPropertyBinding { get; set; }
+
         internal bool IsDeltaOfT
         {
             get
--- a/src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
+++ b/src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
@@ -224,5 +224,14 @@ namespace Microsoft.AspNet.OData.Adapters
         {
             get { return this.innerRequest.GetWriterSettings(); }
         }
+
+        /// <summary>
+        /// Gets the HttpConfiguration associated with the request.
+        /// </summary>
+        /// <returns></returns>
+        public HttpConfiguration Configuration
+        {
+            get { return this.innerRequest.GetConfiguration(); }
+        }
     }
 }
--- a/src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
+++ b/src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
@@ -4,6 +4,7 @@
 using System.Net.Http;
 using System.Web.Http.Controllers;
 using Microsoft.AspNet.OData.Adapters;
+using Microsoft.AspNet.OData.Extensions;
 
 namespace Microsoft.AspNet.OData.Formatter.Deserialization
 {
@@ -25,8 +26,13 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             set
             {
                 _request = value;
-                InternalRequest = _request != null ? new WebApiRequestMessage(_request) : null;
+                WebApiRequestMessage webApiRequestMessage = _request != null ? new WebApiRequestMessage(_request) : null;
+                InternalRequest = webApiRequestMessage;
                 InternalUrlHelper = _request != null ? new WebApiUrlHelper(_request.GetUrlHelper()) : null;
+
+                // We add this setting via CompatibilityOptions
+                CompatibilityOptions options = webApiRequestMessage != null ? webApiRequestMessage.Configuration.GetCompatibilityOptions() : CompatibilityOptions.None;
+                DisableCaseInsensitiveRequestPropertyBinding = options.HasFlag(CompatibilityOptions.DisableCaseInsensitiveRequestPropertyBinding) ? true : false;
             }
         }
 
--- a/src/Microsoft.AspNetCore.OData/Adapters/WebApiRequestMessage.cs
+++ b/src/Microsoft.AspNetCore.OData/Adapters/WebApiRequestMessage.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNet.OData.Adapters
         /// <summary>
         /// The inner request wrapped by this instance.
         /// </summary>
-        internal HttpRequest innerRequest;
+        private HttpRequest innerRequest;
 
         /// <summary>
         /// Initializes a new instance of the WebApiRequestMessage class.
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataDeserializerContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataDeserializerContext.cs
@@ -4,6 +4,7 @@
 using Microsoft.AspNet.OData.Adapters;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNet.OData.Formatter.Deserialization
 {
@@ -25,8 +26,13 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             set
             {
                 _request = value;
-                InternalRequest = _request != null ? new WebApiRequestMessage(_request) : null;
+                WebApiRequestMessage webApiRequestMessage = _request != null ? new WebApiRequestMessage(_request) : null;
+                InternalRequest = webApiRequestMessage;
                 InternalUrlHelper = _request != null ? new WebApiUrlHelper(_request.GetUrlHelper()) : null;
+
+                // We add this setting via CompatibilityOptions
+                CompatibilityOptions options = webApiRequestMessage != null ? webApiRequestMessage.RequestContainer.GetService<ODataOptions>().CompatibilityOptions : CompatibilityOptions.None;
+                DisableCaseInsensitiveRequestPropertyBinding = options.HasFlag(CompatibilityOptions.DisableCaseInsensitiveRequestPropertyBinding) ? true : false;
             }
         }
     }
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
@@ -6,14 +6,20 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Runtime.Serialization;
-using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter;
 using Microsoft.AspNet.OData.Formatter.Deserialization;
 using Microsoft.AspNet.OData.Test.Abstraction;
-using Microsoft.AspNet.OData.Test.Builder.Conventions;
 using Microsoft.AspNet.OData.Test.Common;
 using Microsoft.AspNet.OData.Test.Common.Types;
+#if NETCORE
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+#else
+using System.Net.Http;
+using System.Web.Http;
+#endif
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Moq;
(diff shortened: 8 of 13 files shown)


src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
```diff
@@ -26,7 +26,7 @@ internal class WebApiRequestMessage : IWebApiRequestMessage
         /// <summary>
         /// The inner request wrapped by this instance.
         /// </summary>
-        private HttpRequestMessage innerRequest;
+        internal HttpRequestMessage innerRequest;
```
xuzhg, line 29: >internal HttpRequestMessage innerRequest; [](start = 8, length = 41)

expose it throw an internal property

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -27,6 +28,11 @@ public HttpRequestMessage Request
                 _request = value;
                 InternalRequest = _request != null ? new WebApiRequestMessage(_request) : null;
                 InternalUrlHelper = _request != null ? new WebApiUrlHelper(_request.GetUrlHelper()) : null;
+
+                var request = (WebApiRequestMessage)InternalRequest;
```
xuzhg, line 32: >var request = (WebApiRequestMessage)InternalRequest; [](start = 16, length = 52)

why don't use the "value" here?

xuzhg: besides, "InternalRequest" could be null at line29, right?

---
In reply to: [501472502](https://github.com/OData/WebApi/pull/2311#discussion_r501472502) [](ancestors = 501472502)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -10,18 +10,41 @@
...
-                       
+        {
+            IEdmProperty edmProperty;
+
+            // Extract edmProperty using a Case-Sensitive match
+            edmProperty = resourceType.FindProperty(property.Name);
+
+            if (readContext != null && readContext.EnableCaseInsensitiveModelBinding && edmProperty == null)
+            {
+                // if edmProperty is null, we try a Case-Insensitive match
```
xuzhg, line 32: how can we distinguish case-insensitive property from the dynamic property? 

For example: 

Customer has property name "Name",  it can accept "name" as dynamic property. Now, can we?  

Can we involve @Michael Pizzo

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
KenitoInc: AspNet Core `innerRequest` is also `internal`. So i wanted some consistency.

xuzhg (approved):
:shipit:

xuzhg:
I approved with a design concern.  That's "dynamic property" vs "case insensitive". Hope we can get input from Mike

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
xuzhg: Can we change both? internal field is not good design

---
In reply to: [501473795](https://github.com/OData/WebApi/pull/2311#discussion_r501473795) [](ancestors = 501473795)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: This has been take care of when I fixed @mikepizzo 's comment https://github.com/OData/WebApi/pull/2249#discussion_r489146141

1. We first do a case-sensitive match. This ensures that properties differentiated by case are matched correctly.
2. If a case-sensitive match fails and we have enabled case-insensitive binding, we try a case-insensitive match.
3. If we have multiple case-insensitive matches and no case sensitive matches, we throw an exception

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
KenitoInc: If `InternalRequest` is `null`, `request` will be `null` and `EnableCaseInsensitiveModelBinding` will be `false`

KenitoInc: Let me rename `request` to `webApiRequestMessage`.

src/Microsoft.AspNet.OData/Adapters/WebApiRequestMessage.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
mikepizzo: Just to be clear, if it doesn't match case-sensitive or case-insensitive, then it would be treated as a dynamic property.

Sam is right, though, that if the customer has a property named "Name", previously they could have specified a value "name" and it would have been interpreted as a dynamic property. Now we'll do the case-insensitive match and match it as the defined property "Name".

It would be poor design, and probably an accidental bug, to have dynamic properties that varied from defined property only by case, but having a way to disable case insensitive would address that very edge case.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -10,18 +10,41 @@
...
+                var sProperties = resourceType.StructuralProperties().Where(k => k.Name.Equals(property.Name, StringComparison.OrdinalIgnoreCase));
```
mikepizzo, line 33: >var [](start = 16, length = 3)

We generally try to avoid the use of "var" as it makes it harder to understand what the code is expecting.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -10,18 +10,41 @@
...
+                if (sProperties.Count() > 1)
+                {
+                    throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
+                }
+
+                if (sProperties.Count() == 1)
+                {
+                    var sProperty = sProperties.Single();
+                    edmProperty = resourceType.FindProperty(sProperty.Name);
+                }
```
mikepizzo, line 45: calling count twice against an IQueryable is expensive -- it actually evaluates the expensive LINQ query twice -- then a third time once you do the .Single(). Better would be to loop through the structural properties doing the case insensitive match -- if you find one keep going, if you find a second then error.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -47,6 +47,11 @@ public partial class ODataDeserializerContext
         /// </summary>
         internal IWebApiUrlHelper InternalUrlHelper { get; private set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether request body binding should be case insensitive.
+        /// </summary>
+        internal bool EnableCaseInsensitiveModelBinding { get; set; }
```
mikepizzo, line 53: It is a bit strange to have a different setting for controlling case sensitivity for properties in a payload versus properties in the URL.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -31,5 +31,10 @@ public class ODataOptions
         /// Gets or Sets the set of flags that have options for backward compatibility
         /// </summary>
         public CompatibilityOptions CompatibilityOptions { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether request body binding should be case insensitive.
+        /// </summary>
+        public bool EnableCaseInsensitiveModelBinding { get; set; }
```
mikepizzo, line 38: Name is a bit long, and doesn't really convey that this applies to properties within a request.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataDeserializerContext.cs
KenitoInc: I tested and found that properties in the URL are case-insensitive by default

mikepizzo: yes; that's what makes me think we should consider fixing this as a bug and making it the default (which can be disabled through a compatibility option) rather than providing a separate method to opt-in.

At least in 8.0 we should strongly consider just making this the default and not adding a separate flag for controlling case insensitivity in the body.

---
In reply to: [503683409](https://github.com/OData/WebApi/pull/2311#discussion_r503683409) [](ancestors = 503683409)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: #Resolved

src/Microsoft.AspNetCore.OData/ODataOptions.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataDeserializerContext.cs
KenitoInc: #Resolved

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -27,6 +28,12 @@ public HttpRequestMessage Request
                 _request = value;
                 InternalRequest = _request != null ? new WebApiRequestMessage(_request) : null;
                 InternalUrlHelper = _request != null ? new WebApiUrlHelper(_request.GetUrlHelper()) : null;
+
+                var webApiRequestMessage = (WebApiRequestMessage)InternalRequest;
```
mikepizzo, line 32: >var [](start = 16, length = 3)

We typically try and avoid "var" in our production code as it makes it a bit harder to understand what is intended.

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
mikepizzo: This still seems suspect to me.  

This is a public API, so the user could presumably call set with _request = null. As Sam says, we handle that case in line 29, but line 35 would throw a null ref exception when we tried to deref the Configuration property of the (null) webApiRequestMessage.

---
In reply to: [501514115](https://github.com/OData/WebApi/pull/2311#discussion_r501514115) [](ancestors = 501514115)

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
mikepizzo, line 32: >var webApiRequestMessage = (WebApiRequestMessage)InternalRequest; [](start = 16, length = 65)

We should be able to avoid a cast here.  Rather than set InternalRequest as a WebApiRequestMessage in line 39 and then cast back to WebApiRequestMessage in line 32, we should set webApiRequestMessage as new WebApiRequestMessage(_request) and then set InternalRequest to be that.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataDeserializerContext.cs
mikepizzo: Can you create an issue, assigned to 8.0 and marked "breaking change opportunity" to remove this flag?

---
In reply to: [505930201](https://github.com/OData/WebApi/pull/2311#discussion_r505930201) [](ancestors = 505930201,503683409)

mikepizzo:
I think applying the precedent rules should be fine. If there is a case insensitive match to an existing property it is unlikely it was intended to be a dynamic property of the same name with a different case. If so, we have a flag for disabling this behavior, but in 8.0 if we haven't encountered any issues I think we can remove that flag.

---
In reply to: [705357825](https://github.com/OData/WebApi/pull/2311#issuecomment-705357825) [](ancestors = 705357825)

mikepizzo (changes requested):
:clock1:

mikepizzo:
I resolved/closed the comments that were addressed, and added a few very minor comments.  Those should be easy to address and, once addressed, I'm good to sign off.

---
In reply to: [527475778](https://github.com/OData/WebApi/pull/2311#issuecomment-527475778) [](ancestors = 527475778)

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
KenitoInc: #Resolved

KenitoInc: #Resolved

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -25,8 +26,14 @@ public HttpRequestMessage Request
...
             {
                 _request = value;
-                InternalRequest = _request != null ? new WebApiRequestMessage(_request) : null;
+                WebApiRequestMessage webApiRequestMessage = _request != null ? new WebApiRequestMessage(_request) : null;
+                InternalRequest = webApiRequestMessage;
                 InternalUrlHelper = _request != null ? new WebApiUrlHelper(_request.GetUrlHelper()) : null;
+
+                // We add this setting via CompatibilityOptions
+                // routeBuilder.SetCompatibilityOptions(CompatibilityOptions.DisableCaseInsensitiveRequestPropertyBinding);
+                CompatibilityOptions options = webApiRequestMessage.Configuration.GetCompatibilityOptions();
```
mikepizzo, line 35: >webApiRequestMessage.Configuration.GetCompatibilityOptions(); [](start = 47, length = 61)

This line will still throw a null ref exception if the user tries to set Request to null.

mikepizzo:
Still need to address null ref exception when setting Request to null; then we can merge.

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
mikepizzo, line 35: Can we remove this commented out line?

mikepizzo (approved):
:shipit:

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,9 +19,33 @@ internal static class DeserializationHelpers
...
+            // Extract edmProperty using a Case-Sensitive match
+            edmProperty = resourceType.FindProperty(property.Name);
+
+            if (readContext != null && !readContext.DisableCaseInsensitiveRequestPropertyBinding && edmProperty == null)
+            {
+                // if edmProperty is null, we try a Case-Insensitive match.
+                bool propertyMatched = false;
+
+                IEnumerable<IEdmStructuralProperty> sProperties = resourceType.StructuralProperties();
+                foreach(IEdmStructuralProperty sProperty in sProperties)
```
gathogojr, line 34:
```suggestion
                foreach(IEdmStructuralProperty structuralProperty in structuralProperties)
```
I'd vote for the longer unambiguous names

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -19,9 +19,33 @@ internal static class DeserializationHelpers
...
+                    if(sProperty.Name.Equals(property.Name, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // We throw an exception when we have more than 1 case-insensitive matches and no case sensitive matches.
+                        if (propertyMatched)
+                        {
+                            throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
+                        }
+                        propertyMatched = true;
+                        edmProperty = resourceType.FindProperty(sProperty.Name);
+                    }
```
gathogojr, line 45: @KenitoInc Consider if this might be easier to follow than the `foreach` loop:
```csharp
IList<IEdmStructuralProperty> matchedProperties = structuralProperties.Where(d => d.Name.Equals(property.Name, StringComparison.OrdinalIgnoreCase)).ToList();

if (matchedProperties.Count > 1)
{
	// We throw an exception when we have more than 1 case-insensitive
	throw new ODataException(Error.Format(SRResources.CannotDeserializeUnknownProperty, property.Name, resourceType.Definition));
}
else if (matchedProperties == 1)
{
	edmProperty = resourceType.FindProperty(matchedProperties.First().Name);
}
```

src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
```diff
@@ -18,6 +18,11 @@ public enum CompatibilityOptions
         /// <summary>
         ///  Generate nextlink even if the top value specified in request is less than page size when the request extension method is directly called.
         /// </summary>
-        AllowNextLinkWithNonPositiveTopValue = 0x1
+        AllowNextLinkWithNonPositiveTopValue = 0x1,
+
+        /// <summary>
+        /// Disable case-insensitive request property binding.
+        /// </summary>
+        DisableCaseInsensitiveRequestPropertyBinding = 0x2
```
gathogojr, line 26: If I understand correctly, case-insensitive property binding will be the default? The fact that we have both "disable" and "case-insensitive" in the name makes feel like we could just name the option `CaseSensitiveRequestPropertyBinding` that you can toggle on by assigning a value of `true`?

gathogojr:
@KenitoInc Not 100% sure the delta functionality makes use of the logic in _DeserializationHelpers_. We should verify lest we leave a gap

src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
KenitoInc: I will leave it as it is, since this is a `CompatibilityOption`.
We will be removing this setting in the next breaking change release.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: Calling Count against an IQueryable is expensive.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
KenitoInc: #Resolved

KenitoInc:
> @KenitoInc Not 100% sure the delta functionality makes use of the logic in _DeserializationHelpers_. We should verify lest we leave a gap

That can be a separate work item.

mikepizzo (approved):
:shipit:

Issue #2312, opened 2020-10-05
BindPrefixTemplate does not work for route parameters with type when Endpoint Routing is used for aspnetcore 3.1


t03apt:
Route parameters defined with type are currently not working when Endpoint Routing is used for aspnetcore 3.1.
For example: api/v1/tenants/{tid:guid}

Result: "@odata.context":"https://localhost:5001/api/v1/tenants/%7Btid:guid%7D/$metadata#Foo"
Expected: "@odata.context":"https://localhost:5001/api/v1/tenants/003890b9972e4092aef80e882f1e0999/$metadata#Foo"

### Assemblies affected

Microsoft.AspNet.OData.Extensions

### Reproduce steps

```[csharp]
        [Theory]
        [InlineData("api/v1/tenants/{tid:guid}", "api/v1/tenants/003890b9972e4092aef80e882f1e0999")]
        public void BindPrefixTemplateWithGuidParameterWorksAsExpected(string prefix, string expect)
        {
            // Arrange
            RouteValueDictionary values = new RouteValueDictionary
            (
                new
                {
                    tid = "003890b9972e4092aef80e882f1e0999"
                }
            );

            // Act
            string actual = ODataEndpointLinkGenerator.BindPrefixTemplate(prefix, values, null);

            // Assert
            Assert.Equal(expect, actual);
        }
```

xuzhg:
More, I'd think to use "https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Extensions/LinkGeneratorHelpers.cs#L101-L102" to process the template in the route prefix. Let's use this issue to track on this improvement.

xuzhg:
Merged.

PR #2313, opened 2020-10-05
add support for parameters defined with type

t03apt:
### Issues

#2312 

### Description

Add support for parameters defined with type when Endpoint Routing is used for aspnetcore 3.1.

- [x] *Test cases added*


xuzhg (approved):
Thanks, I am ok for this quick fix. 

More, I'd think to use "https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Extensions/LinkGeneratorHelpers.cs#L101-L102" to process the template in the route prefix.

PR #2314, opened 2020-10-06
Fix failing tests occasioned by use of await expression in test asserts

gathogojr:
Fix failing tests on master occasioned by use of await expression in test asserts

### More details:
When an Xunit assert statement contains an `await` expression as shown below, it'd seem that on ADO, the `@odata.context` property in the response will have the DNS HostName (equivalent to machine name but in upper case) in place of the actual machine name and this causes the assertion to fail.
```c#
Assert.Equal(expectedContent, await response.Content.ReadAsStringAsync());
```
Results into the error:
```
2020-10-06T06:13:49.8104815Z [xUnit.net 00:19:49.5446312]       Expected: Â·Â·Â·ta.context":"http://WIN-D2O1RI1MT21:11129/odata/$metadata#CusÂ·Â·Â·
2020-10-06T06:13:49.8106363Z [xUnit.net 00:19:49.5446408]       Actual:   Â·Â·Â·ta.context":"http://win-d2o1ri1mt21:11129/odata/$metadata#CusÂ·Â·Â·
```
This can be fixed by eagerly awaiting the results of the asynchronous operation as follows:
```c#
var actualContent = response.Content.ReadAsStringAsync().Result;
Assert.Equal(expectedContent, actualContent);
```
Instead of clawing back the gains from use of asynchrony, we instead change the assert statement to verify what is of interest - the substantive body of the response, e.g.,
```c#
Assert.Contains("\"value\":[{\"Id\":1,\"Name\":\"Customer Name 1\"}]",
                await response.Content.ReadAsStringAsync());
```

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DerivedTypes/DerivedTypesTests.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DerivedTypes/DerivedTypesTests.cs
@@ -53,11 +53,9 @@ namespace Microsoft.Test.E2E.AspNet.OData.DerivedTypes
             var response = await this.Client.SendAsync(request);
             Assert.True(response.IsSuccessStatusCode);
 
-            var expectedContent = string.Format("{{\"@odata.context\":\"{0}/odata/$metadata" +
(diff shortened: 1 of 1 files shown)


KenitoInc (approved):
LGTM

ElizabethOkerio (approved):

Issue #2301, opened 2020-09-29
Some of my Get methods that work in Microsoft.AspNetCore.OData 7.4.1 result in 404 in 7.5.0


needo2:
Works in Microsoft.AspNetCore.OData 7.4.1
Breaks in Microsoft.AspNetCore.OData 7.5.0

Get methods in my project defined as

    [EnableQuery]  
    public virtual async Task<IActionResult> Get()

Still work (200 + result)
These are for my directly queried entities

----

Get methods in my project that are defined as

    public Task<ListResult<TModel>> Get([FromQuery] TArgs args, ODataQueryOptions<TModel> queryOptions)

Don't work anymore (404, no exception message or anything)
These are for entities queried by stored procedure
ListResult contains paging information

----

Even if I'm doing something wrong without knowing, this still highlights a breaking change and it should be in the release notes as a breaking change.

Update: Replacing return type with `Task<IActionResult>` did NOT help

henrikdahl8240:
I cannot invoke even one of my Function definitions, all result in 404, Not Found. Now I am back at 7.4.1 and this works again. Has 7.5.0 really passed the test suite before being released?

marabooy:
@henrikdahl8240 this seems to be a regression in 7.5 on how the action selector handles some case. Glad to hear it is working in 7.4 as we work on a fix for this issue.

habbes:
I think the issue could be the `[FromQuery] TArgs args` not being recognized by the action selector as parameters bound to the query options. Let me investigate it.

xuzhg:
Merged at https://github.com/OData/WebApi/commit/dc77dce6ef16235c5f151592a6636e87b064efee

Issue #2305, opened 2020-09-30
EnableQuery attribute changes response to an empty list


frankpeng7:
### Problem Description
We have an API using ODataController with ASP.NET Core 2.1. This is one of the actions:
```csharp
[HttpGet]
[ODataRoute("route_redacted")]
[EnableQuery(AllowedQueryOptions = AllowedQueryOptions.Filter)]
public async Task<IActionResult> name_redacted([FromServices] ODataQueryOptions<type_redacted> oDataQueryOptions)
```
When calling this API using: `<route_redacted>?$filter=<parameter_name_redacted>%20eq%20<value_redacted>`

The response is always an empty list
```json
{
    "@odata.context": "context_redacted",
    "value": []
}
```

However, if I remove the line `[ODataRoute("route_redacted")]` then it works. Response looks like:
```jsonc
{
    "@odata.context": "context_redacted",
    "value": [
        {
            "id": "id_redacted",
            // rest of object redacted
        }
    ]
}
```

Could this behavior be related to the EnableQuery attribute or other Odata stuff?

### Other findings
In our use case, `value_redacted` is an url and is partially encoded. So the actual request URL looks lie 
```
<route_redacted>?filter=SomeUrl%20eq%20https://teams.microsoft.com/l/meetup-join/19%3ameeting_MGQ
```
As described above, this will result in an empty list in response, but if I change the the URL to a fully encoded one
```
https%3A%2F%2Fteams.microsoft.com%2Fl%2Fmeetup-join%2F19%253ameeting_MGQ4MDQ
```
the response turns back to normal (the second one above)

marabooy:
@frankpeng7 Could you kindly share what version of WebApi you are using as well as try with 7.4 as 7.5 has a known regression affecting route selection.

frankpeng7:
Hi @marabooy thanks for the quick response! If you're talking about `Microsoft.AspNet.WebApi` we are using 5.2.7

xuzhg:
@frankpeng7 how do you use "oDataQueryOptions" in your action?

marabooy:
[...]

PR #2315, opened 2020-10-06
Fix routing regression causing 404 on actions with [FromQuery]

habbes:
### Issues

*This pull request fixes issue #2301.*

### Description

The routing was failing in this case because the action selector could not match the parameter decorated with `[FromQuery]` to keys in the url. Since `[FromQuery]` parameters are bound from query options, the action selector should ignore them (it already ignores parameters of type `ODataQueryOptions`). I made the corresponding update to the action selector.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -141,7 +141,7 @@ namespace Microsoft.AspNet.OData.Routing
                         id: c.Id,
                         parameterCount: c.Parameters.Count,
                         filteredParameters: c.Parameters.Where(p => p.ParameterType != typeof(ODataPath) &&
-                        
+                            !IsParameterFromQuery(c as ControllerActionDescriptor, p.Name) &&
                             !ODataQueryParameterBindingAttribute.ODataQueryParameterBinding.IsODataQueryOptions(p.ParameterType)),
                         descriptor: c));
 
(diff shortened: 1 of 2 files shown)


gathogojr (approved):
LGTM but wondering if other than `[FromQuery]` and `ODataQueryOptions` parameter that you made exceptions for there are others that we need to consider. There was [this issue](https://github.com/OData/WebApi/issues/2305) we triaged on Oct 6 where someone had `[FromServices]` that I have never come across before. How do we handle such? Also wondering whether we have good test coverage around action routing

KenitoInc (commented):
A minor change request.
A point to note: This regression came about because you made improvements to the `ActionSelector` and there wasn't adequate test cases to catch the issues. Are there any other scenarios where we don't have test cases and issues might come up in the future?

habbes:
@gathogojr For parameters that we don't know about, the action selector tries to find whether there's a registered model binder that can match that parameter. In your example, the `[FromServices]` decorated parameter was of type `ODataQueryOptions<T>` which is already accounted for by the action selector. If it weren't, then hopefully the user had registered a model binder for it.

The action selector also already checks for parameters in the query options, but if it's an optional query option, then it would fail to match when it's not set. So we should have supported `[FromQuery]` indeed.

@KenitoInc @gathogojr unfortunately I don't have a satisfactory answer regarding the test coverage. While I did add more tests and accounted for more scenarios, the way the action selector works makes it hard to guarantee that we won't miss some corner cases that we don't know about. Maybe we could consider re-designing the action selector and reconsider whether trying matching each parameter to the request is the best approach.

marabooy (approved):
👍

KenitoInc (approved):
LGTM

xuzhg (approved):

xuzhg:
Merged at https://github.com/OData/WebApi/commit/dc77dce6ef16235c5f151592a6636e87b064efee

Issue #2316, opened 2020-10-06
Selecting a complex property with a null value in a feed fails


ificator:
If I have a simple object model like this:
```
    public class Simple
    {
        [DataMember]
        public Complex Complex { get; set; }

        [Key]
        public string Id { get; set; }

        [DataMember]
        public string Name { get; set; }
    }

    public class Complex
    {
        public string Value { get; set; }
    }
```
*Where `Simple` is used for the entity set.*

And I make a query like this:
`.../Simple?$select=Complex`

The request will fail if any of the `Simple` instances have a null value for `Complex`

### Assemblies affected

Microsoft.AspNet.OData 7.5.0

### Reproduce steps

https://github.com/ificator/examples/tree/missingpropertyrepro/OData/MissingProperty

1. Hit `https://localhost:44324/odata/Data?$select=Complex&repro=1` and it will generate an invalid response with console output stating:
```
System.InvalidOperationException: The EDM instance of type '[ODataAnnotations.Complex Nullable=True]' is missing the property 'Value'.
   at Microsoft.AspNet.OData.ResourceContext.GetPropertyValue(String propertyName)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.CreateStructuralProperty(IEdmStructuralProperty structuralProperty, ResourceContext resourceContext)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.CreateStructuralPropertyBag(SelectExpandNode selectExpandNode, ResourceContext resourceContext)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.CreateResource(SelectExpandNode selectExpandNode, ResourceContext resourceContext)
   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSerializer.WriteResource(Object graph, ODataWriter writer, ODataSerializerContext writeContext, IEdmTypeReference expectedType)
```
2. Hit `https://localhost:44324/odata/Data?$select=Complex` and it will work as expected

### Additional detail

It seems that `resourceContext.GetPropertyValue("Complex")` is returning an "empty" `SelectExpandWrapper` instance instead of the expected value of `null`.

ificator:
[...]

PR #2320, opened 2020-10-09
Issue2316: Select a complex with null value fails

ificator:
### Issues

This is a fix for https://github.com/OData/WebApi/issues/2316, please see that issue for a full discussion

### Description

If I have a simple object model like this:

```
    public class Simple
    {
        [DataMember]
        public Complex Complex { get; set; }

        [Key]
        public string Id { get; set; }

        [DataMember]
    }

    public class Complex
    {
        public string Value { get; set; }
    }
```
*Where Simple is used for the entity set.*

And I make a query like this:
`.../Simple?$select=Complex`

The request will fail if any of the `Simple` instances have a null value for `Complex`.

This change fixes the issue by sure the null check is only removed for the scenario that was problematic in the first place. It also does a couple of other things:
1. Rename `None` to `Unknown`
1. Map `System.Linq` as `InMemory`

- [x ] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             _model = _context.Model;
             _modelID = ModelContainer.GetModelID(_model);
             _settings = settings;
-            _dataSourceProviderKind = DataSourceProviderKind.None;
+            _dataSourceProviderKind = DataSourceProviderKind.Unknown;
         }
 
         public static IQueryable Bind(IQueryable queryable, ODataQuerySettings settings, SelectExpandQueryOption selectExpandQuery)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -821,15 +824,7 @@ private Expression CreateTotalCountExpression(Expression source, bool? countOpti
             {
                 if (!structuralProperty.Type.IsCollection())
                 {
-                    // EF5 and EF6 doesn't support to compare the complex with null.
-                    // With the following null check, EF5 and EF6 will throw except similar to:
-                    // "Cannot compare elements of type 'xxxx'.Only primitive types, enumeration types and entity types are supported.
-                    // EF Core has imporvement so the following null won't throw error.
-                    if (_dataSourceProviderKind == DataSourceProviderKind.EFCore ||
```
ificator, line 828: This check is unnecessary - if the condition is true we could already guarantee that `nullCheck` was `null` and so the assignment is essentially a no-op

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -792,9 +792,12 @@ private Expression CreateTotalCountExpression(Expression source, bool? countOpti
                 return;
             }
 
+            // EF5 and EF6 doesn't support to compare the complex with null.
+            // With the following null check, EF5 and EF6 will throw except similar to:
+            // "Cannot compare elements of type 'xxxx'.Only primitive types, enumeration types and entity types are supported.
+            // EF Core has imporvement so the following null won't throw error.
```
xuzhg, line 798: >imporvement [](start = 27, length = 11)

can you help to fix the typo? Sorry, that's the existing typo

xuzhg (approved):
:shipit:

xuzhg:
@ificator  Nightly202010092336 is the nightly version which contains this PR. Can you try it?

ificator:
Confirmed my scenario works with that build

xuzhg (approved):

xuzhg:
Merged at: https://github.com/OData/WebApi/commit/50a759df9d451e002f58d26a2b425207cb8b4221

PR #2332, opened 2020-10-23
Fix invalid next page link in nested collections

### LINKED ISSUE #2321: Issue with nested collection nextlink generation when using API versioning
When using API versioning with ASP.NET Core OData, and enforcing a PageSize limit to enable server-side paging, the nextlink generation for nested collections throws an exception. This seems to be due to the generated nextlink uri containing a URL encoded character '/' that causes the parsing of the generated uri to fail.

### Assemblies affected

Microsoft.AspNetCore.OData 7.5.0 (also reproduced in master branch)

### Reproduce steps

Reproduced by forking the aspnet-api-versioning repo and modifying one of the samples to enable server-side paging and expansion of a child navigation collection with more records than the server-side page size limit. See commit here: https://github.com/gregfullman/aspnet-api-versioning/commit/04a3b26bb998f9f50e9367319c953d30d49a18ec.

Debugging that sample, go to this url in the browser: http://localhost:5000/v1/People?$expand=orders

### Expected result

The OData response would include the first N child records, along with a {childEntity}@odata.nextLink property, providing the nextlink to retrieve additional child records.

### Actual result

In the case of this sample app, the response displayed in the browser is not complete JSON. An exception will show up in the console output (if you're debugging, an unhandled exception will appear).

image

### Additional detail

The commit I made in this branch fixes the issue: https://github.com/gregfullman/WebApi/commit/330fce017823f8ef3ae3501f349d24e161277eca. It seems that if API versioning is being used, the nextlink generation takes a different codepath that results in the URL encoded forward-slash being included in the nextlink uri.

This commit works with all existing tests, and seems to be sufficiently isolated to the nested navigation collection nextlink use-case I'm dealing with. However, I could use some guidance incorporating existing or new tests to cover this change. I'm happy to submit a PR to incorporate this change as well.

gathogojr:
### Issues

*This pull request fixes #2321 .*

### Description

During link generation using `RouteCollection`'s `GetVirtualPath` method, the catch-all parameter escapes the appropriate characters when the route is used to generate a URL, including path separator (/) characters. 
For example, the route `prefix/{*odataPath}` with route values `{ odataPath = "Customers(1)/Orders" }` generates `prefix/Customers(1)%2FOrders` - the forward slash is escaped. 
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.1#url-generation-with-linkgenerator
This causes a problem in some scenarios, e.g. when generating next page link for an expanded collection navigation property

This PR mitigates the problem by substituting the forward slash with a unique token and then substituting the forward slash back after the call to `GetVirtualPath`.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataRoute.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataRoute.cs
@@ -69,7 +69,6 @@ namespace Microsoft.AspNet.OData.Routing
                 {
                     // Try to generate an optimized direct link
                     // Otherwise, fall back to the base implementation
-                    VirtualPathData path;
                     if (CanGenerateDirectLink)
                     {
                         return GenerateLinkDirectly(odataPath);
@@ -96,7 +95,7 @@ namespace Microsoft.AspNet.OData.Routing
                         // Q. Are there scenarios when we'd be happy with having the path separator escaped for us?
                         string token = System.Guid.NewGuid().ToString().Replace("-", "");
                         context.Values[ODataRouteConstants.ODataPath] = odataPath.Replace("/", token);
-                        path = base.GetVirtualPath(context);
+                        VirtualPathData path = base.GetVirtualPath(context);
                         path.VirtualPath = path.VirtualPath.Replace(token, "/");
                         return path;
                     }
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataRoute.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataRoute.cs
@@ -73,11 +73,8 @@ namespace Microsoft.AspNet.OData.Routing
                     {
                         return GenerateLinkDirectly(odataPath);
                     }
-                    else if (!odataPath.Contains("/"))
-                    {
-                        return base.GetVirtualPath(context);
-                    }
-                    else
+
+                    if (odataPath.Contains("/"))
                     {
                         // During link generation using `RouteCollection`'s `GetVirtualPath` method, 
                         // the catch-all parameter escapes the appropriate characters when the route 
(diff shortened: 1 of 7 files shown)


marabooy (approved):
Lgtm

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ServerSidePaging/ServerSidePagingControllers.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Test.E2E.AspNet.OData.ServerSidePaging
+{
+    public class ServerSidePagingCustomersController : TestODataController
+    {
+        private readonly IList<ServerSidePagingCustomer> _serverSidePagingCustomers;
+
+        public ServerSidePagingCustomersController()
+        {
+            _serverSidePagingCustomers = new List<ServerSidePagingCustomer>(
```
g2mula, line 17: Nit: Enumerable.Range(1, 7).Select(i=> new SSPC{ })**.ToList()** :-)

g2mula (commented):
A couple of nitpicks and a question...

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ServerSidePaging/ServerSidePagingControllers.cs
gathogojr: @g2mula I used `.ToList()` at first, then changed to call the constructor instead - my love for `.ToSomething()` is not that much for no good reason.

src/Microsoft.AspNetCore.OData/Routing/ODataRoute.cs
```diff
@@ -69,9 +69,36 @@ public override VirtualPathData GetVirtualPath(VirtualPathContext context)
...
+                        // https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.1#url-generation-with-linkgenerator
+                        // This causes a problem in some scenarios, e.g. when generating 
+                        // next page link for an expanded collection navigation property
+
+                        // HACK! We go round the problem by substituting the forward slash 
+                        // with a unique token and then substituting the forward slash back 
+                        // after the call to `GetVirtualPath`.
+                        // Q. Are there scenarios when we'd be happy with having the path separator escaped for us?
+                        string token = System.Guid.NewGuid().ToString().Replace("-", "");
+                        context.Values[ODataRouteConstants.ODataPath] = odataPath.Replace("/", token);
```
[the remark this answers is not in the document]
gathogojr: Good catch!

KenitoInc (approved):
LGTM

mikepizzo (approved):

PR #2336, opened 2020-10-29
Full async support

mikepizzo:
READY FOR REVIEW -- NOT READY TO MERGE

### Issues

### Description

The OData WebAPI Library operates synchronously. ASP.NET Core is optimized for asynchronous calls and services/libraries like OData WebAPI must set AllowSynchronousIO=true in order to make synchronous calls, which hurts performance. 

To take advantage of (rather than suffer from) ASP.NET Core's optimization around asynchronous libraries, the OData WebAPI library needs to be fully asynchronous.

This PR adds full asynchronous calling through the WebAPI OData stack, enabling us to remove the AllowSynchronousIO=true directive.

- [ ] *Test cases added*

### Additional work necessary

The JSON reader/writer in the low level OData Library still reads/writes synchronously; it's just wrapped in a Task. We should explore making the JSON reader/writer in ODataLibrary fully asychronous.

If the service uses other serializers, such as the default JSON Serializer, which are not async, then they will have to set AllowSychronousIO=true as we are no longer setting it.


src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
```diff
@@ -151,25 +143,27 @@ public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterCo
 
                 ODataDeserializerProvider deserializerProvider = request.GetRequestContainer().GetRequiredService<ODataDeserializerProvider>();
 
-                object result = ODataInputFormatterHelper.ReadFromStream(
+                Task<object> resultTask = ODataInputFormatterHelper.ReadFromStreamAsync(
```
davidfowl, line 146: This needs to await

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
```diff
@@ -151,25 +143,27 @@ public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterCo
...
                     new WebApiRequestMessage(request),
-                    () => ODataMessageWrapperHelper.Create(request.Body, request.Headers, request.GetODataContentIdMapping(), request.GetRequestContainer()),
+                    () => ODataMessageWrapperHelper.Create(new StreamWrapper(request.Body), request.Headers, request.GetODataContentIdMapping(), request.GetRequestContainer()),
                     (objectType) => deserializerProvider.GetEdmTypeDeserializer(objectType),
                     (objectType) => deserializerProvider.GetODataDeserializer(objectType, request),
                     getODataDeserializerContext,
                     (disposable) => toDispose.Add(disposable),
                     logErrorAction);
 
+                resultTask.Wait();
```
davidfowl, line 159: Don't Wait here, just await above.

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
```diff
@@ -299,4 +292,53 @@ private MediaTypeHeaderValue GetContentType(string contentTypeValue)
             return contentType;
         }
     }
+
+    internal class StreamWrapper : Stream
```
davidfowl, line 296: What's this for?

davidfowl, line 296: What's this for?

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
mikepizzo: Originally I just did await above (along with the async modifier on the method), but that caused some of the existing tests to fail (I think primarily in the error cases) so I had to do it this way for backward compatibility (since this is unfortunately a public API I couldn't just fix the tests.)

---
In reply to: [515363752](https://github.com/OData/WebApi/pull/2336#discussion_r515363752) [](ancestors = 515363752)

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
mikepizzo: The WebAPI OData library ends up calling a lower level OData Library with an internal JSON reader/writer.  That reader/writer currently calls synchronous Read/Write methods.  So, by wrapping the body.Stream in this StreamWrapper before passing it to the lower level library we can make sure that the synchronous APIs are never called.

Would love your thoughts here, but seemed an easy/lightweight way to make sure that no synchronous method was called (if the lower level library already calls the asynch methods it's just pass-through).

---
In reply to: [515376424](https://github.com/OData/WebApi/pull/2336#discussion_r515376424) [](ancestors = 515376424)

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
davidfowl: I'm not sure what tests broke but doing this wait basically negates the entire fix.

mikepizzo: Got it. I'll put it back and disable the test for now -- it was verifying an expected exception from the underlying method, and getting back an aggregation exception instead of the expected argument null exception. that difference won't affect the test release, but for the final release I may have to create two versions of the method to avoid a possible breaking change.

---
In reply to: [515383405](https://github.com/OData/WebApi/pull/2336#discussion_r515383405) [](ancestors = 515383405)

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
```diff
@@ -225,7 +219,7 @@ public override Task WriteResponseBodyAsync(OutputFormatterWriteContext context,
 
                 ODataSerializerProvider serializerProvider = request.GetRequestContainer().GetRequiredService<ODataSerializerProvider>();
 
-                ODataOutputFormatterHelper.WriteToStream(
+                return ODataOutputFormatterHelper.WriteToStreamAsync(
```
xuzhg, line 222: shall we call ".ConfigureAwait(false);" after await?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataMetadataSerializer.cs
```diff
@@ -34,5 +35,22 @@ public override void WriteObject(object graph, Type type, ODataMessageWriter mes
             // the model received by this method and the model passed(from configuration) while building ODataMessageWriter is the same (clr object).
             messageWriter.WriteMetadataDocument();
         }
+
+        /// <inheritdoc/>
+        /// <remarks>The metadata written is from the model set on the <paramref name="messageWriter"/>. The <paramref name="graph" />
+        /// is not used.</remarks>
+        public override async Task WriteObjectAsync(object graph, Type type, ODataMessageWriter messageWriter, ODataSerializerContext writeContext)
```
xuzhg, line 42: > public override async Task WriteObjectAsync [](start = 7, length = 44)

shall we use the base "virtual" default implementation?

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
davidfowl: This is a good way to verify that but I would add it to tests, not to the main project.

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
davidfowl: Not required for ASP.NET Core components.

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
davidfowl: What besides a test would be able to observe this breaking change? The formatter is called invoked via MVC.

mikepizzo: There would be a different exception surfaced to the developer -- instead of getting a meaningful exception they would get a generic "AggregateException"

---
In reply to: [515997243](https://github.com/OData/WebApi/pull/2336#discussion_r515997243) [](ancestors = 515997243)

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
mikepizzo: Unfortunately, it's not just a test verification issue.  We *know* that, deep within the lower level OData Library that this library calls, the code that actually reads/writes to the stream currently calls synchronous read/write methods. It will take a while to fix that lower level library to be async all of the way to the wire, so this makes sure that, at the lowest level, even though the underlying OData Library makes a synchronous call to read/write, we are always calling the asynchronous APIs on the stream.

Are we likely to run into issues with this approach? The main idea of this overall fix is to not set the "AllowSynchronousIO=true" flag, which unblocks *other* asynchronous calls.  Is there a better way to handle this until we are able to rev the lower level library?

---
In reply to: [515953270](https://github.com/OData/WebApi/pull/2336#discussion_r515953270) [](ancestors = 515953270)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataMetadataSerializer.cs
mikepizzo: We could, but I like having this as a placeholder for when we implement the WriteMetadataDocumentAsync() method in ODL.

---
In reply to: [515435490](https://github.com/OData/WebApi/pull/2336#discussion_r515435490) [](ancestors = 515435490)

davidfowl:
If your want to preserve that test behavior you can split the async method in 2, one that does argument validation and one that does the async logic. That would preserve the existing behavior of not getting aggregate exceptions here though I question the value since it'll never get called with null arguments since the framework is calling

mikepizzo:
Agree; this one doesn't concern me. I'm more concerned about the ApplyTest -- if I don't comment out that one the tests on the build machine hang.  

---
In reply to: [723464898](https://github.com/OData/WebApi/pull/2336#issuecomment-723464898) [](ancestors = 723464898)

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
mikepizzo: @ - This file is only used in .NET Core, but some of the shared files are used by both .NET Core and .NET Classic.

For the shared files, do we need the .ConfigureAwait(false) everywhere we use await in order to avoid possible deadlocks?  If so, should we use an #IF !NETCORE to add the .ConfigureAwait(false), or should we just always add it to the common calls (i.e., is there any disadvantage to adding in .NET Core)?

---
In reply to: [515953430](https://github.com/OData/WebApi/pull/2336#discussion_r515953430) [](ancestors = 515953430)

davidfowl:
> Agree; this one doesn't concern me. I'm more concerned about the ApplyTest -- if I don't comment out that one the tests on the build machine hang.

Do you need help tracking it down? Is it easily reproducible?

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
mikepizzo: Reverted async logic for .NET Classic. It's not needed there, and introduces unnecessary churn.

---
In reply to: [520190988](https://github.com/OData/WebApi/pull/2336#discussion_r520190988) [](ancestors = 520190988,515953430)

PR #2339, opened 2020-10-30
Ensure that untyped values escaped in ODL are unescaped in OData WebApi during deserialization

### LINKED ISSUE #2292: Double escaping new line character when converting ODataUntypedValue to string
When client posts or patches open type property without specifying its type, and the type of property is string containing a new line, the final result of operation is double escaped new line character - effectively becoming a literal _\_ followed by literal _n_.

### Assemblies affected

Microsoft.AspNet.OData 7.5.0

### Reproduce steps

* Create any service with at leas one entity of open type.
* Send following PATCH request:
`{
    "Memo": "Update of memo\n4\n40"
}`

### Expected result

The PATCH request should have set open property named `Memo` to the value:
<pre>Update of memo
4
40</pre>

### Actual result

The end result is `"Update of memo\\n4\\n40"` that translates to:
<pre>Update of memo\n4\n40</pre>

### Additional detail

I believe that the root cause of this behaviour is how `ConvertPrimitiveValue(string value)` method is implemented in OData Web API in contrast of how `JsonValueUtils` (part of Microsoft.OData.Core) escapes characters.
The Web API code is available at: https://github.com/OData/WebApi/blob/1f48bed2de0da996054f804cfd243057bbb3b2de/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs#L524
And OData Core is: https://github.com/OData/odata.net/blob/c6876f23b78f21407aedeb7e2d985b117d04d3f3/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L715

On can see that WebApi is only removing starting and ending `"` character; but OData Lib has escaped all control characters inside the string as well; and this is not unmade by `ConvertPrimitiveValue `implementation.
I would guess that all previously escaped characters need to be reverted, or at least those specially listed at the end of `CreateSpecialCharToEscapedStringMap()` at https://github.com/OData/odata.net/blob/c6876f23b78f21407aedeb7e2d985b117d04d3f3/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L699.

### Workaround

Clients can specify type of the open attribute and mitigate the whole thing.
`{
    "Memo@odata.type": "#String",
    "Memo": "Update of memo\n4\n40"
}`
But sometimes this is not an option when one can not affect the client development.

gathogojr:
### Issues

*This pull request fixes #2292.*

### Description

Ensure that untyped values escaped in ODL are unescaped in OData WebApi during deserialization

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
@@ -8,6 +8,7 @@ using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.Reflection;
 using System.Runtime.Serialization;
+using System.Text.RegularExpressions;
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.OData;
@@ -553,7 +554,9 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                 throw new ODataException(Error.Format(SRResources.InvalidODataUntypedValue, value));
             }
 
-            return value.Substring(1, value.Length - 2);
+            // `JsonValueUtils`'s `GetEscapedJsonString` method in ODL converts string
+            // to Json-formatted string with specific special characters escaped.
+            return UnescapeString(value.Substring(1, value.Length - 2));
         }
 
         private static object ConvertEnumValue(ODataEnumValue enumValue, ref IEdmTypeReference propertyType,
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -580,6 +583,22 @@ private static object ConvertPrimitiveValue(string value)
...
+        /// <returns>String with any escaped characters converted to their unescaped form.</returns>
+        private static string UnescapeString(string inputString)
+        {
+            if (string.IsNullOrWhiteSpace(inputString))
+            {
+                return inputString;
+            }
+
+            // Make use `Regex`'s `Unescape` method to pick out any escaped special characters and unescape them
+            return Regex.Unescape(inputString);
```
xuzhg, line 599: >Regex [](start = 19, length = 5)

I remember that for some reason (security or performance), we try to remove "Regex" usage in the source code.
Do we have other way to overcome it?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/DeserializationHelpersTest.cs
```diff
@@ -372,6 +372,41 @@ public void ApplyAnnotations_FailsWithUsefulErrorMessageOnUnknownProperty()
...
+        [InlineData('\n', "\\n")]
+        [InlineData('\r', "\\r")]
+        [InlineData('\t', "\\t")]
+        [InlineData('\"', "\"")]
+        [InlineData('\\', "\\\\")]
+        [InlineData('\b', "\\b")]
+        [InlineData('\f', "\\f")]
+        public void ApplyProperty_UntypedValueOfTypeStringUnescaped(char token, string tokenEscaped)
+        {
+            var escapedValue = "\"Update of memo" + tokenEscaped + "4" + tokenEscaped + "40\"";
```
xuzhg, line 385: // Arrange

KenitoInc (approved):
LGTM

marabooy (approved):
LGTM

Issue #2307, opened 2020-10-01
ExpressionBinderBase.GuidCompare is not playing well with EF Core


brunoAltinet:
when calling compare on guids, EF Core complains: InvalidOperationException: The LINQ expression 'DbSet<CatalogElement>
.Where(c => ExpressionBinderBase.GuidCompare(
firstValue: c.CatalogElementGuid,
secondValue: __TypedProperty_0) < 0)' could not be translated

### Assemblies affected
Latest, see guidCompare method:
https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs

### Reproduce steps

Any call to odata that invokes Guid comparison eg:
/odata/CatalogElements/$count?$filter=CatalogElementGuid%20lt%20ae36fb69-01d0-4c67-9cd9-1a7dbcb7dbc9

### Expected result

It would return count for aforementioned method

### Actual result

EF Core exception:

InvalidOperationException: The LINQ expression 'DbSet<CatalogElement>
.Where(c => ExpressionBinderBase.GuidCompare(
firstValue: c.CatalogElementGuid,
secondValue: __TypedProperty_0) < 0)' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to either AsEnumerable(), AsAsyncEnumerable(), ToList(), or ToListAsync(). See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.

brunoAltinet:
I could maybe create a PR for this, let me know if that's necessary

marabooy:
@brunoAltinet your contribution will be greatly appreciated. Do mention me or @Sreejithpin if you need any help.

xuzhg:
@Sreejithpin Please involve Smit Patel into this issue.

PR #2342, opened 2020-11-02
compare Non-nullable guid

xuzhg:
### Issues

*This pull request fixes issue #https://github.com/OData/WebApi/issues/2307.*

### Description

*We have the code to compare non-nullable guid and compare them as "null".

Guid is value type, how can we compare to null?

This PR is to fix that.

- [] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -33,7 +33,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
     public abstract class ExpressionBinderBase
     {
         internal static readonly MethodInfo StringCompareMethodInfo = typeof(string).GetMethod("Compare", new[] { typeof(string), typeof(string) });
-        internal static readonly MethodInfo GuidCompareMethodInfo = typeof(ExpressionBinderBase).GetMethod("GuidCompare", new[] { typeof(Guid), typeof(Guid) });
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1103,27 +1100,6 @@ internal static Expression ConvertNull(Expression expression, Type type)
...
-            {
-                return firstValue.CompareTo(secondValue);
-            }
-
-            if (secondValue != null)
-            {
-                return (-1) * secondValue.CompareTo(firstValue);
-            }
-
-            return 0;
```
mikepizzo, line 1124: This is technically a breaking change to a public interface, although I doubt it is being used we could keep it here and just not use it when building the expression (and remove in 8.0).

I'm also fine removing if we are pretty confident it's not being used externally.

mikepizzo:
See my one comment -- removing GuidCompare is technically a breaking change. 

It seems highly unlikely that anyone is using it, but if we think this could break anyone we could keep it and just not use it when building an expression.

mikepizzo (approved):
:shipit:

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
xuzhg: I do think there's less customer who relies on this method.  It's "bad" method and we should fix before it broadcasts to more customer. Besides, it's a "easy" task at customer side to re-create a similar GUID compare method. So, in my opinion, let's remove it.

Issue #2327, opened 2020-10-20
Batch response passes preflight check but the response is empty


andygjp:
Batch requests fail preflight checks. After using a custom batch middleware, I can pass preflight checks, but the body is empty despite getting a "valid" response.

### Assemblies affected

Nuget package Microsoft.AspNetCore.OData 7.5.0

### Reproduce steps

```
  // application setup in my Startup.cs
  public void ConfigureServices(IServiceCollection services)
  {
            services.AddCors(options => 
                options.AddDefaultPolicy(builder => 
                    builder
                        .WithOrigins("http://localhost:3000")
                        .AllowAnyHeader()
                        .AllowAnyMethod()
                        .AllowCredentials()
                        .WithExposedHeaders("ETag")));

            services.AddOData();
            services.AddSingleton<ODataBatchHandler, DefaultODataBatchHandler>();
            // ...
  }

  public void Configure(IApplicationBuilder app)
  {
            // ...
            app
                .UseMiddleware<CorsAwareODataBatchMiddleware>() 
                .UseRouting()
                .UseCors()
                .UseEndpoints(routeBuilder =>
                {
                    routeBuilder.MapControllers();

                    routeBuilder.Select().Expand().Filter().OrderBy().MaxTop(null).Count();

                    routeBuilder.MapODataRoute(
                        "ODataRoute",
                        "api",
                        containerBuilder =>
                        {
                            containerBuilder
                                .AddService(ServiceLifetime.Singleton,
                                    _ => EdmModelBuilder.GetEdmModel(
                                        new ODataConventionModelBuilder(routeBuilder.ServiceProvider)))
                                .AddService(ServiceLifetime.Singleton,
                                    _ => routeBuilder.ServiceProvider.GetRequiredService<ODataBatchHandler>())
                                .AddService<IEnumerable<IODataRoutingConvention>>(ServiceLifetime.Singleton,
                                    _ => ODataRoutingConventions.CreateDefaultWithAttributeRouting("ODataRoute",
                                        routeBuilder.ServiceProvider))
                                .AddServicePrototype(new ODataMessageReaderSettings(ODataVersion.V401)
                                {
                                    EnableMessageStreamDisposal = false,
                                    MessageQuotas = new ODataMessageQuotas
                                    {
                                        MaxReceivedMessageSize = long.MaxValue
                                    }
                                })
                                .AddServicePrototype(new ODataMessageWriterSettings
                                {
                                    EnableMessageStreamDisposal = false,
[...]

PR #2356, opened 2020-11-12
Fix behaviour for Preflight requests on the ODataBatchMiddleware.

marabooy:
### Issues

*This pull request fixes issue #2327.*

### Description
This introduces a change that allows the batch handler to yield to other middleware's requests that are preflight. This allows the cors middleware to do its job if registered.
For this to work correctly the Cors handler needs to be registered before calling
```csharp
app.UseCors(); //needs to be before batching.
app.UseODataBatching();
```

- [X] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.OData.Batch
         public async Task Invoke(HttpContext context)
         {
             HttpRequest request = context.Request;
-            bool isPreFlight = request.Method.Equals("Options", StringComparison.OrdinalIgnoreCase);
+            bool isPreFlight = HttpMethods.IsOptions(request.Method);
            
             // Attempt to match the path to a batch route.
             ODataBatchPathMapping batchMapping = context.RequestServices.GetRequiredService<ODataBatchPathMapping>();
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
@@ -34,19 +34,13 @@ namespace Microsoft.AspNet.OData.Batch
         /// <returns>A task that can be awaited.</returns>
         public async Task Invoke(HttpContext context)
         {
-            var request = context.Request;
-            var isPreFlight = request.Method.Equals("Options", StringComparison.OrdinalIgnoreCase);
-            if (isPreFlight)
-            {
-                // if the request is a preflight request let the next handler be called as trying to handle the request will result in preflight requests failing.
-                await this.next(context);
-                return;
-            }
-
-            // Attempt to match the path to a bach route.
+            HttpRequest request = context.Request;
+            bool isPreFlight = request.Method.Equals("Options", StringComparison.OrdinalIgnoreCase);
+           
+            // Attempt to match the path to a batch route.
             ODataBatchPathMapping batchMapping = context.RequestServices.GetRequiredService<ODataBatchPathMapping>();
 
-            if (batchMapping.TryGetRouteName(context, out var routeName))
+            if (!isPreFlight && batchMapping.TryGetRouteName(context, out var routeName))
             {
                 // Get the per-route container and retrieve the batch handler.
                 IPerRouteContainer perRouteContainer = context.RequestServices.GetRequiredService<IPerRouteContainer>();
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNet.OData.Batch
             // Attempt to match the path to a batch route.
             ODataBatchPathMapping batchMapping = context.RequestServices.GetRequiredService<ODataBatchPathMapping>();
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -34,13 +34,21 @@ public ODataBatchMiddleware(RequestDelegate next)
...
+            var request = context.Request;
+            var isPreFlight = request.Method.Equals("Options", StringComparison.OrdinalIgnoreCase);
+            if (isPreFlight)
+            {
+                // if the request is a preflight request let the next handler be called as trying to handle the request will result in preflight requests failing.
+                await this.next(context);
+                return;
+            }
+
             // Attempt to match the path to a bach route.
```
ElizabethOkerio, line 46: batch

odero (approved):

src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -34,13 +34,21 @@ public ODataBatchMiddleware(RequestDelegate next)
...
             ODataBatchPathMapping batchMapping = context.RequestServices.GetRequiredService<ODataBatchPathMapping>();
 
-            string routeName;
-            if (batchMapping.TryGetRouteName(context, out routeName))
+            if (batchMapping.TryGetRouteName(context, out var routeName))
```
mikepizzo, line 49: >var [](start = 58, length = 3)

please use string, not var

src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -34,13 +34,21 @@ public ODataBatchMiddleware(RequestDelegate next)
         /// <returns>A task that can be awaited.</returns>
         public async Task Invoke(HttpContext context)
         {
+            var request = context.Request;
```
mikepizzo, line 37: >var [](start = 12, length = 3)

please use types, not var, for improved code readabilty

src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -34,13 +34,21 @@ public ODataBatchMiddleware(RequestDelegate next)
...
+            var isPreFlight = request.Method.Equals("Options", StringComparison.OrdinalIgnoreCase);
+            if (isPreFlight)
+            {
+                // if the request is a preflight request let the next handler be called as trying to handle the request will result in preflight requests failing.
+                await this.next(context);
```
mikepizzo, line 42: >await [](start = 16, length = 5)

in this scenario, you can probably just return this.next(context) to avoid the final await.  see [https://blog.stephencleary.com/2016/12/eliding-async-await.html](https://blog.stephencleary.com/2016/12/eliding-async-await.html) on eliding the final await.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -34,13 +34,15 @@ public ODataBatchMiddleware(RequestDelegate next)
         /// <returns>A task that can be awaited.</returns>
         public async Task Invoke(HttpContext context)
         {
-            // Attempt to match the path to a bach route.
+            HttpRequest request = context.Request;
+            bool isPreFlight = request.Method.Equals("Options", StringComparison.OrdinalIgnoreCase);
```
KenitoInc, line 38: Why not use `bool isPreFlight = HttpMethods.IsOptions(request.Method)`?

KenitoInc (commented):
Looks good.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
marabooy: Updated.

marabooy: Updated

src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -34,13 +34,15 @@ public ODataBatchMiddleware(RequestDelegate next)
...
-            // Attempt to match the path to a bach route.
+            HttpRequest request = context.Request;
+            bool isPreFlight = HttpMethods.IsOptions(request.Method);
+           
+            // Attempt to match the path to a batch route.
             ODataBatchPathMapping batchMapping = context.RequestServices.GetRequiredService<ODataBatchPathMapping>();
 
-            string routeName;
-            if (batchMapping.TryGetRouteName(context, out routeName))
+            if (!isPreFlight && batchMapping.TryGetRouteName(context, out var routeName))
```
gathogojr, line 43:
```suggestion
            if (!isPreFlight && batchMapping.TryGetRouteName(context, out string routeName))
```
I however think that both `out string routeName` and `out var routeName` would be flagged by the Fx analyzer which is why in lots of places we declare the variable first before using it as an out parameter

gathogojr (approved):
LGTM

KenitoInc (approved):
Update the publicApi

ElizabethOkerio (approved):

aulati:
Should the middleware order be updated after this PR?
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-7.0#middleware-order

marabooy:
Yes this should be updated to follow this sequence.

PR #2361, opened 2020-11-13
Enable JSON Metadata

xuzhg:
### Issues

### Description

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Enums/EnumsTest.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Enums/EnumsTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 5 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Enums/EnumsTest.cs
```diff
@@ -57,6 +58,38 @@ public async Task ModelBuilderTest(string modelMode)
             var reader = new ODataMessageReader(message);
             var edmModel = reader.ReadMetadataDocument();
 
+            VerifyEdmModel(edmModel);
+        }
+
+#if NETCOREAPP3_1
```
mikepizzo, line 64: can we do #if !NETSTANDARD2_0?  That way, the test will implicitly be supported on future versions of the framework.

xuzhg: In the test project, we target the real implementation of ".NETStandard".  

.NETCoreApp3.1 is an implementation .NETStandard 2.0.  In my opinion, ".NETStandard 2.0" is used in Library, in the running APP,  real implementation is used either.

mikepizzo: My point is, we should exclude the environments in which the test should not run, rather than include the environments in which it should run. That way, as we add more environments, it will be included by default.

xuzhg: Oh, got it.

mikepizzo (approved):
:shipit:

PR #2368, opened 2020-11-24
Fix bug in attribute convention route values

### LINKED ISSUE #2363: Wrong endpoint be selected if 'get' action methods share the same ActionName in attribute rounting scenario
If two 'Get' action methods share the same name (such as 'Get')  in attribute rounting scenario and both of them are not decorated by [HttpGet], then one of them will be selected to solve two urls, regardless of the path templates specified in [ODataRoute] attributes.

### Assemblies affected

Microsoft.AspNetCore.OData 7.5.1
Microsoft.AspNetCore.OData 7.5.2

### Reproduce steps

Change Get methods in samples/AspNetCore3xEndpointSample.Web/Controllers/CustomersController.cs to:

```

        [ODataRoute("Customers")]
        [EnableQuery]
        public IActionResult GetA() // Use 'GetA' but not 'Get' to exclude convention routing
        {
            // Be noted: without the NoTracking setting, the query for $select=HomeAddress with throw exception:
            // A tracking query projects owned entity without corresponding owner in result. Owned entities cannot be tracked without their owner...
            _context.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;

            return Ok(_context.Customers);
        }

        [ODataRoute("Customers/{key}")]
        [EnableQuery]
        //[HttpGet] // Remove this comment to make this method work
        public IActionResult GetA(int key)
        {
            return Ok(_context.Customers.FirstOrDefault(c => c.Id == key));
        }

```

### Expected result
 
url '{server}/odata/Customers' returns entity collection
url '{server}/odata/Customers/1' returns single entity

### Actual result

both urls return entity collection

### Additional detail

Adding [HttpGet] to the second 'GetA' method will make it work.

When there is a 'key' parameter whose binding source is routepath, the second method is **NEVER** selected to handle '{server}/odata/Customers', it **CAN** be selected to handle '{server}/odata/Customers/1' with [HttpGet]. 

If we remove the routepath 'key' parameter(remove the parameter or add [FromQuery] to change binding source), the second method **CAN** be selected to handle both urls as the first method. Which method is selected to handle both urls depend on:
1. the method with [HttpGet] will be selected.
2. if both methods is not decorated with [HttpGet], the method which has more parameter in signature will be selected.
3. if both methods have same number of parameters, the first method by order will be selected. 

If change one method name(or specify a different name by [ActionName] attribute), then things work correctly.

What is expected is that [ODataRoute] attribute should be respected with highest priority. Method signature and metadata should **NOT** affect which endpoint to choose once methods can be distinguished by route template in [ODataRoute] attibute.

habbes:
### Issues

Fixes #2363 

### Description

**Some background**
Routing conventions that add values from the url to the routing convention store also keep track of the number of values added (by using the `IncrementKeyCount` helper method). The action selector uses this information to match the controller action with the correct number of parameters (amongst many overloads).

**Current issue**
The `AttributeRoutingConvention` adds values to routing convention store directly without using existing helper methods and did not increment the key count. This misleads the action selector to behave as if there were not keys in the url and consequently picks the `Get()` action without parameters over the `Get(int key)` action. The `AttributeRoutingConvention` also had an additional bug that caused it to add extra key-value pairs to the routing store, because each `ODataRoute` template it tried to match could insert values to the store, even if that template would not end up matching the path.

This PR fixes those issues by incrementing the key count for each value added to the convention store and clearing the values left over by route templates that did not match the current odata path.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
@@ -59,6 +59,11 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
 
                     return result;
                 }
+
+                // It's possible that template.TryMatch inserted values in the values dict even if
+                // it did not match the current path. So let's clear the dict before trying
+                // the next template
+                values.Clear();
             }
 
             return null;
@@ -79,6 +84,7 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
                         item.Value is ODataParameterValue)
                     {
                         routingConventionsStore.Add(item);
+                        RoutingConventionHelpers.IncrementKeyCount(routingConventionsStore);
                     }
                     else
                     {
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingModel.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingModel.cs
@@ -32,6 +32,7 @@ namespace Microsoft.AspNet.OData.Test.Routing
             builder.EntitySet<Incident>("Incidents");
             builder.EntitySet<NotFoundCustomer>("NotFoundCustomers");
             builder.EntitySet<NotFoundWithIdCustomer>("NotFoundWithIdCustomers");
+            builder.EntitySet<AttributeCustomer>("AttributeCustomers");
             builder.ComplexType<Dog>();
             builder.ComplexType<Cat>();
             builder.EntityType<SpecialProduct>();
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
```diff
@@ -59,6 +59,11 @@ private AttributeRoutingConvention(string routeName)
 
                     return result;
                 }
+
+                // It's possible that template.TryMatch inserted values in the values dict even if
```
xuzhg, line 63: It's certain ... , right?

xuzhg (approved):

gathogojr (approved):

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
habbes: It's no certain. It depends on whether `TryMatch` was called on path segments that have values, for example `KeySegment`. But if  `TryMatch` was only called on `EntitySetSegment`, then it would not add values to the dict.

KenitoInc (approved):

PR #2385, opened 2020-12-18
Fix #2294: batch processing loses HttpContext for main request.

### LINKED ISSUE #2294: `IHttpContextAccessor.HttpContext` returns null when executing within an odata batch call.
`IHttpContextAccessor.HttpContext` returns null when executing within an odata batch call.

### Assemblies affected

Microsoft.AspNetCore.OData 7.5.0

### Reproduce steps

Simply setup a simple odata batch pipeline

```
var defaultBatchHandler = new DefaultODataBatchHandler();
defaultBatchHandler.MessageQuotas.MaxNestingDepth = 2;
defaultBatchHandler.MessageQuotas.MaxOperationsPerChangeset = 10;
```

and create an action that users the http context accessor:

```
[ODataRoutePrefix("me")]
public class MeController : ODataController
{
    private readonly IHttpContextAccessor httpContextAccessor;

    public MeController(IHttpContextAccessor httpContextAccessor)
    {
        this.httpContextAccessor = httpContextAccessor;
    }

    [EnableQuery]
    [HttpGet]
    public User Get()
    { 
        var user = this.httpContextAccessor.HttpContext.User;
        // this.HttpContext is fine
        return new User { Name = user.Identity.Name };
    }
}
```

the following request
```
POST /$batch
Content-Type: application/json
{
    "requests": [
        {
            "id": "1",
            "method": "GET",
            "url": "/me"
        }
    ]
}
```
Produces the error

### Expected result

In the example above HttpContextAccessor.HttpContext should be equal to this.HttpContext - the http context of the controller/action.

### Actual result

HttpContextAccessor.HttpContext is null.

### Additional detail

I realize from the simple example above I can simply use this.HttpContext but I need to access the httpcontext in a more deeper part of my code.

NicholasNoise:
### Issues

This pull request fixes issue #2294.

### Description

```csharp
HttpContext context = null;
IHttpContextFactory httpContextFactory = originalContext.RequestServices.GetRequiredService<IHttpContextFactory>();
if (httpContextFactory != null)
{
    context = httpContextFactory.Create(features);
}
else
{
    context = new DefaultHttpContext(features);
}
```
`IHttpContextFactory`(`DefaultHttpContextFactory`) service  should not be used to create new `HttpContext`, because it resets current context to new one:
https://github.com/dotnet/aspnetcore/blob/master/src/Hosting/Hosting/src/Http/DefaultHttpContextFactory.cs#L73

- [x] *Test cases added*


src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
```diff
@@ -239,17 +238,9 @@ private static HttpContext CreateHttpContext(HttpContext originalContext)
...
-            {
-                context = httpContextFactory.Create(features);
-            }
-            else
-            {
-                context = new DefaultHttpContext(features);
-            }
+            // Create a context.
+            // IHttpContextFactory should not be used, because it resets IHttpContextAccessor.HttpContext;
+            HttpContext context = new DefaultHttpContext(features);
```
mikepizzo, line 243: >HttpContext context = new DefaultHttpContext(features); [](start = 12, length = 55)

What are the ramifications of not using the context factory?  Doesn't that mean that, if someone adds their own context factory, we don't use it for batch?  Couldn't that be a problem if the developer is expecting their custom httpcontext to be used?  Do we instead need a way to create the context from the factory without resetting the IHttpContextAccessor.HttpContext?

mikepizzo:
Hi @NicholasNoise -- thanks for the contribution!

As per my inline comments, I'm curious what the ramifications are if we don't use the injected HttpContextFactory. Eager to get your (and others') thoughts.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Batch/DefaultODataBatchHandlerTest.cs
```diff
@@ -727,7 +730,7 @@ public async Task SendAsync_CorrectlyHandlesCookieHeader()
             var changesetRef = $"changeset_{Guid.NewGuid()}";
             var endpoint = "http://localhost";
 
-            Type[] controllers = new[] { typeof(BatchTestCustomersController), typeof(BatchTestOrdersController), };
```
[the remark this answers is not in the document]
NicholasNoise: I've added unused controller with #2379, it is just a cleanup.
So yes, it works as expected.

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
NicholasNoise: I do not see any other reason to use the factory for creating `HttpContext` from a set of sub-requests, but dependency injection. And it is unclear why a global factory (that creates context for an actual http request) should be used to instance batch sub-contexts.

If the factory is used we should make a `closure` of original context after processing batch changesets like this one:
```csharp
        /// <inheritdoc />
        public override async Task ProcessBatchAsync(HttpContext context, RequestDelegate nextHandler)
        {
            // Retrieve current httpcontext.
            var httpContextAccessor = context.RequestServices.GetService<IHttpContextAccessor>();
            var beforeContext = httpContextAccessor?.HttpContext;

            await base.ProcessBatchAsync(context, nextHandler);

            if (httpContextAccessor != null)
            {
                httpContextAccessor.HttpContext = beforeContext;
            }
        }
```
But still a closure is more like a workaround than THE solution.

Keyword `new` is kinda bad in common, especially in the name of DI, instead we can introduce an interface `IBatchHttpContextFactory` to create a context of the changeset.

NicholasNoise:
@mikepizzo
How do I make this PR unblocked? =)

src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
NicholasNoise: Hi @marabooy
Yeah, we should give more control of the context creation. But as I've pointed out `IHttpContextFactory` is not a way to do so. I suggest a new abstraction `IBatchHttpContextFactory` for this matter.

Issue #2395, opened 2021-01-12
Concurrent requests with different $filters return invalid responses in OData 7.5+ (Including 8-prev3)


ysmoradi:
Write a simple function which returns in memory fixed data
Call that function concurrently with different $fitlers
You'd expect results filtered based on the provided filter
But you'll see invalid responses

### Assemblies affected

OData 7.5.4+ has this issue (including 8 prev3)
OData 7.5.2 and lower versions are fine

### Reproduce steps

```cs
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddControllers();
        services.AddOData();
    }

    public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
    {
        app.UseRouting();

        app.UseEndpoints(endpoints =>
        {
            endpoints.Select().Filter().OrderBy().Count().Expand().MaxTop(100);
            endpoints.MapODataRoute("odata", "odata", GetEdmModel());
        });
    }

    IEdmModel GetEdmModel()
    {
        ODataConventionModelBuilder builder = new();

        builder.EntitySet<TestModel>("Test");

        return builder.GetEdmModel();
    }
}

public class TestModel
{
    public int Id { get; set; }
}

public class TestController : ODataController
{
    [HttpGet, EnableQuery, ODataRoute("Test")]
    public IQueryable<TestModel> Test()
    {
        return Enumerable.Range(1, 100)
            .Select(i => new TestModel
            {
                Id = i
            }).AsQueryable();
    }
}

[TestClass]
public class Test
{
    [TestMethod]
    public async Task Impl()
    {
        IHostBuilder hostBuilder = new HostBuilder()
            .ConfigureWebHost(webHost =>
            {
                webHost.UseTestServer();
                webHost.UseStartup<Startup>();
            });

        using IHost host = await hostBuilder.StartAsync();

        using HttpClient client = host.GetTestClient();

        var results = await Task.WhenAll(Enumerable.Range(1, 100).Select(i => client.GetCount(i)));

        foreach (var result in results)
        {
            Assert.AreEqual(100 - result.i, result.length);
        }
    }
}

public static class HttpClientExtensions
{
    public static async Task<(int i, int length)> GetCount(this HttpClient client, int i)
    {
        return (i, (await (await client.GetAsync($"odata/test?$filter=Id gt {i}")).EnsureSuccessStatusCode().Content.ReadFromJsonAsync<JsonElement>()).GetProperty("value").GetArrayLength());
    }
[...]

PR #2397, opened 2021-01-14
Moving EnableQuery request level query data storage to HttpContext

RamjotSingh:
### Issues

*This pull request fixes issue #2395*

### Description

Moving query data to HttpContext instead of instance variable since Attribute instances are not thread-safe.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
@@ -54,7 +54,11 @@ namespace Microsoft.AspNet.OData
 
             base.OnActionExecuting(context);
 
-            RequestQueryData requestQueryData = new RequestQueryData();
+            RequestQueryData requestQueryData = new RequestQueryData()
+            {
+                QueryValidationRunBeforeActionExecution = false,
+            };
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -420,5 +407,29 @@ public virtual void ValidateQuery(HttpRequest request, ODataQueryOptions queryOp
             Contract.Assert(model != null);
             return model;
         }
+
+        /// <summary>
+        /// Holds request level query information.
+        /// </summary>
+        private class RequestQueryData
```
xuzhg, line 414: >RequestQueryData [](start = 22, length = 16)

can we use the "ODataFeature" ?

xuzhg (approved):
:shipit:

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
RamjotSingh: Their is no intent to allow customers to change this value, thus exposing it in ODataFeature might give a sense that callers can change it, however changing it will result into break in queries.

PR #2400, opened 2021-01-21
Generate correct odata.nextlink when expanding multi-level contained entities

KenitoInc:
### Issues

*This pull request fixes issue #2262* 
*Additional work after PR #2287*
### Description

We have been generating an incorrect `navigationLink` when generating `nextLink` in multi-level expand in contained entities. This PR fixes that.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
@@ -101,10 +101,9 @@ namespace Microsoft.AspNet.OData.Builder
         /// </summary>
         /// <param name="resourceContext">The <see cref="ResourceContext"/> representing the entity for which the navigation link needs to be generated.</param>
         /// <param name="navigationProperty">The EDM navigation property.</param>
-        /// <param name="includeCast">Represents whether the generated link should have a cast segment representing a type cast.</param>
         /// <returns>The navigation link following the OData URL conventions.</returns>
         public static Uri GenerateContainedNavigationPropertyLink(this ResourceContext resourceContext,
-            IEdmNavigationProperty navigationProperty, bool includeCast)
+            IEdmNavigationProperty navigationProperty)
         {
             if (resourceContext == null)
             {
@@ -116,11 +115,6 @@ namespace Microsoft.AspNet.OData.Builder
             }
 
             IList<ODataPathSegment> navigationPathSegments = GenerateODataPathSegmentsForContainedEntity(resourceContext);
-
-            if (includeCast)
-            {
-                navigationPathSegments.Add(new TypeSegment(resourceContext.StructuredType, navigationSource: null));
-            }
 
             navigationPathSegments.Add(new NavigationPropertySegment(navigationProperty, navigationSource: null));
 
@@ -137,40 +131,39 @@ namespace Microsoft.AspNet.OData.Builder
             ResourceContext resourceContext)
         {
             IList<ODataPathSegment> odataPathSegments = new List<ODataPathSegment>();
-            ODataPath path = resourceContext.SerializerContext.Path;
-
-            // We first add segments in the Path.
-            var segments = path.Segments;
-            int length = segments.Count;
-
-            for (int i = 0; i < length; i++)
-            {
-                ODataPathSegment pathSegment = segments[i];
-                EntitySetSegment entitySetPathSegment = pathSegment as EntitySetSegment;
-                IEdmEntityType entityType = entitySetPathSegment.EntitySet.EntityType();
-
-                odataPathSegments.Add(entitySetPathSegment);
-                odataPathSegments.Add(new KeySegment(ConventionsHelpers.GetEntityKey(resourceContext), entityType, null));
-
-            }
-
             Stack<ODataPathSegment> internalOdataPath = new Stack<ODataPathSegment>();
-            IEdmNavigationSource currentNavigationSource = resourceContext.NavigationSource;
 
-            while (currentNavigationSource != null)
+            ResourceContext currentLevelResourceContext = resourceContext;
+
+            // Loop through the all ExpandedResource(s).
+            // We generate OData Path Segments for all levels.
+            while (currentLevelResourceContext != null)
             {
-                IEdmContainedEntitySet containedEntitySet = currentNavigationSource as IEdmContainedEntitySet;
-                if (containedEntitySet == null)
+                IEdmNavigationSource navigationSource = currentLevelResourceContext.NavigationSource;
+                IEdmEntitySet entitySet = navigationSource as IEdmEntitySet;
+                IEdmContainedEntitySet containedEntitySet = navigationSource as IEdmContainedEntitySet;
+
+                // For Contained entity sets
+                if(containedEntitySet != null)
                 {
-                    break;
+                    NavigationPropertySegment navigationPropertySegment = new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource);
+                    internalOdataPath.Push(new KeySegment(ConventionsHelpers.GetEntityKey(currentLevelResourceContext), currentLevelResourceContext.StructuredType as IEdmEntityType, null));
+                    internalOdataPath.Push(navigationPropertySegment);
                 }
-                NavigationPropertySegment navigationPropertySegment = new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource);                
-                internalOdataPath.Push(new KeySegment(ConventionsHelpers.GetEntityKey(resourceContext), resourceContext.StructuredType as IEdmEntityType, null));
-                internalOdataPath.Push(navigationPropertySegment);
 
-                currentNavigationSource = containedEntitySet.ParentNavigationSource;
+                // For Non-Contained entity sets
+                if(entitySet != null)
+                {
+                    EntitySetSegment entitySetSegment = new EntitySetSegment(entitySet);
+                    internalOdataPath.Push(new KeySegment(ConventionsHelpers.GetEntityKey(currentLevelResourceContext), currentLevelResourceContext.StructuredType as IEdmEntityType, null));
+                    internalOdataPath.Push(entitySetSegment);
+                }
 
+                currentLevelResourceContext = currentLevelResourceContext.SerializerContext.ExpandedResource;
             }
+
+            // We have added the path segments from the last to the first
+            // We then need to order them from the first to the last.
             foreach (var i in internalOdataPath)
             {
                 odataPathSegments.Add(i);
--- a/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
@@ -96,6 +96,90 @@ namespace Microsoft.AspNet.OData.Builder
         }
 
         /// <summary>
+        /// Generates a contained navigation link following the OData URL conventions for the entity represented by <paramref name="resourceContext"/> and the given
+        /// navigation property.
+        /// </summary>
+        /// <param name="resourceContext">The <see cref="ResourceContext"/> representing the entity for which the navigation link needs to be generated.</param>
+        /// <param name="navigationProperty">The EDM navigation property.</param>
+        /// <param name="includeCast">Represents whether the generated link should have a cast segment representing a type cast.</param>
+        /// <returns>The navigation link following the OData URL conventions.</returns>
+        public static Uri GenerateContainedNavigationPropertyLink(this ResourceContext resourceContext,
+            IEdmNavigationProperty navigationProperty, bool includeCast)
+        {
+            if (resourceContext == null)
+            {
+                throw Error.ArgumentNull("resourceContext");
+            }
+            if (resourceContext.InternalUrlHelper == null)
+            {
+                throw Error.Argument("resourceContext", SRResources.UrlHelperNull, typeof(ResourceContext).Name);
+            }
+
+            IList<ODataPathSegment> navigationPathSegments = GenerateODataPathSegmentsForContainedEntity(resourceContext);
+
+            if (includeCast)
+            {
+                navigationPathSegments.Add(new TypeSegment(resourceContext.StructuredType, navigationSource: null));
+            }
+
+            navigationPathSegments.Add(new NavigationPropertySegment(navigationProperty, navigationSource: null));
+
+            string link = resourceContext.InternalUrlHelper.CreateODataLink(navigationPathSegments);
+            if (link == null)
+            {
+                return null;
+            }
+
+            return new Uri(link);
+        }
+
+        private static IList<ODataPathSegment> GenerateODataPathSegmentsForContainedEntity(
+            ResourceContext resourceContext)
+        {
+            IList<ODataPathSegment> odataPathSegments = new List<ODataPathSegment>();
+            ODataPath path = resourceContext.SerializerContext.Path;
+
+            // We first add segments in the Path.
+            var segments = path.Segments;
+            int length = segments.Count;
+
+            for (int i = 0; i < length; i++)
+            {
+                ODataPathSegment pathSegment = segments[i];
+                EntitySetSegment entitySetPathSegment = pathSegment as EntitySetSegment;
+                IEdmEntityType entityType = entitySetPathSegment.EntitySet.EntityType();
+
+                odataPathSegments.Add(entitySetPathSegment);
+                odataPathSegments.Add(new KeySegment(ConventionsHelpers.GetEntityKey(resourceContext), entityType, null));
+
+            }
+
+            Stack<ODataPathSegment> internalOdataPath = new Stack<ODataPathSegment>();
+            IEdmNavigationSource currentNavigationSource = resourceContext.NavigationSource;
+
+            while (currentNavigationSource != null)
+            {
+                IEdmContainedEntitySet containedEntitySet = currentNavigationSource as IEdmContainedEntitySet;
+                if (containedEntitySet == null)
+                {
+                    break;
+                }
+                NavigationPropertySegment navigationPropertySegment = new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource);                
+                internalOdataPath.Push(new KeySegment(ConventionsHelpers.GetEntityKey(resourceContext), resourceContext.StructuredType as IEdmEntityType, null));
+                internalOdataPath.Push(navigationPropertySegment);
+
+                currentNavigationSource = containedEntitySet.ParentNavigationSource;
+
+            }
+            foreach (var i in internalOdataPath)
+            {
+                odataPathSegments.Add(i);
+            }
+
+            return odataPathSegments;
+        }
+
+        /// <summary>
         /// Generates an action link following the OData URL conventions for the action <paramref name="action"/> and bound to the
         /// collection of entity represented by <paramref name="resourceSetContext"/>.
         /// </summary>
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
@@ -486,10 +486,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet)
             {
                 // Contained navigation.
-                Uri idlink = linkBuilder.BuildIdLink(writeContext.ExpandedResource);
-
-                var link = idlink.ToString() + "/" + writeContext.NavigationProperty.Name;
-                navigationLink = new Uri(link);
+                navigationLink = writeContext.ExpandedResource.GenerateContainedNavigationPropertyLink(writeContext.NavigationProperty, false);
             }
             else
             {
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,83 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
         }
 
+        /// <summary>
+        /// Generates a contained navigation link following the OData URL conventions for the entity represented by <paramref name="resourceContext"/> and the given
+        /// navigation property.
+        /// </summary>
+        /// <param name="resourceContext">The <see cref="ResourceContext"/> representing the entity for which the navigation link needs to be generated.</param>
+        /// <param name="navigationProperty">The EDM navigation property.</param>
+        /// <returns>The navigation link following the OData URL conventions.</returns>
+        public static Uri GenerateContainedNavigationPropertyLink(this ResourceContext resourceContext,
```
xuzhg, line 105: >GenerateContainedNavigationPropertyLink [](start = 26, length = 39)

Can we use the above "GenerateNavigationPropertyLink" method with some changes?

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,83 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+            IEdmNavigationProperty navigationProperty)
```
xuzhg, line 106: >ionProperty [](start = 42, length = 11)

the above method has "includeCast", shall we add such in this new method?

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,83 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                // For Contained entity sets
+                if(containedEntitySet != null)
+                {
+                    NavigationPropertySegment navigationPropertySegment = new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource);
+                    internalOdataPath.Push(new KeySegment(ConventionsHelpers.GetEntityKey(currentLevelResourceContext), currentLevelResourceContext.StructuredType as IEdmEntityType, null));
+                    internalOdataPath.Push(navigationPropertySegment);
+                }
+
+                // For Non-Contained entity sets
+                if(entitySet != null)
```
xuzhg, line 155: >if(entitySet != null) [](start = 16, length = 21)

else?

xuzhg, line 155: >if(entitySet != null) [](start = 16, length = 21)

how about if the path starts with singleton? the navigation source is "IEdmSingleton"

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,83 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+            // Loop through the all ExpandedResource(s).
+            // We generate OData Path Segments for all levels.
+            while (currentLevelResourceContext != null)
+            {
+                IEdmNavigationSource navigationSource = currentLevelResourceContext.NavigationSource;
+                IEdmEntitySet entitySet = navigationSource as IEdmEntitySet;
+                IEdmContainedEntitySet containedEntitySet = navigationSource as IEdmContainedEntitySet;
+
+                // For Contained entity sets
+                if(containedEntitySet != null)
```
xuzhg, line 147: >if(containedEntitySet != null) [](start = 16, length = 30)

It could have the type cast in the path? for example:

~/customers/NS.VipCustomer/ContainedOrders/NS.VipOrder

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: Added `includeCast`
#Resolved

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: Added `includeCast`
#Resolved

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: Added capabilities for Singleton
#Resolved

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: Added type cast.
#Resolved

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: We don't need else here.
We are checking if any of the following is not null and adding the appropriate Path Segments
- entitySet
- containedEntitySet
- singleton

habbes: Also, is it possible for there to be a valid segment other than the 3 you have tested for? Like for example a property segment?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -486,10 +486,7 @@ private static Uri GetNestedNextPageLink(ODataSerializerContext writeContext, in
             if (writeContext.NavigationSource.NavigationSourceKind() == EdmNavigationSourceKind.ContainedEntitySet)
             {
                 // Contained navigation.
-                Uri idlink = linkBuilder.BuildIdLink(writeContext.ExpandedResource);
-
-                var link = idlink.ToString() + "/" + writeContext.NavigationProperty.Name;
-                navigationLink = new Uri(link);
+                navigationLink = writeContext.ExpandedResource.GenerateContainedNavigationPropertyLink(writeContext.NavigationProperty, false);
```
habbes, line 489: Maybe add `includeCast: false` for clarity?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/LinkGenerationHelpersTest.cs
```diff
@@ -54,6 +54,24 @@ public void GenerateNavigationLink_WorksToGenerateExpectedNavigationLink_ForEnti
             Assert.Equal(expectedNavigationLink, uri.AbsoluteUri);
         }
 
+        [Theory]
+        [InlineData(false, "http://localhost/MyOrders(42)/OrderLines")]
+        [InlineData(true, "http://localhost/MyOrders(42)/NS.MyOrder/OrderLines")]
```
habbes, line 59: Could you also add a test where the last path segment is a type cast?

habbes (changes requested):

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,102 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                // For Singletons
+                if(singleton != null)
+                {
+                    SingletonSegment singletonSegment = new SingletonSegment(singleton);
+                    internalOdataPath.Push(singletonSegment);
+                }
+
+                currentLevelResourceContext = currentLevelResourceContext.SerializerContext.ExpandedResource;
+
+                if (includeCast & currentLevelResourceContext != null)
```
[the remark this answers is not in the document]
habbes: Just checked the docs and realized that `&` is also used as an AND operator that does not short-circuit when the first operand is false. But in this case, is it necessary to evaluate `currentLevelResourceContext != null` if `includeCast` is false?

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,102 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                }
+
+                // For Singletons
+                if(singleton != null)
+                {
+                    SingletonSegment singletonSegment = new SingletonSegment(singleton);
+                    internalOdataPath.Push(singletonSegment);
+                }
+
+                currentLevelResourceContext = currentLevelResourceContext.SerializerContext.ExpandedResource;
```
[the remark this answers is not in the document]
KenitoInc: Actually, it's the reverse. In `Employee`, the expanded resource is `Store`.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSetSerializer.cs
KenitoInc: I have updated to ` /*includeCast*/ false`

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: 1. I have update the code to use pattern matching
2. I don't think there are other valid segments in this use case. I will be interested in getting suggestions.

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,102 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                    internalOdataPath.Push(new TypeSegment(currentLevelResourceContext.StructuredType, navigationSource: null));
+                }
+            }
+
+            // We have added the path segments from the last to the first
+            // We then need to order them from the first to the last.
+            foreach (var i in internalOdataPath)
+            {
+                odataPathSegments.Add(i);
+            }
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/LinkGenerationHelpersTest.cs
KenitoInc: Actually in this test the last segment has a typecast

habbes (approved):

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+        /// <param name="includeCast">Represents whether the generated link should have a cast segment representing a type cast.</param>
+        /// <returns>The navigation link following the OData URL conventions.</returns>
+        public static Uri GenerateContainedNavigationPropertyLink(this ResourceContext resourceContext,
+            IEdmNavigationProperty navigationProperty, bool includeCast)
+        {
+            if (resourceContext == null)
+            {
+                throw Error.ArgumentNull("resourceContext");
+            }
+            if (resourceContext.InternalUrlHelper == null)
```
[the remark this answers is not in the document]
mikepizzo: Yes; the empty line should be there per style guidelines, and will be required once we enforce style guidelines.

---
In reply to: [572719349](https://github.com/OData/WebApi/pull/2400#discussion_r572719349) [](ancestors = 572719349)

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+
+            navigationPathSegments.Add(new NavigationPropertySegment(navigationProperty, navigationSource: null));
+
+            string link = resourceContext.InternalUrlHelper.CreateODataLink(navigationPathSegments);
+            if (link == null)
+            {
+                return null;
+            }
+
+            return new Uri(link);
```
mikepizzo, line 133: >link [](start = 27, length = 4)

will link always be an absolute Uri?

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                    reversedOdataPathSegments.Push(new KeySegment(ConventionsHelpers.GetEntityKey(currentLevelResourceContext), currentLevelResourceContext.StructuredType as IEdmEntityType, null));
+                    reversedOdataPathSegments.Push(entitySetSegment);
+                }
+
+                // For Singletons
+                if(navigationSource is IEdmSingleton singleton)
+                {
+                    SingletonSegment singletonSegment = new SingletonSegment(singleton);
+                    reversedOdataPathSegments.Push(singletonSegment);
+                }
```
mikepizzo, line 171: if we encounter a navigation source that is a singleton, or non-contained entity set (and we have a key), shouldn't we reset the reversedOdatapathSegments in order to return the canonical navigation path?  i.e., if we have an indirect path like users(user1)/groups(group1)/members(user5)/address, where address is a contained navigation property and members is a navigation path to a user (bound to a non-contained entity set "users") then shouldn't we ignore the users(user1)/groups(group1) and just return users(user5)/address?  And similarly, if company is a top level singleton, shouldn't users(user1)/bestFriend/friends(user7)/company/address just return company/address?

What happens if we have groups(group1)/owner/address (where "owner is a single-value non-contained nav prop on a group), would we have to return the (non-canonical) groups(group1)/owner/address, or would we have the id of the owner so that we could return users(userId)/address?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/LinkGenerationHelpersTest.cs
```diff
@@ -54,6 +54,24 @@ public void GenerateNavigationLink_WorksToGenerateExpectedNavigationLink_ForEnti
             Assert.Equal(expectedNavigationLink, uri.AbsoluteUri);
         }
 
+        [Theory]
+        [InlineData(false, "http://localhost/MyOrders(42)/OrderLines")]
```
mikepizzo, line 58: >http://localhost/MyOrders(42)/OrderLines [](start = 28, length = 40)

what if you have a single-value, non-contained, navigation source in the path.  i.e., assuming a Customer has a non-contained single-valued navigation property "mostRecentOrder" and the request was:
[http://localhost/Customers(42)/mostRecentOrder/OrderLines](http://localhost/Customers(42)/mostRecentOrder/OrderLines)

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+        /// <summary>
+        /// Generates a contained navigation link following the OData URL conventions for the entity represented by <paramref name="resourceContext"/> and the given
+        /// navigation property.
+        /// </summary>
+        /// <param name="resourceContext">The <see cref="ResourceContext"/> representing the entity for which the navigation link needs to be generated.</param>
+        /// <param name="navigationProperty">The EDM navigation property.</param>
+        /// <param name="includeCast">Represents whether the generated link should have a cast segment representing a type cast.</param>
+        /// <returns>The navigation link following the OData URL conventions.</returns>
+        public static Uri GenerateContainedNavigationPropertyLink(this ResourceContext resourceContext,
+            IEdmNavigationProperty navigationProperty, bool includeCast)
```
mikepizzo, line 107: >includeCast [](start = 60, length = 11)

how does the caller know whether to includeCast? Shouldn't we determine this based on whether the navigationProperty is defined on the same type as the navigationSource?  If so, we don't need the cast, if not we have to include it.

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                // For Singletons
+                if(navigationSource is IEdmSingleton singleton)
+                {
+                    SingletonSegment singletonSegment = new SingletonSegment(singleton);
+                    reversedOdataPathSegments.Push(singletonSegment);
+                }
+
+                currentLevelResourceContext = currentLevelResourceContext.SerializerContext.ExpandedResource;
+
+                if (includeCast && currentLevelResourceContext != null)
```
mikepizzo, line 175: >includeCast [](start = 20, length = 11)

we should determine this for each step depending on whether the type that the segment we are appending is defined on matches the type of the previous segment.

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+            IList<ODataPathSegment> odataPathSegments = new List<ODataPathSegment>();
+            Stack<ODataPathSegment> reversedOdataPathSegments = new Stack<ODataPathSegment>();
+
+            ResourceContext currentLevelResourceContext = resourceContext;
+
+            // Loop through the all ExpandedResource(s).
+            // We generate OData Path Segments for all levels.
+            while (currentLevelResourceContext != null)
+            {
+                IEdmNavigationSource navigationSource = currentLevelResourceContext.NavigationSource;
```
mikepizzo, line 148: >NavigationSource [](start = 84, length = 16)

is there ever a case where navigationSource is null? BadEntitySet? Are those cases properly handled?

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                // For Contained entity sets
+                if (navigationSource is IEdmContainedEntitySet containedEntitySet)
+                {
+                    NavigationPropertySegment navigationPropertySegment = new NavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet.ParentNavigationSource);
+                    reversedOdataPathSegments.Push(new KeySegment(ConventionsHelpers.GetEntityKey(currentLevelResourceContext), currentLevelResourceContext.StructuredType as IEdmEntityType, null));
+                    reversedOdataPathSegments.Push(navigationPropertySegment);
+                }
+
+                // For Non-Contained entity sets
+                if(navigationSource is IEdmEntitySet entitySet)
```
mikepizzo, line 159: What happens if we have groups(group1)/owner/address (where "owner is a single-value non-contained nav prop on a group), would we have to return the (non-canonical) groups(group1)/owner/address, or would rely on being able to get the id of the user from ResourceContext so that we could return users(userId)/address?

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
```diff
@@ -95,6 +95,99 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
...
+                {
+                    EntitySetSegment entitySetSegment = new EntitySetSegment(entitySet);
+                    reversedOdataPathSegments.Push(new KeySegment(ConventionsHelpers.GetEntityKey(currentLevelResourceContext), currentLevelResourceContext.StructuredType as IEdmEntityType, null));
```
mikepizzo, line 162: >GetEntityKey [](start = 85, length = 12)

is there ever a case where we won't be able to get the entity key in the resource context (i.e., if the key properties weren't selected)? How do we handle such a case? Or do we enforce that the key properties are always selected in this case?

mikepizzo (changes requested):
:clock1:

mikepizzo:
Bunch of questions about how this works for edge cases. Perhaps those edge cases never apply due to how this is called, but want to be sure we think through and make sure they are not applicable and that we have relevant assertions in the code.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/LinkGenerationHelpersTest.cs
```diff
@@ -89,6 +107,24 @@ public void GenerateNavigationLink_WorksToGenerateExpectedNavigationLink_ForSing
             Assert.Equal(expectedNavigationLink, uri.AbsoluteUri);
         }
 
+        [Theory]
```
mikepizzo, line 110: >Theory [](start = 9, length = 6)

also add a test case for a request that includes a function in the path.

src/Microsoft.AspNet.OData.Shared/Builder/LinkGenerationHelpers.cs
KenitoInc: `CreateODataLink` will always generate an absolute uri.

Issue #328, opened 2015-05-05
[Feature] Add Stream Support to WebApi


liqian19891011:
Add contents for Stream support in OData V4 on odata.github.io/WebApi.

We have related questions on [StackOverflow](http://stackoverflow.com/questions/30043442/stream-support-in-odata-v4)

congysu:
Consider to support named stream (aka stream property) or media link entry (i.e., media entry)
If named stream, we need to change OData entry serializer.
```c#
public class CustomODataEntityTypeSerializer : ODataEntityTypeSerializer
{
    public CustomODataEntityTypeSerializer(IEdmEntityTypeReference entityType, ODataSerializerProvider serializerProvider)
        : base(entityType, serializerProvider)
    {
    }

    public override ODataProperty CreateStructuralProperty(IEdmStructuralProperty structuralProperty, object entityInstance, ODataSerializerContext writeContext)
    {
        if (structuralProperty.Type.IsStream())
        {
            return new ODataProperty { Name = structuralProperty.Name, Value = new ODataStreamReferenceValue { ReadLink = new Uri("http://image") } };
        }

        return base.CreateStructuralProperty(structuralProperty, entityInstance, writeContext);
    }
}
```

If media link entry, we need to create an entity type with HasStream set to true in OData v4.  E.g., use this ctor:
```c#
public EdmEntityType(string namespaceName, string name, IEdmEntityType baseType, bool isAbstract, bool isOpen, bool hasStream);
```
The change is in model builder.

sammym1982:
thanks @congysu  Is this issue accepted for next release of libraries? 

For interim do you have any suggestion on how we can go ahead and add the stream support.

congysu:
@sammym1982 The related issue is https://github.com/OData/WebApi/issues/149.  Can start from a blog: http://blogs.msdn.com/b/mrtechnocal/archive/2013/12/31/10462258.aspx, although it is for OData v3 Web API.

sammym1982:
@congysu 

I do tried porting that code to V4 but the issue as I mentioned in my original post on stackoverflow is there is currently no way available to mark an entity as HasStream = true. The original IEdmModel.SetHasDefaultStream  seems like no longer available. 

Model we have is as followos:

Entity: Incident
NavigationProperties Collection: Attachments

Therefore We want all our attachment posts to happen as below:

POST: http://foo.com/incidents(1L)/Attachments

VikingsFan:
[...]

PR #2401, opened 2021-01-28
Support write resource with stream property and Support query stream property

xuzhg:
…property

### Issues

*This pull request fixes issue #328*

### Description

* Make the resource serialize successful if the resource has stream property
* Support the stream property query

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataStreamMediaTypeMapping.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataStreamMediaTypeMapping.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.OData.Formatter
+{
+    /// <summary>
+    /// Media type mapping that associates requests with stream property.
+    /// </summary>
+    public partial class ODataStreamMediaTypeMapping
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataStreamMediaTypeMapping"/> class.
+        /// </summary>
+        public ODataStreamMediaTypeMapping()
+            : base("application/octet-stream")
+        {
+        }
+    }
+}
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -592,6 +592,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                     else
                     {
                         writer.WriteStart(resource);
+                        WriteStreamProperties(selectExpandNode, resourceContext, writer);
                         WriteComplexProperties(selectExpandNode, resourceContext, writer);
                         WriteDynamicComplexProperties(resourceContext, writer);
                         WriteNavigationLinks(selectExpandNode, resourceContext, writer);
@@ -633,6 +634,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                     else
                     {
                         await writer.WriteStartAsync(resource);
+                        await WriteStreamPropertiesAsync(selectExpandNode, resourceContext, writer);
                         await WriteComplexPropertiesAsync(selectExpandNode, resourceContext, writer);
                         await WriteDynamicComplexPropertiesAsync(resourceContext, writer);
                         await WriteNavigationLinksAsync(selectExpandNode, resourceContext, writer);
@@ -1269,6 +1271,58 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             }
         }
 
+        private void WriteStreamProperties(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
+        {
+            Contract.Assert(selectExpandNode != null);
+            Contract.Assert(resourceContext != null);
+            Contract.Assert(writer != null);
+
+            if (selectExpandNode.SelectedStructuralProperties != null)
+            {
+                IEnumerable<IEdmStructuralProperty> structuralProperties = selectExpandNode.SelectedStructuralProperties;
+
+                foreach (IEdmStructuralProperty structuralProperty in structuralProperties)
+                {
+                    if (structuralProperty.Type != null && structuralProperty.Type.IsStream())
+                    {
+                        ODataStreamPropertyInfo property = CreateStreamProperty(structuralProperty, resourceContext);
+
+                        if (property != null)
+                        {
+                            writer.WriteStart(property);
+                            writer.WriteEnd();
+                        }
+                    }
+                }
+            }
+        }
+
+        private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
+        {
+            Contract.Assert(selectExpandNode != null);
+            Contract.Assert(resourceContext != null);
+            Contract.Assert(writer != null);
+
+            if (selectExpandNode.SelectedStructuralProperties != null)
+            {
+                IEnumerable<IEdmStructuralProperty> structuralProperties = selectExpandNode.SelectedStructuralProperties;
+
+                foreach (IEdmStructuralProperty structuralProperty in structuralProperties)
+                {
+                    if (structuralProperty.Type != null && structuralProperty.Type.IsStream())
+                    {
+                        ODataStreamPropertyInfo property = CreateStreamProperty(structuralProperty, resourceContext);
+
+                        if (property != null)
+                        {
+                            await writer.WriteStartAsync(property);
+                            await writer.WriteEndAsync();
+                        }
+                    }
+                }
+            }
+        }
+
         private IEnumerable<KeyValuePair<IEdmStructuralProperty, PathSelectItem>> GetPropertiesToWrite(SelectExpandNode selectExpandNode, ResourceContext resourceContext)
         {
             IDictionary<IEdmStructuralProperty, PathSelectItem> complexProperties = selectExpandNode.SelectedComplexTypeProperties;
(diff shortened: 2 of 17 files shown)


src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
```diff
@@ -291,6 +291,21 @@ public override Task<object> ReadFromStreamAsync(Type type, Stream readStream, H
...
             {
+                if (typeof(Stream).IsAssignableFrom(type))
+                {
+                    // Ideally, it should go into the "ODataRawValueSerializer",
+                    // However, OData lib doesn't provide the method to overwrite/copyto stream
+                    // So, Here's the workaround
+                    Stream objStream = value as Stream;
+                    if (objStream != null)
+                    {
+                        objStream.CopyToAsync(writeStream);
```
NicholasNoise, line 302: Do we need to await asynchronous operations to ensure its completion?
```csharp
                    if (objStream != null)
                    {
                        await objStream.CopyToAsync(writeStream);
                        await writeStream.FlushAsync();
                    }
```

xuzhg: No, we don't.
And C# compile doesn't allow to use await in a non-async method.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ODataStreamPropertyTest.cs
```diff
@@ -0,0 +1,222 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+#if NETCORE
```
marabooy, line 4: These seem duplicated

src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
mikepizzo: Yes, but if we don't wait for CopyToAsync to complete, then FlushAsync could interrupt the CopyTo.  So I think we should either mark this method as async and do the awaits (which is the right thing once we make ODL fully async), or we should call the synchronous objectStream.CopyTo(writeStream).

---
In reply to: [566634550](https://github.com/OData/WebApi/pull/2401#discussion_r566634550) [](ancestors = 566634550)

xuzhg: `Task WriteResponseBodyAsync` is a vritual method, we can't change the syntax of this method.

I updated the codes using await. please take a look again.

src/Microsoft.AspNet.OData/Formatter/ODataMediaTypeFormatter.cs
```diff
@@ -335,6 +344,16 @@ public override Task<object> ReadFromStreamAsync(Type type, Stream readStream, H
             }
         }
 
+        private static async Task CopyStreamAsync(Stream source, Stream destination)
+        {
+            if(source != null)
```
NicholasNoise, line 349: `if (source != null)` - space is missing.

xuzhg: Fixed, thanks @NicholasNoise

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ODataStreamPropertyTest.cs
xuzhg: what do you mean "duplicated"? @marabooy 

---
In reply to: [566650769](https://github.com/OData/WebApi/pull/2401#discussion_r566650769) [](ancestors = 566650769)

marabooy: I meant that the two blocks of the namespace imports are similar for both net core & non net core namespaces

xuzhg: Yes. We separate them and avoid multiple #if #else #endif 

---
In reply to: [568498223](https://github.com/OData/WebApi/pull/2401#discussion_r568498223) [](ancestors = 568498223)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1561,6 +1615,12 @@ private IEnumerable<ODataProperty> CreateStructuralPropertyBag(SelectExpandNode
 
                 foreach (IEdmStructuralProperty structuralProperty in structuralProperties)
                 {
+                    if (structuralProperty.Type != null && structuralProperty.Type.IsStream())
+                    {
+                        // skip the stream property, the stream property is written in its own logic
+                        continue;
+                    }
```
xuzhg, line 1622: @mikepizzo It's a little bit confusing.

For the stream property serializer, what 's the correct way to do?

1) Create a normal "ODataProperty" and build its value as "ODataStreamReferenceValue"   or
2) Crete a "ODataStreamPropertyInfo" and call "WriteStart(),  WriteEnd()" 

It's "ODataStreamPropertyInfo" only for $expand=photo ?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1572,6 +1632,43 @@ private IEnumerable<ODataProperty> CreateStructuralPropertyBag(SelectExpandNode
...
+
+            if (structuralProperty.Type == null || !structuralProperty.Type.IsStream())
+            {
+                return null;
+            }
+
+            if (resourceContext.SerializerContext.MetadataLevel != ODataMetadataLevel.FullMetadata)
+            {
+                return null;
+            }
```
mikepizzo, line 1661: If the stream property does not exactly match convention (i.e., we know the mediatype of the stream) then we should explicitly write the stream property info rather than let ODL generate it with default values.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1572,6 +1632,43 @@ private IEnumerable<ODataProperty> CreateStructuralPropertyBag(SelectExpandNode
...
+
+            // TODO: we need to return ODataStreamReferenceValue if
+            // 1) If we have the EditLink link builder
+            // 2) If we have the ReadLink link builder
+            // 3) If we have the Core.AcceptableMediaTypes annotation associated with the Stream property
+
+            // So far, let's return null and let OData.lib to calculate the ODataStreamReferenceValue by conventions.
+            return null;
```
mikepizzo, line 1669: If the user has an instance annotation associated with this property that specifies the media type, then we should explicitly write the streamreferencevalue and not let ODL fill it in.

src/Microsoft.AspNet.OData.Shared/Formatter/ODataStreamMediaTypeMapping.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Media type mapping that associates requests with stream property.
+    /// </summary>
+    public partial class ODataStreamMediaTypeMapping
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataStreamMediaTypeMapping"/> class.
+        /// </summary>
+        public ODataStreamMediaTypeMapping()
```
mikepizzo, line 14: >ODataStreamMediaTypeMapping [](start = 15, length = 27)

If I know the media type is something other than application/octet-stream, can I specify that somewhere?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ODataStreamPropertyTest.cs
```diff
@@ -0,0 +1,222 @@
...
+        {
+            IList<StreamCustomer> customers = CreateCustomers();
+            StreamCustomer customer = customers.FirstOrDefault(c => c.Id == key);
+            if (customer == null)
+            {
+                return NotFound();
+            }
+
+            return Ok(customer.Photo);
+        }
```
mikepizzo, line 189: We need a way to return a media-type other than application/octet-stream here (i.e., image/jpeg)

mikepizzo:
Looks good to me, with 2 caveats:
1) In most cases, application/octet-stream won't be the best media-type to return. We need to understand, document, and test returning a stream property value with a type other than application/octet-stream.  How do we make this easy/intuitive?
2) The user should be able to specify a content-type for an instance of a stream property (or media entity, if supported), and we should return that in the streamreferencepropertyinfo.

xuzhg:
> Looks good to me, with 2 caveats:
> 
> 1. In most cases, application/octet-stream won't be the best media-type to return. We need to understand, document, and test returning a stream property value with a type other than application/octet-stream.  How do we make this easy/intuitive?
> 2. The user should be able to specify a content-type for an instance of a stream property (or media entity, if supported), and optionally custom read/edit links (and perhaps etag values), and we should return that in the streamreferencepropertyinfo.

Thanks for caveats. I'd like to unblock customers first and move these as continuing tasks.

src/Microsoft.AspNet.OData.Shared/Formatter/ODataStreamMediaTypeMapping.cs
xuzhg: So far, customer can create the Formatter and set the content type.
Later, we'd figure out a solution for customer to set up the content type.

---
In reply to: [574808287](https://github.com/OData/WebApi/pull/2401#discussion_r574808287) [](ancestors = 574808287)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
xuzhg: Yes, we'd figure out the solution to let customer set the value. It's next step.

---
In reply to: [574804726](https://github.com/OData/WebApi/pull/2401#discussion_r574804726) [](ancestors = 574804726)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
xuzhg: Yes. It's the next step. I am working on this design. 

---
In reply to: [574805575](https://github.com/OData/WebApi/pull/2401#discussion_r574805575) [](ancestors = 574805575)

PR #2418, opened 2021-02-23
Check for empty Namespace

### LINKED ISSUE #952: StructuralTypeConfiguration class's FullName property does not check if the Namespace is an empty string.
StructuralTypeConfiguration class's FullName property does not check if the Namespace is an empty string.

### Assemblies affected
Microsoft.AspNet.OData.6.0.0

### Reproduce steps
```
// set the namespace to an empty string
ODataModelBuilder builder = new ODataConventionModelBuilder();
EntitySetConfiguration<Order> entitySetConfiguration = builder.EntitySet<Order>("Orders");
entitySetConfiguration.EntityType.Namespace = string.Empty;

// what is FullName?
throw new Exception(entitySetConfiguration.FullName);
```

### Expected result
`System.Exception: Order`

### Actual result
`System.Exception: .Order`

### Additional details
Notice the preceding '.'

There's no current work-around AFAIK because I can't seem to inject my own custom `StructuralTypeConfiguration`.
`builder.AddEntityType` will always use a structural type of `EntityTypeConfiguration`.

icnocop:
### Issues

Fixes #952

### Description

Check if the `Namespace` is an empty string when returning the `FullName` of `EnumTypeConfiguration` and `StructuralTypeConfiguration`.

- [x] *Test cases added*

### DIFF
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationOfTStructuralTypeTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationOfTStructuralTypeTest.cs
@@ -4,6 +4,7 @@
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Test.Common;
 using Moq;
+using System;
 using Xunit;
 
 namespace Microsoft.AspNet.OData.Test.Builder
@@ -28,13 +29,13 @@ namespace Microsoft.AspNet.OData.Test.Builder
         [Fact]
         public void Property_Name_RoundTrips()
         {
-            ReflectionAssert.Property(_configuration, c => c.Name, "Name", allowNull: false, roundTripTestValue: _name);
+            ReflectionAssert.Property<StructuralTypeConfiguration<object>, string, ArgumentException>(_configuration, c => c.Name, "Name", allowNull: false, roundTripTestValue: _name);
         }
 
         [Fact]
         public void Property_Namespace_RoundTrips()
         {
-            ReflectionAssert.Property(_configuration, c => c.Namespace, "Namespace", allowNull: false, roundTripTestValue: _namespace);
+            ReflectionAssert.Property<StructuralTypeConfiguration<object>, string, ArgumentException>(_configuration, c => c.Namespace, "Namespace", allowNull: false, roundTripTestValue: _namespace);
         }
 
         [Fact]
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationOfTStructuralTypeTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationOfTStructuralTypeTest.cs
@@ -1,10 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Test.Common;
 using Moq;
-using System;
 using Xunit;
 
 namespace Microsoft.AspNet.OData.Test.Builder
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
@@ -214,7 +214,7 @@ namespace Microsoft.AspNet.OData.Test.Builder
         [InlineData(null)]
         [InlineData("")]
         [InlineData(" ")]
-        public void NameAssignment_WithNullOrWhiteSpace_ThrowsArgumentException(string @namespace)
+        public void NameAssignment_WithNullOrWhiteSpace_ThrowsArgumentException(string name)
         {
             // Arrange
             ODataConventionModelBuilder modelBuilder = new ODataConventionModelBuilder();
@@ -222,7 +222,7 @@ namespace Microsoft.AspNet.OData.Test.Builder
             EntityTypeConfiguration<MyOrder> entityTypeConfiguration = modelBuilder.EntitySet<MyOrder>("orders").EntityType;
 
             // Act and Assert
-            ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Name = @namespace, "value");
+            ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Name = name, "value");
         }
     }
 }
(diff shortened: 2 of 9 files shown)


gathogojr:
@icnocop Based on [the specification](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Toc453752522), should the value for the namespace ever be null or empty?

icnocop:
The spec indicates the namespace must be unique.

It doesn't explicitly say it must not be empty.

Does the spec need to be updated?

Currently, if you set the namespace to an empty string, the 'FullName' will begin with a period, which is not expected.

If the spec needs to be updated to clarify that requirement, an exception should occur if the namespace is being set to an empty string.

mikepizzo:
@icnocop -- thanks for the contribution!

Clearly the existing logic is flawed, as it allows setting the namespace to an empty string and then builds a qualified name prefixed with ".".

From Section 6 of the [abnf](http://docs.oasis-open.org/odata/odata/v4.01/os/abnf/odata-abnf-construction-rules.txt):
qualifiedEnumTypeName = namespace "." enumerationTypeName
namespace  = namespacePart *( "." namespacePart )
namespacePart = odataIdentifier
; Note: this pattern is overly restrictive, the normative definition is type TSimpleIdentifier in OData EDM XML Schema
odataIdentifier = identifierLeadingCharacter *127identifierCharacter

Which all means that namespace must have at least one character.

So the real issue in the code is that it lets the user set the namespace to an empty string, which should not be allowed.

icnocop:
Thank you for referencing the ABNF.

I've updated the code to throw an `ArgumentException` if the `StructuralTypeConfiguration.Namespace` or `StructuralTypeConfiguration.Name` properties are set to null, empty, or consists of only white-space characters.

I reverted the changes I had made to check the same properties for `EnumTypeConfiguration` because it seems to behave differently.

Thank you.

icnocop:
Let me know if I should squash my commits into one.

mikepizzo:
Hi @icnocop  -- thanks for the follow-up. No need to squash, we'll do that when we merge.

Slight concern that this is a breaking change (i.e, if someone were previously setting namespace to null it wouldn't throw an exception), but given the fact that it would almost certainly break further down the line, I think it's probably safe to accept.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationOfTStructuralTypeTest.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.AspNet.OData.Builder;
 using Microsoft.AspNet.OData.Test.Common;
 using Moq;
+using System;
```
mikepizzo, line 7: Nit: System using statements should come before other using statements.

mikepizzo (approved):
:shipit:

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationOfTStructuralTypeTest.cs
icnocop: Fixed. Thank you.

KenitoInc (approved):
LGTM

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -188,5 +188,39 @@ public void NamespaceAssignment_AutoAssignsNamespaceToStructuralType_NamespaceAs
             Assert.Equal(expectedNamespace, modelBuilder.EntityType<OrderHeader>().Namespace);
             Assert.Equal(expectedNamespace, modelBuilder.ComplexType<ZipCode>().Namespace);
         }
+
+        /// <summary>
+        /// Tests the Namespace property setter logic with a null or white space value throws an argument exception.
```
gathogojr, line 193: @icnocop Instead of having 3 asserts for 3 different scenarios, we could use:
```csharp
[Theory]
[InlineData(null)]
[InlineData("")
[InlineData(" ")]
public void NamespaceAssignment_WithNullOrWhiteSpace_ThrowsArgumentException(string namespace)
{
    // ...
    ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Namespace = namespace, "value");
}
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -188,5 +188,39 @@ public void NamespaceAssignment_AutoAssignsNamespaceToStructuralType_NamespaceAs
...
+            ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Namespace = null, "value");
+            ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Namespace = string.Empty, "value");
+            ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Namespace = " ", "value");
+        }
+
+        /// <summary>
+        /// Tests the Name property setter logic with a null or white space value throws an argument exception.
+        /// </summary>
+        [Fact]
+        public void NameAssignment_WithNullOrWhiteSpace_ThrowsArgumentException()
```
gathogojr, line 213: @icnocop Instead of having 3 asserts for 3 different scenarios, we could use:
```csharp
[Theory]
[InlineData(null)]
[InlineData("")
[InlineData(" ")]
public void NameAssignment_WithNullOrWhiteSpace_ThrowsArgumentException(string name)
{
    // ...
    ExceptionAssert.ThrowsArgument(() => entityTypeConfiguration.Name = name, "value");
}
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
icnocop: Fixed. Thank you.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
icnocop: Fixed. Thank you.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/StructuralTypeConfigurationTest.cs
```diff
@@ -188,5 +188,41 @@ public void NamespaceAssignment_AutoAssignsNamespaceToStructuralType_NamespaceAs
...
+        }
+
+        /// <summary>
+        /// Tests the Name property setter logic with a null or white space value throws an argument exception.
+        /// </summary>
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData(" ")]
+        public void NameAssignment_WithNullOrWhiteSpace_ThrowsArgumentException(string @namespace)
```
gathogojr, line 217: @icnocop Maybe use `name` to avoid confusion...

icnocop: Done. Thank you.

gathogojr (approved):

Issue #2384, opened 2020-12-17
Scale is not working on decimal property


komdil:
*I am trying to round decimal property using Scale, but it is not working*

### Assemblies affected

*Microsoft.AspNetCore.OData 7.4.1

### Reproduce steps

**Create an entity:
```
    public class Student 
    {
        public Guid Id { get; set; }
        public decimal Decimal { get; set; } = 0.000000m;
    }
```
configuration:

```
            builder.EntitySet<Student>("Student");
            var student = builder.EntityType<Student>();
            student.Property(s => s.Decimal).Scale = 2;
```
Send request: https://localhost:44383/OData/Student

### Expected result

*I am expecting result will be *

```
{
    "@odata.context": "https://localhost:44383/OData/$metadata#Student",
    "value": [
        {
            "Decimal": 0.00,
            "Id": "efb93e81-2729-40e2-bd0b-44dc338217f9"
        }
    ]
}
```

### Actual result

Result is:
```

{
    "@odata.context": "https://localhost:44383/OData/$metadata#Student",
    "value": [
        {
            "Decimal": 0.000000,
            "Id": "efb93e81-2729-40e2-bd0b-44dc338217f9"
        }
    ]
}
```

komdil:
Workarround provied here: https://stackoverflow.com/questions/65351595/how-to-set-decimal-precision-in-odata/65878710#65878710

Reena-Patel:
@komdil  Do we need to set CustomODataPrimitiveSerializer to the EDM model? or just adding the below code will work which is shared [here](https://stackoverflow.com/questions/65351595/how-to-set-decimal-precision-in-odata?answertab=active#tab-top)?

```
public static decimal CalculateDecimalPrecision(decimal value, int precision)
    {
        if (ShouldCalculatePrecision(value, precision, out bool shouldRound))
        {
            if (shouldRound)
            {
                value = Math.Round(value, precision);
            }
            else
            {
                var calculatedPrecision = decimal.Divide(1, Convert.ToDecimal(Math.Pow(10, precision)));
                var zeroAfterPoint = calculatedPrecision - calculatedPrecision;
                value += zeroAfterPoint;
            }
        }
        return value;
    }

public class CustomODataPrimitiveSerializer : ODataPrimitiveSerializer
{
    /// <summary>
    /// Rewriting decimal value
    /// </summary>
[...]

PR #2420, opened 2021-02-25
Scale is not working on decimal property

komdil:
### Issues

*This pull request fixes issue #2384*

### Description

If scale of decimal property was set, ODataPrimitiveSerializer should calculate it. 
One more note: It is confusing what is difference between Scale and Precision? Should one them be removed?

- [x] *Test cases added*


komdil:
Test is passing on my local

KenitoInc:
@komdil 
This is the failing test 
failingAspNetTest

komdil:
> @komdil
> This is the failing test
> failingAspNetTest

It is passing on my local. Is it possible requeue build?

chrisspre:
> One more note: It is confusing what is difference between Scale and Precision? Should one them be removed?
> 

Informally, precision is the maximum number of digits allow on the left and right of the decimal point. scale it the number of digits to the right of the decimal point. For floating point numbers they are the same. For fixed point numbers not. 
more details [here](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Precision)

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ScaleDecimalTest.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNet.OData.Test.Builder.TestModels;
+using Microsoft.OData.Edm;
+using Newtonsoft.Json;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Xunit;
+using System.Globalization;
```
xuzhg, line 12: can you add the copy right

and sorting the using

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ScaleDecimalTest.cs
```diff
@@ -0,0 +1,98 @@
...
+            IEnumerable<Order> orders = await ReadEntitiesFromResponse<Order>(response);
+            foreach (Order order in orders)
+            {
+                Assert.Equal("10.00", order.TotalPrice.ToString(CultureInfo.InvariantCulture));
+            }
+        }
+
+        private static HttpClient GetClient()
+        {
+            var controllers = new[] { typeof(MetadataController), typeof(OrdersController) };
```
xuzhg, line 40: >typeof(MetadataController), [](start = 38, length = 27)

we can exclude MetadataController if no test for metadata

xuzhg (dismissed):
:shipit:

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ScaleDecimalTest.cs
komdil: added and sorted

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ScaleDecimalTest.cs
komdil: excluded

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/ScaleDecimalTest.cs
```diff
@@ -0,0 +1,101 @@
...
+            HttpClient client = GetClient();
+
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, Uri);
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json;odata.metadata=full"));
+
+            // Act
+            HttpResponseMessage response = await client.SendAsync(request);
+
+            // Assert
+            IEnumerable<Order> orders = await ReadEntitiesFromResponse<Order>(response);
```
gathogojr, line 34: @komdil Let's assume decimal values in the raw response appear as 10:00. The moment the payload is deserialized into CLR objects, there's no guarantee the scale is respected or preserved as the `TotalPrice` property is materialized. Consequently, `order.TotalPrice.ToString(CultureInfo.InvariantCulture)` won't necessarily yield "10.00"

An alternative approach (hopefully foolproof) is to read the response stream into a string and base your assert on that, i.e.
```csharp
var response = await client.SendAsync(request);
var responseContent = await response.Content.ReadAsStringAsync();
Assert.Equal("EXPECTED", responseContent);
```

If your fix is good then the array part of the payload would be expected to be something like this: 
`[{"OrderId":1,"TotalPrice":10.00},{"OrderId":2,"TotalPrice":10.00},{"OrderId":3,"TotalPrice":10.00}]`

Kindly let me know if you're in a position to try this to fix the failing test(s)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataPrimitiveSerializer.cs
```diff
@@ -151,21 +152,40 @@ internal static object ConvertPrimitiveValue(object value, IEdmPrimitiveTypeRefe
...
         }
 
+        static decimal CalculateDecimalScale(decimal value, int scale)
+        {
+            string scaleForConverting = "0.";
+            for (int i = 0; i < scale; i++)
+            {
+                scaleForConverting += "0";
+            }
+            return decimal.Parse(value.ToString(scaleForConverting, CultureInfo.InvariantCulture), CultureInfo.InvariantCulture);
```
habbes, line 186: Since the final size of the `scaleForConverting` string is known in advance (scale + 2), I wonder whether some string allocations can be avoided here. But maybe this is premature optimization:
```c#
decimal value = 10.0040m;
int scale = 2;
// we could have threshold for scale size beyond which we allocate an array
Span<char> scaleForConverting = stackalloc char[scale + 2];
scaleForConverting[0] = '0';
scaleForConverting[1] = '.';
for (int i = 0; i < scale; i++) {
    scaleForConverting[2 + i] = '0';
}
decimal result = decimal.Parse(
    value.ToString(new String(scaleForConverting), CultureInfo.InvariantCulture),
    CultureInfo.InvariantCulture);
```

habbes: Alternatively, would `decimal.Round` work here?

I have compared your implementation with `decimal.Round(value, scale)`. The two use different rounding strategies when rounding a midpoint value. In the case of your method, it seems to always round up (i.e. 10.0050 will round up to 10.01 and 10.0150 will round up to 10.02 when the scale is 2). `decimal.Round`  uses the "round half to even strategy" (i.e. 10.0050 will round down to 10.00 and 10.0150 will round up to 10.02).

I'm not sure what the desired rounding strategy is, or what the spec says about this. But you can use an overload of `decimal.Round` that accepts a round strategy. The following snippet is functionally equivalent to your method (I also tried it with negative numbers):
```c#
decimal.Round(value, scale, MidpointRouting.AwayFromZero)
```

habbes (changes requested):

corranrogue9 (dismissed):

pfedotovsky:
When will this PR be merged?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataPrimitiveSerializer.cs
mikepizzo: If we can replace the above code with decimal.Round(), that would definitely be my preferred approach.

The [round function](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_round), at least, is defined in OData as rounding away from zero, so it would probably be consistent to do that here as well.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Resources/DecimalFullMetadata.json
```diff
@@ -1,3 +1,3 @@
 ﻿{
-  "@odata.context":"http://localhost/$metadata#Edm.Decimal","@odata.type":"#Decimal","value":12345.99999
+  "@odata.context":"http://localhost/$metadata#Edm.Decimal","@odata.type":"#Decimal","value":12345
```
habbes, line 2: Why was this change made?

komdil: Guys, I think first we need to define expectation. Currently, we  have a Scale and Precision. So here is explanation at picture: 
image

In my opinion, first we need to validate: scale cannot be higher then precision, because it does not make sense 
I created this unit test for some use cases:

```
        [TestCase(3, 0, 123, "123")]
        [TestCase(7, 5, 123, "12.30000")]
        [TestCase(3, 2, 123, "1.23")]
        [TestCase(6, 3, 123, "123.000")]
        [TestCase(6, 3, 123.123, "123.123")]
        [TestCase(6, 4, 123.123, "12.3123")]
        public void DecimalConverterTest(int precision, int scale, decimal decimalValue, string expectedDecimalString)
        {
            var converter = new MyDecimalConverter(scale, precision);
            var actualDecimal = converter.ConvertDecimal(decimalValue).ToString();
            Assert.That(actualDecimal, Is.EqualTo(expectedDecimalString));
        }
```
Since it is not working currently, this change make surprise some people after updating package. I would be careful on making this change because it will affect highly in some use cases. For example, from our use case, we have Amount in transaction which we use Scale and Precision. So, our client may see different amount of transaction. 
So, it should be included in documentation and in release notes also.

habbes: @komdil, in the test cases you've shared above, I have some confusions: Is the converter converting 123 to 1.23 when precision is 3 and scale is 2? Doesn't that actually change the value of the decimal? Same concern for 123.123 -> 12.3123.

komdil: @habbes Actually, I don't know what is expected here. Look at the picture above, when precision is 3 and scale is 2 and decimal value is 1 it will be 1.00 and it looks OK. So, in the case when the decimal value is 123 I don't know what would be expected. Do you have any idea? What did you decide in your stand-up meeting?

komdil: In my opinion, in OData having only scale without precision is make sense. Scale and precision are mostly will be used in database

habbes: @komdil my view is that OData can either:
-  correct data to ensure it matches the specification of the Precision and Scale values
- or it can validate values check whether they match the specification and throw an exception if they don't
- or both

I think the second case is easier to deal with. We look at the input value and check whether it matches the expected precision and scale without trying to do any rounding or modification. In this scenario, it is the responsibility of the service owner or client to ensure the values are valid before serialization. Would this be feasible for you?

But if we want to support the request scenario were OData actual "coerces" the values to match the specified Precision and Scale, we should be aware that the input values might be impossible to represent with the required precision+scale combination. How do we handle such cases? I don't think we should try to force the values to fit the precision when that would effectively change the value of the input (e.g. transforming 123.123 to 12.123). In this case either OData should throw an exception, or it should do a best-effort and let the values through even if they exceed the expected precision.

Because of these considerations, I agree that it's better to just support Scale for now where we don't worry about precision. That way we only focus on rounding. We can deal with Precision when support for it is required.

habbes:
@komdil could you make the following changes to your PR:
- add tests to verify that rounding works correctly for negative numbers. As @mikepizzo proposed, the rounding strategy should be consistent with the [round function](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_round). The midpoint should be rounded away from zero (your current implementation always rounds up)
- Consider rounding using `decimal.Round()` with `MidpointRouting.AwayFromZero` since it already implement the request behaviour

habbes:
It seems the current tests are failing on CI possibly to Culture differences. Could the tests be updated to not be Culture-dependent?

komdil:
@habbes the reason for failing test is the scale of decimal being set to 0 from 

`Microsoft.OData.Edm.Csdl.CsdlConstants.Default_Scale = 0 `

When it rounds 12345.99999 with scale it will be 12346. Changing expected behavior may fix the failed test, but I am not sure it is the correct way. Is there any suggestion here?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataPrimitiveSerializer.cs
```diff
@@ -155,19 +156,54 @@ internal static object ConvertPrimitiveValue(object value, IEdmPrimitiveTypeRefe
...
+        static decimal CalculateScaleOfDecimal(decimal value, int scale)
+        {
+            string leftSide = GetLeftSideFromPoint(value);
+            string rightSide = GetRightSideFromPoint(value);
+            var numberOfZeroToAddToTheEndOfDecimalValue = scale - rightSide.Length;
+            for (int i = 0; i < numberOfZeroToAddToTheEndOfDecimalValue; i++)
             {
-                TimeOfDay tod = (TimeSpan)value;
-                return tod;
+                rightSide += "0";
```
habbes, line 190: Since you know the size of the string in advance (i.e. `numberOfZeroAtoAddToTheEndOfDecimalValue`), you could pre-allocate block of chars on the stack using `stackalloc` or rent an array from the array pool if the size exceeds a certain threshold. Then you can add the `'0'` chars to the block and append to `rightSide` to avoid allocating all those temporary strings when doing `rightSide += "0"` inside of a loop.

e.g.
```c#
chart[] rentedArray = null;
Span<char>rightSideBuffer;
if (scale <= 256) {
   buffer = stackalloc char[scale];
}
else {
   rentedArray = ArrayPool<chart>.Rent(scale);
   buffer = rentedArray;
}

for (int i = 0; i < scale && i < rightSideLength; i++) {
    rightSideBuffer[rightSideStartPos + i] = valueAsString[rightSideStartPos + i];
}

for (int i = rightSideLength; i < scale; i++) {
   rightSdeBuffer[rightSideLength + i] = '0';
}

string rightSide = new string(rightSideBuffer.Slice(0, slice));

if (rentedArray != null) {
   ArrayPool<char>.Return(rentedArray);
}
```
Not sure if that ties all the screws, but hopefully the general idea is clear. Maybe we can probably also avoid allocating `leftSide`. 

We want to avoid "avoidable" allocations in methods that could be executed in a hot path, in this method will be executed for every decimal property when Scale is provided.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataPrimitiveSerializer.cs
```diff
@@ -155,19 +156,54 @@ internal static object ConvertPrimitiveValue(object value, IEdmPrimitiveTypeRefe
...
             }
+            value = decimal.Parse($"{leftSide}.{rightSide}", CultureInfo.InvariantCulture);
```
habbes, line 192: Why is this necessary? Why is calling `decimal.Round(...)` not sufficient? Is this to ensure that we always have `scale` number of digits after the decimal point by padding `0`'s at the end if the original value has fewer digits?

If that's the case, I feel like that's unnecessary work. According to the [spec](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Scale), Scale specifies the **maximum** number of digits allowed after the decimal point, **not** the **exact** number of digits. So a value like `10.1` is valid for Scale = 2, we don't need to coerce it to `10.10` in order to be valid.

If the service requires that values have an exact number of decimal places, maybe the service should ensure the values adhere to that specification before being processed by OData. By forcing an exact number of decimal places, I think we would be deviating from the expectations of the library. I think we should do `decimal.Round(value, scale, ...)` and leave the rest up to the service or client. My 2 cents.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DecimalTest.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        private static IList<Order> _orders;
+
+        static DecimalModelContext()
+        {
+            _orders = new List<Order>()
+            {
+                new Order(){ OrderId=1,TotalPrice=10 },
+                new Order(){ OrderId=2, TotalPrice=10.000m },
+                new Order(){ OrderId=3, TotalPrice=10.00m }
```
habbes, line 86: Could you also add tests with negative decimal values, also values at the midpoint (e.g. 10.335 => 10.34, -10.335 => -10.33)

habbes (changes requested):

komdil:
> 

 I fixed the comments. Please take a look at my comment. It seems scale is 0 by default, which looks wrong

habbes:
@komdil thanks for the updates. Let me look into it.

habbes:
@komdil you are right. Not sure why the default scale is set to 0. It makes more sense to default to `null`. We can make that change in EdmLib

habbes:
But there seems to be other test cases that failed in the tests that you added:

image

habbes:
@komdil  I've created PR to set the default scale to `null`: https://github.com/OData/odata.net/pull/2346, however I just realized [the spec](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Scale) says that the scale facet should default to 0 if not set.

One workaround that I can think of is to ignore the scale if Precision is null and Scale is 0 or null.  But then what if someone wants the precision to be 0? Let me consult on how that should be handled.

corranrogue9:
Is this working now the @habbes fix has been merged into ODL?

KenitoInc:
> Is this working now the @habbes fix has been merged into ODL?

We need to update the ODL version in the WebAPI repo then rebase this PR. That should unblock this PR

komdil:
@KenitoInc What are the next steps here? I updated Microsoft.OData.Core and Microsoft.OData.Edm packages from 7.10.0 to 7.11.0 and run my tests. They are still failing because of default scale is 0. It seems that changes are not in this version (7.11.0) of NuGet packages. When they will be updated?

komdil:
> > @KenitoInc What are the next steps here? I updated Microsoft.OData.Core and Microsoft.OData.Edm packages from 7.10.0 to 7.11.0 and run my tests. They are still failing because of default scale is 0. It seems that changes are not in this version (7.11.0) of NuGet packages. When they will be updated?
> 
> Kindly rebase your branch and see if the tests will pass

No, they are still failing. version of Microsoft.OData.Core and Microsoft.OData.Edm is still 7.11.0.  [This ](https://github.com/OData/odata.net/pull/2346) change is not on that version

KenitoInc:
@komdil 
This is the changelog with the ODL fix https://docs.microsoft.com/en-us/odata/changelog/odatalib-7x#odatalib-7110-release

FYI, there are several areas in WebAPI you need to change the version to 7.11.0 for the build to pass(see PR https://github.com/OData/WebApi/pull/2643)

That's why I requested you to rebase your branch

KenitoInc:
This is a test that is passing in #2643 to confirm that the scale fix was released in `v7.11.0`
scale-webapi-test

komdil:
> @komdil This is the changelog with the ODL fix https://docs.microsoft.com/en-us/odata/changelog/odatalib-7x#odatalib-7110-release
> 
> FYI, there are several areas in WebAPI you need to change the version to 7.11.0 for the build to pass(see PR #2643)
> 
> That's why I requested you to rebase your branch

I did, but the test is still failing because the default scale is 0. See my configuration below. I just set the scale for my `ActualDecimalValue ` property : 
image

But for some reason, it is setting scale of another decimal property (`ExpectedDecimalValue`) to 0:

image

komdil:
> This is a test that is passing in #2643 to confirm that the scale fix was released in `v7.11.0` scale-webapi-test

Done!

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataPrimitiveSerializer.cs
```diff
@@ -158,16 +159,25 @@ internal static object ConvertPrimitiveValue(object value, IEdmPrimitiveTypeRefe
...
-            }
 
-            if (primitiveType != null && primitiveType.IsTimeOfDay() && TypeHelper.IsTimeSpan(type))
+            if (primitiveType != null)
             {
-                TimeOfDay tod = (TimeSpan)value;
-                return tod;
+                if (primitiveType.IsDate() && TypeHelper.IsDateTime(type))
+                {
+                    Date dt = (DateTime)value;
```
KenitoInc, line 167:
```suggestion
                    Date dt = (DateTime)value;
                    
```

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataPrimitiveSerializer.cs
```diff
@@ -158,16 +159,25 @@ internal static object ConvertPrimitiveValue(object value, IEdmPrimitiveTypeRefe
...
+                    return dt;
+                }
+
+                if (primitiveType.IsTimeOfDay() && TypeHelper.IsTimeSpan(type))
+                {
+                    TimeOfDay tod = (TimeSpan)value;
```
KenitoInc, line 173:
```suggestion
                    TimeOfDay tod = (TimeSpan)value;
                    
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DecimalTest.cs
```diff
@@ -0,0 +1,98 @@
+﻿using Microsoft.AspNet.OData.Builder;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.AspNet.OData.Test.Builder.TestModels;
+using Microsoft.OData.Edm;
+using Newtonsoft.Json;
+using System.Collections.Generic;
```
KenitoInc, line 7: `System.*`should appear first

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DecimalTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        private static HttpClient GetClient()
+        {
+            var controllers = new[] { typeof(MetadataController), typeof(DecimalRoundTestEntityController) };
+            var server = TestServerFactory.Create(controllers, (config) =>
+            {
+                config.Count().OrderBy().Filter().Expand().MaxTop(null).Select();
+                config.MapODataServiceRoute("odata", "odata", GetEdmModel());
+
+            });
```
KenitoInc, line 46:
```suggestion
            });
            
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DecimalTest.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            ODataConventionModelBuilder builder = ODataConventionModelBuilderFactory.Create();
+            builder.EntitySet<DecimalRoundTestEntity>("DecimalRoundTestEntity");
+            builder.EntityType<DecimalRoundTestEntity>().Property(s => s.ActualDecimalValue).Scale = 2;
+            return builder.GetEdmModel();
+        }
+
+        async Task<IEnumerable<T>> ReadEntitiesFromResponse<T>(HttpResponseMessage httpResponseMessage) where T : class
+        {
+            string jsonContent = await httpResponseMessage.Content.ReadAsStringAsync();
```
KenitoInc, line 60:
```suggestion
            string jsonContent = await httpResponseMessage.Content.ReadAsStringAsync();
            
```

KenitoInc (commented):
Thanks for this PR @komdil 
Just a few style changes then I will approve it.

Note: We usually leave an empty line before `return`

KenitoInc (approved):
LGTM

habbes (approved):

KenitoInc:
Thanks for the PR @komdil

Issue #2259, opened 2020-08-21
Deep OrderBy is not working


komdil:
I am trying to use $orderBy inside of $expand. When I add two properties in $orderBy query, the second is not working

### Assemblies affected

Microsoft.AspNetCore.OData

### Reproduce steps

Model:
```
 public class Student
    {
        public Guid Id { get; set; }
        public List<Backpack> Backpacks { get; set; }
    }

    public class Backpack
    {
        public Guid Id { get; set; } = Guid.NewGuid();
        public Address Address { get; set; }
        public string Name { get; set; }
    }

    public class Address
    {
        public Guid Id { get; set; } = Guid.NewGuid();
        public int PlaceNumber { get; set; }
    }
```
Controller:

```
  [EnableQuery(MaxExpansionDepth = 0)]
    public class BaseController<TEntity> : ODataController where TEntity : class
    {
        [HttpGet]
        public IActionResult Get(ODataQueryOptions<TEntity> queryOptions, CancellationToken cancellationToken)
        {
            var list = new List<Student>
            {
                CreateNewStudent(),
            };
            return Ok(list);
        }

        static Student CreateNewStudent()
        {
            return new Student
            {
                Id = Guid.NewGuid(),
                Backpacks = new List<Backpack>
                {
                    new Backpack{Name="Backpack",Address=new Address(){ PlaceNumber=342} },
                    new Backpack{Name="Backpack",Address=new Address(){ PlaceNumber=123} },
                    new Backpack{Name="Backpack",Address=new Address(){ PlaceNumber=987} },
                }
            };
        }
    }
```
My query is` https://localhost:44383/odata/Student?$expand=Backpacks($expand=Address;$orderBy=Name,Address/PlaceNumber)`
### Expected result

I should get a student with sorted Backpacks like:

```
  "Backpacks": [
                {
                    "Id": "2f4e43b4-87ee-4d21-a328-b23d5369f5e4",
                    "Name": "Backpack",
                    "Address": {
                        "Id": "b11cb104-4a16-4686-9067-570e2878d159",
                        "PlaceNumber": 123
                    }
                },
                {
                    "Id": "2343d709-47a3-49b2-9124-ea042715416e",
                    "Name": "Backpack",
                    "Address": {
                        "Id": "24b78278-618f-42dd-ac1c-4ef83d1238dd",
                        "PlaceNumber": 342
                    }
                },
                {
                    "Id": "ce8aae50-ca91-4a10-869d-0673884dffee",
[...]

PR #2426, opened 2021-03-02
Multiple level of OrderBy

komdil:
### Issues

This pull request fixes issue #2259.

### Description

Multilevel ordering is included

- [x] *Test cases added*


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OrderByTest.cs
```diff
@@ -0,0 +1,139 @@
...
+using Microsoft.AspNet.OData.Test.Abstraction;
+using Microsoft.OData.Edm;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Xunit;
+using System.Linq;
```
xuzhg, line 12: copyright and sorting the using

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OrderByTest.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Microsoft.AspNet.OData.Test
+{
+    public class OrderByTest
```
xuzhg, line 16: >OrderByTest [](start = 17, length = 11)

we don't have the orderby unit test file?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OrderByTest.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        [Fact]
+        public async Task OrderBy_MultipleLevelOrderByTest()
```
xuzhg, line 19: >OrderBy_MultipleLevelOrderByTest [](start = 26, length = 32)

Such test case is better to put in the e2e

xuzhg:
the product change looks good to me. it's better to make the test code clean

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OrderByTest.cs
komdil: This file has been removed

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OrderByTest.cs
komdil: We have, I added unit test to SelectExpandBinderTest

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/OrderByTest.cs
komdil: This file was removed

Issue #2435, opened 2021-03-12
Excessive memory is used due to empty lists -ODataConventionModeBuilder is using 9.54 mb 


Sreejithpin:
*Short summary (3-5 sentences) describing the issue.*

ODataConventionModeBuilder is using 9.54 mb of memory, mostly due to Dictionary<Type, List<Type>>.

Here is the code generating it:

        private static Dictionary<Type, List<Type>> BuildDerivedTypesMapping(IAssembliesResolver assemblyResolver)
        {
            IEnumerable<Type> enumerable =
                from t in TypeHelper.GetLoadedTypes(assemblyResolver)
                where t.IsVisible && t.IsClass && t != typeof(object)
                select t;
            Dictionary<Type, List<Type>> dictionary = enumerable.ToDictionary((Type k) => k, (Type k) => new List<Type>());

            foreach (Type current in enumerable)
            {
                List<Type> list;
                if (current.BaseType != null && dictionary.TryGetValue(current.BaseType, out list))
                {
                    list.Add(current);
                }
            }
            return dictionary;
        }

Do you see the problem now? The dictionary is allocated to contain all visible reference types which is not system.Object, each with an empty List<Type> allocated. The list is used to keep derived types. There are only about 10% types with derived types, that explains why we have such large dictionaries and so many empty lists.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData WebApi lib 6.1.0

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*

### Expected result

*What would happen if there wasn't a bug.*

### Actual result

*What is actually happening.*

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

PR #2436, opened 2021-03-13
Excessive memory in ODataConventionModelBuilder BuildDerivedTypesMapping

xuzhg:
### Issues

*This pull request fixes issue #2435.*

### Description

Here is the code generating it:

```C#
        private static Dictionary<Type, List<Type>> BuildDerivedTypesMapping(IAssembliesResolver assemblyResolver)
        {
            IEnumerable<Type> enumerable =
                from t in TypeHelper.GetLoadedTypes(assemblyResolver)
                where t.IsVisible && t.IsClass && t != typeof(object)
                select t;
            Dictionary<Type, List<Type>> dictionary = enumerable.ToDictionary((Type k) => k, (Type k) => new List<Type>());

            foreach (Type current in enumerable)
            {
                List<Type> list;
                if (current.BaseType != null && dictionary.TryGetValue(current.BaseType, out list))
                {
                    list.Add(current);
                }
            }
            return dictionary;
        }
```
The dictionary is allocated to contain all visible reference types which is not system.Object, each with an empty List<Type> allocated. The list is used to keep derived types. 

This change is to make the dictionary as small as possible.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
@@ -86,7 +86,7 @@ namespace Microsoft.AspNet.OData.Builder
         private bool _isQueryCompositionMode;
 
         // build the mapping between type and its derived types to be used later.
-        private Lazy<IDictionary<Type, List<Type>>> _allTypesWithDerivedTypeMapping;
+        private Lazy<IDictionary<Type, Type[]>> _allTypesWithDerivedTypeMapping;
 
         /// <summary>
         /// Initializes a new <see cref="ODataConventionModelBuilder"/>.
@@ -133,7 +133,7 @@ namespace Microsoft.AspNet.OData.Builder
             _mappedTypes = new HashSet<StructuralTypeConfiguration>();
             _ignoredTypes = new HashSet<Type>();
             ModelAliasingEnabled = true;
-            _allTypesWithDerivedTypeMapping = new Lazy<IDictionary<Type, List<Type>>>(
+            _allTypesWithDerivedTypeMapping = new Lazy<IDictionary<Type, Type[]>>(
                 () => BuildDerivedTypesMapping(assembliesResolver),
                 isThreadSafe: false);
         }
@@ -603,7 +603,7 @@ namespace Microsoft.AspNet.OData.Builder
                 StructuralTypeConfiguration baseType = typeToBeVisited.Dequeue();
                 visitedTypes.Add(baseType.ClrType);
 
-                List<Type> derivedTypes;
+                Type[] derivedTypes;
                 if (_allTypesWithDerivedTypeMapping.Value.TryGetValue(baseType.ClrType, out derivedTypes))
                 {
                     foreach (Type derivedType in derivedTypes)
@@ -825,7 +825,7 @@ namespace Microsoft.AspNet.OData.Builder
                 Type currentType = typeToBeVisited.Dequeue();
                 visitedTypes.Add(currentType);
 
-                List<Type> derivedTypes;
+                Type[] derivedTypes;
                 if (_allTypesWithDerivedTypeMapping.Value.TryGetValue(currentType, out derivedTypes))
                 {
                     foreach (Type derivedType in derivedTypes)
@@ -865,7 +865,7 @@ namespace Microsoft.AspNet.OData.Builder
                 Type currentType = typeToBeVisited.Dequeue();
                 visitedTypes.Add(currentType);
 
-                List<Type> derivedTypes;
+                Type[] derivedTypes;
                 if (_allTypesWithDerivedTypeMapping.Value.TryGetValue(currentType, out derivedTypes))
                 {
                     foreach (Type derivedType in derivedTypes)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -1080,21 +1080,48 @@ private void ApplyPropertyConvention(IEdmPropertyConvention propertyConvention,
...
-                List<Type> derivedTypes;
-                if (TypeHelper.GetBaseType(type) != null && allTypeMapping.TryGetValue(TypeHelper.GetBaseType(type), out derivedTypes))
+                if (type != null && verifyType(type))
                 {
-                    derivedTypes.Add(type);
+                    Type baseType = TypeHelper.GetBaseType(type);
+                    if (baseType != null)
+                    {
+                        List<Type> list;
+                        if (!temp.TryGetValue(baseType, out list))
```
gathogojr, line 1100: @xuzhg There might be a bug here...
Supposing `GetLoadedTypes` method returned the types in the following order:
1. `A : Object` - _Object is the base type of A_
2. `C : B` - _B is the based type of C_
3. `B : A` - _A is the base type of B_

When "processing" type `C`, the `TryGetValue` call here will evaluate to false and we'll end up not establishing the relationship between `B` and `C`.

Maybe this can be remedied by also checking for any types derived from `type` that might already be in the `temp` dictionary and adding them to `temp[type]` list

gathogojr: Unless `GetLoadedTypes()` always returns base types ahead of derived types...

gathogojr: Thanks @xuzhg for the explanation. Makes sense. That we're using `baseType` as the key and adding `type` to the `list` (value) is the statement that didn't register in my mind properly

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -416,11 +416,9 @@ internal static Type GetImplementedIEnumerableType(Type type)
         // This code is copied from DefaultHttpControllerTypeResolver.GetControllerTypes.
         internal static IEnumerable<Type> GetLoadedTypes(IWebApiAssembliesResolver assembliesResolver)
         {
-            List<Type> result = new List<Type>();
-
             if (assembliesResolver == null)
             {
-                return result;
+                yield return null;
```
mikepizzo, line 421: >yield return null; [](start = 16, length = 18)

Is this the same as returning an empty list, or does it return a list with a null entry?

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -449,11 +447,15 @@ internal static IEnumerable<Type> GetLoadedTypes(IWebApiAssembliesResolver assem
 
                 if (exportedTypes != null)
                 {
-                    result.AddRange(exportedTypes.Where(t => t != null && TypeHelper.IsVisible(t)));
+                    foreach (var t in exportedTypes)
```
mikepizzo, line 450: >var [](start = 29, length = 3)

prefer "Type" to "var"

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -1080,21 +1080,48 @@ private void ApplyPropertyConvention(IEdmPropertyConvention propertyConvention,
...
         }
 
-        private static Dictionary<Type, List<Type>> BuildDerivedTypesMapping(IWebApiAssembliesResolver assemblyResolver)
+        private static Dictionary<Type, Type[]> BuildDerivedTypesMapping(IWebApiAssembliesResolver assemblyResolver)
         {
-            IEnumerable<Type> allTypes = TypeHelper.GetLoadedTypes(assemblyResolver).Where(t => TypeHelper.IsVisible(t) && TypeHelper.IsClass(t) && t != typeof(object));
-            Dictionary<Type, List<Type>> allTypeMapping = allTypes.Distinct().ToDictionary(k => k, k => new List<Type>());
-
-            foreach (Type type in allTypes)
+            Predicate<Type> verifyType = t => TypeHelper.IsVisible(t) && TypeHelper.IsClass(t) && t != typeof(object);
```
mikepizzo, line 1085: >TypeHelper.IsVisible(t) [](start = 46, length = 23)

Note: this is already checked in GetLoadedTypes -- don't need to do it again here.  Can any of these other checks be pushed to GetLoadedTypes? i.e., don't we awlays want to exclued typeof(object)?

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -1080,21 +1080,48 @@ private void ApplyPropertyConvention(IEdmPropertyConvention propertyConvention,
...
                 }
             }
 
-            return allTypeMapping;
+            // We can throw away all lists and keep the dictionary as small as possible
+            Dictionary<Type, Type[]> map = new Dictionary<Type, Type[]>(temp.Count);
+            foreach (var kv in temp)
+            {
+                map[kv.Key] = kv.Value.ToArray();
+            }
```
mikepizzo, line 1122: rather than copying to arrays to reduce unused space, could you just use temp.TrimExcess()?

mikepizzo (changes requested):
:clock1:

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
xuzhg: That's the suggestion from Feng. "As such data structure is normally generated once and kept in memory forever, I would event change it to use array. Now we’re throwing away all lists and keep the dictionary minimum. Notice here ToArray is a real method on List, not LINQ extension method"


---
In reply to: [598868250](https://github.com/OData/WebApi/pull/2436#discussion_r598868250) [](ancestors = 598868250)

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
xuzhg: yield return null is returning a list with one item (null) in the list.

```C#
        static void Main(string[] args)
        {
            var types = GetLoadedTypes();
            Console.WriteLine(types.Count());
            Console.WriteLine(types.ElementAt(0) == null ? "null" : "not null");
        }

        internal static IEnumerable<Type> GetLoadedTypes()
        {
            yield return null;
        }
```

Returns:
```C#
1
null
``

---
In reply to: [598852813](https://github.com/OData/WebApi/pull/2436#discussion_r598852813) [](ancestors = 598852813)

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
xuzhg: Yes, IsVisible can remove. 

TypeHelper.GetLoadedTypes, as its name, is to loaded types, doesn't care about it's class or object. So, let's keep the method atomic.

---
In reply to: [598862955](https://github.com/OData/WebApi/pull/2436#discussion_r598862955) [](ancestors = 598862955)

xuzhg: "verifyType" action is also used to verify the base type. 
BaseType is loaded at Line 1096, that's not from "GetLoadedTypes".

So, all these checks are needed here.

---
In reply to: [598862955](https://github.com/OData/WebApi/pull/2436#discussion_r598862955) [](ancestors = 598862955)

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
mikepizzo: That was my assumption.  But that's not the semantics we want here, right? Don't we want to return an empty list for this case?

---
In reply to: [598898800](https://github.com/OData/WebApi/pull/2436#discussion_r598898800) [](ancestors = 598898800,598852813)

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
mikepizzo: Looks like TrimExcess() does a reallocate/copy, so ToArray() is likely a better alternative.

---
In reply to: [598894104](https://github.com/OData/WebApi/pull/2436#discussion_r598894104) [](ancestors = 598894104,598868250)

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
xuzhg: I changed the codes as discussed.

mikepizzo (approved):
:shipit:

PR #2438, opened 2021-03-17
Extend runtime Type methods of OperationConfiguration

bdebaere:
### Issues

*This pull request fixes issue https://github.com/OData/WebApi/issues/2439.

### Description

Extend runtime Type methods of OperationConfiguration. Currently only Parameter(Type,string) exists, this pull requests adds CollectionParameter(Type,string), CollectionEntityParameter(Type,string), EntityParameter(Type,string).


xuzhg (approved):
:shipit:

Issue #1828, opened 2019-06-10
ODataResourceDeserializer Doesn't Preserve Request on ODataDeserializerContext for Nested Resource


SuricateCan:
I am subclassing ODataResourceDeserializer and I noticed that when deserializing nested properties, the read context does not contain the Request information.
When deserializing a root level entity, the read context contains the information as expected.

### Assemblies affected

Microsoft.AspNetCore.OData, Version=7.0.1.20718, Culture=neutral, PublicKeyToken=31bf3856ad364e35

### Reproduce steps

Subclass ODataResourceDeserializer and Post an entity that contains references to another entity. Observe the Read methods to access the ODataDeserializerContext.

### Expected result

Th nested deserialization should occur with the same request information on ODataDeserializerContext.

### Actual result

No request information is present on nested contexts.

### Additional detail

I believe a change in these lines would solve the issue:

[ODataResourceDeserializer.cs:481](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs#L481)

[ODataResourceDeserializer.cs:589](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs#L589)

```csharp
            var nestedReadContext = new ODataDeserializerContext
            {
                Path = readContext.Path,
                Model = readContext.Model,
                //missing the following lines
                InternalRequest = readContext.InternalRequest,
                Request = readContext.Request,
            };
```

raheph:
@SuricateCan Thanks for pointing this out. While we will add this to our backlog to fix, it would be great if you can contribute this change to the library, we would be happy to review that.

SuricateCan:
@raheph I'd b happy to, but I'm having trouble to sign the CLA.

korygin:
Hello,

Any ETA on the fix? It's been almost a year. We experience the same problem and it's critical for us as we use nested resources quite extensively.

Thanks,

Alex

xuzhg:
Merged. Thanks.

PR #2445, opened 2021-03-22
ODataResourceDeserializer Preserves Request on ODataDeserializerContext for Nested Resource

pdbruno:
### Issues

*This pull request fixes issue #1828.*

### Description

*Trying to acces the readContext.Request property yields a null value when deserializing nested resources

The readContext.Request property is now being copied to the nestedReadContext*

- [X] *Test cases added*

### Additional work necessary

*No Additional work necessary.*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -492,6 +492,8 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             {
                 Path = readContext.Path,
                 Model = readContext.Model,
+                Request = readContext.Request,
+                RequestContext = readContext.RequestContext,
             };
 
             Type clrType = null;
@@ -600,6 +602,8 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             {
                 Path = readContext.Path,
                 Model = readContext.Model,
+                Request = readContext.Request,
+                RequestContext = readContext.RequestContext,
             };
 
             if (readContext.IsUntyped)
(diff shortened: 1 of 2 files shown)


xuzhg (approved):
:shipit:

xuzhg:
Make sure the test can pass.

---
In reply to: [804220690](https://github.com/OData/WebApi/pull/2445#issuecomment-804220690) [](ancestors = 804220690)

pdbruno:
> Make sure the test can pass.
> 
> In reply to: [804220690](https://github.com/OData/WebApi/pull/2445#issuecomment-804220690) [](ancestors = 804220690)

In order for the new test to pass, the project would need additional dependencies (System.Web.Http, System.Net.Http) because of the new property being testes. Should I add those dependecies myself? If so, how can I do it?

xuzhg:
UnitTest shared project is referenced by ".NET Classic" and ".NET Core" Unit test project.
you can open the corresponding solution, use the "nuget package manager" to install the new dependencies.

But, before including the new dependencies (most of time it's rejected to involve new dependencies), let's ask us "Can we avoid introducing new dependencies by changing the test cases"? 
Let's Keep the unit test as unique as possible.

---
In reply to: [804256211](https://github.com/OData/WebApi/pull/2445#issuecomment-804256211) [](ancestors = 804256211)

pdbruno:
> UnitTest shared project is referenced by ".NET Classic" and ".NET Core" Unit test project.
> you can open the corresponding solution, use the "nuget package manager" to install the new dependencies.
> 
> But, before including the new dependencies (most of time it's rejected to involve new dependencies), let's ask us "Can we avoid introducing new dependencies by changing the test cases"?
> Let's Keep the unit test as unique as possible.
> 
> In reply to: [804256211](https://github.com/OData/WebApi/pull/2445#issuecomment-804256211) [](ancestors = 804256211)

Hello! Is this ready to be merged? I fixed the test

PR #2448, opened 2021-03-25
Feature/add support for odatabind

ElizabethOkerio:
### Issues

### Description
If you have such a model: 
```
public class Item
{
    [Key]
    public int ID { get; set; }

    [Required]

    [Required]
    public virtual Group Group { get; set; }
}

public class Group
{
    [Key]
    public int ID { get; set; }

    [Required]
}
```

And you want to create a new Item and reference an existing Group in your database or in whichever data source that you have, then as per the OData spec http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_LinktoRelatedEntitiesWhenCreatinganE , you should be able to do so using any of the following below formats depending on the OData Version that you are using: 

For v4.0 clients 
```
{
  "@odata.type":"#Item",
  "ID": 1,
  "Name": "ItemA",
  "Group@odata.bind":"http://host/service/Groups(4)"
}
```
For v4.01 clients
```
{
  "@odata.type":"#Item",
  "ID": 1,
  "Name": "ItemA",
  "Group":{"@odata.id":"http://host/service/Groups(4)"}
}
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs
@@ -40,5 +40,6 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         /// Gets the nested items that are part of this nested resource info.
         /// </summary>
         public IList<ODataItemBase> NestedItems { get; private set; }
+
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -404,7 +404,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         }
 
         private ODataResourceWrapper CreateResourceWrapper(IEdmTypeReference edmPropertyType, ODataEntityReferenceLinkBase refLink, ODataDeserializerContext readContext)
-        {
+        { 
             Contract.Assert(readContext != null);
 
             ODataResource resource = new ODataResource
@@ -432,7 +432,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         /// <param name="readContext">The read context.</param>
         /// <returns>The resource wrapper.</returns>
         private ODataResourceWrapper UpdateResourceWrapper(ODataResourceWrapper resourceWrapper, ODataDeserializerContext readContext)
-        {
+        { 
             Contract.Assert(readContext != null);
 
             if (resourceWrapper?.Resource?.Id == null)
@@ -473,7 +473,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         /// <param name="id">The key Id.</param>
         /// <param name="readContext">The reader context.</param>
         /// <returns>The key properties.</returns>
-        private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext)
+        private IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext)
         {
             Contract.Assert(id != null);
             Contract.Assert(readContext != null);
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -426,7 +426,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         }
 
         /// <summary>
-        /// Update the resource wrapper if it's have the "Id" value.
+        /// Update the resource wrapper if it has the "Id" value.
         /// </summary>
         /// <param name="resourceWrapper">The resource wrapper.</param>
         /// <param name="readContext">The read context.</param>
@@ -483,36 +483,29 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                 return properties;
             }
 
-            try
-            {
-                IODataPathHandler pathHandler = readContext.InternalRequest.PathHandler;
-                IWebApiRequestMessage internalRequest = readContext.InternalRequest;
-                IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;    
-                string serviceRoot = urlHelper.CreateODataLink(
-                    internalRequest.Context.RouteName,
-                    internalRequest.PathHandler,
-                    new List<ODataPathSegment>());
+            IODataPathHandler pathHandler = readContext.InternalRequest.PathHandler;
+            IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+            IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
+            string serviceRoot = urlHelper.CreateODataLink(
+                internalRequest.Context.RouteName,
+                internalRequest.PathHandler,
+                new List<ODataPathSegment>());
 
-                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
-                IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
-                foreach (KeySegment keySegment in keySegments)
-                {            
-                    foreach (KeyValuePair<string, object> key in keySegment.Keys)
+            ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+            IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+            foreach (KeySegment keySegment in keySegments)
+            {
+                foreach (KeyValuePair<string, object> key in keySegment.Keys)
+                {
+                    properties.Add(new ODataProperty
                     {
-                        properties.Add(new ODataProperty
-                        {
-                            Name = key.Key,
-                            Value = key.Value
-                        });
-                    }
+                        Name = key.Key,
+                        Value = key.Value
+                    });
                 }
+            }
 
-                return properties;
-            }
-            catch (Exception)
-            {
-                throw;
-            }
+            return properties;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -384,8 +384,8 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         }
 
 
-        private ODataResourceSetWrapper CreateResourceSetWrapper(IEdmCollectionTypeReference edmPropertyType,
-    IList<ODataEntityReferenceLinkBase> refLinks, ODataDeserializerContext readContext)
+        private ODataResourceSetWrapper CreateResourceSetWrapper(IEdmCollectionTypeReference edmPropertyType, 
+            IList<ODataEntityReferenceLinkBase> refLinks, ODataDeserializerContext readContext)
         {
             ODataResourceSet resourceSet = new ODataResourceSet
             {
@@ -473,7 +473,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
         /// <param name="id">The key Id.</param>
         /// <param name="readContext">The reader context.</param>
         /// <returns>The key properties.</returns>
-        private IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext)
+        private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext)
         {
             Contract.Assert(id != null);
             Contract.Assert(readContext != null);
@@ -486,26 +486,32 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             IODataPathHandler pathHandler = readContext.InternalRequest.PathHandler;
             IWebApiRequestMessage internalRequest = readContext.InternalRequest;
             IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
-            string serviceRoot = urlHelper.CreateODataLink(
-                internalRequest.Context.RouteName,
-                internalRequest.PathHandler,
-                new List<ODataPathSegment>());
-
-            ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
-            KeySegment keySegment = odataPath.Segments.OfType<KeySegment>().LastOrDefault();
-            if (keySegment != null)
+            try
             {
-                foreach (KeyValuePair<string, object> key in keySegment.Keys)
+                string serviceRoot = urlHelper.CreateODataLink(
+                    internalRequest.Context.RouteName,
+                    internalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+                KeySegment keySegment = odataPath.Segments.OfType<KeySegment>().LastOrDefault();
+                if (keySegment != null)
                 {
-                    properties.Add(new ODataProperty
+                    foreach (KeyValuePair<string, object> key in keySegment.Keys)
                     {
-                        Name = key.Key,
-                        Value = key.Value
-                    });
+                        properties.Add(new ODataProperty
+                        {
+                            Name = key.Key,
+                            Value = key.Value
+                        });
+                    }
                 }
-            }
 
-            return properties;
+                return properties;
+            }
+            catch (Exception)
+            {
+                return properties;
+            }          
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -492,8 +492,8 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                 new List<ODataPathSegment>());
 
             ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
-            IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
-            foreach (KeySegment keySegment in keySegments)
+            KeySegment keySegment = odataPath.Segments.OfType<KeySegment>().LastOrDefault();
+            if (keySegment != null)
             {
                 foreach (KeyValuePair<string, object> key in keySegment.Keys)
                 {
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -11,9 +11,12 @@ using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Common;
-using Microsoft.AspNet.OData.Formatter.Serialization;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.AspNet.OData.Routing;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
+using Microsoft.OData.UriParser;
+using ODataPath = Microsoft.AspNet.OData.Routing.ODataPath;
 
 namespace Microsoft.AspNet.OData.Formatter.Deserialization
 {
@@ -96,6 +99,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
 
             // Recursion guard to avoid stack overflows
             RuntimeHelpers.EnsureSufficientExecutionStack();
+            resourceWrapper = UpdateResourceWrapper(resourceWrapper, readContext);
 
             return ReadResource(resourceWrapper, edmType.AsStructured(), readContext);
         }
@@ -297,7 +301,39 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                 }
             }
 
-            foreach (ODataItemBase childItem in resourceInfoWrapper.NestedItems)
+            IList<ODataItemBase> nestedItems;
+            ODataEntityReferenceLinkBase[] referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkBase>().ToArray();
+            if (referenceLinks.Length > 0)
+            {
+                // Be noted:
+                // 1) OData v4.0, it's "Orders@odata.bind", and we get "ODataEntityReferenceLinkWrapper"(s) for that.
+                // 2) OData v4.01, it's {"odata.id" ...}, and we get "ODataResource"(s) for that.
+                // So, in OData v4, if it's a single, NestedItems contains one ODataEntityReferenceLinkWrapper,
+                // if it's a collection, NestedItems contains multiple ODataEntityReferenceLinkWrapper(s)
+                // We can use the following code to adjust the `ODataEntityReferenceLinkWrapper` to `ODataResourceWrapper`.
+                // In OData v4.01, we will not be here.
+                // Only supports declared property
+                Contract.Assert(edmProperty != null);
+
+                nestedItems = new List<ODataItemBase>();
+                if (edmProperty.Type.IsCollection())
+                {
+                    IEdmCollectionTypeReference edmCollectionTypeReference = edmProperty.Type.AsCollection();
+                    ODataResourceSetWrapper resourceSetWrapper = CreateResourceSetWrapper(edmCollectionTypeReference, referenceLinks, readContext);
+                    nestedItems.Add(resourceSetWrapper);
+                }
+                else
+                {
+                    ODataResourceWrapper resourceWrapper = CreateResourceWrapper(edmProperty.Type, referenceLinks[0], readContext);
+                    nestedItems.Add(resourceWrapper);
+                }
+            }
+            else
+            {
+                nestedItems = resourceInfoWrapper.NestedItems;
+            }
+
+            foreach (ODataItemBase childItem in nestedItems)
             {
                 // it maybe null.
                 if (childItem == null)
@@ -312,13 +348,6 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                     {
                         ApplyResourceInNestedProperty(edmProperty, resource, null, readContext);
                     }
-                }
-
-                ODataEntityReferenceLinkBase entityReferenceLink = childItem as ODataEntityReferenceLinkBase;
-                if (entityReferenceLink != null)
-                {
-                    // ignore entity reference links.
-                    continue;
                 }
 
                 ODataResourceSetWrapper resourceSetWrapper = childItem as ODataResourceSetWrapper;
@@ -351,6 +380,138 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                         ApplyResourceInNestedProperty(edmProperty, resource, resourceWrapper, readContext);
                     }
                 }
+            }
+        }
+
+
+        private ODataResourceSetWrapper CreateResourceSetWrapper(IEdmCollectionTypeReference edmPropertyType,
+    IList<ODataEntityReferenceLinkBase> refLinks, ODataDeserializerContext readContext)
+        {
+            ODataResourceSet resourceSet = new ODataResourceSet
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            IEdmTypeReference elementType = edmPropertyType.ElementType();
+            ODataResourceSetWrapper resourceSetWrapper = new ODataResourceSetWrapper(resourceSet);
+            foreach (ODataEntityReferenceLinkBase refLinkWrapper in refLinks)
+            {
+                ODataResourceWrapper resourceWrapper = CreateResourceWrapper(elementType, refLinkWrapper, readContext);
+                resourceSetWrapper.Resources.Add(resourceWrapper);
+            }
+
+            return resourceSetWrapper;
+        }
+
+        private ODataResourceWrapper CreateResourceWrapper(IEdmTypeReference edmPropertyType, ODataEntityReferenceLinkBase refLink, ODataDeserializerContext readContext)
+        {
+            Contract.Assert(readContext != null);
+
+            ODataResource resource = new ODataResource
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            resource.Properties = CreateKeyProperties(refLink.EntityReferenceLink.Url, readContext);
+         
+            if (refLink.EntityReferenceLink.InstanceAnnotations != null)
+            {
+                foreach (ODataInstanceAnnotation instanceAnnotation in refLink.EntityReferenceLink.InstanceAnnotations)
+                {
+                    resource.InstanceAnnotations.Add(instanceAnnotation);
+                };
+            }
+
+            return new ODataResourceWrapper(resource);
+        }
+
+        /// <summary>
+        /// Update the resource wrapper if it's have the "Id" value.
+        /// </summary>
+        /// <param name="resourceWrapper">The resource wrapper.</param>
+        /// <param name="readContext">The read context.</param>
+        /// <returns>The resource wrapper.</returns>
+        private ODataResourceWrapper UpdateResourceWrapper(ODataResourceWrapper resourceWrapper, ODataDeserializerContext readContext)
+        {
+            Contract.Assert(readContext != null);
+
+            if (resourceWrapper?.Resource?.Id == null)
+            {
+                return resourceWrapper;
+            }
+
+            IEnumerable<ODataProperty> keys = CreateKeyProperties(resourceWrapper.Resource.Id, readContext);
+            if (keys == null)
+            {
+                return resourceWrapper;
+            }
+
+            if (resourceWrapper.Resource.Properties == null)
+            {
+                resourceWrapper.Resource.Properties = keys;
+            }
+            else
+            {
+                IDictionary<string, ODataProperty> newPropertiesDic = resourceWrapper.Resource.Properties.ToDictionary(p => p.Name, p => p);
+                foreach (ODataProperty key in keys)
+                {
+                    if (!newPropertiesDic.ContainsKey(key.Name))
+                    {
+                        newPropertiesDic[key.Name] = key;
+                    }
+                }
+
+                resourceWrapper.Resource.Properties = newPropertiesDic.Values;
+            }
+
+            return resourceWrapper;
+        }
+
+        /// <summary>
+        /// Do uri parsing to get the key values.
+        /// </summary>
+        /// <param name="id">The key Id.</param>
+        /// <param name="readContext">The reader context.</param>
+        /// <returns>The key properties.</returns>
+        private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext)
+        {
+            Contract.Assert(id != null);
+            Contract.Assert(readContext != null);
+            IList<ODataProperty> properties = new List<ODataProperty>();
+            if (readContext.Request == null)
+            {
+                return properties;
+            }
+
+            try
+            {
+                IODataPathHandler pathHandler = readContext.InternalRequest.PathHandler;
+                IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+                IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;    
+                string serviceRoot = urlHelper.CreateODataLink(
+                    internalRequest.Context.RouteName,
+                    internalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+                IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+                foreach (KeySegment keySegment in keySegments)
+                {            
+                    foreach (KeyValuePair<string, object> key in keySegment.Keys)
+                    {
+                        properties.Add(new ODataProperty
+                        {
+                            Name = key.Key,
+                            Value = key.Value
+                        });
+                    }
+                }
+
+                return properties;
+            }
+            catch (Exception)
+            {
+                throw;
             }
         }
 
@@ -492,7 +653,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             {
                 Path = readContext.Path,
                 Model = readContext.Model,
-                Request = readContext.Request,
+                Request = readContext.Request
             };
 
             Type clrType = null;
@@ -518,7 +679,6 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                 : clrType;
             return deserializer.ReadInline(resourceWrapper, edmType, nestedReadContext);
         }
-
         private void ApplyResourceSetInNestedProperty(IEdmProperty nestedProperty, object resource,
             ODataResourceSetWrapper resourceSetWrapper, ODataDeserializerContext readContext)
         {
@@ -601,7 +761,7 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
             {
                 Path = readContext.Path,
                 Model = readContext.Model,
-                Request = readContext.Request,
+                Request = readContext.Request
             };
 
             if (readContext.IsUntyped)
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Formatter/ODataEntityReferenceLinkE2ETests.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Formatter/ODataEntityReferenceLinkE2ETests.cs
@@ -1,0 +1,276 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Data.Entity;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Microsoft.AspNet.OData;
+using Microsoft.AspNet.OData.Extensions;
+using Microsoft.AspNet.OData.Routing;
+using Microsoft.OData.Edm;
+using Microsoft.Test.E2E.AspNet.OData.Common.Controllers;
+using Microsoft.Test.E2E.AspNet.OData.Common.Execution;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.Test.E2E.AspNet.OData.Formatter
+{
+    public class ODataEntityReferenceLinkE2ETests : WebHostTestBase
+    {
+        public ODataEntityReferenceLinkE2ETests(WebHostTestFixture fixture)
+            : base(fixture)
+        {
+        }
+        protected override void UpdateConfiguration(WebRouteConfiguration configuration)
+        {
+            var controllers = new[] { typeof(BooksController)};
+            configuration.AddControllers(controllers);
+            configuration.Count().Filter().OrderBy().Expand().MaxTop(null).Select();
+            configuration.MapODataServiceRoute("odata", "odata", BuildEdmModel(configuration));
+        }
+        private static IEdmModel BuildEdmModel(WebRouteConfiguration configuration)
+        {
+            var builder = configuration.CreateConventionModelBuilder();
+            builder.EntitySet<Book>("Books");
+            builder.EntitySet<Author>("Authors");
+            builder.Action("ResetDataSource");
+            builder.Action("RelateToExistingEntityAndUpdate").ReturnsFromEntitySet<Book>("Books").EntityParameter<Book>("book");
+
+            return builder.GetEdmModel();
+        }
+
+        [Fact]
+        public async Task CanCreate_ANewEntityAndRelateToAnExistingEntity_UsingODataBind()
+        {
+            await ResetDataSource();
+            // Arrange
+            const string Payload = "{" +
+            "\"Id\":\"1\"," +
+            "\"Name\":\"BookA\"," +
+            "\"Author@odata.bind\":\"Authors(1)\"}";
+
+            string Uri = BaseAddress + "/odata/Books";
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, Uri);
+
+            request.Content = new StringContent(Payload);
+            request.Content.Headers.ContentType = MediaTypeWithQualityHeaderValue.Parse("application/json");
+
+            // Act
+            HttpResponseMessage response = await Client.SendAsync(request);
+
+            // Assert
+            Assert.True(response.IsSuccessStatusCode);
+            Assert.Equal(HttpStatusCode.Created, response.StatusCode);
+            Assert.NotNull(response.Content);
+
+            //Get the above saved entity from the database
+            //and expand the navigation property to see if
+            //it was correctly created with the existing entity
+            //attached to it. 
+            string query = string.Format("{0}/odata/Books?$expand=Author", BaseAddress);
+            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Get, query);
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json"));
+            
+            // Act
+            HttpResponseMessage res = await Client.SendAsync(requestMessage);
+
+            // Assert
+            Assert.True(res.IsSuccessStatusCode);
+            var responseObject = JObject.Parse(await res.Content.ReadAsStringAsync());
+            var result = responseObject["value"] as JArray;
+            var expandProp = result[0]["Author"] as JObject;
+            Assert.Equal(1, expandProp["Id"]);
+
+        }
+
+        [Fact]
+        public async Task CanUpdate_TheRelatedEntitiesProperties()
+        {
+            await ResetDataSource();
+            // Arrange
+            const string Payload = "{" +
+               "\"book\":{" +
+                   "\"Id\":\"1\"," +
+                   "\"Name\":\"BookA\"," +
+                   "\"Author\":{" +
+                       "\"@odata.id\":\"Authors(1)\"," +
+                       "\"Name\":\"UpdatedAuthor\"}}}";
+
+            string Uri = BaseAddress + "/odata/RelateToExistingEntityAndUpdate";
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, Uri);
+
+            request.Content = new StringContent(Payload);
+            request.Content.Headers.ContentType = MediaTypeWithQualityHeaderValue.Parse("application/json");
+
+            // Act
+            HttpResponseMessage response = await Client.SendAsync(request);
+
+            // Assert
+            Assert.True(response.IsSuccessStatusCode);
+            Assert.Equal(HttpStatusCode.Created, response.StatusCode);
+            Assert.NotNull(response.Content);
+
+            //Get the above saved entity from the database
+            //and expand its navigation property to see if it was created with
+            //the existing entity correctly.
+            //Also note that we were able to update the name property 
+            //of the existing entity
+            string query = string.Format("{0}/odata/Books?$expand=Author", BaseAddress);
+            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Get, query);
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json"));
+
+            // Act
+            HttpResponseMessage res = await Client.SendAsync(requestMessage);
+
+            // Assert
+            Assert.True(res.IsSuccessStatusCode);
+            var responseObject = JObject.Parse(await res.Content.ReadAsStringAsync());
+            var result = responseObject["value"] as JArray;
+            var expandProp = result[0]["Author"] as JObject;
+            Assert.Equal(1, expandProp["Id"]);
+            Assert.Equal("UpdatedAuthor", expandProp["Name"]);
+        }
+
+        private async Task ResetDataSource()
+        {
+            string requestUri = BaseAddress + "/odata/ResetDataSource";
+            HttpClient client = new HttpClient();
+            HttpResponseMessage response = await client.GetAsync(requestUri);
+            response.EnsureSuccessStatusCode();
+        }
+    }
+
+    public class BooksController : TestODataController
+#if NETCORE
+        , IDisposable
+#endif
+    { 
+        private ODataEntityReferenceLinkContext db = new ODataEntityReferenceLinkContext();
+
+        [EnableQuery]
+        public ITestActionResult Get()
+        {        
+            return Ok(db.Books);
+        }
+        public ITestActionResult Post([FromBody] Book book)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest();
+            }
+
+            db.Authors.Attach(book.Author);
+            db.Books.Add(book);
+            db.SaveChanges();
+
+            return Created(book);
+        }
+
+        [HttpPost]
+        [ODataRoute("RelateToExistingEntityAndUpdate")]
+        public ITestActionResult RelateToExistingEntityAndUpdate(ODataActionParameters odataActionParameters)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest();
+            }
+
+            Book book = (Book)odataActionParameters["book"];
+            string authorName = book.Author.Name;
+            Author author = new Author()
+            {
+                Id = book.Author.Id
+            };
+            db.Authors.Attach(author);
+            book.Author = author;
+            book.Author.Name = authorName;
+            db.Books.Add(book);
+
+            db.SaveChanges();
+            return Created(book);
+        }
+
+        [HttpGet]
+        [ODataRoute("ResetDataSource")]
+        public ITestActionResult ResetDataSource()
+        {
+            db.Database.Delete();  // Start from scratch so that tests aren't broken by schema changes.
+            CreateDatabase();
+            return Ok();
+        }
+
+        private static void CreateDatabase()
+        {
+            using (ODataEntityReferenceLinkContext db = new ODataEntityReferenceLinkContext())
+            {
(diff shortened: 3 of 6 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkDeserializer.cs
```diff
@@ -3,16 +3,22 @@
...
 using Microsoft.AspNet.OData.Common;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.AspNet.OData.Routing;
 using Microsoft.OData;
+using Microsoft.OData.Edm;
 
 namespace Microsoft.AspNet.OData.Formatter.Deserialization
 {
     /// <summary>
     /// Represents an <see cref="ODataDeserializer"/> that can read OData entity reference link payloads.
```
habbes, line 19: Maybe update the doc comment to match the argument type

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -369,6 +469,30 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, resourceWrapper.Resource.Id.ToString(), request.RequestContainer);
+                KeySegment keys = odataPath.Segments.OfType<KeySegment>().FirstOrDefault();
+
+                if (keys == null)
+                {
+                    throw new InvalidOperationException("The link does not contain a key.");
+                }
+
+                var pk = keys.Keys.First().Key;
+                var pkv = keys.Keys.First().Value;
```
habbes, line 492: I think calling `keys.Keys.First()` twice is wasteful. May be change to:
```
var keyValPair = keys.Keys.First();
keyValPair.Key;
keyValPair.Value;
```

habbes: Also, why only use the first key? We don't support composite keys for this feature?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -314,10 +318,43 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
-                if (entityReferenceLink != null)
+                ODataEntityReferenceLinksWrapper referenceLinksWrapper = childItem as ODataEntityReferenceLinksWrapper;
+                if (referenceLinksWrapper != null)
                 {
-                    // ignore entity reference links.
+                    IEdmCollectionTypeReference edmCollectionTypeReference = edmProperty.Type.AsCollection();
+                    ODataResourceSetWrapper rSetWrapper = CreateResourceSetWrapper(edmCollectionTypeReference, referenceLinksWrapper.EntityReferenceLinks, readContext);
+                    if (rSetWrapper != null)
+                    {
+                        if (edmProperty == null)
```
marabooy, line 328: The inner if looks repeated and could be refactored.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +391,69 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            Uri serviceRootUri = null;
+            string serviceRoot = "";
+            DefaultODataPathHandler pathHandler = new DefaultODataPathHandler();
+            if (!refLink.EntityReferenceLink.Url.IsAbsoluteUri)
+            {
+                IWebApiRequestMessage request1 = readContext.InternalRequest;
```
marabooy, line 428:
```suggestion
                IWebApiRequestMessage internalRequest = readContext.InternalRequest;
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -314,10 +318,43 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
-                    // ignore entity reference links.
+                    IEdmCollectionTypeReference edmCollectionTypeReference = edmProperty.Type.AsCollection();
+                    ODataResourceSetWrapper rSetWrapper = CreateResourceSetWrapper(edmCollectionTypeReference, referenceLinksWrapper.EntityReferenceLinks, readContext);
+                    if (rSetWrapper != null)
+                    {
+                        if (edmProperty == null)
+                        {
+                            ApplyDynamicResourceSetInNestedProperty(resourceInfoWrapper.NestedResourceInfo.Name,
+                                resource, structuredType, rSetWrapper, readContext);
+                        }
```
xuzhg, line 332: dynamic reference link? I think we should exclude dynamic property reference link

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinksWrapper.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Text;
+using Microsoft.OData;
+
+namespace Microsoft.AspNet.OData.Formatter.Deserialization
+{
+
+    /// <summary>
+    /// Encapsulates an <see cref="ODataResourceSet"/> and the <see cref="ODataEntityReferenceLinks"/>'s that are part of it.
+    /// </summary>
+    public class ODataEntityReferenceLinksWrapper : ODataItemBase
```
xuzhg, line 12: >ODataEntityReferenceLinksWrapper [](start = 17, length = 32)

we don't need this class.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkWrapper.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Gets the wrapped <see cref="ODataEntityReferenceLink"/>.
+        /// </summary>
+        public ODataEntityReferenceLink EntityReferenceLink
+        {
+            get
+            {
+                return Item as ODataEntityReferenceLink;
+            }
+        }
```
xuzhg, line 32: change it as 

public ODataEntityReferenceLink EntityReferenceLink {get;}

in the ctor, set this value as:

EntityReferenceLink  = item;

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs
```diff
@@ -40,5 +42,22 @@ public ODataNestedResourceInfo NestedResourceInfo
         /// Gets the nested items that are part of this nested resource info.
         /// </summary>
         public IList<ODataItemBase> NestedItems { get; private set; }
+
+        /// <summary>
+        /// Gets/set the nested entity reference(s) that are part of this nested resource info.
+        /// A nested resource info for a singleton nested property can only contain one ODataEntityReferenceLink.
+        /// A nested resource info for a collection nested property can contain any number of ODataEntityReferenceLink
+        /// </summary>
+        public IList<ODataEntityReferenceLinkWrapper> NestedLinks => _nestedLinks;
```
xuzhg, line 51: >NestedLinks [](start = 54, length = 11)

we don't need this, use NestedItems

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataReaderExtensions.cs
```diff
@@ -158,12 +158,26 @@ private static void ReadCollectionItem(ODataReader reader, Stack<ODataItemBase>
...
+                        {
+                            ODataEntityReferenceLinksWrapper entityReferenceLinksWrapper = new ODataEntityReferenceLinksWrapper(null);
+                            foreach (ODataEntityReferenceLinkWrapper dataEntityReferenceLinkWrapper in parentNavigationLink.NestedLinks)
+                            {
+                                entityReferenceLinksWrapper.EntityReferenceLinks.Add(dataEntityReferenceLinkWrapper);
+                                parentNavigationLink.NestedItems.Remove(dataEntityReferenceLinkWrapper);
+                            }
+
+                            parentNavigationLink.NestedItems.Add(entityReferenceLinksWrapper);
+                        }
```
xuzhg, line 180: append to NesteItems

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +401,143 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                return null;
+            }
+
+            resource.Properties = keySegment.Keys.Select(k => new ODataProperty
+            {
+                Name = k.Key,
+                Value = k.Value
+            });
+
+            return new ODataResourceWrapper(resource);
```
xuzhg, line 476: remember to copy the instance annotation from entity reference link to odata resource if any

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +401,143 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            Contract.Assert(readContext != null);
+
+            ODataResource resource = new ODataResource
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            Uri serviceRootUri = null;
+            string serviceRoot = "";
+            DefaultODataPathHandler pathHandler = new DefaultODataPathHandler();
```
xuzhg, line 447: >pathHandler [](start = 36, length = 11)

in 7.x, we can retrieve the path handler from DI

xuzhg:
no test cases?

xuzhg (changes requested):
:clock1:

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs
ElizabethOkerio: resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkWrapper.cs
ElizabethOkerio: resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkDeserializer.cs
ElizabethOkerio: The doc is okay..The ODataEntityReferenceLinkDeserializer is a representation of the ODataDeserializer.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: fixed

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: @xuzhg is there any reason for that?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataReaderExtensions.cs
ElizabethOkerio: @xuzhg you mean  parentNavigationLink.NestedItems.Append(entityReferenceLinksWrapper);

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: solved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: solved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinksWrapper.cs
ElizabethOkerio: resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs
ElizabethOkerio: resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,158 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+        {
+            Contract.Assert(readContext != null);
+
+            ODataResource resource = new ODataResource
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            Uri serviceRootUri = null;
+            string serviceRoot = "";
```
KenitoInc, line 439:
```suggestion
            string serviceRoot = String.Empty;
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,158 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            DefaultODataPathHandler pathHandler = (DefaultODataPathHandler)readContext.InternalRequest.PathHandler;
+            if (!refLink.EntityReferenceLink.Url.IsAbsoluteUri)
```
KenitoInc: From the spec, the referenceLink can be a relativeUri. Is that supported in this feature?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,158 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+        {
+            Contract.Assert(readContext != null);
+
+            ODataResource resource = new ODataResource
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            Uri serviceRootUri = null;
+            string serviceRoot = "";
```
KenitoInc, line 491:
```suggestion
            string serviceRoot = String.Empty;
```

KenitoInc (changes requested):
A few change requests and comments!
Where are the tests?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: @KenitoInc  I can't see any of this..I think you have not reviewed the latest files.. Also there is a test file.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs
```diff
@@ -40,5 +40,6 @@ public ODataNestedResourceInfo NestedResourceInfo
         /// Gets the nested items that are part of this nested resource info.
         /// </summary>
         public IList<ODataItemBase> NestedItems { get; private set; }
+
```
gathogojr, line 43:
```suggestion
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,40 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
             }
 
-            foreach (ODataItemBase childItem in resourceInfoWrapper.NestedItems)
+            IList<ODataItemBase> nestedItems;
+            var referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkWrapper>().ToArray();
+            if (referenceLinks.Length > 0)
+            {
+                // Be noted:
+                // 1) OData v4.0, it's "Orders@odata.bind", and we get "ODataEntityReferenceLinkWrapper"(s) for that.
+                // 2) OData v4.01, it's {"odata.id" ...}, and we get "ODataResource"(s) for that.
```
gathogojr, line 310: @ElizabethOkerio The example provided in the [spec](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_LinktoRelatedEntitiesWhenCreatinganE) for v4.01 is:
```
{
  "DirectReports": [
    {"@id": "Employees(5)"},
    {"@id": "Employees(6)"}
  ]
}
```
Does that mean that we'll support both `@id` and `@odata.id`?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,40 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                    ODataResourceWrapper resourceWrapper = CreateResourceWrapper(edmProperty.Type, referenceLinks[0], readContext);
+                    nestedItems.Add(resourceWrapper);
+                }
+            }
+            else
+            {
+                nestedItems = resourceInfoWrapper.NestedItems;
+            }
+
+
```
gathogojr, line 336:
```suggestion
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkWrapper.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.Collections.Generic;
+using Microsoft.OData;
+
+namespace Microsoft.AspNet.OData.Formatter.Deserialization
+{
+    /// <summary>
+    /// Encapsulates an <see cref="ODataEntityReferenceLink"/> .
+    /// </summary>
+    public sealed class ODataEntityReferenceLinkWrapper : ODataItemBase
```
gathogojr, line 12: @ElizabethOkerio This class looks similar to `ODataEntityReferenceLinkBase`. Do we really need to introduce a new class?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -331,7 +361,23 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
                     }
                     else
                     {
-                        ApplyResourceSetInNestedProperty(edmProperty, resource, resourceSetWrapper, readContext);
+                        ODataResourceSetWrapper resSetWrapper = resourceSetWrapper;
```
gathogojr, line 364: @ElizabethOkerio Maybe I'm missing something... What is the reason for introducing the variable `resSetWrapper` instead of just using `resourceSetWrapper`?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -331,7 +361,23 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                        IList<ODataResourceWrapper> resourceWrappers  = resourceSetWrapper.Resources.ToList();
+                        IList<Uri> resourceIds = new List<Uri>();
+                        foreach (ODataResourceWrapper resWrapper in resourceWrappers)
```
gathogojr, line 367:
```suggestion
                        foreach (ODataResourceWrapper resWrapper in resourceSetWrapper.Resources)
```
I don't think there's any benefit in introducing the `resourceWrappers` variable two lines above this one

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            Contract.Assert(readContext != null);
+
+            ODataResource resource = new ODataResource
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            Uri serviceRootUri = null;
+            string serviceRoot = "";
+            DefaultODataPathHandler pathHandler = new DefaultODataPathHandler();
```
gathogojr, line 440: Follow up the Sam's earlier comment, shouldn't you be retrieving the `ODataPathHandler` from the DI container since you already have a way of referencing it using `readContext.InternalRequest.RequestContainer`?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+
+            IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+            IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
+            serviceRoot = readContext.InternalUrlHelper.CreateODataLink(
```
gathogojr, line 444:
```suggestion
            serviceRoot = urlHelper.CreateODataLink(
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                readContext.InternalRequest.Context.RouteName,
```
gathogojr, line 445:
```suggestion
                internalRequest.Context.RouteName,
```
Since you already have `internalRequest` variable...

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                readContext.InternalRequest.PathHandler,
```
gathogojr, line 446:
```suggestion
                internalRequest.PathHandler,
```
Since you already have `internalRequest` variable...

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                new List<ODataPathSegment>());
+
+            serviceRootUri = new Uri(serviceRoot, UriKind.Absolute);
+
+            var request = readContext.Request;
+            IEdmModel model = readContext.Model;
+
+            ODataPath odataPath = pathHandler.Parse(serviceRoot, refLink.EntityReferenceLink.Url.OriginalString,readContext.InternalRequest.RequestContainer);
```
gathogojr, line 454:
```suggestion
            ODataPath odataPath = pathHandler.Parse(serviceRoot, refLink.EntityReferenceLink.Url.OriginalString, internalRequest.RequestContainer);
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            DefaultODataPathHandler pathHandler = new DefaultODataPathHandler();
+
+            IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+            IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
+            serviceRoot = readContext.InternalUrlHelper.CreateODataLink(
+                readContext.InternalRequest.Context.RouteName,
+                readContext.InternalRequest.PathHandler,
+                new List<ODataPathSegment>());
+
+            serviceRootUri = new Uri(serviceRoot, UriKind.Absolute);
```
gathogojr, line 449: Where do we apply this `serviceRootUri` variable?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+
+            var request = readContext.Request;
+            IEdmModel model = readContext.Model;
+
+            ODataPath odataPath = pathHandler.Parse(serviceRoot, refLink.EntityReferenceLink.Url.OriginalString,readContext.InternalRequest.RequestContainer);
+            IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+            if (keySegments == null)
```
gathogojr, line 456: Firstly, I don't think `odataPath.Segments.OfType<KeySegment>().ToList()` will ever return a `null`. It'll be an empty list at worst.

Secondly, if we find no key segments, should we return null or throw an exception? To find ourselves here, I'd think that the reference to the related entity was most probably invalid - I could be wrong, plus I'm not sure we'd get this far if it was..., but if it was wrong, we can't successfully create a relationship between the entity targeted for creation and the related entity. According to the [spec](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_LinktoRelatedEntitiesWhenCreatinganE), _the service MUST never create an entity in a partially valid state (with the navigation property unset)_

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                foreach (ODataInstanceAnnotation instanceAnnotation in refLink.EntityReferenceLink.InstanceAnnotations)
+                {
+                    resource.InstanceAnnotations.Add(instanceAnnotation);
+                };
+            }
+            
+            return new ODataResourceWrapper(resource);
+        }
+
+        private ODataResourceWrapper UpdateResourceWrapper(IEdmTypeReference edmPropertyType, Uri resourceId, ODataDeserializerContext readContext)
```
gathogojr, line 481: A fair amount of logic in this method is a mirror of the logic in `CreateResourceWrapper` method. You should refactor the shared logic into a method that both methods can call

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            if (!resourceId.IsAbsoluteUri)
+            {
+                IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+                IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
+                serviceRoot = readContext.InternalUrlHelper.CreateODataLink(
+                    readContext.InternalRequest.Context.RouteName,
+                    readContext.InternalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+
+                serviceRootUri = new Uri(serviceRoot, UriKind.Absolute);
```
gathogojr, line 502: Again, not sure where this variable is applied

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+            IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
+            serviceRoot = readContext.InternalUrlHelper.CreateODataLink(
+                readContext.InternalRequest.Context.RouteName,
+                readContext.InternalRequest.PathHandler,
+                new List<ODataPathSegment>());
+
+            serviceRootUri = new Uri(serviceRoot, UriKind.Absolute);
+
+            var request = readContext.Request;
```
gathogojr, line 451: Where are the `request` and `model` properties applied

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;
+                serviceRoot = readContext.InternalUrlHelper.CreateODataLink(
+                    readContext.InternalRequest.Context.RouteName,
+                    readContext.InternalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+
+                serviceRootUri = new Uri(serviceRoot, UriKind.Absolute);
+            }
+
+            var request = readContext.Request;
```
gathogojr, line 505: Again, I'm not sure where are the `request` and `model` properties applied

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            IEdmModel model = readContext.Model;
+
+
+            ODataPath odataPath = pathHandler.Parse(serviceRoot, resourceId.ToString(), readContext.InternalRequest.RequestContainer);
+            IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+            if (keySegments == null)
```
gathogojr, line 511: See my comment [here](https://github.com/OData/WebApi/pull/2448/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.projitems#r612192552)

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -348,12 +394,156 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                {
+                    Name = k.Key,
+                    Value = k.Value
+                });
+            }
+
+            return new ODataResourceWrapper(resource);
+        }
+
+        private ODataResourceSetWrapper UpdateResourceSetWrapper(IEdmCollectionTypeReference edmPropertyType,
```
gathogojr, line 528: This method is almost a mirror of `CreateResourceSetWrapper`. The key difference being a single argument plus the loop within the body that calls either `CreateResourceWrapper` or `UpdateResourceWrapper` in the different methods. You could substitute the two methods with a single one that takes an `Action` argument in place of `IList<Uri>` and `IList<ODataEntityReferenceLinkWrapper> refLinks` that differentiates them. The logic in the loop would go into the action.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,40 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            IList<ODataItemBase> nestedItems;
+            var referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkWrapper>().ToArray();
+            if (referenceLinks.Length > 0)
+            {
+                // Be noted:
+                // 1) OData v4.0, it's "Orders@odata.bind", and we get "ODataEntityReferenceLinkWrapper"(s) for that.
+                // 2) OData v4.01, it's {"odata.id" ...}, and we get "ODataResource"(s) for that.
+                // So, in OData v4, if it's a single, NestedItems contains one ODataEntityReferenceLinkWrapper,
+                // if it's a collection, NestedItems contains multiple ODataEntityReferenceLinkWrapper(s)
+                // We can use the following codes to adjust the `ODataEntityReferenceLinkWrapper` to `ODataResourceWrapper`.
```
habbes, line 313:
```suggestion
                // We can use the following code to adjust the `ODataEntityReferenceLinkWrapper` to `ODataResourceWrapper`.
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,40 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                {
+                    IEdmCollectionTypeReference edmCollectionTypeReference = edmProperty.Type.AsCollection();
+                    ODataResourceSetWrapper resourceSetWrapper = CreateResourceSetWrapper(edmCollectionTypeReference, referenceLinks, readContext);
+                    nestedItems.Add(resourceSetWrapper);
+                }
+                else
+                {
+                    ODataResourceWrapper resourceWrapper = CreateResourceWrapper(edmProperty.Type, referenceLinks[0], readContext);
+                    nestedItems.Add(resourceWrapper);
+                }
```
habbes, line 329: Could you give an example of what each of these cases look like in a JSON payload?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,40 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            }
+            else
+            {
+                nestedItems = resourceInfoWrapper.NestedItems;
+            }
```
habbes, line 334: Does this means that if the `resourceInfoWrapper.NestedItems` contains reference links, then it cannot contain other types of items and vice-versa? You can have both reference links and other kind of contents in the nested items?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -314,13 +351,6 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
                     }
                 }
 
-                ODataEntityReferenceLinkBase entityReferenceLink = childItem as ODataEntityReferenceLinkBase;
-                if (entityReferenceLink != null)
-                {
-                    // ignore entity reference links.
-                    continue;
-                }
-
```
habbes, line 323: Curious to know why the logic of the reference links happens before the loop instead of at this point?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -331,7 +361,23 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                            if (resWrapper.Resource.Id != null)
+                            {
+                                resourceIds.Add(resWrapper.Resource.Id);
+                            }
+                        }
+
+                        if (resourceIds.Count() > 0)
+                        {
+                            IEdmCollectionTypeReference edmCollectionTypeReference = edmProperty.Type.AsCollection();
+                            resSetWrapper = UpdateResourceSetWrapper(edmCollectionTypeReference, resourceIds, readContext);
```
habbes, line 378: Could you explain what this statement achieves?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: yes

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: Currently @id is not supported in WEB API so we'll be using "@odata.id" which is supported.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkWrapper.cs
ElizabethOkerio: we wanted some form of consistency in the naming..Most of the wrappers have the wrapper word on them.. Can do away with the other class.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: solved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataEntityReferenceLinkWrapper.cs
gathogojr: That would be a breaking change since the class is `public`...

gathogojr: But you could leave the old class until a major release opportunity presents itself

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +384,146 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                    new List<ODataPathSegment>());
+
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+                IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+                if (keySegments != null)
+                {
+                    foreach (KeySegment keySegment in keySegments)
+                    {
+                        properties = new List<ODataProperty>();
+                        foreach (var key in keySegment.Keys)
```
gathogojr, line 508: Swap `var` with `KeyValuePair<?, ?>`

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +384,146 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            }
+
+            if (resourceWrapper.Resource.Properties == null)
+            {
+                resourceWrapper.Resource.Properties = keys;
+            }
+            else
+            {
+                IDictionary<string, ODataProperty> newPropertiesDic = resourceWrapper.Resource.Properties.ToDictionary(p => p.Name, p => p);
+                foreach (var key in keys)
```
gathogojr, line 459: Swap `var` with `KeyValuePair<?, ?>`

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,40 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
                 }
             }
 
-            foreach (ODataItemBase childItem in resourceInfoWrapper.NestedItems)
+            IList<ODataItemBase> nestedItems;
+            var referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkWrapper>().ToArray();
```
gathogojr, line 305: Swap `var` with `ODataEntityReferenceLinkWrapper[]`

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
gathogojr: @ElizabethOkerio You marked this as resolved yet I still see two blank lines...

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +384,146 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+            Contract.Assert(readContext != null);
+
+            if (readContext.Request == null)
+            {
+                return null;
+            }
+
+            try
+            {
+                DefaultODataPathHandler pathHandler = (DefaultODataPathHandler)readContext.InternalRequest.PathHandler;
```
gathogojr, line 491:
```suggestion
                IODataPathHandler pathHandler = readContext.InternalRequest.PathHandler;
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +384,146 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                IList<ODataProperty> properties = null;
+
+                string serviceRoot = urlHelper.CreateODataLink(
+                    internalRequest.Context.RouteName,
+                    internalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+                IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+                if (keySegments != null)
```
gathogojr, line 503: I don't think this check for null is necessary. I believe at worst it will be an empty list

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +384,146 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+        private ODataResourceWrapper CreateResourceWrapper(IEdmTypeReference edmPropertyType, ODataEntityReferenceLinkWrapper refLink, ODataDeserializerContext readContext)
+        {
+            Contract.Assert(readContext != null);
+
+            ODataResource resource = new ODataResource
+            {
+                TypeName = edmPropertyType.FullName(),
+            };
+
+            resource.Properties = CreateKeyProperties(refLink.EntityReferenceLink.Url, readContext);
```
gathogojr, line 416: `CreateKeyProperties` can return a null. Is the value of `Properties` `null` or empty collection by default? If its empty collection by default, would there be a problem with assign `null` as the value?

ElizabethOkerio: The value is an empty collection by default.. resolved.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,143 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+        /// Update the resource wrapper if it's have the "Id" value.
+        /// </summary>
+        /// <param name="resourceWrapper">The resource wrapper.</param>
+        /// <param name="readContext">The read context.</param>
+        /// <returns>The resource wrapper.</returns>
+        private ODataResourceWrapper UpdateResourceWrapper(ODataResourceWrapper resourceWrapper, ODataDeserializerContext readContext)
+        {
+            Contract.Assert(readContext != null);
+
+            if (resourceWrapper == null ||
```
gathogojr, line 438: A little bit unconventional but we could rewrite this expression this way:
```
if (resourceWrapper?.Resource?.Id == null)
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
gathogojr: If the default is an empty collection, then there's a danger in assigning a `null` value to the property since existing logic might not anticipate that. For your piece of mind you might want to verify that, or only assign the property if `CreateKeyProperties` doesn't return `null`

gathogojr: Never mind. I see you changed `CreateKeyProperties` so it doesn't return `null`

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,143 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                        {
+                            Name = key.Key,
+                            Value = key.Value
+                        });
+                    }
+                }
+
+                return properties;
+            }
+            catch (Exception)
```
gathogojr, line 517: I'm not particularly sure about the use of try/catch here. Is there a specific exception we anticipate to be thrown from the logic in the try block and if yes, why would it be okay to ignore it and move on? In addition, `Exception` is too general and broad. I'm not convinced we need the try/catch

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,143 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                    {
+                        properties.Add(new ODataProperty
+                        {
+                            Name = key.Key,
+                            Value = key.Value
+                        });
+                    }
+                }
+
+                return properties;
```
gathogojr, line 515: This method will return `null` if there are no `KeySegment`s. Maybe you should initialize `properties` to an empty list at declaration

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,143 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                string serviceRoot = urlHelper.CreateODataLink(
+                    internalRequest.Context.RouteName,
+                    internalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+                IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
+                foreach (KeySegment keySegment in keySegments)
+                {
+                    properties = new List<ODataProperty>();
```
gathogojr, line 504: Bug here? We re-initialize the properties list after each iteration? Can we have more than one `KeySegment`? If yes, then this logic is definitely **not** right. If no, then the following statements:
```
IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
foreach (KeySegment keySegment in keySegments)
```
should be replaced with
```
KeySegment keySegment = odataPath.Segments.OfType<KeySegment>().FirstOrDefault();
if (keySegment != null)
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
ElizabethOkerio: parsing the link using the pathhandler might fail..

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,138 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                return resourceWrapper;
+            }
+
+            if (resourceWrapper.Resource.Properties == null)
+            {
+                resourceWrapper.Resource.Properties = keys;
+            }
+            else
+            {
+                IDictionary<string, ODataProperty> newPropertiesDic = resourceWrapper.Resource.Properties.ToDictionary(p => p.Name, p => p);
```
ElizabethOkerio, line 455: @gathogojr  @habbes  Mike confirmed this. If the key Id exists then we should not update it with what was sent as a reference link.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
gathogojr: @ElizabethOkerio This part of my earlier comment: **why would it be okay to ignore the exception and move on?**

ElizabethOkerio: resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,138 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                {
+                    resource.InstanceAnnotations.Add(instanceAnnotation);
+                };
+            }
+
+            return new ODataResourceWrapper(resource);
+        }
+
+        /// <summary>
+        /// Update the resource wrapper if it's have the "Id" value.
```
habbes, line 429:
```suggestion
        /// Update the resource wrapper if it has the "Id" value.
```

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -492,7 +653,7 @@ private object ReadNestedResourceInline(ODataResourceWrapper resourceWrapper, IE
             {
                 Path = readContext.Path,
                 Model = readContext.Model,
-                Request = readContext.Request,
+                Request = readContext.Request
```
marabooy, line 656:
```suggestion
                Request = readContext.Request,
```
So that in future updates addition won't show 2 changes.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -601,7 +761,7 @@ private object ReadNestedResourceInline(ODataResourceWrapper resourceWrapper, IE
...
```
marabooy, line 764:
```suggestion
                Request = readContext.Request,
```
Add trailing comma

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,138 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                IODataPathHandler pathHandler = readContext.InternalRequest.PathHandler;
+                IWebApiRequestMessage internalRequest = readContext.InternalRequest;
+                IWebApiUrlHelper urlHelper = readContext.InternalUrlHelper;    
+                string serviceRoot = urlHelper.CreateODataLink(
+                    internalRequest.Context.RouteName,
+                    internalRequest.PathHandler,
+                    new List<ODataPathSegment>());
+
+                ODataPath odataPath = pathHandler.Parse(serviceRoot, id.OriginalString, internalRequest.RequestContainer);
+                IList<KeySegment> keySegments = odataPath.Segments.OfType<KeySegment>().ToList();
```
habbes, line 497: I think we only expect to have one key segment right? If we have more than one key segments, then that means the url has navigation properties and not all the keys belong to the same resource, e.g. `http://localhost/odata/Books(1)/Authors(2)`.

I assume we don't expect such a scenario to occur in `@odata.bind` or `@odata.id`. Or am I wrong?

And in such cases, it wouldn't make sense to add both the keys of `Book` and `Author` to the same resource. So maybe you should pick the first key segment instead of going through all of them. Or just pick the last key segment if it makes sense to have a url with multiple key segments.

gathogojr: @habbes Question, maybe unrelated to your point here, how does OData handle a composite key?

habbes: @gathogojr in the case of a composite key, the `KeySegment` will have multiple keys. `keySegment.Keys` is a collection of key-value pairs, mapping the key property names to their values.

gathogojr: @habbes Composite key question out of the way, you raise an important question here. If something like `/Books(1)/Authors(2)` is allowed, then we'd probably need to target the last key segment. We might need some guidance here @mikepizzo

odero:
> 

```suggestion
public ODataNestedResourceInfo NestedResourceInfo { get; }
```
??

---
Refers to: src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataNestedResourceInfoWrapper.cs:38 in 0907c49. [](commit_id = 0907c49c5a662920938a4426607babb239d71906, deletion_comment = False)

odero (changes requested):

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,39 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
                 }
             }
 
-            foreach (ODataItemBase childItem in resourceInfoWrapper.NestedItems)
+            IList<ODataItemBase> nestedItems;
+            ODataEntityReferenceLinkBase[] referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkBase>().ToArray();
+            if (referenceLinks.Length > 0)
+            {
```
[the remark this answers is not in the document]
ElizabethOkerio: so we get all nesteditems that are of type <code>ODataEntityReferenceLinkBase</code> and create an array out of them.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -297,7 +301,39 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
...
+                // if it's a collection, NestedItems contains multiple ODataEntityReferenceLinkWrapper(s)
+                // We can use the following code to adjust the `ODataEntityReferenceLinkWrapper` to `ODataResourceWrapper`.
+                // In OData v4.01, we will not be here.
+                // Only supports declared property
+                Contract.Assert(edmProperty != null);
+
+                nestedItems = new List<ODataItemBase>();
+                if (edmProperty.Type.IsCollection())
+                {
+                    IEdmCollectionTypeReference edmCollectionTypeReference = edmProperty.Type.AsCollection();
```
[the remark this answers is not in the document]
ElizabethOkerio: @odero <code> edmProperty.Type.IsCollection </code> this is what is being used in the codebase to check for whether an edmproperty is a collection. Is there any reason why you feel the change will be important?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -354,6 +383,131 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
             }
         }
 
+
+        private ODataResourceSetWrapper CreateResourceSetWrapper(IEdmCollectionTypeReference edmPropertyType,
+    IList<ODataEntityReferenceLinkBase> refLinks, ODataDeserializerContext readContext)
```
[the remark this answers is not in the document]
ElizabethOkerio: resolved

xuzhg (approved):

gathogojr (approved):
:shipit:

marabooy (approved):

xuzhg:
+11111111111111

---
In reply to: [633225192](https://github.com/OData/WebApi/pull/2448#issuecomment-633225192) [](ancestors = 633225192)

xuzhg:
Please add E2E test cases to covert the scenarios.

---
In reply to: [818824797](https://github.com/OData/WebApi/pull/2448#issuecomment-818824797) [](ancestors = 818824797)

xuzhg (changes requested):
:clock1:

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNet/Formatter/ODataEntityReferenceLinkE2ETests.cs
```diff
@@ -0,0 +1,276 @@
...
+        [HttpPost]
+        [ODataRoute("RelateToExistingEntityAndUpdate")]
+        public ITestActionResult RelateToExistingEntityAndUpdate(ODataActionParameters odataActionParameters)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest();
+            }
+
+            Book book = (Book)odataActionParameters["book"];
```
marabooy, line 183:
```suggestion
            Book book = odataActionParameters["book"] as Book; 
```
then check for nulls, 

Interested in how nulls would be treated in this pr. or if someone sent a wrong payload.

marabooy: I could have missed it during an earlier review.

PR #2450, opened 2021-03-25
Add the OData version to reader settings when specified in the request

### LINKED ISSUE #2449: ODataInputFormatterHelper does not respect the OData version set in headers by OData-Max-Version/OData-Version
The OData reader in the framework does not respect the OData-Max-Version/OData-Version header in the request when reading the request body. This should be propagated to the ODataMessageReaderSettings so the OData framework can deserialize using OData version 4.01. By default, it uses version 4.0 which will break any scenarios that use 4.01 annotations.

### Assemblies affected

All OData versions since the issue is in master.

### Reproduce steps

Pass \@type as part of the request as specified in the OData docs. This can be used when you have an abstract type and the type information is passed as part of the request.

### Expected result

OData version 4.01 annotations should be respected.

### Actual result

OData version 4.01 annotations are not respected and will fail internally in the framework.

### Additional detail

This is already done for the writer settings, it just needs to be propagated in the same way to the reader settings.

Writer settings (already implemented)
https://github.com/OData/WebApi/blob/1d6ddf999cb27a8c74fb06ec5a73fee64edaa46b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs#L155

Reader settings (not implemented)
https://github.com/OData/WebApi/blob/1d6ddf999cb27a8c74fb06ec5a73fee64edaa46b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs#L76

owenhuyn:
Fixes #2449

The OData reader in the framework does not respect the OData-Max-Version/OData-Version header in the request when reading the request body. This should be propagated to the ODataMessageReaderSettings so the OData framework can deserialize using OData version 4.01. By default, it uses version 4.0 which will break any scenarios that use 4.01 annotations.

### Assemblies affected

All OData versions since the issue is in master.

### Reproduce steps

Pass \@type as part of the request as specified in the OData docs. This can be used when you have an abstract type and the type information is passed as part of the request.

### Expected result

OData version 4.01 annotations should be respected.

### Actual result

OData version 4.01 annotations are not respected and will fail internally in the framework.

### Additional detail

This is already done for the writer settings, it just needs to be propagated in the same way to the reader settings.

Writer settings (already implemented)
https://github.com/OData/WebApi/blob/1d6ddf999cb27a8c74fb06ec5a73fee64edaa46b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs#L155

Reader settings (not implemented)
https://github.com/OData/WebApi/blob/1d6ddf999cb27a8c74fb06ec5a73fee64edaa46b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs#L76

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs
@@ -51,6 +51,7 @@ namespace Microsoft.AspNet.OData.Formatter
             Type type,
             object defaultValue,
             IEdmModel model,
+            ODataVersion version,
             Uri baseAddress,
             IWebApiRequestMessage internalRequest,
             Func<IODataRequestMessage> getODataRequestMessage,
@@ -74,6 +75,7 @@ namespace Microsoft.AspNet.OData.Formatter
                 ODataMessageReaderSettings oDataReaderSettings = internalRequest.ReaderSettings;
                 oDataReaderSettings.BaseUri = baseAddress;
                 oDataReaderSettings.Validations = oDataReaderSettings.Validations & ~ValidationKinds.ThrowOnUndeclaredPropertyForNonOpenType;
+                oDataReaderSettings.Version = version;
 
                 IODataRequestMessage oDataRequestMessage = getODataRequestMessage();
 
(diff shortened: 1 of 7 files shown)


xuzhg (approved):
:shipit:

PR #2456, opened 2021-03-31
Built-in support for deeply nested paths

habbes:
### Issues

*This pull request fixes issue #2433 *

### Description

The aim of this feature is to provide built-in support for handling arbitrary nested paths like `/Customers(1)/Orders(2)/Products`

We introduce a `[EnableNestedPaths]` attribute that can be applied to a `Get()` controller action that returns a base IQueryable, and the rest works automatically.

```c#
class CustomersController: ODataController {
   /.../
   [EnableNestedPaths]
   [EnableQuery]
   public IActionResult<IQueryable<Customer>> Get() {
     return Ok(_customers);
   }

   /.../

}
```

That single Get() method will handle all GET requests (but there are some limitations currently). The `[EnableNestedPaths]` attribute transforms the returned `IQueryable` to match the sequence of navigations in the path. It can work alongside `[EnableQuery]` where `[EnableQuery]` is applied after `[EnableNestedPaths]`.

You can use it on an entity set controller as well as a singleton controller, in the case of a singleton, you should return `SingleResult` object that wraps the single object in an `IQueryable`.

If you controller still has other conventional actions like `Get(int key)`, `GetProperty(int key)`, or actions using `ODataRoute`, those will take precedence over the action with `[EnableNestedPaths]`

Current limitations:
- Does not support paths with actions and functions
- Does not support `$ref` requests
- Does not support paths with type cast segments (this is potentially not too hard to add)
- Does not support non-OData controllers
- Only supports AspNetCore

**Implementation**:

The implementation of this feature revolves around the following classes:

- `ODataPathQueryBuilder` this takes an ODataPath and an IQueryable as inputs and applies a series of transformation to the IQuerable based on the path segments, and returns the resulting IQueryable
- `EnableNestedPathsAttribute` this is an action filter that when applied to a controller action, transforms the result of the action by applying the `ODataPathQueryBuilder` to it. If the result is supposed to be a single object, it unwraps the IQueryable using SingleOrDefault.
- `NestedPathsRoutingConvention`: a new routing convention that routes supported GET requests to the `Get()` or `Get<NavigationSourceName>` method that has a `[EnableNestedPaths]` attribute applied to it. This is added towards the end of the default list of routing conventions to make sure that it can be overriden by existing conventions
- `ODataActionSelector` changes made here allow the action select to select methods with `[EnableNestedPaths]` even if their arguments don't match the key values in the path (it's expected that such actions won't have arguments).

** Handling 404s

Currently, when a path segment resolves to null (e.g. Customers(1)/Address/City) where address is null, a 404 is returned. I'm not sure whether this is always desirable, maybe it might be desirable to return an empty content response instead of a 404? Or should this be configurable?

** Future work
- Adding support for TypeCast segments
- Support for non-OData controllers
- Porting feature to AspNet classic
- Provide ways for user to configure things like maximum depth, supported/accessible properties, handling 404s, etc.

- [x] *Test cases added*

### Additional work necessary

Documentation/blog post needed.


src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using Microsoft.AspNet.OData.Adapters;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Edm;
+using Microsoft.OData.UriParser;
+using System.Collections.Generic;
+using System.Linq;
```
xuzhg, line 12: sorting

gathogojr:
@habbes In the PR description, this is how you've illustrated how the `EnableNestedPaths` attribute can be applied to a controller action:
```csharp
[EnableNestedPaths]
[EnableQuery]
public IActionResult<IQueryable<Customer>> Get()
{
    return Ok(_customers);
}
```
Curious if it would work for these typical controller actions:
1.
```csharp
[EnableNestedPaths]
[EnableQuery]
public ActionResult Get()
{
    return Ok(_customers);
}
```
2.
```csharp
[EnableNestedPaths]
[EnableQuery]
public ActionResult<Customer> Get()
{
    return Ok(_customers);
}
```
3.
```csharp
[EnableNestedPaths]
[EnableQuery]
public IQueryable<Customer> Get()
{
    return _customers;
}
```

In addition, I'm curious if it behaves okay against in-memory collections (e.g. `List<T>`)

src/Microsoft.AspNet.OData/Adapters/WebApiActionMap.cs
```diff
@@ -41,5 +41,29 @@ public bool Contains(string name)
...
+        public IWebApiActionDescriptor GetActionDescriptor(string actionName)
+        {
+            var group = this.innerMap.FirstOrDefault(g => g.Key == actionName);
+
+            if (group == null)
+            {
+                return null;
+            }
+
+            HttpActionDescriptor descriptor = group.FirstOrDefault();
```
gathogojr, line 59: @habbes You can get rid of the `if` block above by changing this line to
```suggestion
            HttpActionDescriptor descriptor = group?.FirstOrDefault();
```

src/Microsoft.AspNetCore.OData/Adapters/WebApiActionDescriptor.cs
```diff
@@ -112,7 +112,7 @@ public string ActionName
         /// <returns>A list of attributes of type T.</returns>
         public IEnumerable<T> GetCustomAttributes<T>(bool inherit) where T : Attribute
         {
-            return this.innerDescriptor.ControllerTypeInfo.GetCustomAttributes<T>(inherit);
+            return this.innerDescriptor.MethodInfo.GetCustomAttributes<T>(inherit);
```
gathogojr, line 115: @habbes Can you explain this change?

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+            if (!descriptor.GetCustomAttributes<EnableNestedPathsAttribute>(/* inherit */ true).Any())
+            {
+                return null;
+            }
+
+            return descriptor.ActionName;
+        }
+    }
+}
+#endif
```
gathogojr, line 86: Add newline maybe?

src/Microsoft.AspNet.OData/Adapters/WebApiActionMap.cs
```diff
@@ -41,5 +41,29 @@ public bool Contains(string name)
...
         }
+
+        /// <summary>
+        /// Gets the action descriptor of the specified action
+        /// </summary>
+        /// <param name="actionName">The name of the action</param>
+        /// <returns>The <see cref="IWebApiActionDescriptor"/> if it exists, otherwise null</returns>
+        public IWebApiActionDescriptor GetActionDescriptor(string actionName)
+        {
+            var group = this.innerMap.FirstOrDefault(g => g.Key == actionName);
```
gathogojr, line 52: @habbes Confirm if a scenario where `g.Key` and `actionName` can differ by case can arise here

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -212,12 +213,27 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+            if (actionDescriptor is ControllerActionDescriptor controllerActionDescriptor)
+            {
+                if (controllerActionDescriptor.MethodInfo.GetCustomAttributes(true)
+                    .OfType<EnableNestedPathsAttribute>().Any())
+                {
+                    return true;
+                }
+            }
+
+
```
gathogojr, line 236:
```suggestion
```

src/Microsoft.AspNetCore.OData/Routing/Conventions/ODataRoutingConventions.cs
```diff
@@ -77,6 +77,7 @@ public static IList<IODataRoutingConvention> CreateDefault()
                 new RefRoutingConvention(),
                 new ActionRoutingConvention(),
                 new FunctionRoutingConvention(),
+                new NestedPathsRoutingConvention(),
```
gathogojr, line 80: @habbes With reference to your comment on the PR description that the `NestedPathsRoutingConvention` should appear later in the list, what would happen if its promoted to close to the top of the list? For instance, if the user opts out of configuring the routing in their service using the `CreateDefault`, what risk do they face if this routing convention appears before `EntitySetRoutingConvention` or `EntityRoutingConvention` for example? If mis-ordering can occasion any risk can we catch that as the service is bootstrapping and throw an exception?

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+                sourceName = singletonSegment.Singleton.Name;
+            }
+            else
+            {
+                // this only supports paths starting with an entity set or singleton
+                return null;
+            }
+
+
+            // if we did no find a matching action amongst the conventional user-defined methods
```
gathogojr, line 68:
```suggestion
            // if we did not find a matching action amongst the conventional user-defined methods
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+        public EnableNestedPathsAttribute(): base()
+        {
+            // ensures this filter is executed before [EnableQuery] which has default order 0
+            // since the OnActionExecuted method is executed after the action,
+            // a filter with a lower order will be executed after a filter with a higher order
+            Order = 1;
+        }
+        
+        /// <summary>
+        ///
```
gathogojr, line 41: Provide docstring

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+                throw Error.Argument("actionExecutedContext", SRResources.ActionExecutedContextMustHaveRequest);
+            }
+
+            ActionDescriptor actionDescriptor = actionExecutedContext.ActionDescriptor;
+            if (actionDescriptor == null)
+            {
+                throw Error.Argument("actionExecutedContext", SRResources.ActionContextMustHaveDescriptor);
+            }
+
+
```
gathogojr, line 63:
```suggestion
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+
```
gathogojr, line 64:
```suggestion
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <param name="actionExecutedContext"></param>
+        public override void OnActionExecuted(ActionExecutedContext actionExecutedContext)
+        {
+            if (actionExecutedContext == null)
+            {
+                throw Error.ArgumentNull("actionExecutedContext");
+            }
+
+            HttpRequest request = actionExecutedContext.HttpContext.Request;
+            if (request == null)
```
gathogojr, line 52: Is this an actual possibility? Can `request` ever be null?

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                throw Error.Argument("actionExecutedContext", SRResources.ActionExecutedContextMustHaveRequest);
+            }
+
+            ActionDescriptor actionDescriptor = actionExecutedContext.ActionDescriptor;
```
gathogojr, line 57: Maybe
```suggestion
            ControllerActionDescriptor actionDescriptor = actionExecutedContext.ActionDescriptor as ControllerActionDescriptor`;
```
Just to emphasize we're handling a controller action descriptor...

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+                // on the property and get the one with the most parameters.
+                PropertyInfo propInfo = responseContent.Value.GetType().GetProperties()
+                    .OrderBy(p => p.GetIndexParameters().Count())
+                    .Where(p => p.Name.Equals("Queryable"))
+                    .LastOrDefault();
+
+                result = propInfo.GetValue(singleResult) as IQueryable;
+            }
+
+
```
gathogojr, line 84:
```suggestion
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                throw Error.Argument("actionExecutedContext", SRResources.ActionContextMustHaveDescriptor);
+            }
+
+
+
+            //base.OnActionExecuted(context);
+            HttpResponse response = actionExecutedContext.HttpContext.Response;
+
+            ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
```
gathogojr, line 68: @habbes Remember that return types of `ActionResult`, `ActionResult<T>`, `Queryable<T>` and even `Task<T>` for asynchronous action methods are very common. Will your logic here work with those?
We seem to handle them a certain type of way in `EnableQueryAttribute`
https://github.com/OData/WebApi/blob/5a5c585c139cddc3c63488f03daeb97a4b1ba868/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs#L121

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+                    actionExecutedContext.Result = new NotFoundObjectResult(null);
+                }
+                else if (path.EdmType.TypeKind == EdmTypeKind.Collection || transformedResult.HasCountSegment)
+                {
+                    responseContent.Value = transformedResult.Result;
+                }
+                else
+                {
+                    Type elementType = transformedResult.Result.ElementType;
+                    object singleValue = EnableQueryAttribute.SingleOrDefault(transformedResult.Result, new WebApiActionDescriptor(actionDescriptor as ControllerActionDescriptor));
```
gathogojr, line 105: @habbes Not sure I understand what is happening here...

habbes: The result returned by `ODataPathQueryBuilder.BuildQuery()` is an IQueryable, if you return it as-is, then it will always return a collection response. So here I'm checking whether the result of the path is expected to be a single item or a collection, if it's a single item hen I call the `SingleOrDefault` helper method to extract the single item from the `IQueryable` and return it.

It's a helper method used by `EnableQueryAttribute`, but now I think I should extract it to a common helper class.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+        /// <returns>The result of the query transformtions or null if the path contained unsupported segments</returns>
+        public ODataPathQueryResult BuildQuery()
+        {
+            var result = new ODataPathQueryResult();
+
+            var segments = path.Segments;
+            // assume first segment is entitySet
+            var firstSegment = segments.FirstOrDefault() as EntitySetSegment;
+
+            //var edmType = firstSegment.EdmType;
```
marabooy, line 51: remove commented out code👀 
```suggestion
```

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+                    var filterParam = Expression.Parameter(currentType, "entity");
+                    var conditions = keySegment.Keys.Select(kvp =>
+                        Expression.Equal(
+                            Expression.Property(filterParam, kvp.Key),
+                            Expression.Constant(kvp.Value)));
+                    var filterBody = conditions.Aggregate((left, right) => Expression.AndAlso(left, right));
+                    var filterPredicate = Expression.Lambda(filterBody, filterParam);
+
+                    queryable = Where(queryable, filterPredicate, currentType);
+
```
marabooy, line 78:
```suggestion
```

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+
+#if NETCORE
+
+// TODO: [EnableNestedPaths] feature has not yet been ported to AspNet classic
+
+using System.Linq;
+using Microsoft.AspNet.OData.Interfaces;
+using Microsoft.OData.UriParser;
+
+
```
marabooy, line 12:
```suggestion
```

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+
+            ODataRequestMethod method = controllerContext.Request.GetRequestMethodOrPreflightMethod();
+
+            if (method != ODataRequestMethod.Get)
+            {
+                // [EnableNestedPaths] only supports GET requests
+                return null;
+            }
+
```
marabooy, line 42:
```suggestion
```

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                sourceName = singletonSegment.Singleton.Name;
+            }
+            else
+            {
+                // this only supports paths starting with an entity set or singleton
+                return null;
+            }
+
+
```
marabooy, line 67:
```suggestion
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: I think a potential concern here is potentially materializing the IQueryable prematurely, if the controller action has both `[EnableQuery]` and `[EnableNestedPaths]`. If the `EnableNestedPathsAttribute` calls `SingleOrDefault` and sets it as result, I believe this materializes the IQueryable (i.e. executes the db query), now when this single object is later processed by `[EnableQuery]`, it will be applying the query options like select, expand to the single object in memory, as opposed to the deferred IQueryable. In the case of $select, this means that we would have fetched all the fields from the database, then doing the field selection on the object in memory. In the case of $expand, I'm not sure it would have fetched the necessary related entity, so I have to verify whether this will actually expand the expected property or just set it to null.

In this case, I think I should check whether the controller action also has `[EnableQuery]`, if it has it, then instead of calling `SingleOrDefault`, I create a `SingleResult` wrapper for the returned `IQueryable` then `[EnableQuery]` will do the rest, if the action does not have `[EnableQuery]` then I just call `SingleOrDefault`

src/Microsoft.AspNetCore.OData/Adapters/WebApiActionDescriptor.cs
marabooy: @habbes could this be a set of maybe controller and the method attributes so as to retain the old behaviour?

habbes: This method is supposed to return the custom attributes of the underlying action descriptor. In the case of AspNetCore the underlying descriptor is a `ControllerActionDescriptor`. How it was returning the attributes of the controller type, rather than of the action method. I believe that was a bug, since the `WebApiActionDescriptor` is a wrapper around the action descriptor of an action method, not the controller. So I changed it to return the attributes of the method instead.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,454 @@
...
+                config.MapODataServiceRoute("odata", "odata", _model);
+                config.Count().OrderBy().Filter().Expand().MaxTop(null).Select();
+            });
+
+            _client = TestServerFactory.CreateClient(_server);
+
+            _db = new EnableNestedPathsDatabase();
+        }
+
+        [Fact]
```
marabooy, line 80: Add some doc strings for future generations to understand these tests.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/Conventions/NestedPathsRoutingConventionTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+
+        }
+
+        void MethodWithoutNestedPathsAttribute()
+        {
+
+        }
+
+
```
marabooy, line 159:
```suggestion
```

marabooy, line 159:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/Conventions/NestedPathsRoutingConventionTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        [EnableNestedPaths]
+        void MethodWithNestedPathsAttribute()
+        {
+
+        }
+
+        void MethodWithoutNestedPathsAttribute()
+        {
+
```
marabooy, line 156:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/Conventions/NestedPathsRoutingConventionTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        MethodInfo GetMethodWithNestedPathsAttribute()
+        {
+            return this.GetType().GetMethod("MethodWithNestedPathsAttribute", BindingFlags.Instance | BindingFlags.NonPublic);
+        }
+
+        MethodInfo GetMethodWithoutNestedPathsAttribute()
+        {
+            return this.GetType().GetMethod("MethodWithoutNestedPathsAttribute", BindingFlags.Instance | BindingFlags.NonPublic);
+        }
+
```
marabooy, line 146:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/Conventions/NestedPathsRoutingConventionTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        [EnableNestedPaths]
+        void MethodWithNestedPathsAttribute()
+        {
+
```
marabooy, line 151:
```suggestion
```

marabooy (commented):
Surface scan looks good.

src/Microsoft.AspNetCore.OData/Routing/Conventions/ODataRoutingConventions.cs
habbes: Consider the following controller:
```c#
class CustomersController {
   Get(int key) () { / ... }
  
  GetOrders(int key) () {}

  [EnableNestedPaths]
   Get() {}

}
```

In the current setup of routing conventions, this is how the following paths would be routed:

Path    |    Action  | Convention
-------|------------|-------------
/Customers | `Get()`  | `EntitySetConvention`
/Customers(1) | `Get(int key)` | `EntityConvention`
/Customers(1)/Orders | `GetOrders(int key)` | `NavigationPropertyConvention`
/Customers(1)/Name  | `Get()` | `NestedPathsConvention`
/Customers(1)/Orders(2)/Name | `Get()` | `NestedPathsConvention`

So the "standard" conventions take precedence. Meaning the `NestedPathsConvention` is not triggered if there's a user-defined method created to handle that specific path.

If the conventions are miss-ordered, such that the `NestedPathsConvention` appears above the "standard" conventions, then all applicable paths will be routed to the `Get()` method (if it has `[EnableNestedPaths]`) even if there are other user-defined methods that could handle such paths:

Path    |    Action  | Convention
-------|------------|-------------
/Customers | `Get()`  | `NestedPathsConvention`
/Customers(1) | `Get()` | `NestedPathsConvention`
/Customers(1)/Orders | `Get()` | `NestedPathsConvention`
/Customers(1)/Name  | `Get()` | `NestedPathsConvention`
/Customers(1)/Orders(2)/Name | `Get()` | `NestedPathsConvention`

So the danger is that the user would not be able to override the `Get()` method with `[EnableNestedPaths]`.

I think this should be documented, but I'm not sure if it should go as far as throwing an exception. The same concern would follow if the user miss-ordered the attribute routing convention, making it appear later in the list. I think the user should be aware of the consequences, but not necessarily get an exception.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+            // if we did no find a matching action amongst the conventional user-defined methods
+            // then let's check if the controller has a Get method with [EnableNestedPaths] attribute
+            // which should be used to catch any any nested GET request
+            IWebApiActionDescriptor descriptor = actionMap.GetActionDescriptor($"Get{sourceName}") ?? actionMap.GetActionDescriptor("Get");
+            if (descriptor == null)
+            {
+                return null;
+            }
+
+            if (!descriptor.GetCustomAttributes<EnableNestedPathsAttribute>(/* inherit */ true).Any())
```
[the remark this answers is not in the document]
habbes: I think `descriptor.GetCustomAttributes<EnableNestedPathsAttribute>()` is the preferred way, or at least that what's been used elsewhere in the code. I'm not sure about the performance difference (maybe one calls the other behind the scenes).

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: I think in all those cases, `responseContent.Value` would still be the actual value in your result, i.e. whether you return `Ok(customers)` or just `customers`, the `responseContent.Value` will still be the `customers` object. But let me confirm whether this always holds, including in the case of `Task<T>` which I haven't tested with.

src/Microsoft.AspNetCore.OData/Adapters/WebApiActionDescriptor.cs
habbes: @marabooy, my suspicion is that the old behaviour was a bug. @xuzhg  could you comment on this?

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
KenitoInc: He is using /// <inherit/>

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
KenitoInc: Without null check, we MAY break some tests in WebAPI

KenitoInc (changes requested):
A few comments/questions

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+        /// Transforms the source <see cref="IQueryable"/> based on the sequence of path segments in the <see cref="ODataPath"/>
+        /// </summary>
+        /// <returns>The result of the query transformtions or null if the path contained unsupported segments</returns>
+        public ODataPathQueryResult BuildQuery()
+        {
+            var result = new ODataPathQueryResult();
+
+            var segments = path.Segments;
+            // assume first segment is entitySet
+            var firstSegment = segments.FirstOrDefault() as EntitySetSegment;
```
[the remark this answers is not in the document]
habbes: Thanks for catching this, actually `firstSegment` variable is not used. The code works fine even if first segment is a singleton since query expressions are applied starting from the second. I will remove this statement.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            //var edmType = firstSegment.EdmType;
+
+            var queryable = source;
```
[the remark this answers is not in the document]
habbes: Do we also avoid the use of `var` when creating new instances and the actual type can be seen on the right hand side?
e.g.:
```c#
var result = new ODataPathQueryResult();
```

Or we should always be explicit even when repetitive?:
```c#
ODataPathQueryResult result = new ODataPathQueryResult();
```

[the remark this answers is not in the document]
gathogojr: @habbes That was the advise from Mike. Use of `var` is only permitted in test projects

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+        private static MethodInfo QueryableSelectGeneric { get; } = GenericMethodOf(_ => Queryable.Select(default(IQueryable<int>), i => i));
+
+        private static MethodInfo QueryableSelectManyGeneric { get; } = GenericMethodOf(_ => Queryable.SelectMany(default(IQueryable<int>), i => default(IQueryable<int>)));
+
+        public static MethodInfo QueryableOfTypeGeneric { get; } = GenericMethodOf(_ => Queryable.OfType<int>(default(IQueryable)));
+
+        private static IQueryable Where(IQueryable query, LambdaExpression where, Type type)
+        {
+            var whereMethod = QueryableWhereGeneric.MakeGenericMethod(type);
+            return whereMethod.Invoke(null, new object[] { query, where }) as IQueryable;
```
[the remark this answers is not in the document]
habbes: Here `whereMethod` is a `MethodInfo` that represents the [Queryable.Where<T>()](https://docs.microsoft.com/en-us/dotnet/api/system.linq.queryable.where) method. We want to call that method using reflection, and to do that we use the `Invoke()` method.

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+            // a filter with a lower order will be executed after a filter with a higher order
+            Order = 1;
+        }
+        
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="actionExecutedContext"></param>
+        public override void OnActionExecuted(ActionExecutedContext actionExecutedContext)
+        {
```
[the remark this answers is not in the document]
habbes: Good point. I will how much effort it takes to add support for that. If it's not a quick fix, then we can work on it on a follow-up PR.

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: EnableQuery has [similar checks](https://github.com/OData/WebApi/blob/5a5c585c139cddc3c63488f03daeb97a4b1ba868/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs#L213)

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: My suspicion was correct, using SingleOrDefault() to return a single result when the action also has `[EnableQuery]` results in query options like `$expand` returning null. If fixed this by using `SingleOrDefault()` if the action **does not** have `[EnableQuery]` and use `SingleResult<T>` if the action has `[EnableQuery]`

src/Microsoft.AspNet.OData/Adapters/WebApiActionMap.cs
habbes: Since actions are named after their controller methods, I don't think we should perform case-insensitive lookup.

habbes: But I guess this depends on the implementation of the `innerMap.Contains()` method. It implements `ILookup.Contains()`, but from that alone we can't tell whether it's case-sensitive or not. I can try to test it in the debugger to see if it's case-insensitive. But I'd be surprised if it is.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+            }
+
+            IEnumerable<ODataPathSegment> remainingSegments = segments.Skip(1);
+
+            foreach (ODataPathSegment segment in remainingSegments)
+            {
+                if (segment is KeySegment keySegment)
+                {
+                    Dictionary<string, object> keys = new Dictionary<string, object>();
+                    foreach (var kvp in keySegment.Keys)
```
gathogojr, line 65: Replace `var` with `KeyValuePair<?, ?>`?
Or do we excuse `var` in a `foreach`?

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+                    {
+                        keys.Add(kvp.Key, kvp.Value);
+                    }
+
+                    // filterPredicate = entity => (entity.KeyProp1 == Val1) && (entity.keyProp2 == Val2) && ...
+                    ParameterExpression filterParam = Expression.Parameter(queryable.ElementType, "entity");
+                    IEnumerable<BinaryExpression> conditions = keySegment.Keys.Select(kvp =>
```
gathogojr, line 72: Why do we use `keySegment.Keys` here instead of `keys` variable? Or where do we use `keys`?

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// <summary>
+        /// Gets the element type "T" of the given type
+        /// if it implements <see cref="IEnumerable{T}"/>
+        /// </summary>
+        /// <param name="enumerableType"></param>
+        /// <returns>
+        /// The element type if <paramref name="enumerableType"/> implements <see cref="IEnumerable{T}"/>
+        /// otherwise returns <paramref name="enumerableType"/> itself
+        /// </returns>
+        private static Type GetEnumerableItemType(Type enumerableType)
```
gathogojr, line 204: I'd check across the repo whether there's a method that already does this...

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// <param name="type">
+        /// The subject type.
+        /// </param>
+        /// <param name="definition">
+        /// The generic definition to check with.
+        /// </param>
+        /// <returns>
+        /// The base type or the interface found; otherwise, <c>null</c>.
+        /// </returns>
+        private static Type FindGenericType(Type type, Type definition)
```
gathogojr, line 228: A better home for this method would be a static TypeHelpers class - one might exist

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// Checks whether <paramref name="type"/> conforms to a generic
+        /// definition of the given generic type <paramref name="definition"/>
+        /// </summary>
+        /// <param name="type">The type to test</param>
+        /// <param name="definition">The generic type definition to test against</param>
+        /// <returns>
+        /// True if <paramref name="type"/> conforms to a generic
+        /// definition of the given generic type <paramref name="definition"/> otherwise false
+        /// </returns>
+        private static bool IsGenericDefinition(Type type, Type definition)
```
gathogojr, line 282: Same opinion as this https://github.com/OData/WebApi/pull/2456/files#r614589604

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// </param>
+        /// <param name="definition">
+        /// The generic definition to check with.
+        /// </param>
+        /// <returns>
+        /// The base type or the interface found; otherwise, <c>null</c>.
+        /// </returns>
+        private static Type FindGenericType(Type type, Type definition)
+        {
+            if (type == null)
```
gathogojr, line 230: A `Debug.Assert` wouldn't suffice here?

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,83 @@
...
+            else
+            {
+                // this only supports paths starting with an entity set or singleton
+                return null;
+            }
+
+            // if we did not find a matching action amongst the conventional user-defined methods
+            // then let's check if the controller has a Get method with [EnableNestedPaths] attribute
+            // which should be used to catch any nested GET request
+            IWebApiActionDescriptor descriptor = actionMap.GetActionDescriptor($"Get{sourceName}") ?? actionMap.GetActionDescriptor("Get");
```
gathogojr, line 68: Can you use the `FindMatchingAction` method like we do in [`EntitySetRoutingConvention`](https://github.com/OData/WebApi/blob/757631faf10b494e55117904196e8aa45d4c764d/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs#L53)?

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
+            IQueryable result = responseContent.Value as IQueryable;
+            SingleResult singleResult = responseContent.Value as SingleResult;
+            if (singleResult != null)
+            {
+                // This could be a SingleResult, which has the property Queryable.
+                // But it could be a SingleResult() or SingleResult<T>. Sort by number of parameters
+                // on the property and get the one with the most parameters.
+                PropertyInfo propInfo = responseContent.Value.GetType().GetProperties()
```
gathogojr, line 78:
```suggestion
                PropertyInfo propInfo = singleResult.GetType().GetProperties()
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,150 @@
...
+                    .OrderBy(p => p.GetIndexParameters().Count())
+                    .Where(p => p.Name.Equals("Queryable"))
+                    .LastOrDefault();
```
gathogojr, line 81: You could use `OrderByDescending` then `FirstOrDefault()`...
I'd assume that `LastOrDefault()` had to perform a count

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                ODataPath path = actionExecutedContext.HttpContext.ODataFeature().Path;
+                IEdmModel model = actionExecutedContext.HttpContext.Request.GetModel();
+
+                var queryBuilder = new ODataPathQueryBuilder(result, model, path);
+                ODataPathQueryResult transformedResult = queryBuilder.BuildQuery();
+
+                if (transformedResult == null)
+                {
+                    // TODO: is this the best way to return 404
```
gathogojr, line 96: TODO

gathogojr: Plus I'm not sure how `EnableQuery` (if it also exists on the action method) will handle the `Result` if the value is `NotFoundObjectResult`...

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,150 @@
...
+                            {
+                                // TODO: is this the best way to return 404
+                                actionExecutedContext.Result = new NotFoundObjectResult(null);
+                            }
+                            else
+                            {
+                                responseContent.Value = singleValue;
+                            }
+                        }
+                        catch (NullReferenceException)
```
gathogojr, line 141: Which cases would give us this outcome?

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: Thanks for the catch, will refactor that.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: Hmm, good point. I don't need both. I can stick to `keySegment.Keys` since I only iterate once over them.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: Alright, I'll try to find whether one exists.

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
```diff
@@ -0,0 +1,150 @@
...
+                    // If the action has [EnableQuery], then we return a SingleResult<T> which allows query options to be
+                    // applied to the IQueryable. If we returned the materialized object, then query options like $expand
+                    // would return null because navigation properties are not fetched from relational databases by default.
+                    // If the action does not have [EnableQuery], then we run the query using SingleOrDefault and return the
+                    // actual object. If we returned a SingleResult<T> in this case, we'd get an error because the serializer
+                    // does not recognize it.
+                    if (actionDescriptor.MethodInfo.GetCustomAttributes<EnableQueryAttribute>().Any())
+                    {
+                        // calls SingleResult.Create<T>(IQueryable<T>)
+                        MethodInfo singleResultCreate = typeof(SingleResult)
```
gathogojr, line 120: I think you could move the logic for getting `SingleResult` `Create` method to `ExpressionHelperMethods` so its reusable

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
habbes: In this case we're not looking for the action name, but the action descriptor. I believe `FindMatchingAction` only returns a string. And I didn't think it would be necessary to create another helper to get the action descriptor unless it's something that we do commonly.

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: EnableQuery checks whether the result is successful before processing. If we set the result to a 404, then the user will get a 404 response. What I wasn't sure is whether it's okay to just create the not found result without passing an exception or error message to it. I don't know if there's any appropriate exception to pass at this point other than to just indicate that the resource does not exist.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
gathogojr: Fair point, but consider if it'd be better to do this:
```csharp
string action = actionMap.FindMatchingAction("Get" + sourceName, "Get");
if (action == null)
{
    return null;
}

IWebApiActionDescriptor descriptor = actionMap.GetActionDescriptor(action);

// No need to check for if descriptor is null
```

gathogojr: Another point, arguably, `$"Get{sourceName}"` should be more expensive than `"Get" + sourceName` since the former is syntactical sugar for `string.Format`...

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
gathogojr: Consider the following current behaviors:
- `/Sales(13)` - You'll get a 404 if a resource with key 13 does not exist
- `/Sales?$filter=Id eq 13` - You'll get an empty array if a resource with Id equal to 13 does not exist
Since nested paths feature key segments, a 404 would appear like the logic result to return in this scenario.

However, consider this nested path scenario also
`/Sales(1)/Remark`
Supposing in whatever data source I'm using the value for `Remark` is `null`. Is your logic treating that as a value or responding with a `NotFound`?

gathogojr (approved):
Left a few comments but otherwise the feature worked as expected after I cloned your branch and tested the functionality

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+namespace Microsoft.AspNet.OData.Query
+{
+
+    /// <summary>
+    /// This class transforms an <see cref="IQueryable"/> based on
+    /// the key and property accesses in the segments of an <see cref="ODataPath"/>
+    /// </summary>
+    internal class ODataPathQueryBuilder
+    {
+        readonly IQueryable source;
```
marabooy, line 22: Are these used outside this class? if not kindly give them the right access modifier

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
marabooy: I believe for less than 3 segements String.Concat is faster or String.join with string.Empty

src/Microsoft.AspNet.OData/Adapters/WebApiActionMap.cs
```diff
@@ -41,5 +41,17 @@ public bool Contains(string name)
...
```
marabooy, line 52: Use concrete types so as to catch errors in future public api changes.

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
marabooy: What would happen if a collection with mutliple nulls is encountered? would it filter out the nulls? return them? or simply end up having a 404 bubble up?

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
marabooy: Would this be ever triggered by collection having nulls?

habbes: I removed some of the null-checks from the query builder since they cause errors in EF when constructing the query. But this also leaves the chance that nulls might be encountered when materializing the query, for example when accessing the property of a null item, e.g. /Customers/1/Address/City where `Address` is null. In that case a `NullReferenceException` will be thrown, we catch it here and return a 404.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: I found `TypeHelper.GetImplementedEnumerableType` which does the same thing. Thanks for the tip.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: I removed this by reusing methods in `TypeHelper`, thanks.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+            builder.EntityType<EnableNestedPathsVipCustomer>();
+
+            builder.EntityType<EnableNestedPathsVipCustomer>()
+                .Function("GetMostPurchasedProduct")
+                .ReturnsFromEntitySet<EnableNestedPathsProduct>("EnableNestedPathsProduct");
+
+            builder.EntityType<EnableNestedPathsProduct>()
+                .Collection
+                .Action("SetDiscountRate");
+
```
marabooy, line 59:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+            _db = new EnableNestedPathsDatabase();
+        }
+
+        [Fact]
+        public async Task EnableNestedPaths_ReturnsEntiySetCollection()
+        {
+            // Arrange
+            string url = $"{_baseUrl}EnableNestedPathsCustomers";
+
```
marabooy, line 85:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+                readCustomers.Cast<EnableNestedPathsCustomer>(),
+                new EnableNestedPathsCustomerComparer());
+        }
+
+        [Fact]
+        public async Task EnableNestedPaths_ReturnsEntityById()
+        {
+            // Arrange
+            string url = $"{_baseUrl}EnableNestedPathsCustomers(2)";
+
```
marabooy, line 105:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+                readCustomer,
+                new EnableNestedPathsCustomerComparer());
+        }
+
+        [Fact]
+        public async Task EnableNestedPaths_ReturnsSingleton()
+        {
+            // Arrange
+            string url = $"{_baseUrl}EnableNestedPathsTopCustomer";
+
```
marabooy, line 127:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+                readCustomer,
+                new EnableNestedPathsCustomerComparer());
+        }
+
+        [Fact]
+        public async Task EnableNestedPaths_ReturnsPropertyOfSingleton()
+        {
+            // Arrange
+            string url = $"{_baseUrl}EnableNestedPathsTopCustomer/FavoriteProduct";
+
```
marabooy, line 149:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+                readProduct,
+                new EnableNestedPathsProductComparer());
+        }
+
+        [Fact]
+        public async Task EnableNestedPaths_ReturnsPrimitiveResults()
+        {
+            // Arrange
+            string url = $"{_baseUrl}EnableNestedPathsCustomers(1)/Name";
+
```
marabooy, line 171:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+            HttpResponseMessage response = await _client.GetAsync(url);
+
+            var stream = await response.Content.ReadAsStreamAsync();
+            var readName = ReadPrimitiveResponse<string>(stream, _model);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(_db.Customers.First().Name, readName);
+        }
+
```
marabooy, line 183:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+            // Assert
+            Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
+        }
+
+        [Fact]
+        public async Task EnableNestedPaths_AppliedBeforeEnableQuery()
+        {
+            // Arrange
+            string url = $"{_baseUrl}EnableNestedPathsCustomers(1)/Products?$orderby=Id desc";
+
```
marabooy, line 233:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+        private T ReadPrimitiveResponse<T>(Stream stream, IEdmModel model)
+        {
+            ODataMessageWrapper message = new ODataMessageWrapper(stream);
+            ODataMessageReader messageReader = new ODataMessageReader(message as IODataResponseMessage, new ODataMessageReaderSettings(), model);
+            ODataDeserializerContext readContext = new ODataDeserializerContext() { Model = model };
+            object readValue = _primitiveDeserializer.Read(messageReader, typeof(T), readContext);
+            return (T)readValue;
+        }
+
+
```
marabooy, line 279:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+                    Id = 3,
+                    Name = "Cust3",
+                    Products = new List<EnableNestedPathsProduct>(),
+                    Emails = new List<string>(),
+                    FavoriteProduct = null,
+                    HomeAddress = addresses[2],
+                    Addresses = new List<EnableNestedPathsAddress> { addresses[1], addresses[2] }
+                }
+            };
+
```
marabooy, line 345:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+        }
+
+        public IList<EnableNestedPathsProduct> Products { get; set; }
+        public IList<EnableNestedPathsCustomer> Customers { get; set; }
+
```
marabooy, line 350:
```suggestion
```

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/EnableNestedPathsTest.cs
```diff
@@ -0,0 +1,453 @@
...
+    {
+        readonly EnableNestedPathsDatabase _db = new EnableNestedPathsDatabase();
+
+        [EnableNestedPaths]
+        public SingleResult<EnableNestedPathsProduct> Get()
+        {
+            return new SingleResult<EnableNestedPathsProduct>(_db.Products.Where(p => p.Id == 1).AsQueryable());
+        }
+    }
+
```
marabooy, line 423:
```suggestion
```

marabooy (approved):
Just need to fix some whitespaces issues.

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
```diff
@@ -611,7 +611,7 @@ public virtual object ApplyQuery(object entity, ODataQueryOptions queryOptions)
                 {
                     IQueryable queryable = singleResultCollection as IQueryable;
                     queryable = ApplyQuery(queryable, queryOptions);
-                    return SingleOrDefault(queryable, actionDescriptor);
+                    return QueryHelpers.SingleOrDefault(queryable, actionDescriptor);
```
ElizabethOkerio, line 614: Is this change necessary? Having the QueryHelpers class? The implementation of SingleOrDefault in this class and what was there before is similar.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,208 @@
...
+                return null;
+            }
+
+            IEnumerable<ODataPathSegment> remainingSegments = segments.Skip(1);
+
+            foreach (ODataPathSegment segment in remainingSegments)
+            {
+                if (segment is KeySegment keySegment)
+                {
+                    // filterPredicate = entity => (entity.KeyProp1 == Val1) && (entity.keyProp2 == Val2) && ...
```
ElizabethOkerio, line 62: remove commented code

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,208 @@
...
+                    ParameterExpression filterParam = Expression.Parameter(queryable.ElementType, "entity");
+                    IEnumerable<BinaryExpression> conditions = keySegment.Keys.Select(kvp =>
+                        Expression.Equal(
+                            Expression.Property(filterParam, kvp.Key),
+                            Expression.Constant(kvp.Value)));
+                    BinaryExpression filterBody = conditions.Aggregate((left, right) => Expression.AndAlso(left, right));
+                    LambdaExpression filterPredicate = Expression.Lambda(filterBody, filterParam);
+
+                    // queryable = queryable.Where(entity => (entity.KeyProp1 == Val1) && (entity.KeyProp2 == Val2) ...)
```
ElizabethOkerio, line 71: remove commented code

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: I've also moved the Where filter before the sort.

src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
habbes: I moved the method to a helper class (`QueryHelpers`) because now it's also used by the `EnableNestedPathsAttribute` class. So instead of calling `EnableQueryAttribute.SingleOrDefault` from `EnableNestedPathsAttribute`, they now both call `QueryHelpers.SingleOrDefault`.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: This was not commented code, it was meant to explain what the following statement was trying to achieve using reflection:
```c#
queryable = ExpressionHelpers.Where(queryable, filterPredicate, queryable.ElementType);
```
maybe I should make that more clear.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: This is not commented code but an explanation of what the statement after it is trying to achieve using reflection. I think I will rewrite is follows to make it more clear:
```c#
// The following block is equivalent to
// filterPredicate = entity => (entity.KeyProp1 == Val1) && (entity.keyProp2 == Val2) && ...
ParameterExpression filterParam = Expression.Parameter(queryable.ElementType, "entity");
IEnumerable<BinaryExpression> conditions = keySegment.Keys.Select(kvp =>
    Expression.Equal(
        Expression.Property(filterParam, kvp.Key),
        Expression.Constant(kvp.Value)));
BinaryExpression filterBody = conditions.Aggregate((left, right) => Expression.AndAlso(left, right));
LambdaExpression filterPredicate = Expression.Lambda(filterBody, filterParam);
```

src/Microsoft.AspNetCore.OData/EnableNestedPathsAttribute.cs
habbes: @marabooy since you're accessing resources by their ids or properties, I'm struggling to picture the scenario you're laying out.

Suppose `Customer(1)/Orders` returns a collection. If the collection has null items, the response here should still be an array. If we want to access an entity in that collection, we'd access it by ID: `Customer(1)/Orders(2)` and in this case it doesn't matter what other items are null, we only care about the `Order` with id 2. If that doesn't exist, we return 404.

habbes: The tricky part here is that there's currently no consistency in how nulls and 404 are handled. I've explained that in detail in this issue: https://github.com/OData/WebApi/issues/2479

My view is that we should address the null/404 issue in a separate PR because it seems to be an issue that already exists outside of this PR, rather than trying to handle it here.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        private readonly IQueryable source;
+        private readonly Routing.ODataPath path;
+
+        /// <summary>
+        /// Creates an instance of <see cref="ODataPathQueryBuilder"/>
+        /// </summary>
+        /// <param name="source">The original <see cref="IQueryable"/> to be transformed</param>
+        /// <param name="model">The <see cref="IEdmModel"/></param>
+        /// <param name="path">The <see cref="ODataPath"/> based on which to transform the <see cref="source"/></param>
+        public ODataPathQueryBuilder(IQueryable source, IEdmModel model, Routing.ODataPath path)
```
KenitoInc, line 31: Remove the `IEdmModel model` since it's not being used anywhere

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        /// <summary>
+        /// The result of the query transformations applied to the original <see cref="IQueryable"/>
+        /// by the <see cref="ODataPathQueryBuilder"/>
+        /// </summary>
+        public IQueryable Result { get; set; }
+        /// <summary>
+        /// Whether the path has a $count segment
+        /// </summary>
+        public bool HasCountSegment { get; set; }
```
KenitoInc, line 204: How is this property used in computing the count value?

KenitoInc (commented):
A few comments.
Otherwise LGTM

xuzhg (approved):

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            if (method != ODataRequestMethod.Get)
+            {
+                // [EnableNestedPaths] only supports GET requests
+                return null;
+            }
+
+
+            // unsupported path segments
+            if (odataPath.PathTemplate.EndsWith("$ref")
```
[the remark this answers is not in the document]
habbes: @xuzhg This means that paths that end with `$ref` will no be handled by this convention. The reason for this is that the feature does not automatically generate responses for `$ref` requests. So for that, the user would have to manually create a controller action. `$ref` is still supported by the `RefRoutingConvention`.

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/NestedPathsRoutingConvention.cs
```diff
@@ -0,0 +1,89 @@
...
+                return null;
+            }
+
+
+
+
+            // if we did no find a matching action amongst the conventional user-defined methods
+            // then let's check if the controller has a Get method with [EnableNestedPaths] attribute
+            // which should be used to catch any any nested GET request
+            IWebApiActionDescriptor descriptor = actionMap.GetActionDescriptor($"Get{sourceName}") ?? actionMap.GetActionDescriptor("Get");
```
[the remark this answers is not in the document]
habbes: I'm not sure I fully understand the question, but assuming we have both a `Get()` and a `Get(int key)` method. Then the `Get(int key)` will be handled by the `EntityRoutingConvention`. Because this convention runs after the other conventions have been tried, the overloads should have been taken care by this point.

src/Microsoft.AspNet.OData.Shared/Query/ODataPathQueryBuilder.cs
habbes: This flag is set to true if the path contains a `$count` segment. This informs as that the result of the query should be treated as a collection.

KenitoInc (approved):
LGTM

Issue #2469, opened 2021-04-20
Solution does not build under Linux because of wrong lettercase in filenames


obohaciak:
*When building the WebApiOData.AspNetCore.sln under Linux, several files have wrong letter cases causing the compilation to fail. Case-insensitive OS like Windows is more permissive and does not fail to compile the solution.*

### Assemblies affected

Current master, as of 2020/04/20.

### Reproduce steps

Run `dotnet build ./sln/WebApiOData.AspNetCore.sln` in shell.

### Expected result

Solution compiles.

### Actual result

Compiler reports following errors:
```
##[error]CSC(0,0): Error CS2001: Source file '/home/vsts/work/1/s/src/Microsoft.AspNet.OData.Shared/Builder/NonbindingParameterConfiguration.cs' could not be found.
##[error]CSC(0,0): Error CS2001: Source file '/home/vsts/work/1/s/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/PlainTextODataQueryOptionsParserTest.cs' could not be found.
##[error]CSC(0,0): Error CS1566: Error reading resource 'Microsoft.AspNet.OData.Test.Resources.ProductsCsdl.xml' -- 'Could not find file '/home/vsts/work/1/s/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Resources/ProductsCsdl.xml'.'
```
### Additional detail

Clearly, the offending filenames were supposed to be Pascal-cased but either a part of the filename was not capitalized or an acronym did not follow .NET casing guideline.

obohaciak:
The fix is trivial, I'll prepare a PR immediately.

habbes:
@obohaciak  Thanks for raising this and for the quick PR. We'll review it.

PR #2470, opened 2021-04-20
Fix the casing so that the project compiles under case-sensitive OS 

obohaciak:
Fix the casing so that the project compiles under case-sensitive OS like Linux

### Issues

*This pull request fixes issue #2469.*

### Description

*Corrects filename casing.*

- [ ] *Test cases added - no test cases needed*

### Additional work necessary

*No tests were added as there's no change in code. No docs needed.*


habbes (approved):

KenitoInc (approved):
LGTM

marabooy (approved):

habbes:
I've found that while the file is named `NonBindingParameterConfiguration.cs`, the class itself is actually named `NonbindingParameterConfiguration` (`binding` is no capitalized). But renaming the class would be a breaking changing since it's a public class.

KenitoInc:
> I've found that while the file is named `NonBindingParameterConfiguration.cs`, the class itself is actually named `NonbindingParameterConfiguration` (`binding` is no capitalized). But renaming the class would be a breaking changing since it's a public class.

We can ensure all files are consistent with the class names!

obohaciak:
Fix the casing so that the project compiles under case-sensitive OS

obohaciak:
Misclicked :)

habbes:
@obohaciak in order to maintain consistency between the filenames and class names, maybe it would be better to rename the file instead to `NonbindingParameterConfiguration.cs` to match the class name.

obohaciak:
@habbes I suppose you're right. Apologies, but @xuzhg was faster and the PR is already merged. I could open up another PR, but I don't want to waste your time reviewing trivial changes.

PR #2483, opened 2021-05-01
Delta does not allow UpdatableProperties to be changed

### LINKED ISSUE #2478: Delta does not allow UpdatableProperties to be changed
The framework creates a Delta for Patch() but in many scenarios the controller wants to protect from the client
over-posting. When searching StackOverflow questions such as [this](https://stackoverflow.com/questions/28248785/odata-delta-patch-security) show that the solutions are very complex.

### Assemblies affected

OData WebApi lib 7.x

### Reproduce steps

Use the sample controller's Patch code and try to restrict the properties that are updated.

### Expected result

Only update the properties that are specified by the controller.

### Actual result

All modified properties are updated.

### Additional detail

When a Delta is constructed, it is possible to specify the `updatableProperties` parameter to restrict which properties can be set in the Delta, and also limit which properties will be copied on `Patch` and `Put`. However, once the Delta is constructed the list cannot be changed. Furthermore, because the framework constructs the Delta it cannot be specified by the controller.

Workarounds on the Internet typically involve one of the following:

- Reconstructing the Delta through extensions to find all changed properties and only copying the selected
- Replacing the Put/Patch method with an equivalent method that explicitly does the copying

A much simpler solution is to expose the property:

```
public IEnumerable<string> UpdatableProperties { get; set; } 
```

The getter simply returns the existing `_updatableProperties`.

The setter replaces `_updatableProperties` with the `value` and additionally updates the `_changedProperties` to remove any properties that are no longer in the `_updatableProperties`.

Subsequently the Put() and Patch() methods work as expected, only the specified properties are updated.

Example:
```
public void Patch(int productId, Delta<Product> product)
{
    var productFromDb = // get product from db by productId
    // Restrict the updatable properties to prevent over posting
    product.UpdatableProperties =  new[] { "SafeProperty1", "SafeProperty2"};
    product.Patch(productFromDb);
    // some other stuff
}
```

dxrdxr:
### Issues

This pull request fixes #2478

### Description

Adds a new settable property `UpdatableProperties` that allows the developer to change the list of updatable properties (`_updatableProperties`) after construction. This enables the developer to easily control which properties are updated in `Patch` and `Put` to prevent the caller from over-posting.

- [ x] *Test cases added*
- [ x] *Build and test with one-click build and test script passed*

### Additional work necessary

Docs Needed:

- Update the documentation pages to include the new property.

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -107,6 +107,23 @@ namespace Microsoft.AspNet.OData
             get { return typeof(TStructuralType); }
         }
 
+        /// <summary>
+        /// The list of property names that can be updated.
+        /// </summary>
+        /// <remarks>When the list is modified, any modified properties that were removed from the list are no longer
+        /// considered to be changed.</remarks>
+        public IEnumerable<string> UpdatableProperties
+        {
+            get => _updatableProperties;
+            set
+            {
+                _updatableProperties = InitializeUpdatableProperties(value);
+
+                // Remove any un-updatable properties from the changed list
+                _changedProperties.IntersectWith(_updatableProperties);
+            }
+        }
+
         /// <inheritdoc/>
         public override void Clear()
         {
@@ -503,20 +520,28 @@ namespace Microsoft.AspNet.OData
                     .Select<PropertyInfo, PropertyAccessor<TStructuralType>>(p => new FastPropertyAccessor<TStructuralType>(p))
                     .ToDictionary(p => p.Property.Name));
 
+            _updatableProperties = InitializeUpdatableProperties(updatableProperties);
+        }
+
+        private HashSet<string> InitializeUpdatableProperties(IEnumerable<string> updatableProperties)
+        {
+            HashSet<string> newUpdatableProperties;
             if (updatableProperties != null)
             {
-                _updatableProperties = new HashSet<string>(updatableProperties);
-                _updatableProperties.IntersectWith(_allProperties.Keys);
+                newUpdatableProperties = new HashSet<string>(updatableProperties);
+                newUpdatableProperties.IntersectWith(_allProperties.Keys);
             }
             else
             {
-                _updatableProperties = new HashSet<string>(_allProperties.Keys);
+                newUpdatableProperties = new HashSet<string>(_allProperties.Keys);
             }
 
             if (_dynamicDictionaryPropertyinfo != null)
             {
-                _updatableProperties.Remove(_dynamicDictionaryPropertyinfo.Name);
+                newUpdatableProperties.Remove(_dynamicDictionaryPropertyinfo.Name);
             }
+
+            return newUpdatableProperties;
         }
 
         // Copy changed dynamic properties and leave the unchanged dynamic properties
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -4,13 +4,11 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
-using Microsoft.AspNet.OData.Builder.Conventions.Attributes;
 using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Formatter;
 
@@ -24,11 +22,11 @@ namespace Microsoft.AspNet.OData
     public class Delta<TStructuralType> : TypedDelta, IDelta where TStructuralType : class
     {
         // cache property accessors for this type and all its derived types.
-        private static ConcurrentDictionary<Type, Dictionary<string, PropertyAccessor<TStructuralType>>> _propertyCache
+        private static readonly ConcurrentDictionary<Type, Dictionary<string, PropertyAccessor<TStructuralType>>> _propertyCache
             = new ConcurrentDictionary<Type, Dictionary<string, PropertyAccessor<TStructuralType>>>();
 
         private Dictionary<string, PropertyAccessor<TStructuralType>> _allProperties;
-        private HashSet<string> _updatableProperties;
+        private List<string> _updatableProperties;
 
         private HashSet<string> _changedProperties;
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -107,6 +107,23 @@ public override Type ExpectedClrType
             get { return typeof(TStructuralType); }
         }
 
+        /// <summary>
+        /// The list of property names that can be updated.
+        /// </summary>
+        /// <remarks>When the list is modified, any modified properties that were removed from the list are no longer
+        /// considered to be changed.</remarks>
+        public IEnumerable<string> UpdatableProperties
```
xuzhg, line 115: I'd prefer to have an overload of "Patch of Put " to accept the list of updatable properties.
Because we have the overload of constructor to accept the list of updatable properties. This properties doesn't make sense here.

@mikepizzo What do you think?

xuzhg (changes requested):

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
habbes: @xuzhg @dxrdxr shared reasons why they believe an overload would not be suitable in the issue discussion here: https://github.com/OData/WebApi/issues/2478#issuecomment-828337073

dxrdxr: @xuzhg In addition to my comments in #2478, the existing `updatableProperties` is set only through the constructor, a very strange pattern. It is a write-only state of the class that cannot be examined through any interface on the class. Operations such as `TrySetPropertyValue()` can fail but there is no way to determine if it failed because it was not updatable, or another reason (e.g. type mismatch), this is a developer headache (note that the current docs do not specify that it will fail if not updatable).

Additionally, the current ctor argument only initialization goes against the much more common practice of allowing state (properties) to be set either as an argument or as with an object initializer.
```
var delta = new Delta<Student>(typeof(Student), myUpdatableProperties);
```
```
var delta = new Delta<Student>()
{
    UpdatableProperties = myUpdatableProperties
};
```
Lastly as I said in my other comments, #3, for consistent semantics it is not just `Put` and `Patch` it impacts `GetChangedPropertyNames` and `GetUnchangedPropertyNames`. Having the developer maintain and wrap all of the Get/Set methods to filter properties versus pushing it close to the object that is already managing the state, is a cleaner approach.

habbes: @xuzhg could you provide your views on this one?

marabooy:
@xuzhg, @habbes, @KenitoInc, @gathogojr  Could you give this another review?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
xuzhg: I think it's ok to expose the set for property. Or maybe we can add a new method to update the "UpdatableProperties".

dxrdxr: @xuzhg 
> Or maybe we can add a new method to update the "UpdatableProperties".

To create a new method to update the field it would require a `GetUpdateableProperties` and a `SetUpdateableProperties`. How is that different that the getter/setter on the property `UpdateableProperties`?

It is important to fix the problem that the existing `_updatableProperties` has no way retrieve its current value, a write-only field. So a Set method or property setter must be part of the solution.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/PublicApi/PublicApiTest.cs
```diff
@@ -31,12 +32,25 @@ public void TestPublicApi()
             string baselineString = GetBaseLineString();
 
             // Assert
-            Assert.True(String.Compare(baselineString, outputString, StringComparison.Ordinal) == 0,
-                String.Format("Base line file {1} and output file {2} do not match, please check.{0}" +
-                "To update the baseline, please run:{0}{0}" +
-                "copy /y \"{2}\" \"{1}\"", Environment.NewLine,
-                BaseLineFileFolder + BaseLineFileName,
-                outputFile));
+            if (string.Compare(baselineString, outputString, StringComparison.Ordinal) != 0)
```
gathogojr, line 35: @dxrdxr Was it necessary to include this change in this PR?

gathogojr (approved):
LGTM

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/PublicApi/PublicApiTest.cs
dxrdxr: I was getting testing errors and this simply allowed me to have better diagnostics on what was actually wrong.

I am ambivalent on if it is included, but it seemed to be a generally positive improvement to the test environment.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
habbes: I think keeping it as a property makes more sense.

dxrdxr: > I'd prefer to have an overload of "Patch of Put " to accept the list of updatable properties.
> Because we have the overload of constructor to accept the list of updatable properties. This properties doesn't make sense here.
> 
> @mikepizzo What do you think?

@xuzhg Do you still need changes to be made?

KenitoInc (approved):

xuzhg (approved):

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
dxrdxr: Excellent question! 

I was following the existing implementation being an `IEnumerable` so left it as a set the whole property.

It makes mores sense to expose it as an `IList` so that it can be updated instead of just replaced.

I will reimplement this and see the impact.

Thanks!

gathogojr: Making it `IList` and readonly does make a lot of sense

dxrdxr: I updated the Pull Request to change to having it be an `IList` and readonly.

This change invalidates an existing assumption that the contents of the `_updatableProperties` field only contains the names of properties that are also in the `allProperties` dictionary. As an `IList`, any arbitrary string can be added, so the code needed to be updated in places to filter against invalid properties.

Additionally, the internal use of `HashSet` must be changed to `List` as `HashSet` does not implement `IList`. This will have a minor performance impact on the use of `Contains` going from O(1) to O(n), however the number of properties is always small so should not be a factor.

gathogojr:
@dxrdxr Any particular reason you opted to substitute `var` with specific types in the tests? It ended up increasing the diff unnecessarily. The additional tests are welcome, however, having as few changes to the existing tests as possible would have served as a validation that the changes to `Delta<TStructuralType>` class didn't materially change how the feature works .

dxrdxr:
@gathogojr I was getting warnings in VS about the code style, which is inconsistent within the file. So I normalized it for consistency and code hygiene.

The project really should have a .editorconfig in the root of the directory tree to enforce a common code style.

PR #2487, opened 2021-05-05
Tests using TimeZoneInfo fail due to race conditons

### LINKED ISSUE #2484: Test ODataModelBinderProvider_Works_DateTime randomly fails
When running the test ODataModelBinderProvider_Works_DateTime in WebAPI it often "randomly" fails.

The symptom is that this Assert fails:

```
    Assert.Equal(
        value,
        await response.Content.ReadAsAsync(value.GetType(), _configuration.Formatters));
```
The expected value is correct as expected, however the response value is off by a whole number of hours, always less than 24. This is an indication that it is likely a TimeZone problem.

_Other tests may also be failing in a similar manner._

### Assemblies affected

WebApi 7.5.7

### Reproduce steps

- When the test is repeatedly run by itself in Visual Studio Test Explorer it never fails.
- When the full Microsoft.AspNet.OData.Test 8000+ tests are run in Visual Studio with "Run Tests in Parallel" it will frequently fail.
- When run as part of the rolling tests or build.cmd/build.ps1 it will fail more often than not.

### Expected result

Test reliably passes every time.

### Actual result

Test frequently fails.

### Additional detail

In debugging the failures a number of things came up:

1. Testing DateTime is inherently difficult because the conversion from DateTimeOffset to DateTime causes the TimeZone information to be [lost](https://docs.microsoft.com/en-us/odata/webapi/datetime-support#time-zone-configuration). Attempts to [fix](https://docs.microsoft.com/en-us/aspnet/web-api/overview/formats-and-model-binding/json-and-xml-serialization#dates) the TimeZone to a specific value on both the client and server did not help the problem.
2. In examining the test, it is discovered that the `DateTime.Kind` is always properly set to `Local`, but in the `GetDateTime()` controller method it is `Unspecified` and that is returned in the Response. This is an indication that the parsing is needing to use the configured TimeZone information to set value.
3. The symptoms made it appear as if the Controller was running in a different data center, plausible for a Azure test but impossible on a local Dev machine. Thus it is highly unlikely to be a system TimeZone issue.
4. The `TimeZoneInfoHelper.TimeZone` property is used in 30+ places to set the TimeZone. However, in the shipping `src` it is  only set in `SetTimeZoneInfo()` a Public API that is likely only called during Http configuration. But, both the property and method are called numerous places in the `test` code.
5. One problem with the `TimeZoneInfoHelper.TimeZone` property is that it is a static property, something that is inherently not thread-safe. While not documented as thread-unsafe, it is a reasonable assumption given it's Public API is usually a configuration option, it is likely not designed to be thread-safe. One option to improve the situation is to **add a lock** to protect the `_defaultTimeZoneInfo` field.
6. Making the static property thread-safe will likely help some test failure conditions, but the parallelism in xUnit will still allow race conditions where a test sets the TimeZoneInfo to be a specific value (e.g. on the client side) while another test sets it to be a different value by the time the first test uses the now changed value (e.g. on the server controller). There is no reasonable locking scheme to handle this scenario. However, it is possible to force certain [xUnit tests to not be run in parallel](https://xunit.net/docs/running-tests-in-parallel) which should prevent the inter-test races.

My suggested fixes:

- As stated in 5 above, add a lock and make the static property `TimeZoneInfoHelper.TimeZone`  thread-safe.
- As stated in 6 above, identity all tests that manipulate the `TimeZone` through the property or `SetTimeZoneInfo()` and put them in their own "Custom Test Collection" via the `CollectionAttribute`. This looks easier than trying to split those tests into a single Class.

### Root Cause
xUnit v2 introduced parallelism. If tests were written for an earlier (or different) framework they were likely single-threaded.

dxrdxr:
### Issues

Fixes #2484

### Description

When running the test ODataModelBinderProvider_Works_DateTime in WebAPI it often "randomly" fails.

Details in the issue.

- [X] *Test cases added*

Testcases run repeatedly run after the changes and the test reliably passes.

### DIFF
--- a/src/Microsoft.AspNet.OData/GlobalSuppressions.cs
+++ b/src/Microsoft.AspNet.OData/GlobalSuppressions.cs
@@ -60,3 +60,4 @@ using System.Diagnostics.CodeAnalysis;
 [assembly: SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Scope = "member", Target = "Microsoft.AspNet.OData.Builder.FunctionConfiguration.#HasDerivedTypeConstraintForReturnType`1()")]
 [assembly: SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Scope = "member", Target = "Microsoft.AspNet.OData.Builder.IODataInstanceAnnotationContainer.#GetAllResourceAnnotations()")]
 [assembly: SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Scope = "member", Target = "Microsoft.AspNet.OData.Builder.IODataInstanceAnnotationContainer.#GetResourceAnnotations()")]
+[assembly: SuppressMessage("Microsoft.Usage", "CA2208:InstantiateArgumentExceptionsCorrectly", Scope = "member", Target = "Microsoft.AspNet.OData.Common.Error.#PropertyNullOrWhiteSpace()")]
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DateAndTimeOfDay/DateAndTimeOfDayTest.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DateAndTimeOfDay/DateAndTimeOfDayTest.cs
@@ -22,6 +22,7 @@ using Xunit;
 
 namespace Microsoft.Test.E2E.AspNet.OData.DateAndTimeOfDay
 {
+    [Collection("TimeZoneTests")] // TimeZoneInfo is not thread-safe. Tests in this collection will be executed sequentially 
     public class DateAndTimeOfDayTest : WebHostTestBase
     {
         public DateAndTimeOfDayTest(WebHostTestFixture fixture)
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DateTimeSupport/DateTimeTest.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DateTimeSupport/DateTimeTest.cs
@@ -21,6 +21,7 @@ using Xunit;
 
 namespace Microsoft.Test.E2E.AspNet.OData.DateTimeSupport
 {
+    [Collection("TimeZoneTests")] // TimeZoneInfo is not thread-safe. Tests in this collection will be executed sequentially 
     public class DateTimeTest : WebHostTestBase
     {
         public DateTimeTest(WebHostTestFixture fixture)
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBoundQuerySettings/PageAttributeTest/SkipTokenTest.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBoundQuerySettings/PageAttributeTest/SkipTokenTest.cs
@@ -12,6 +12,7 @@ using Xunit;
 
 namespace Microsoft.Test.E2E.AspNet.OData.ModelBoundQuerySettings.PageAttributeTest.SkipTokenTest
 {
+    [Collection("TimeZoneTests")] // TimeZoneInfo is not thread-safe. Tests in this collection will be executed sequentially 
     public class SkipTokenTest : WebHostTestBase
     {
         private const string CustomerBaseUrl = "{0}/skiptokentest/Customers";
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DateTimeTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DateTimeTest.cs
@@ -37,6 +37,7 @@ using Xunit;
 
 namespace Microsoft.AspNet.OData.Test
 {
+    [Collection("TimeZoneTests")] // TimeZoneInfo is not thread-safe. Tests in this collection will be executed sequentially 
     public class DateTimeTest
     {
         // Note: the product uses a static TimezoneHelper class to store the timezone. As a result,
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/DefaultODataETagHandlerTests.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/DefaultODataETagHandlerTests.cs
@@ -12,6 +12,7 @@ using Xunit;
 
 namespace Microsoft.AspNet.OData.Test.Formatter
 {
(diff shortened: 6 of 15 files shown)


src/Microsoft.AspNet.OData.Shared/TimeZoneInfoHelper.cs
```diff
@@ -9,18 +9,30 @@ internal class TimeZoneInfoHelper
...
         private static TimeZoneInfo _defaultTimeZoneInfo;
 
+        private static object _timeZoneLock = new object();
+
         public static TimeZoneInfo TimeZone
         {
             get
             {
-                if (_defaultTimeZoneInfo == null)
+                var timeZone = TimeZoneInfo.Local;
```
xuzhg, line 18: I am worry about the timezone lock.  We need to refactor this static.

I did the refactor at our 8.0. Maybe we can do it here.

xuzhg (changes requested):

src/Microsoft.AspNet.OData.Shared/TimeZoneInfoHelper.cs
dxrdxr: In 8.0 you refactored the code to move the `TimeZoneInfo` into an instance of `ODataOptions` so `SetTimeZoneInfo` becomes an instance method.

In WebApi, `SetTimeZoneInfo` is a public `static` method on `ODataRouteBuilderExtensions` (`IRouteBuilder`) and it is using `TimeZoneInfoHelper.TimeZone` to save the `TimeZoneInfo` in a static field. The difficulty of a **refactor** is in figuring out where to store a per-instance value through an extension method (extension properties don't exist). The `IRouteBuilder` has a size that is fixed at construction so space can't be added to it. The proper technique is through inheritance, but that will break the existing public API.

So I agree that it should be refactored as you have done in 8.0, but I don't see how to do that without going to WebAPI 8.0.

I agree that locks are _ugly_, but static fields in a static class are inherently not thread-safe (and strongly discouraged). So a lock is required.

However, the impact of the lock on real-world usage is trivial. Multi-threaded calls to Get/SetTimeZoneInfo is extremely rare, the common case is to set it only once during configuration.

xuzhg: Is there any solution that we can resolve the problem without introducing lock?

chrisspre: Maybe AsyncLocal<T>. It  essentially gives each thread its own copy and ensures that changes are propagated through the call chain even if thread context changes happen. 

```CSHARP
        private static AsyncLocal<TimeZoneInfo> _defaultTimeZoneInfo = ...;

        public static TimeZoneInfo TimeZone { get => _defaultTimeZoneInfo.Value; set => _defaultTimeZoneInfo.Value = value}
```

More research and more testing would be required to see if that achieves what is desired.

dxrdxr: @chrisspre `AsyncLocal` will not work, conceptually it is analogous to Thread Local Storage with the added property that the values "flow" across Task execution to another thread if during an `await` the thread is blocked and the flow has to continue on another thread. Internally it is using the `ExecutionContext` to store the "local" values. Once a thread's "flow" completes (Thread Abort or Join; ThreadPool completion) the AsyncLocal data is cleared.

So the semantics are not correct:

- Any `TimeZoneInfo` set and saved in an `AsyncLocal` cannot be read by any other thread unless it is a thread that "flowed" through Async from the original thread.
- Any `TimeZoneInfo` set and saved in an `AsyncLocal` will be cleared when the Thread exits or is returned to the ThreadPool. For a ASP.Net OData server that means the `TimeZoneInfo`  will only persist for the duration of the HttpRequest.
- The current semantics of `SetTimeZoneInfo` is that the value persists for the live of the service

The question @xuzhg  posed is can we avoid a lock? Underlying that question is the more fundamental question of **why a lock should be avoided**? 

Locks are _heavyweight_ (TLB shootdowns) but in this case it is written once and read many which is the optimal case.

In this case the `TimeZoneInfo` is rarely called (1-2 times per request), not in any critical path or loop.

In this case the lock's hold time is extremely short.

habbes: @dxrdxr @xuzhg  @chrisspre  what if the tests that use the TimeZoneInfo were set to run sequentially?

In production, is the TimeZone changed per request, per route or once during the entire lifetime of the app? If it's meant to be set only once in production, then maybe it might be sufficient to let the tests run sequentially and avoid the locks in production. What do you think?

dxrdxr: @habbes I have put all the `TimeZoneInfo` tests in the same "Collection" so that the do now run sequentially. But that is necessary even with the locked access. The problem is that 2 tests running in parallel one sets it to UTC-8 and another one sets it to UTC, then the first accesses it getting UTC and fails the test.

I have done some deeper C#/CLR research. Based on [Atomicity of variable references](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/variables#atomicity-of-variable-references) "Reads and writes of the following data types are atomic: [...] reference types." I mistakenly thought that a `TimeZoneInfo` was a structure type (like `DateTime`) which is a Value type and not atomic.

So I agree and withdraw the lock changes.

dxrdxr:
Removal of the lock accepted.

habbes (approved):
Looks good to me.

I wonder whether it would be necessary to add a comment explaining why it's necessary for those tests to run sequentially.

dxrdxr:
> Looks good to me.
> 
> I wonder whether it would be necessary to add a comment explaining why it's necessary for those tests to run sequentially.

Done, comments added.

KenitoInc (approved):

gathogojr (approved):
@dxrdxr I'm not sure the comment you added is that helpful to someone who's not that familiar with what effect adding the attribute has. Consider revising it, e.g.
```csharp
// TimeZoneInfo is not thread-safe. Tests decorated with this attribute will be executed sequentially
```
Otherwise, LGTM

dxrdxr:
> @dxrdxr I'm not sure the comment you added is that helpful to someone who's not that familiar with what effect adding the attribute has. Consider revising it, e.g.
> 
> ```cs
> // TimeZoneInfo is not thread-safe. Tests decorated with this attribute will be executed sequentially
> ```
> 
> Otherwise, LGTM

I agree.

I made a slight rewording, emphasizing that it is about the collection, which is what the attribute creates:
`// TimeZoneInfo is not thread-safe. Tests in this collection will be executed sequentially`

xuzhg (approved):

mikepizzo:
Thanks for the improvement David!

Issue #2490, opened 2021-05-11
Private reflection breaks .NET 6 


davidfowl:
As part of .NET 6, we will want to move the `ServiceCollection` type to **Microsoft.Extensions.DependencyInjection.Abstractions** https://github.com/dotnet/runtime/pull/52284. This broke OData since it uses private reflection to make this work 1.0 and 2.0 work without having to re-compile.

https://github.com/OData/WebApi/blob/757631faf10b494e55117904196e8aa45d4c764d/src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs#L80-L92

It might be time to drop **Microsoft.Extensions.DependencyInjection** 1.0 support. That would clean this code up nicely.

davidfowl:
That's the change I plan to make.

PR #2491, opened 2021-05-11
Make the reflection more robust

davidfowl:
- This was caught when moving ServiceCollection to another assembly (type forwarded).

### Issues

Fixes https://github.com/OData/WebApi/issues/2490

### Description

Today, OData tries to make MEDI (Microsoft.Extensions.DependencyInjection) 1.0 and 2.0 work side by side (original change here https://github.com/OData/WebApi/pull/1082), to make this work it uses reflection to call BuildServiceProvider (since there was a breaking change between 1.0 and 2.0). This change makes the reflection code a bit more robust by looking for the method on the type and method where it's defined.

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
@@ -85,9 +85,7 @@ namespace Microsoft.AspNet.OData
             * More info at https://github.com/OData/WebApi/pull/1082
             */
 
-            Assembly microsoftExtensionsDependencyInjectionAssembly = services.GetType().GetTypeInfo().Assembly;
-            TypeInfo serviceCollectionContainerBuilderExtensionsType = microsoftExtensionsDependencyInjectionAssembly.GetType(typeof(ServiceCollectionContainerBuilderExtensions).GetTypeInfo().FullName).GetTypeInfo();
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
```diff
@@ -85,9 +85,7 @@ public virtual IServiceProvider BuildContainer()
             * More info at https://github.com/OData/WebApi/pull/1082
             */
 
-            Assembly microsoftExtensionsDependencyInjectionAssembly = services.GetType().GetTypeInfo().Assembly;
-            TypeInfo serviceCollectionContainerBuilderExtensionsType = microsoftExtensionsDependencyInjectionAssembly.GetType(typeof(ServiceCollectionContainerBuilderExtensions).GetTypeInfo().FullName).GetTypeInfo();
-            MethodInfo buildServiceProviderMethod = serviceCollectionContainerBuilderExtensionsType.GetMethod("BuildServiceProvider", new[] { typeof(IServiceCollection) });
+            MethodInfo buildServiceProviderMethod = typeof(ServiceCollectionContainerBuilderExtensions).GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) });
```
eerhardt, line 88: (nit) how often does this get called? I assume only once usually. So it probably doesn't pay to cache it.

eerhardt (approved):
Looks great!

xuzhg (approved):

mikepizzo (approved):
:shipit:

Issue #2493, opened 2021-05-13
Problem when generating OData nextlink with a function parameter alias


rajni291:
*Short summary (3-5 sentences) describing the issue.*

There is a problem when using Odata Web API with function parameter alias. 

Odata Original Request:
https://localhost:44374/odata/v3/HealthCare/Changes(startDate=@startDate)?@startDate=2020-06-06T14:03:00.6753255Z

OdataNextLink is generated as below:-
"@odata.nextLink":"https://localhost:44374/odata/v3/HealthCare/Changes(startDate=@startDate)?%40startdate=2020-06-06T14%3A03%3A00.6753255Z&$skip=1000"
The nextlink that is generated has lowercased the **startDate** in query (after ?). 

### Expected result
NextLink should maintain case sensitivity of any user defined parameter aliases.

### Actual result
NextLink is lower casing any user defined parameter aliases.

### Additional detail

In Class GetNextPageHelper, all query parameter keys are converted to lowercase. Can this be skipped if the key starts with '@'
signifying parameter alias.

marabooy:
Hi @rajni291, This does appear to be a bug which was intended to normalize the query options from OData so they are compared in a case insensitive manner. A good fix for this would be to only change the OData query options to lowercase and leave everything else in the same case. Would you be interested in contributing to a fix for this? If so do let me know if you need any help.

rajni291:
Hi @marabooy , I've created PR for the same, Could you look into that.
https://github.com/OData/WebApi/pull/2501

PR #2501, opened 2021-06-02
Update GetNextPageHelper.cs (Issue #2493)

rajni291:
Converted key to lowercase only if it's odata query options (starting with $).

### Issues

*This pull request fixes issue #2493.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNet.OData
 
             foreach (KeyValuePair<string, string> kvp in queryParameters)
             {
-                string key = kvp.Key.ToLowerInvariant();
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.OData.Shared/GetNextPageHelper.cs
```diff
@@ -29,7 +29,7 @@ internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<str
 
             foreach (KeyValuePair<string, string> kvp in queryParameters)
             {
-                string key = kvp.Key.ToLowerInvariant();
+                string key = kvp.Key.StartsWith("$")? kvp.Key.ToLowerInvariant() : kvp.Key;
```
habbes, line 32: Is this a reliable way to tell whether the query param is not a standard OData query param. I think in 4.01, OData allows "official" query params without the `$` prefix.

habbes (changes requested):
Could you add tests?

KenitoInc (changes requested):
Add tests case to validate that this fix works for both dollar and no-dollar query options

Issue #2511, opened 2021-06-17
Cannot return errors in WebApi with EnableFilter on.


marabooy:
*Short summary (3-5 sentences) describing the issue.*

Trying to return an error object from an endpoint that has EnableQuery results in an error.
### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData WebApi lib 7.6.5

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*
```csharp
[EnableQuery(AllowedQueryOptions = AllowedQueryOptions.Filter)]
public IActionResult Get()
{
        return this.StatusCode(500, new { Error = new { Code = "500", Message ="Some message"} });
}
```
### Expected result
*What would happen if there wasn't a bug.*

```json 

{
    "error": {
        "code": "500",
        "message": "Some message"
    }
}
```

### Actual result

*What is actually happening.*
```js
{
    "error": {
        "code": "",
        "message": "The query specified in the URI is not valid. The requested resource is not a collection. Query options $filter, $orderby, $count, $skip, and $top can be applied only on collections.",
        "details": [],
        "innererror": {
            "message": "The requested resource is not a collection. Query options $filter, $orderby, $count, $skip, and $top can be applied only on collections.",
            "type": "Microsoft.OData.ODataException"
      }
}
```
### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*
EnableQueryAttribute

```csharp
 // actionExecutedContext.Result might also indicate a status code that has not yet
// been applied to the result; make sure it's also successful.
StatusCodeResult statusCodeResult = actionExecutedContext.Result as StatusCodeResult;
// Potential bug as the this.StatusCode(int, obj) returns an ObjectResult instead of a Status code result. Pre core 2.2 did not have IStatusCodeActionResult.
// StatusCodeResult & ObjectResult all implement IStatusCodeActionResult

```

frankpeng7:
Hi @marabooy thanks for opening this! Is there a workaround I could use to show the correct error message before the fix gets released?

marabooy:
Hi @frankpeng7 you could use the API for OData error result which should format the error message to what you need unless if you have special formatters.

`this.ODataErrorResult(string code, string message)`

frankpeng7:
[...]

PR #2512, opened 2021-06-18
Fix behaviour on status code requests.

marabooy:
### Issues

*This pull request fixes issue #2511 .*

### Description

Fixes the usage of enable query for status code requests as they are not used within AspNetCore OData.
The api ` this.StatusCode(int, obj)` returns an `ObjectResult` instead of a `StatusCodeResult`. and Object result is not a child of status code result.

The original intent seems to be when the ObjectResult is not null proceed to apply the query options.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
@@ -140,7 +140,7 @@ namespace Microsoft.AspNet.OData
                 {
                     elementType = TypeHelper.GetImplementedIEnumerableType(returnType);
                 }
-                else if(TypeHelper.IsGenericType(returnType) && returnType.GetGenericTypeDefinition() == typeof(Task<>))
+                else if (TypeHelper.IsGenericType(returnType) && returnType.GetGenericTypeDefinition() == typeof(Task<>))
                 {
                     elementType = returnType.GetGenericArguments().First();
                 }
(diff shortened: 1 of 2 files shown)


KenitoInc (approved):
LGTM

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -228,46 +228,44 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
             {
                 // actionExecutedContext.Result might also indicate a status code that has not yet
                 // been applied to the result; make sure it's also successful.
-                StatusCodeResult statusCodeResult = actionExecutedContext.Result as StatusCodeResult;
-                if (statusCodeResult == null || IsSuccessStatusCode(statusCodeResult.StatusCode))
+                ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
```
habbes, line 231: Are there other return result types other than `ObjectResult` that we should care about? I see that [`IStatusCodeActionResult`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.infrastructure.istatuscodeactionresult?view=aspnetcore-5.0) is the common interface of all action results that have a status code. Why not use that?

habbes: Looking further, it seems that there are a few other result types that have a `Value`, such as `JsonResult`, and there's also `ContentResult` which has `Content`. Is the assumption here that `ObjectResult` is the only one that has a `Value` that can be processed by `[EnableQuery]`?

marabooy: The interface exists for  net core 2.2 and above. I just went through the code and patched the behaviour as we don't look at other items such as JsonResult and Content result.

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -228,46 +228,44 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
...
+
+                if (responseContent != null && (responseContent.StatusCode == null || IsSuccessStatusCode(responseContent.StatusCode.Value)))
                 {
-                    ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
-                    if (responseContent != null)
+
+                    //throw Error.Argument("actionExecutedContext", SRResources.QueryingRequiresObjectContent,
+                    //    actionExecutedContext.Result.GetType().FullName);
```
habbes, line 237: Maybe you could remove this commented code.

It seems hat it as meant to throw an exception if the result was not an ObjectResult. I wonder why it was commented out. But I think it should be removed if it's not used.

habbes (approved):

xuzhg (approved):

Issue #2518, opened 2021-07-04
ExpressionBinderBase issue with .NET 6


ysmoradi:
**This exception should occur in ASP.NET Core OData 8 too, because I'm talking about the code which exists in Shared section of  repository codes**

I've updated my owin / web api 2 project to .NET 6 and I'm getting a Sequence contains more than one matching element at following:

```cs
internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
                        .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
```

https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs#L44-L45

### Assemblies affected

OData WebApi lib 7.5.8

### Reproduce steps

Run following in .NET 6 console app:

```cs
typeof(Enum).GetMethods()
                        .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2)
```

### Expected result

TryParse method should get reflected.

### Actual result

Invalid operation exception

### Additional detail

In .NET 5 and older versions, we only have TryParse method which accepts a ```string``` value.
In .NET 6, a new overload has been added which accepts ```ReadOnlySpan<char>```

habbes:
Hi @ysmoradi thanks for reporting this. Would you be willing to contribute a PR to fix this? We'd be happy to review your PR and support you in the process.

ysmoradi:
Thanks, @habbes
I've linked PR https://github.com/OData/WebApi/pull/2520

Treit:
@habbes I am hitting this issue attempting to update some projects to .NET 6. The problem I am having is that the newest NuGet package for this project does not support .NET Standard 2.0.

We have a mixed codebase that consists of both .NET Framework and .NET Core projects and the code that is using the Microsoft.AspNetCore.OData package (and hitting this bug) is a .NET Standard 2.0 project that is shared by both .NET Framework and .NET Core projects.

Is it possible to backport this fix to a version of the NuGet package that support .NET Standard 2.0?

Treit:
I just discovered that the 7.5.12 version of the package, which is .NET Standard compatible, appears to have this fix so I am unblocked. Thanks!

PR #2520, opened 2021-07-05
Fixed Enum.Try parse issue with .NET 6

ysmoradi:
### Issues

*This pull request fixes issue #2518.*

### Description

Changes the way OData finds Enum.TryParse using reflection which works against both .NET 6 and older versions

### Additional work necessary

N/A

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
@@ -30,10 +30,7 @@ namespace Microsoft.AspNet.OData.Formatter
         Justification = "Relies on many ODataLib classes.")]
     public static class ODataModelBinderConverter
     {
-#if NETSTANDARD2_0
-        internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
-            .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
-#else
+#if NETCOREAPP3_1
         // .NET 6 adds a new overload: TryParse<TEnum>(ReadOnlySpan<Char>, TEnum)
         // Now, with `TryParse<TEnum>(String, TEnum)`, there will have two versions with two parameters
         // So, the previous Single() will throw exception.
@@ -43,6 +40,9 @@ namespace Microsoft.AspNet.OData.Formatter
                 typeof(string),
                 Type.MakeGenericMethodParameter(0).MakeByRefType()
             });
+#else
+        internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
+            .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
 #endif
 
         private static readonly MethodInfo CastMethodInfo = typeof(Enumerable).GetMethod("Cast");
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNet.OData.Formatter
     public static class ODataModelBinderConverter
     {
         private static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
-            .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
+                        .Single(m => m.Name == "TryParse" && m.IsGenericMethod && m.GetParameters().Length == 2 && m.GetParameters().First().ParameterType == typeof(string));
 
         private static readonly MethodInfo CastMethodInfo = typeof(Enumerable).GetMethod("Cast");
 
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
@@ -30,8 +30,20 @@ namespace Microsoft.AspNet.OData.Formatter
         Justification = "Relies on many ODataLib classes.")]
     public static class ODataModelBinderConverter
     {
-        private static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
-                        .Single(m => m.Name == "TryParse" && m.IsGenericMethod && m.GetParameters().Length == 2 && m.GetParameters().First().ParameterType == typeof(string));
+#if NETSTANDARD2_0
+        internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
+            .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
+#else
+        // .NET 6 adds a new overload: TryParse<TEnum>(ReadOnlySpan<Char>, TEnum)
+        // Now, with `TryParse<TEnum>(String, TEnum)`, there will have two versions with two parameters
+        // So, the previous Single() will throw exception.
+        internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethod("TryParse",
+            new[]
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
```diff
@@ -30,8 +30,20 @@ namespace Microsoft.AspNet.OData.Formatter
...
+#else
+        // .NET 6 adds a new overload: TryParse<TEnum>(ReadOnlySpan<Char>, TEnum)
+        // Now, with `TryParse<TEnum>(String, TEnum)`, there will have two versions with two parameters
+        // So, the previous Single() will throw exception.
+        internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethod("TryParse",
+            new[]
+            {
+                typeof(string),
+                Type.MakeGenericMethodParameter(0).MakeByRefType()
+            });
```
habbes, line 45: Would there be a problem if you used this method regardless of the .net version, i.e. (without the `#if NETSTANDARD2_0`)?

habbes: Actually, I've seen that it uses methods not supported in netstandard2.0

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
```diff
@@ -30,8 +30,20 @@ namespace Microsoft.AspNet.OData.Formatter
         Justification = "Relies on many ODataLib classes.")]
     public static class ODataModelBinderConverter
     {
-        private static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
+#if NETSTANDARD2_0
```
habbes, line 33: I think you should have he `#if` macro checking for new versions. The solution is currently failing to build.

So maybe `#if NETSTANDARD2_1` or something specific to NET 5.x and above. Then the "legacy" implementation would remain in the `else` block.

ysmoradi: Web API OData is not supporting anything older than .NET Standard 2.0
So we've got the following table:
1- .NET frameworks older than 4.6.1 is not supported at all
2- .NET framework 4.6.1+ (Up to .NET framework 4.8) will choose #if NETSTANDARD2_0 part automatically.
3- .NET Core older than .net core 2.0 are not supported at all.
4- .NET Core 2.0 up to .NET Core 3.0 will choose #if NETSTANDARD2_0 part automatically.
5- .NET Core 3.1 will choose #else part
6- .NET 5 and 6 and upcoming versions will choose #else part.

I think everything is alright unless I'm getting something wrong

ysmoradi: Oh, I see the build error. Let me check

src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
```diff
@@ -30,8 +30,20 @@ namespace Microsoft.AspNet.OData.Formatter
         Justification = "Relies on many ODataLib classes.")]
     public static class ODataModelBinderConverter
     {
-        private static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
+#if NETCOREAPP3_1
```
habbes, line 33: Does this also apply to .Net 6?

habbes (approved):

marabooy (approved):

gathogojr:
@ysmoradi @habbes Wouldn't this do the job?
```csharp
typeof(Enum).GetMethods()
  .Single(d1 =>
  d1.Name.Equals("TryParse", StringComparison.Ordinal)
  && d1.GetParameters().Length == 2
  && d1.GetParameters()[0].ParameterType.Equals(typeof(string)))
```
What are the odds of an overload of `TryParse` that takes 2 parameters where the first is a `string` and the second something else other than the `out Enum` parameter ever being introduced?

ysmoradi:
> @ysmoradi @habbes Wouldn't this do the job?
> 
> ```cs
> typeof(Enum).GetMethods()
>   .Single(d1 =>
>   d1.Name.Equals("TryParse", StringComparison.Ordinal)
>   && d1.GetParameters().Length == 2
>   && d1.GetParameters()[0].ParameterType.Equals(typeof(string)))
> ```
> 
> What are the odds of an overload of `TryParse` that takes 2 parameters where the first is a `string` and the second something else other than the `out Enum` parameter ever being introduced?

The first time I wrote the code, I wrote something like what you recommendted.
But in the end, I decided to make it similar to the code which is located in asp.net core odata
See https://github.com/OData/WebApi/pull/2520/commits/86d601379ec368d45b09883b66fc209b7e8f5aaf

ysmoradi:
@gathogojr I fixed the commit link which I was talking about

https://github.com/OData/AspNetCoreOData/commit/ddf493c1b2df5e655ee03821a135a59a2b0d8535

gathogojr:
> @gathogojr I fixed the commit link which I was talking about
> 
> [OData/AspNetCoreOData@ddf493c](https://github.com/OData/AspNetCoreOData/commit/ddf493c1b2df5e655ee03821a135a59a2b0d8535)

@ysmoradi Any reason you can't substitute:
```csharp
internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
    .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
```
with
```csharp
internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethod("TryParse",
    new[]
    {
        typeof(string),
        Type.MakeGenericMethodParameter(0).MakeByRefType()
    });
```
instead of introducing the `#if...#else...#endif` blocks?
The latter will work across all frameworks, no?

ysmoradi:
> > @gathogojr I fixed the commit link which I was talking about
> > [OData/AspNetCoreOData@ddf493c](https://github.com/OData/AspNetCoreOData/commit/ddf493c1b2df5e655ee03821a135a59a2b0d8535)
> 
> @ysmoradi Any reason you can't substitute:
> 
> ```cs
> internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethods()
>     .Single(m => m.Name == "TryParse" && m.GetParameters().Length == 2);
> ```
> 
> with
> 
> ```cs
> internal static readonly MethodInfo EnumTryParseMethod = typeof(Enum).GetMethod("TryParse",
>     new[]
>     {
>         typeof(string),
>         Type.MakeGenericMethodParameter(0).MakeByRefType()
>     });
> ```
> 
> instead of introducing the `#if...#else...#endif` blocks?
> The latter will work across all frameworks, no?

I'm afraid, it doesn't work for .net standard 2.0
```Type.MakeGenericMethodParameter(0).MakeByRefType()``` has been introduced in .net standard 2.1 & netcoreapp3.0+

**Review by KenitoInc — APPROVED** [review]

**Review by gathogojr — APPROVED** [review]

**Comment by ysmoradi** [comment]
This PR won't solve the issue, so I'm closing it.
Microsoft.AspNet.OData is being built using .net framework 4.5
Microsoft.AspNetCore.OData is being built using both .net core 3.1 and .net standard 2.0
If someone uses Microsoft.AspNet.OData using Microsoft.AspNetCore.Owin in .NET 6 asp.net core app, app will load Microsoft.AspNet.OData dll which is built using .net framework 4.5
Based on current #if #else #endif which is something like followings:

```cs
#if netcoreapp3_1
...
#else
...
#endif
```

The else part will be executed which throws the original exception which was described in issue #2518 
So, I'm gonna use @gathogojr 's approach which works fine in all runtimes which are supported by OData at the moment.
It covers asp.net framework apps that target .net framework 4.5+ and it also supports asp.net core apps that target .net core 2.0+ or .net 5+

Issue #2516, opened 2021-06-30
AmbiguousMatchException for properties with new modifier (Fixed in OData/AspNetCoreOData)


Aarabika:
This line throw AmbiguousMatchException 

https://github.com/OData/WebApi/blob/757631faf10b494e55117904196e8aa45d4c764d/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs#L178

**Case** 
```
public class RequestReply
{
        public Guid Id { get; set; }
}

public class ChildRequestReply : RequestReply
{
        public Guid Id { get; set; }
}

public class Request
{
        public Guid Id { get; set; }
        public ICollection<RequestReply> Replies { get; set; }
}

// odata serializer will throw exception
public class ChildRequest : Request
{
        public Guid Id { get; set; }
        public new ICollection<ChildRequestReply> Replies { get; set; }
}
```

Fixed in https://github.com/OData/AspNetCoreOData/pull/203, any opportunities to fix this there and add to 7.x release?

habbes:
Hi @Aarabika thanks for reporting this. I see that you're the one who fixed the issue in 8.x. Would you be willing to port the fix to this version as well? We'll be happy to review your PR.

habbes:
@Aarabika regarding [the fix that you provided in 8.x](https://github.com/OData/AspNetCoreOData/pull/203/files#diff-f80bb1a01968c0256936bad0b2feb0f3376c194be2466793d5c7d1e5d71905e2R173), there's some concern on whether we can rely on `GetProperties(p => p.Name == propertyName).First()` always returning the current class's implementation and not the base class's (assuming both properties are returned by` GetProperties()`). You mention in your PR description that the first property in a collection is always from the current class (i.e. the one the `Type` object refers to) and not the base class. Is this behavior guaranteed?

The [official documentation of the method](https://docs.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=net-5.0#System_Type_GetProperties) states the following

> The [GetProperties](https://docs.microsoft.com/en-us/dotnet/api/system.type.getproperties) method does not return properties in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which properties are returned, because that order varies.

It doesn't mention inheritance hierarchy explicitly, and it seems reasonable to expect methods declared on the type to appear before those hidden in base types, but this might just be an implementation detail. Are you able to point to any documentation that guarantees that the current type's property is listed first?

[...]

PR #2523, opened 2021-07-07
Fix AmbiguousMatchException

Aarabika:
### Issues

This pull request fixes issue #2516 

### Description

Fixed AmbiguousMatchException for properties with new modifier 

- [x] *Test cases added*

**Be noted:** There are 2 build pipelines currently. The [.NET Foundation pipeline](https://dev.azure.com/dotnet/OData/_build?definitionId=172) is failing as a result of discontinued support for .NET 4.5. The [other pipeline](https://identitydivision.visualstudio.com/OData/_build?definitionId=1101) is completing successfully

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -179,13 +179,8 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
             // Expression:  source.Property
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            
-            PropertyInfo propertyInfo = source.Type.GetProperty(propertyName, BindingFlags.DeclaredOnly);
-            if (propertyInfo == null)
-            {
-                propertyInfo = source.Type.GetProperty(propertyName);
-            }
-            
+            PropertyInfo propertyInfo = TypeHelper.GetProperty(source.Type, propertyName);
+
             Expression propertyValue = Expression.Property(source, propertyInfo);
             Type nullablePropertyType = TypeHelper.ToNullable(propertyValue.Type);
             Expression nullablePropertyValue = ExpressionHelpers.ToNullable(propertyValue);
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -179,7 +179,13 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
             // Expression:  source.Property
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            PropertyInfo propertyInfo = source.Type.GetProperty(propertyName);
+            
+            PropertyInfo propertyInfo = source.Type.GetProperty(propertyName, BindingFlags.DeclaredOnly);
+            if (propertyInfo == null)
+            {
+                propertyInfo = source.Type.GetProperty(propertyName);
+            }
+            
             Expression propertyValue = Expression.Property(source, propertyInfo);
             Type nullablePropertyType = TypeHelper.ToNullable(propertyValue.Type);
             Expression nullablePropertyValue = ExpressionHelpers.ToNullable(propertyValue);
--- a/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
@@ -491,5 +491,43 @@ namespace Microsoft.AspNet.OData
 
             return null;
         }
+
+        /// <summary>
+        /// Searches the public property with the specified name from the target type and its base types safely,
+        /// circumventing <see cref="AmbiguousMatchException"/> where a property has been redeclared in a derived type.
+        /// </summary>
+        /// <param name="targetType">The type to search from.</param>
+        /// <param name="propertyName">The name of the public property to get.</param>
+        /// <returns>An object representing the public property with the specified name from the most derived class,
+        /// if found; otherwise, null.</returns>
+        internal static PropertyInfo GetProperty(Type targetType, string propertyName)
+        {
+            PropertyInfo[] properties = targetType.GetProperties().Where(d => d.Name.Equals(propertyName)).ToArray();
+
+            if (properties.Length == 0)
+            {
+                return null;
+            }
+
+            if (properties.Length == 1)
+            {
+                return properties[0];
+            }
+
+            Type declaringType = targetType;
+            PropertyInfo propertyInfo = null;
+
+            do
+            {
+                // Declaring type not expected to be null
+                Contract.Assert(declaringType != null, $"Declaring type for {propertyName} property not expected to be null");
+
+                propertyInfo = properties.FirstOrDefault(d => d.DeclaringType.Equals(declaringType));
+                declaringType = declaringType.BaseType;
+            }
+            while (propertyInfo == null);
+
+            return propertyInfo;
+        }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
@@ -498,36 +498,73 @@ namespace Microsoft.AspNet.OData
         /// </summary>
         /// <param name="targetType">The type to search from.</param>
         /// <param name="propertyName">The name of the public property to get.</param>
-        /// <returns>An object representing the public property with the specified name from the most derived class,
-        /// if found; otherwise, null.</returns>
+        /// <returns>
+        /// An object representing the public property with the specified name from the most derived class,
+        /// if found; otherwise, null.
+        /// </returns>
         internal static PropertyInfo GetProperty(Type targetType, string propertyName)
         {
-            PropertyInfo[] properties = targetType.GetProperties().Where(d => d.Name.Equals(propertyName)).ToArray();
+            // NOTE: Designed to avoid costly closure lambda allocations since method might be used in hot path
+            PropertyInfo matchedProperty = null;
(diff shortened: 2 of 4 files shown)


habbes:
Thanks for the PR @Aarabika. Could you add some test cases for this?

Also, based on some trials, using `BindingFlags.DeclaredOnly` alone will not work. You should combine it at least with `BindingFlags.Public | BindingFlags.NonStatic`.

Could you also take a look at the impementation proposed by @gathogojr (https://github.com/OData/WebApi/issues/2516#issuecomment-875398484). It seems to take care of more edge cases.

KenitoInc:
Hey @Aarabika 
Kindly look into the review comments!

KenitoInc:
Hey @Aarabika 
Kindly look into the review comments!

xuzhg (approved):

gathogojr (changes requested):
@Aarabika Kindly add tests covering different scenarios

marabooy:
@Aarabika any changes or help required to help move this along? I would be happy to see this go into the next release.

Aarabika:
@marabooy sorry, was so busied with the main projects, I will try to rewrite code + add tests at this weekends, ok?

marabooy:
Thanks @Aarabika

Aarabika:
@marabooy If we redefine property in child class, new property will't be described in odata metadata of this child class edm type, and redefine property that has been already defined on the base type is prohibited by spec to. So I don't know what to do next

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -175,7 +175,13 @@ internal Expression CreatePropertyValueExpression(IEdmStructuredType elementType
 
             // Expression:  source.Property
             string propertyName = EdmLibHelpers.GetClrPropertyName(property, _model);
-            PropertyInfo propertyInfo = source.Type.GetProperty(propertyName);
+            
+            PropertyInfo propertyInfo = source.Type.GetProperty(propertyName, BindingFlags.DeclaredOnly);
```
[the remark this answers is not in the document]
gathogojr: Did a test to confirm that this:
```csharp
PropertyInfo propertyInfo = source.Type.GetProperty(propertyName, BindingFlags.DeclaredOnly);
if (propertyInfo == null)
{
    propertyInfo = source.Type.GetProperty(propertyName);
}
```
doesn't work for a deeper inheritance hierarchy
```csharp
[Fact]
public void ThrowsExceptionForDeeperInheritanceHierarchy()
{
    var property = "Category";

    var modelBuilder = new ODataConventionModelBuilder();
    modelBuilder.EntitySet<QueryPentagon>("Shapes");
    var model = modelBuilder.GetEdmModel();

    var queryPentagonEntityType = model.SchemaElements.OfType<IEdmEntityType>().First(c => c.Name == "QueryPentagon");
    var querySubCategoryEntityType = model.SchemaElements.OfType<IEdmEntityType>().First(c => c.Name == "QuerySubCategory");

    var settings = new ODataQuerySettings { HandleNullPropagation = HandleNullPropagationOption.False };
    var context = new ODataQueryContext(model, typeof(QueryPentagon)) { RequestContainer = new MockContainer() };
    var binder = new SelectExpandBinder(settings, context);

    // Arrange
    Expression source = Expression.Constant(new QueryPentagon());

    IEdmNavigationProperty navProperty = queryPentagonEntityType.NavigationProperties().Single(c => c.Name == property);
    Assert.NotNull(navProperty);

    // Act
    Assert.Throws<AmbiguousMatchException>(() => binder.CreatePropertyValueExpression(querySubCategoryEntityType, navProperty, source, null));
}

public class QueryShape
{
    public int Id { get; set; }
    public QueryCategory Category { get; set; }
}

public class QueryPolygon : QueryShape
{
    public new int Id { get; set; }
    public new QuerySubCategory Category { get; set; }
}

public class QueryPentagon : QueryPolygon
{
}

public class QueryCategory
{
    public int Id { get; set; }
}

public class QuerySubCategory : QueryCategory
{
}
```

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -491,5 +491,43 @@ private static Type GetInnerGenericType(Type interfaceType)
...
+        /// Searches the public property with the specified name from the target type and its base types safely,
+        /// circumventing <see cref="AmbiguousMatchException"/> where a property has been redeclared in a derived type.
+        /// </summary>
+        /// <param name="targetType">The type to search from.</param>
+        /// <param name="propertyName">The name of the public property to get.</param>
+        /// <returns>An object representing the public property with the specified name from the most derived class,
+        /// if found; otherwise, null.</returns>
+        internal static PropertyInfo GetProperty(Type targetType, string propertyName)
+        {
+            PropertyInfo[] properties = targetType.GetProperties().Where(d => d.Name.Equals(propertyName)).ToArray();
```
habbes, line 505: nitpick: Is this method used in a ho path? If so consider using using plain loops to avoid the closure lambda allocations.

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -491,5 +491,43 @@ private static Type GetInnerGenericType(Type interfaceType)
...
+
+            do
+            {
+                // Declaring type not expected to be null
+                Contract.Assert(declaringType != null, $"Declaring type for {propertyName} property not expected to be null");
+
+                propertyInfo = properties.FirstOrDefault(d => d.DeclaringType.Equals(declaringType));
+                declaringType = declaringType.BaseType;
+            }
+            while (propertyInfo == null);
```
habbes, line 528: question: Is it guaranteed that you will find the declaring type in the hierarchy and `propertyInfo` will be non-null before `declaringType.Basetype` is null?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -1881,6 +1881,66 @@ public void CorrelatedSubqueryIncludesToListIfBufferingOptimizationIsTrueAndPage
...
+            var propertyName = "Category";
+            IEdmNavigationProperty navProperty = queryPentagonEntityType.NavigationProperties().Single(c => c.Name == propertyName);
+            Assert.NotNull(navProperty);
+
+            // Act
+            Expression propertyValue = binder.CreatePropertyValueExpression(querySubCategoryEntityType, navProperty, source, null);
+
+            // Assert
+            Assert.Equal(ExpressionType.MemberAccess, propertyValue.NodeType);
+            Assert.Equal(TypeHelper.GetProperty(typeof(QueryPentagon), propertyName), (propertyValue as MemberExpression).Member);
```
habbes, line 1940: Here you're verifying that the property type is `QuerySubCategory` and not `QueryCategory` since that's the type of `QueryPolygon.Category`?

habbes (approved):

KenitoInc (approved):

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
gathogojr: Fixed by rewriting the method for better performance and added unit tests

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandBinderTest.cs
gathogojr: Correct

gathogojr (approved):
:shipit:

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -491,5 +491,75 @@ private static Type GetInnerGenericType(Type interfaceType)
...
+        /// </returns>
+        internal static PropertyInfo GetProperty(Type targetType, string propertyName)
+        {
+            // NOTE: Designed to avoid costly closure lambda allocations since method might be used in hot path
+            List<PropertyInfo> matchedProperties = null;
+
+            PropertyInfo[] properties = targetType.GetProperties();
+            for (int i = 0; i < properties.Length; i++)
+            {
+                PropertyInfo propertyInfo = properties[i];
```
habbes, line 513: Since `properties` is an array, I think using `foreach` is better since it will be compiled down to the same thing.

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
```diff
@@ -491,5 +491,75 @@ private static Type GetInnerGenericType(Type interfaceType)
...
+                // Property declared on target type - opportunity to exit early
+                if (propertyInfo.DeclaringType.Equals(targetType))
+                {
+                    return propertyInfo;
+                }
+
+                // We defer initialization of the list since we (arguably) don't get this far
+                if (matchedProperties == null)
+                {
+                    matchedProperties = new List<PropertyInfo>(4);
```
habbes, line 529: nitpick: This is probably overkill. I think in the majority of the cases, I think only one property will be returned. In that case maybe we can avoid creating the list and store it in a variable and only create the list when the second property is found. But I don't know if that is worth the code complexity. In any case, not a blocker.

habbes (approved):

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
gathogojr: @habbes Since the for loop in this case is fairly straightforward, also owing to the fact that not everyone will be familiar with the compiler optimization, I'm inclined towards leave it be. I have however implemented the optimization fix such that we don't initialize the list unless 2 or more properties are matched

src/Microsoft.AspNet.OData.Shared/TypeHelper.cs
gathogojr: Fixed. See my comment [here](https://github.com/OData/WebApi/pull/2523#discussion_r809903958)

PR #2524, opened 2021-07-08
PR For Bulk Operation and ODataBind Review for  Beta Release

Sreejithpin:
### Issues

PR For Bulk Operation and ODataBind Review for  Beta Release

### Description

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs
```diff
@@ -42,6 +42,13 @@ internal static string SelectActionImpl(ODataPath odataPath, IWebApiControllerCo
                         "Post" + entitySet.EntityType().Name,
                         "Post");
                 }
+                else if (ODataRequestMethod.Patch == controllerContext.Request.Method)
+                {
+                    // e.g. Try PatchCustomer first, then fall back to Patch action name
```
xuzhg, line 47: >PatchCustomer [](http://example.com/codeflow?start=32&length=13)

PatchCustomers

xuzhg:
>                         "PostFrom" + entityType.Name);

do we support patch from cast?

PATCH ~/Customers/NS.VipCustomer


---
Refers to: src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs:83 in fbe081a. [](commit_id = fbe081ab32b0399f5fb5bbcc0f3fce7fe4c1538f, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerHelper.cs
```diff
@@ -0,0 +1,111 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using Microsoft.AspNet.OData.Builder;
+using Microsoft.OData;
+using Microsoft.OData.Edm;
+using Org.OData.Core.V1;
+using System;
+using System.Collections.Generic;
+using System.Text;
```
xuzhg, line 10: sort

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataSerializerHelper.cs
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Microsoft.AspNet.OData.Formatter.Serialization
+{
+    /// <summary>
+    /// Helper class for OData Serialization
+    /// </summary>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1053:StaticHolderTypesShouldNotHaveConstructors")]
+    public class ODataSerializerHelper
```
xuzhg, line 18: >ODataSerializerHelper [](http://example.com/codeflow?start=17&length=21)

if it's an internal class (I noticed no public API in this class), why don't mark the class as interal, and mark API as public.

In current design, we have a public class, but nothing in the public class from customer perspetive.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceWrapper.cs
```diff
@@ -10,20 +12,27 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
...
+        /// <summary>
+        /// Gets the wrapped <see cref="ODataResourceBase"/>.
+        /// </summary>
+        public ODataResourceBase ResourceBase {get;}
 
         /// <summary>
-        /// Gets the wrapped <see cref="ODataResource"/>.
+        /// Gets the wrapped <see cref="ODataResource"/>. This will return null for deleted resources.
         /// </summary>
+        [Obsolete("Please use ResourceBase instead")]
```
xuzhg, line 35: >[Obsolete("Please use ResourceBase instead")] [](http://example.com/codeflow?start=8&length=45)

This obsolete seems unnecessary.

Sreejithpin: The idea was to make it obselete but not to remove, right @mikepizzo . but I guess in 8.0 its removed right?

Sreejithpin:
>                         "PostFrom" + entityType.Name);

Normal Patch (Delta<T>) supports inheritance/casting right?

---
In reply to: [880162857](https://github.com/OData/WebApi/pull/2524#issuecomment-880162857) [](http://example.com/codeflow?ancestors=880162857,880162857)

---
Refers to: src/Microsoft.AspNet.OData.Shared/Routing/Conventions/EntitySetRoutingConvention.cs:83 in fbe081a. [](commit_id = fbe081ab32b0399f5fb5bbcc0f3fce7fe4c1538f, deletion_comment = False)

src/Microsoft.AspNet.OData.Shared/EdmPatchMethodHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.OData
+{    
+   
+    /// <summary>
+    /// Handler Class to handle users methods for create, delete and update
+    /// </summary>
+    public abstract class EdmPatchMethodHandler
```
mikepizzo, line 17: >EdmPatchMethodHandler [](http://example.com/codeflow?start=26&length=21)

Since this is not just used for Patch, should we consider renaming to something like "UpdateHandler" or "DataModificationHandler"?

src/Microsoft.AspNet.OData.Shared/PatchMethodHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// Failure Status
+        /// </summary>
+        Failure,
+        /// <summary>
+        /// Resource Not Found
+        /// </summary>
+        NotFound
+    }
+
```
mikepizzo, line 80: This should be in a separate file

src/Microsoft.AspNet.OData.Shared/PatchMethodHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.OData
+{    
+    /// <summary>
+    /// Base Interface for PatchMethodHandler
+    /// </summary>
+    public interface IPatchMethodHandler
```
mikepizzo, line 14: >IPatchMethodHandler [](http://example.com/codeflow?start=21&length=19)

I'm sure I asked this before, but why do we need an empty marker interface?  Can you add to the comments so that it's clear to others in the future?

Also, this should be in a separate file so that it is easy to find.

src/Microsoft.AspNet.OData.Shared/DefaultPatchHandler.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.OData
+{    
+    internal class DefaultPatchHandler<TStructuralType> : PatchMethodHandler<TStructuralType> where TStructuralType :class
```
mikepizzo, line 11: >DefaultPatchHandler [](http://example.com/codeflow?start=19&length=19)

Add comments describing this class, its behavior, and where/when it is appropriate to use.

src/Microsoft.AspNet.OData.Shared/EdmPatchMethodHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.OData
+{    
+   
+    /// <summary>
+    /// Handler Class to handle users methods for create, delete and update
```
mikepizzo, line 15: >Handler Class to handle users methods for create, delete and update [](http://example.com/codeflow?start=8&length=67)

Add that this is for the case where there is no CLR type -- PatchMethodHandler is used when there is a clr type.

src/Microsoft.AspNet.OData.Shared/PatchMethodHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+
+    }
+
+    /// <summary>
+    /// Handler Class to handle users methods for create, delete and update
```
mikepizzo, line 20: >Handler Class to handle users methods for create, delete and update [](http://example.com/codeflow?start=8&length=67)

add to the description that this is for the clr-backed objects; EdmPatchMethodHandler is used for the non-CLR backed case.

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceWrapper.cs
mikepizzo: In 8.0 the Resource property should be typed as ODataResourceBase.

src/Microsoft.AspNet.OData.Shared/IDeltaSet.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Basic interface to reperesent a deltaset which is a collection of Deltas.
```
mikepizzo, line 11: >Basic interface to reperesent a deltaset which is a collection of Deltas. [](http://example.com/codeflow?start=8&length=73)

Please describe why we have this marker interface with no members.

src/Microsoft.AspNet.OData.Shared/DeltaSetOfT.cs
```diff
@@ -0,0 +1,328 @@
...
+            return CopyChangedValues(patchHandler);
+        }
+
+        /// <summary>
+        /// Patch for DeltaSet, a collection for Delta<typeparamref name="TStructuralType"/>
+        /// </summary>     
+        /// <returns>DeltaSet response</returns>
+        public DeltaSet<TStructuralType> Patch(IPatchMethodHandler patchHandler)
+        {            
+            return CopyChangedValues(patchHandler as PatchMethodHandler<TStructuralType>);
```
mikepizzo, line 71: >patchHandler as PatchMethodHandler<TStructuralType> [](http://example.com/codeflow?start=37&length=51)

what happens if this cast fails?

PR #2530, opened 2021-07-22
Fix bug in Filter Any within Expand

KenitoInc:
### Issues

*This pull request fixes issue https://github.com/OData/AspNetCoreOData/issues/233*

### Description

Issue was raised in the WebAPI v8 repo. We will fix in this repo and that one too.

Generally in Select and Expand binder in ODL, we add a `$this` range variable. When binding AnyNode within expand/select in WebApi, we create parameter expression for range variables which are not `$it`. We also need to ignore the `$this` range variables. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
@@ -940,7 +940,9 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             foreach (RangeVariable rangeVariable in rangeVariables)
             {
                 ParameterExpression parameter;
-                if (!_lambdaParameters.TryGetValue(rangeVariable.Name, out parameter))
+
+                // Create a Parameter Expression when for rangeVariables which are not $it Lambda parameters or $this.
+                if (!_lambdaParameters.TryGetValue(rangeVariable.Name, out parameter) && rangeVariable.Name != ODataThisParameterName)
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs
```diff
@@ -940,7 +940,9 @@ private ParameterExpression HandleLambdaParameters(IEnumerable<RangeVariable> ra
             foreach (RangeVariable rangeVariable in rangeVariables)
             {
                 ParameterExpression parameter;
-                if (!_lambdaParameters.TryGetValue(rangeVariable.Name, out parameter))
+
+                // Create a Parameter Expression for rangeVariables which are not $it Lambda parameters or $this.
+                if (!_lambdaParameters.TryGetValue(rangeVariable.Name, out parameter) && rangeVariable.Name != ODataThisParameterName)
```
gathogojr, line 945: In the comment you talk about creating parameter expression for range variables that are neither $it lambda parameters or $this but your change only added `&& rangeVariable.Name != ODataThisParameterName`. Will that suffice?

KenitoInc: This code `!_lambdaParameters.TryGetValue(rangeVariable.Name, out parameter)` ensures we don't create Parameter expressions for lambda parameters

gathogojr (approved):
LGTM

xuzhg (approved):

PR #2536, opened 2021-08-03
Cache EDMType so that we can avoid costly linq O(n) lookup.

HarieSrinivasa:
### Issues

When we get the EDM type we look at all the schema elements of the model. This lookup is not only costly but also scales with O(n). Since, we use OData API for our service we have a lot of schema elements in the model and we would benefit from this scaling better. In our CPU profiling we spent close to 12% of our on CPU cycles in this method.

image

### Description

In this change we are caching the result of this method and returning the cached value in future invocations. 

I did perf testing with unit tests where I added 96 complex types to model.
Then fetched first and last object 10000 times.

Old code ran in the range 350-400 ms.
New code ran in the range 5-12 ms.

- [ ] *Test cases added* Not sure what tests to add here.

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 #if NETFX // System.Data.Linq.Binary is only supported in the AspNet version.
 using System.Data.Linq;
@@ -30,6 +31,8 @@ namespace Microsoft.AspNet.OData.Formatter
     internal static class EdmLibHelpers
     {
         private static readonly EdmCoreModel _coreModel = EdmCoreModel.Instance;
+
+        private static readonly ConcurrentDictionary<IEdmModel, ConcurrentDictionary<Type, IEdmType>> _cache = new ConcurrentDictionary<IEdmModel, ConcurrentDictionary<Type, IEdmType>>();
 
         private static readonly Dictionary<Type, IEdmPrimitiveType> _builtInTypesMapping =
             new[]
@@ -161,6 +164,18 @@ namespace Microsoft.AspNet.OData.Formatter
                     clrType = underlyingType;
                 }
 
+                if (_cache.TryGetValue(edmModel, out var cachedClrTypes))
+                {
+                    if (cachedClrTypes.TryGetValue(clrType, out var cachedType))
+                    {
+                        return cachedType;
+                    }
+                }
+                else
+                {
+                    _cache.TryAdd(edmModel, new ConcurrentDictionary<Type, IEdmType>());
+                }
+
                 // search for the ClrTypeAnnotation and return it if present
                 IEdmType returnType =
                     edmModel
@@ -179,6 +194,8 @@ namespace Microsoft.AspNet.OData.Formatter
                     // go up the inheritance tree to see if we have a mapping defined for the base type.
                     returnType = returnType ?? GetEdmType(edmModel, TypeHelper.GetBaseType(clrType), testCollections);
                 }
+
+                _cache[edmModel].TryAdd(clrType, returnType);
                 return returnType;
             }
         }
@@ -735,7 +752,7 @@ namespace Microsoft.AspNet.OData.Formatter
             if (edmModel == null)
             {
                 throw Error.ArgumentNull("edmModel");
-            }         
+            }
 
             DynamicPropertyDictionaryAnnotation annotation =
                 edmModel.GetAnnotationValue<DynamicPropertyDictionaryAnnotation>(edmType);
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
@@ -164,9 +164,9 @@ namespace Microsoft.AspNet.OData.Formatter
                     clrType = underlyingType;
                 }
 
-                if (_cache.TryGetValue(edmModel, out var cachedClrTypes))
+                if (_cache.TryGetValue(edmModel, out ConcurrentDictionary<Type, IEdmType> cachedClrTypes))
                 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -161,6 +164,18 @@ private static IEdmType GetEdmType(IEdmModel edmModel, Type clrType, bool testCo
                     clrType = underlyingType;
                 }
 
+                if (_cache.TryGetValue(edmModel, out var cachedClrTypes))
```
Sreejithpin, line 167: Please use ConcurrentDictionary<Type, IEdmType> instead of var

HarieSrinivasa: ack!

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -161,6 +164,18 @@ private static IEdmType GetEdmType(IEdmModel edmModel, Type clrType, bool testCo
...
+                {
+                    if (cachedClrTypes.TryGetValue(clrType, out var cachedType))
```
Sreejithpin, line 169: Please use IEdmType instead of var. We usually try to avoid using var in our libs

HarieSrinivasa: will replace all var with solid types.

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
HarieSrinivasa: Done.

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
HarieSrinivasa: Done.

Sreejithpin (commented):
Can you please also check the memory , is it increasing heavily since we are putting the edmmodels to the cache

HarieSrinivasa:
> Can you please also check the memory , is it increasing heavily since we are putting the edmmodels to the cache

Thank you Sreejith. The memory should not increase drastically because
1. The Model.SchemaElements are already in memory.
2. These are object references so they only have pointers in the ConcurrentDictionary and that should be ok.

HarieSrinivasa:
> > Can you please also check the memory , is it increasing heavily since we are putting the edmmodels to the cache
> 
> Thank you Sreejith. The memory should not increase drastically because
> 
> 1. The Model.SchemaElements are already in memory.
> 2. These are object references so they only have pointers in the ConcurrentDictionary and that should be ok.

I can however measure memory increase.

HarieSrinivasa:
> Can you please also check the memory , is it increasing heavily since we are putting the edmmodels to the cache

Old code memory used 66 MB for 30 of the 96 classes and 10,000 fetch.
image

New code memory used was 64 MB with same scenario
image

Sreejithpin (approved):

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -31,6 +32,8 @@ internal static class EdmLibHelpers
     {
         private static readonly EdmCoreModel _coreModel = EdmCoreModel.Instance;
 
+        private static readonly ConcurrentDictionary<IEdmModel, ConcurrentDictionary<Type, IEdmType>> _cache = new ConcurrentDictionary<IEdmModel, ConcurrentDictionary<Type, IEdmType>>();
```
xuzhg, line 35: >ConcurrentDictionary<IEdmModel, ConcurrentDictionary<Type, IEdmType>> [](http://example.com/codeflow?start=32&length=69)

We already have a Cache for the mapping between CLR Type and Edm Type. So, I don't understand why do we need another cache.

You can find cache :

[https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs)

Would you please share me more usage about the GetEdmType(...)?

And By the way, it's recommended to use 
[https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs#L22](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs#L22)

and 
[https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs#L34](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs#L34)

to get CLR type or Edm Type.

xuzhg (changes requested):
:clock1:

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
HarieSrinivasa: Sorry I am not very familiar with OData library. While looking at my service which is hosted using OData library I noticed that the GetEdmType was taking up 12% of our CPU.

image


I see that the version we are using does not have the cache the new version of the code is seems to be doing the right caching. Thanks for pointing it out.

mikepizzo:
If I understand @HarieSrinivasa's latest reply to @xuzhg -- this is already addressed in the latest code.  Is that correct?  If so, can we close this issue (with thanks to @HarieSrinivasa for the excellent investigation, and in hopes of further contributions in the future)?

HarieSrinivasa:
> If I understand @HarieSrinivasa's latest reply to @xuzhg -- this is already addressed in the latest code. Is that correct? If so, can we close this issue (with thanks to @HarieSrinivasa for the excellent investigation, and in hopes of further contributions in the future)?

Yes we can close this. Thank you.

mikepizzo:
Thank you @HarieSrinivasa!  Hope to hear more from you in the future!

Issue #2532, opened 2021-07-25
Unmapped properties are in Delta<T>


dxrdxr:
In my OData convention based EDM model I have properties that I exclude from the model by using the `[NotMapped]` attribute.

In an entity that has a property with `[NotMapped]`, when I call the `Patch` method I see that the property is in the `_allProperties` and is accessible (although it is always empty).

### Assemblies affected

OData WebApi and Microsoft.AspNetCore.OData

### Reproduce steps

- Create a Class with a `Collection` property:

```
public class DiveGroup
{
    public int DiveGroupID { get; set; }
    public ICollection<DiveGroupMember> DiveGroupMembers { get; } = new Collection<DiveGroupMember>();
    [NotMapped]
    public IList<Diver> Assistants => DiveGroupMembers.Where(dg => dg.Role == MemberRole.Assistant).Select(dgm => dgm.Diver)
} 
```
- Build the EDM Model and examine it, note that `Assistants` is not there.
- Build a OData controller with a `Patch` method
- Send a `PATCH` request using PostMan
- Examine the `Delta<DiveGroup>` that is passed as an argument. Note that in `_allProperties` the property `Assistants` exists

Correct behavior: `Assistants` should not be in  `_allProperties`

### Root Cause

In the code [DeltaOfTStructuralType.cs](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs) method `InitializeProperties` the  `_allProperties` is set:

```
    _allProperties = _propertyCache.GetOrAdd(
        _structuredType,
        (backingType) => backingType
            .GetProperties(BindingFlags.Instance | BindingFlags.Public)
            .Where(p => (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
            .Select<PropertyInfo, PropertyAccessor<TStructuralType>>(p => new FastPropertyAccessor<TStructuralType>(p))
            .ToDictionary(p => p.Property.Name));
```

The `Where()` clause is incomplete, there should be an additional check such as (proposed):

```
            .Where(p => !NotMappedAttributeSet(p) && (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
```

PR #2544, opened 2021-08-13
Fix to ignore Unmapped properties in Delta<T>

Sreejithpin:
### Issues

This pull request fixes issue #2532 

### Description

Fix to ignore Unmapped properties in Delta<T>

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -612,6 +612,11 @@ namespace Microsoft.AspNet.OData
 
         private bool IsIgnoredProperty(bool isTypeDataContract, PropertyInfo propertyInfo)
         {
+            //This is for Ignoring the property that matches below criteria
+            //1. Its marked as NotMapped
+            //2. Its a datacontract type but property is not marked as datamember
+            //3. Its marked with IgnoreDataMember (but not where types datacontract and property marked with datamember)
+
             bool hasNotMappedAttr = propertyInfo.GetCustomAttributes(typeof(NotMappedAttribute), inherit: true).Any();
 
             if (hasNotMappedAttr)
@@ -619,14 +624,12 @@ namespace Microsoft.AspNet.OData
                 return true;
             }
 
-            bool ignoreDataMember = propertyInfo.GetCustomAttributes(typeof(IgnoreDataMemberAttribute), inherit: true).Any();
-
-            if (ignoreDataMember)
+            if (isTypeDataContract)
             {
-                return !(isTypeDataContract && propertyInfo.GetCustomAttributes(typeof(DataMemberAttribute), inherit: true).Any());
+                return !propertyInfo.GetCustomAttributes(typeof(DataMemberAttribute), inherit: true).Any();
             }
 
-            return false;
+            return propertyInfo.GetCustomAttributes(typeof(IgnoreDataMemberAttribute), inherit: true).Any();           
         }
 
         // Copy changed dynamic properties and leave the unchanged dynamic properties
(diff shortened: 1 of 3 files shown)


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/TestCommon/Models/TestDataModels.cs
```diff
@@ -54,4 +57,26 @@ public class AutoExpandChoiceOrder
         public int Id { get; set; }
         public string Name { get; set; }
     }
+
+    public class NewCustomerUnmapped
```
mikepizzo, line 61: >NewCustomerUnmapped [](http://example.com/codeflow?start=17&length=19)

maybe add a [DataMember] attribute to a property that also has [IgnoreDataMember] and validate that property is serialized if [DataContract] is not applied to the type.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/TestCommon/Models/TestDataModels.cs
```diff
@@ -54,4 +57,26 @@ public class AutoExpandChoiceOrder
...
+    }
+
+    [DataContract]
+    public class NewCustomerDataContract
+    {
+        [Key]
+        public int Id { get; set; }
+        [IgnoreDataMember]
+        [DataMember]
+        public string Name { get; set; }
```
mikepizzo, line 78: Just to cover all of the possibilities, should probably have a property with just the [DataMember] attribute and a property with no attributes.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -519,6 +520,25 @@ private void InitializeProperties(IEnumerable<string> updatableProperties)
             }
         }
 
+        private bool IsIgnoredProperty(bool isTypeDataContract, PropertyInfo propertyInfo)
+        {
```
mikepizzo, line 524: Add a comment to explain the logic here.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -519,6 +520,25 @@ private void InitializeProperties(IEnumerable<string> updatableProperties)
             }
         }
 
+        private bool IsIgnoredProperty(bool isTypeDataContract, PropertyInfo propertyInfo)
```
mikepizzo, line 523: >isTypeDataContract [](http://example.com/codeflow?start=44&length=18)

If the type is marked with [DataContract], but the property is not marked with [DataMember], do we include or ignore it?

mikepizzo (changes requested):
:clock1:

mikepizzo (approved):
:shipit:

mikepizzo (changes requested):
:clock1:

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/DeltaTest.cs
```diff
@@ -185,6 +185,38 @@ public void CanPatch_OpenType()
...
+        public void TestDelta_IgnoredMember()
+        {
+            //Arrange
+            var delta = new Delta<NewCustomerDataContract>();
+
+            //Act
+            var properties = delta.GetUnchangedPropertyNames().ToList();
+
+            //Assert
+            Assert.Equal(2, properties.Count);
```
mikepizzo, line 214: >            Assert.Equal(2, properties.Count);   [](http://example.com/codeflow?start=0&length=48)

shouldn't "Id" be returned?

KenitoInc (approved):
LGTM

mikepizzo (approved):
:shipit:

PR #2559, opened 2021-09-01
Fix routing error with attribute routing when action parameters don't match route keys

### LINKED ISSUE #2534: Action implementations with fewer parameters than route now fail
In PR https://github.com/OData/WebApi/pull/2168 an assumption was introduced that an action implementation is only relevant if it contains at least the number of parameters provided by the route. That is an invalid assumption.

It's perfectly valid for an action to contain only a single parameter, but have it derived from any number of parameters from the route. We do this a lot to coalesce various forms of identity into a singular `ResourceHandle`:

```
[ODataRoute("/drives({driveId}/items/{itemId}/oneDrive.getByPath(path={pathKey})")]
public Task<IHttpActionResult> GetItemAsync([ModelBinder] IResourceHandle resourceHandle)
{
  ...
}
```

The implementation of the `IModelBinder` that handles `IResourceHandle` internally pulls out the various route data using `bindingContext.ValueProvider.GetValue(name)`. So all route data is consumed, but the action implementation only has a single parameter.

This issue will block us from upgrading from `7.5.1` (which didn't have the issue since `AttributeRoutingConvention` was not incrementing the key count).

### Assemblies affected

`Microsoft.AspNet.OData` >= `7.5.4`

### Reproduce steps

1. Create a model binder that consumes multiple pieces of route data
2. Use this model binder on a method that is bound to a path that generates multiple pieces of route data
3. Make a request

### Expected result

The action is invoked

### Actual result

The action is not invoked, and the request fails.

habbes:
### Issues

* Fixes #2534 

### Description

In brief this PR makes the following changes:
- Selects action annotated with `[ODataRoute]` regardless of its parameters (by passing the `MethodInfo` to the action selector)
- Ensures that the action with `[ODataRoute]` is selected in case of other candidate overloads

Below I share more context behind the changes in this PR:

### Context behind the changes

The `ODataActionSelector` has become strict when deciding which controller action will handle the request. In particular, it tries to ensure that the parameters of the controller action match the keys in the route/URL. The reason for this is to disambiguate between different overloads and to avoid sending the request to the wrong overload.

However, the increase in "strictness" over a number of PR has led some scenarios to fail, specifically when the user intentionally creates controller actions whose parameters don't match the keys in the route. This PR addresses this issue by ignoring the parameter matching constraints when the action was selected via attribute routing (i.e. using `ODataRoute` attribute).

The main issue that has made `ODataActionSelector` complex and brittle is the fact that it doesn't receive enough information to disambiguate between overloads in a simple way. Since attribute routing can figure out exactly which method to route to before it even reaches the action selector, I've pass this method (via its `MethodInfo` object) down to the `ODataActionSelector` so it doesn't have to do much additional work, and to make it more resilient to edge cases.

With this change in place, we can guarantee that when using `ODataRoute` attribute, the action will be selected regardless of its parameters. We assume that the user knows exactly what they want and we don't try to verify whether the parameters make sense.

This is however not enough to allow the action selector to disambiguate between overloads. The routing conventions return a list of potential action descriptors, which are usually overloads with the same name. This is also the case with `AttributeRoutingConvention`: even if the `ODataRoute` attribute is applied to a single action, the routing convention will return all actions with the same name. Let's take the following example:

```c#
public class CustomersController
{
   [ODataRoute("Customers({custId})/Products({prodId})"]
   public IActionResult GetProduct([ModelBinder] RequestData data)  { /*  ... */ }

   public IActionResult GetProduct(int custId, int prodId) { /*  ... */ }
}
```
The request `GET /Customers(10)/Products(2)` should go to the `GetProduct(RequestData)` method because it's explicitly annotated with `ODataRoute`. 

However what would happen is that the `AttributeRoutingConvention` will indeed match `GetProduct(RequestData)` action with the route. But it will return a list containing both actions. This is due to the design of routing conventions and the way it tries to reuse some logic between AspNet classic and AspNetCore: most routing conventions have 2 partial classes, one is framework agnostic that returns only the name of the selection (as a string) (via the `SelectActionImpl` method) and the other is a framework-specific class that takes this action name received from the framework-agnostic class and returns a list of framework-specific action descriptors that match that name.

Now `ODataActionSelector` will receive both actions and have to figure out which to return (this happens inside the `SelectBestCandidate` method). Before this PR, it would return the action `GetProduct(custId, prodId)` because it has more parameters and its parameters more closely match the route keys. But I've added a step to prioritize actions that have `[ODataRoute]` over actions that do not.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
@@ -59,6 +59,9 @@ namespace Microsoft.AspNet.OData.Routing.Conventions
                 if (action.IsHttpMethodSupported(request.GetRequestMethodOrPreflightMethod()) && template.TryMatch(odataPath, values))
                 {
                     values["action"] = action.ActionName;
+                    // store the method info to allow the action selector to distinguish
+                    // between this action and other overloads
+                    values[ODataRouteConstants.MethodInfo] = action.GetMethodInfo();
                     SelectControllerResult result = new SelectControllerResult(action.ControllerName, values);
 
                     return result;
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -249,7 +249,6 @@ namespace Microsoft.AspNet.OData.Routing
                     }
                 }
 
-
                 // if action has [EnableNestedPaths] attribute, then it doesn't
                 // need to match parameters, since this action is expected to
                 // match arbitrarily nested paths even if it doesn't have any parameters
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
@@ -129,7 +129,8 @@ namespace Microsoft.AspNet.OData.Routing
                 IList<string> availableKeys = routeData.Values.Keys
                     .Where((key) => !RoutingConventionHelpers.IsRouteParameter(key)
                         && key != ODataRouteConstants.Action
-                        && key != ODataRouteConstants.ODataPath)
+                        && key != ODataRouteConstants.ODataPath
+                        && key != ODataRouteConstants.MethodInfo)
                     .Select(k => k.ToLowerInvariant())
                     .ToList();
 
@@ -176,9 +177,18 @@ namespace Microsoft.AspNet.OData.Routing
                 // if there are still multiple candidate actions at this point, let's try some tie-breakers
                 if (matchedCandidates.Count > 1)
                 {
+                    // prioritize actions with explicit [ODataRoute] (i.e. attribute routing)
+                    ActionIdAndParameters bestCandidate = matchedCandidates.FirstOrDefault(candidate =>
+                        candidate.ActionDescriptor is ControllerActionDescriptor action
+                        && action.MethodInfo.GetCustomAttributes<ODataRouteAttribute>().Any());
+                    if (bestCandidate != null)
+                    {
+                        return bestCandidate.ActionDescriptor;
+                    }
+
                     // prioritize actions which explicitly declare the request method
                     // e.g. using [AcceptVerbs("POST")], [HttpPost], etc.
-                    ActionIdAndParameters bestCandidate = matchedCandidates.FirstOrDefault(candidate =>
+                    bestCandidate = matchedCandidates.FirstOrDefault(candidate =>
                         ActionAcceptsMethod(candidate.ActionDescriptor as ControllerActionDescriptor, context.HttpContext.Request.Method));
                     if (bestCandidate != null)
                     {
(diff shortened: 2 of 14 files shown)


xuzhg (approved):
:shipit:

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -226,11 +236,23 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+                // the method match the keys in the route
+                if (context.RouteData.Values.TryGetValue(ODataRouteConstants.MethodInfo, out object method) && method is MethodInfo methodInfo)
+                {
+                    if (controllerActionDescriptor.MethodInfo == methodInfo)
+                    {
+                        return true;
+                    }
+                }
+
+
```
gathogojr, line 252:
```suggestion
```

src/Microsoft.AspNetCore.OData/Routing/ODataActionSelector.cs
```diff
@@ -176,9 +177,18 @@ public ActionDescriptor SelectBestCandidate(RouteContext context, IReadOnlyList<
...
+                    // prioritize actions with explicit [ODataRoute] (i.e. attribute routing)
+                    ActionIdAndParameters bestCandidate = matchedCandidates.FirstOrDefault(candidate =>
+                        candidate.ActionDescriptor is ControllerActionDescriptor action
+                        && action.MethodInfo.GetCustomAttributes<ODataRouteAttribute>().Any());
+                    if (bestCandidate != null)
+                    {
+                        return bestCandidate.ActionDescriptor;
+                    }
+
                     // prioritize actions which explicitly declare the request method
```
gathogojr, line 189:
```suggestion
                    // Next in priority, actions which explicitly declare the request method
```

src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs
```diff
@@ -59,6 +59,9 @@ private AttributeRoutingConvention(string routeName)
                 if (action.IsHttpMethodSupported(request.GetRequestMethodOrPreflightMethod()) && template.TryMatch(odataPath, values))
                 {
                     values["action"] = action.ActionName;
+                    // store the method info to allow the action selector to distinguish
+                    // between this action and other overloads
+                    values[ODataRouteConstants.MethodInfo] = action.GetMethodInfo();
```
gathogojr, line 64: What if instead of storing an entire `MethodInfo` object, you stored a boolean value with a value of true if the method is decorated with `ODataRouteAttribute`; otherwise false?
I'm also curious about the existing method `public IEnumerable<T> GetCustomAttributes<T>(bool inherit) where T : Attribute`. Does it return the attributes on the action method?
If the existing `GetCustomAttribute` method can't do the job, you could add a method or property in the `WebApiActionDescriptor` class that returns true if the action method contains `ODataRouteAttribute`

habbes: @gathogojr you can tell whether a method is decorated with `ODataRouteAttribute` by using the `actionDescriptor.GetCustomAttributes` method. The problem is that just relying on the `ODataRoute` attribute is not sufficient. I actually tried that first, and it causes tests to fail. The problem is that you can have two different overloads with `ODataRouteAttribute`, e.g. `Get()` and `Get(int key)`, or realistic, actions handling OData functions with the same name but different parameters. In this case the action selector would need to match parameters to know which action to select. But because the attribute routing has already done the heavy lifting, we could just use the exact method that it narrowed down to, which is what I was trying to achieve by passing the method info from the `AttributeRoutingConvention` to the action selector.

habbes: Just to be clear, I'm not storing the `MethodInfo` so that I can access its attributes, I'm storing the `MethodInfo` so that I can compare it directly to method info instances in the action selector to find a match. I guess the alternative would to compare two parameters, to check whether two methods have the same sequence of parameter types.

gathogojr (approved):

PR #2565, opened 2021-09-07
Consider the OperationSegment when determining the TargetStructuredType

ificator:
### Issues

Fixes https://github.com/OData/WebApi/issues/1610

### Description

If you have an auto-expanded property on an object:
```
public class A
{
  [AutoExpand]
  public B B { get; set; }
}
```
And a function `F` bound to `A` that returns type `C`, the auto-expand logic incorrectly attempts to expand `B` on `C` if a (any) query string is provided:
```
GET .../A/F?a=b

The query specified in the URI is not valid. Could not find a property named 'B' on type 'C'.
```
Since the fix appeared to be pretty straight forward (make sure `OperationSegment` is handled in the path when trying to determine the `TargetStructuredType`) I've decided to push this PR.

NOTE: I'm also fixing a couple of bonus issues
1. There's a bug whereby the existing `EnableQuery_Works_WithAutoExpand` test wasn't validating auto-expand because `EnableQueryCategory` was not being registered as an `EntityType`.
2. VisualStudio complained about the string interpolation in `EdmTypeCacheItemComparer.GetHashCode`, but methods used for comparisons (e.g. `Equals` and `GetHashCode`) should NOT allocate so I rewrote it to use a hash combining algorithm instead.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
@@ -71,8 +71,15 @@ namespace Microsoft.AspNet.OData.Formatter
 
             public int GetHashCode(EdmTypeCacheItem obj)
             {
-                string combined = $"{obj.EdmType.FullTypeName()}~{obj.Nullable}";
-                return combined.GetHashCode();
+                unchecked
+                {
+                    int hashCode = 17;
+                    hashCode = (hashCode * 31) + obj.EdmType.FullTypeName().GetHashCode(); // Ideally we'd avoid FullTypeName since it
+                                                                                           // allocates internally, but we cannot trust
+                                                                                           // EdmType.GetHashCode().
+                    hashCode = (hashCode * 31) + (obj.Nullable ? 1 : 0);
+                    return hashCode;
+                }
             }
         }
     }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
@@ -74,10 +74,8 @@ namespace Microsoft.AspNet.OData.Formatter
                 unchecked
                 {
                     int hashCode = 17;
-                    hashCode = (hashCode * 31) + obj.EdmType.FullTypeName().GetHashCode(); // Ideally we'd avoid FullTypeName since it
-                                                                                           // allocates internally, but we cannot trust
-                                                                                           // EdmType.GetHashCode().
-                    hashCode = (hashCode * 31) + (obj.Nullable ? 1 : 0);
+                    hashCode = (hashCode * 31) + obj.EdmType.GetHashCode();
+                    hashCode = (hashCode * 31) + obj.Nullable.GetHashCode();
                     return hashCode;
                 }
             }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
@@ -637,15 +637,16 @@ namespace Microsoft.AspNet.OData.Formatter
         {
             property = null;
             structuredType = null;
-            name = String.Empty;
-            string typeCast = String.Empty;
+            name = string.Empty;
+
             if (segments != null)
             {
+                string typeCast = string.Empty;
+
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
```diff
@@ -71,8 +71,15 @@ public bool Equals(EdmTypeCacheItem x, EdmTypeCacheItem y)
 
             public int GetHashCode(EdmTypeCacheItem obj)
             {
-                string combined = $"{obj.EdmType.FullTypeName()}~{obj.Nullable}";
-                return combined.GetHashCode();
+                unchecked
+                {
+                    int hashCode = 17;
+                    hashCode = (hashCode * 31) + obj.EdmType.FullTypeName().GetHashCode(); // Ideally we'd avoid FullTypeName since it
```
ificator, line 77: Or can we trust it? the `Equals` implementation assumes that `Equals` is implemented, which it isn't, so that would suggest we're dealing with instance equality...

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -654,6 +657,19 @@ public static IEdmType GetElementType(IEdmTypeReference edmTypeReference)
                         return;
                     }
 
+                    if (segment is OperationSegment operationSegment)
+                    {
+                        IEdmFunction function = operationSegment.Operations.SingleOrDefault() as IEdmFunction;
```
ificator, line 662: Shouldn't be limited to functions

ificator: I've fixed this locally, but I'll hold off pushing a new iteration until we've resolved my previous question re: `GetHashCode`

src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
ificator: I've convinced myself that the default `GetHashCode` implementation will work for us.

ificator:
@KenitoInc / @gathogojr could you take a look at this PR?

src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
```diff
@@ -71,8 +71,13 @@ public bool Equals(EdmTypeCacheItem x, EdmTypeCacheItem y)
 
             public int GetHashCode(EdmTypeCacheItem obj)
             {
-                string combined = $"{obj.EdmType.FullTypeName()}~{obj.Nullable}";
-                return combined.GetHashCode();
+                unchecked
+                {
+                    int hashCode = 17;
+                    hashCode = (hashCode * 31) + obj.EdmType.GetHashCode();
+                    hashCode = (hashCode * 31) + obj.Nullable.GetHashCode();
```
gathogojr, line 78: @ificator I'm not sure I understand the logic here... What value does line 77 add if the assigned value is overwritten here?

gathogojr:
Hi @ificator. Thank you for your contribution. Looks great. I agree with you that ideally methods used for comparisons should not allocate but I've left some comments around the changes you introduced to `GetHashCode` method.

src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
gathogojr: argh, never mind. I just realized that the value is not exactly being overwritten

gathogojr (approved):
LGTM

KenitoInc (approved):

src/Microsoft.AspNet.OData.Shared/Formatter/ClrTypeCache.cs
ificator: Yeah, this is a somewhat standard algorithm for generating hashcodes that's taken from "Effective Java" (https://stackoverflow.com/questions/11537005/why-chose-31-to-do-the-multiplication-in-the-hashcode-implementation). I became less concerned with my change once I found the unit tests for the cache and they were all passing.

mikepizzo (approved):
:shipit:

ificator:
So who gets to push the merge button? 😁

xuzhg (approved):
:shipit:

ificator:
Thanks everyone!

mikepizzo:
Thank you Brad!

PR #2567, opened 2021-09-09
PR For Bulk Operation and ODataBind Review for Beta Release

Sreejithpin:
### Issues
Used instead of PR #2524 
PR For Bulk Operation and ODataBind Review for Beta Release

### Description

PR For Bulk Operation and ODataBind Review for Beta Release

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.AspNet.OData.Shared/EdmPatchMethodHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.OData
+{    
+   
+    /// <summary>
+    /// Handler Class to handle users methods for create, delete and update.
+    /// This is the handler for data modification where there is no CLR type.
+    /// </summary>
+    public abstract class EdmODataAPIHandler
```
mikepizzo, line 18: >EdmODataAPIHandler [](http://example.com/codeflow?start=26&length=18)

please rename file to match class name

src/Microsoft.AspNet.OData.Shared/DefaultEdmPatchMethodHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Linq;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// This is the default Patch Handler for non CLR type. This calss has default Get, Create and Update
+    /// and will do these actions. This will be used when the original collection to be Patched is provided.
+    /// </summary>
+    internal class DefaultEdmODataAPIHandler : EdmODataAPIHandler
```
mikepizzo, line 16: >DefaultEdmODataAPIHandler [](http://example.com/codeflow?start=19&length=25)

Do we really have both a DefaultEdmPatchMethodHandler and a DefaultODataAPIHandler?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Collections.ObjectModel;
+using System.Linq;
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Navigation Path of an OData ID
+    /// </summary>
+    public class NavigationPath
```
mikepizzo, line 15: >NavigationPath [](http://example.com/codeflow?start=17&length=14)

Maybe call it "ODataNavigationPath" for consistency?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        private string navigationPathName;
+        private ReadOnlyCollection<ODataPathSegment> _pathSegments;
+        private ConcurrentDictionary<string, PathItem[]> _pathItemCache = new ConcurrentDictionary<string, PathItem[]>();
+
+        /// <summary>
+        /// Constructor which takes and odataId and creates PathItems
```
mikepizzo, line 22: >which takes and odataId and creates PathItems [](http://example.com/codeflow?start=24&length=45)

please update description

src/Microsoft.AspNet.OData.Shared/PathItem.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// Name of the Path Item (eg: entity name, entity set name)
+        /// </summary>
+        public string Name { internal set; get; }
+
+        /// <summary>
+        /// Whether the PathItem is a cast segment
+        /// </summary>
+        public bool IsCastType { internal set; get; }
```
mikepizzo, line 21: >IsCastType [](http://example.com/codeflow?start=20&length=10)

Maybe have a PathItemType that is "castSegment", "propertySegment", "navigationPropertySegment"

src/Microsoft.AspNet.OData.Shared/PathItem.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// If the item is a cast segment, Name of the Path Item (eg: derived entity name, entity set name)
+        /// </summary>
+        public string CastTypeName { internal set; get; }
+
+        /// <summary>
+        /// List of Key properties of that entity
+        /// </summary>
+        public Dictionary<string, object> KeyProperties { internal set; get; }
```
mikepizzo, line 31: >KeyProperties [](http://example.com/codeflow?start=42&length=13)

Should we define a derived type with keyProperties or just have null when not relevant?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// </summary>
+        /// <param name="navigationPath">ODataId in string format</param>
+        /// <param name="pathSegments">Pathsegment collection</param>
+        public NavigationPath(string navigationPath, ReadOnlyCollection<ODataPathSegment> pathSegments)
+        {
+            navigationPathName = navigationPath;
+            _pathSegments = pathSegments;           
+        }
```
mikepizzo, line 38: why do we have a public constructor that takes both string and parsed form? when would a developer use this? When would we?

src/Microsoft.AspNet.OData.Shared/DefaultEdmPatchMethodHandler.cs
Sreejithpin: No we dont have any DefaultEdmPatchMethodHandler 
 we have 2 defaults, 1 for typed another for typeles

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
xuzhg: ODL has ODataExpandPath and ODataSelectPath, does it can meet your requirement?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
xuzhg: Check other class:

mainly use the sentence like: "Initializes a new instance of the <see cref="NavigationPath"/> class."

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
```
xuzhg, line 3: Update the head since we moved to .NET Foundation

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Constructor which takes and odataId and creates PathItems
+        /// </summary>
+        /// <param name="navigationPath">ODataId in string format</param>
+        /// <param name="pathSegments">Pathsegment collection</param>
+        public NavigationPath(string navigationPath, ReadOnlyCollection<ODataPathSegment> pathSegments)
+        {
+            navigationPathName = navigationPath;
+            _pathSegments = pathSegments;
```
xuzhg, line 37: Need the paramater null check

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Navigation Path of an OData ID
+    /// </summary>
+    public class NavigationPath
+    {
+        private string navigationPathName;
```
xuzhg, line 17: >navigationPathName [](http://example.com/codeflow?start=23&length=18)

code style.

Other fields using '_',  why "navigationPathName" without '_'.

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+        private ConcurrentDictionary<string, PathItem[]> _pathItemCache = new ConcurrentDictionary<string, PathItem[]>();
+
+        /// <summary>
+        /// Constructor which takes and odataId and creates PathItems
+        /// </summary>
+        public NavigationPath()
+        {
+           
+        }
+
```
xuzhg, line 28: Why do we have the default constructor?
And if we do have this, if customers use it, _pathSegments is not initializated?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                if (_pathSegments != null)
+                {
+                    pathItems = ParseODataId();
+                    _pathItemCache.TryAdd(navigationPathName, pathItems);
+                }
+            }
+
+            return pathItems;
+        }
```
xuzhg, line 63: This method looks werid. 

It looks to retrieve a PathItem[] using a path name, but, the path name is private field? It's not from "outside".  Straigh?

In this case, why do we need a ConcurrencyDictionary?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <param name="pathSegments">Pathsegment collection</param>
+        public NavigationPath(string navigationPath, ReadOnlyCollection<ODataPathSegment> pathSegments)
+        {
+            navigationPathName = navigationPath;
+            _pathSegments = pathSegments;           
+        }
+
+        
+        /// <summary>
+        /// NavigationPath/ODataId in string
```
xuzhg, line 42: >NavigationPath/ODataId in string [](http://example.com/codeflow?start=12&length=32)

Code style.

Following other class comments.  It's better:

/// <summary>
/// Gets the navigation path name.
/// </summary>

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+        private PathItem[] ParseODataId()
+        {
+            List<PathItem> pathItems = new List<PathItem>();
+            PathItem currentPathItem = null;
+
+            foreach (ODataPathSegment segment in _pathSegments)
+            {
+                if (segment is EntitySetSegment || segment is NavigationPropertySegment)
+                {
+                    pathItems.Add(new PathItem());
```
xuzhg, line 74: >pathItems.Add(new PathItem()); [](http://example.com/codeflow?start=20&length=30)

why donot we consider "ComplexProperty" and "TypeCase"?

samples/AspNetODataSample.Web/Web.config
```diff
@@ -4,85 +4,89 @@
...
+			<dependentAssembly>
+				<assemblyIdentity name="System.Buffers" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+				<bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
+			</dependentAssembly>
+			<dependentAssembly>
+				<assemblyIdentity name="System.ValueTuple" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+				<bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
+			</dependentAssembly>
+			<dependentAssembly>
+				<assemblyIdentity name="Microsoft.Spatial" publicKeyToken="31bf3856ad364e35" culture="neutral" />
```
xuzhg, line 71: it seems we don't need this redirect?

If no, why don't add OData.Edm, and OData.Core redirect?

samples/AspNetODataSample.Web/packages.config
```diff
@@ -15,9 +15,9 @@
   <package id="Microsoft.Bcl.AsyncInterfaces" version="1.0.0" targetFramework="net461" />
   <package id="Microsoft.CodeDom.Providers.DotNetCompilerPlatform" version="1.0.0" targetFramework="net461" />
   <package id="Microsoft.Net.Compilers" version="1.0.0" targetFramework="net461" developmentDependency="true" />
-  <package id="Microsoft.OData.Core" version="7.9.0" targetFramework="net461" />
```
xuzhg, line 18: any reason to update the dependency version

src/Microsoft.AspNet.OData.Shared/DataModificationExceptionType.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <summary>
+        ///  Initializes a new instance of the <see cref="DataModificationExceptionType"/> class.
+        /// </summary>
+        public DataModificationExceptionType(DataModificationOperationKind failedOperation)
+        {
+            this.FailedOperation = failedOperation;
+        }
+
+        /// <summary>
+        /// Represents king of <see cref="DataModificationOperationKind"/> type of operation
```
xuzhg, line 65: >king [](http://example.com/codeflow?start=23&length=4)

kind

src/Microsoft.AspNet.OData.Shared/DataModificationExceptionType.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        /// <summary>
+        /// Represents king of <see cref="DataModificationOperationKind"/> type of operation 
+        /// </summary>        
+        public DataModificationOperationKind FailedOperation { get; }
+
+        /// <summary>
+        /// Represents response code
+        /// </summary>        
+        public Int16 ResponseCode { get; set; }
```
xuzhg, line 72: >Int16 [](http://example.com/codeflow?start=15&length=5)

any reason to use Int16?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            return pathItems;
+        }
+
+        private PathItem[] ParseODataId()
+        {
+            List<PathItem> pathItems = new List<PathItem>();
+            PathItem currentPathItem = null;
+
+            foreach (ODataPathSegment segment in _pathSegments)
```
mikepizzo, line 70: >_pathSegments [](http://example.com/codeflow?start=49&length=13)

what if _pathSegments is null?

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+    public class NavigationPath
+    {
+        private string navigationPathName;
+        private ReadOnlyCollection<ODataPathSegment> _pathSegments;
+        private ConcurrentDictionary<string, PathItem[]> _pathItemCache = new ConcurrentDictionary<string, PathItem[]>();
+
+        /// <summary>
+        /// Constructor which takes and odataId and creates PathItems
+        /// </summary>
+        public NavigationPath()
```
mikepizzo, line 24: >NavigationPath [](http://example.com/codeflow?start=15&length=14)

I would have expected a single constructor that took the navigationPathName and parsed it into pathSegments in order construct the PathItem[].  If the navigationPathName could not be parsed it would return null for GetNavigationPathItems, and the consuming logic would have to fall back to the navigationPathName.

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
```diff
@@ -0,0 +1,98 @@
...
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Navigation Path of an OData ID
+    /// </summary>
+    public class NavigationPath
+    {
+        private string navigationPathName;
+        private ReadOnlyCollection<ODataPathSegment> _pathSegments;
+        private ConcurrentDictionary<string, PathItem[]> _pathItemCache = new ConcurrentDictionary<string, PathItem[]>();
```
mikepizzo, line 19: >_pathItemCache [](http://example.com/codeflow?start=57&length=14)

maybe something to take as a future work item, but we would have a much higher cache hit ratio if we cached the path without the key.

mikepizzo, line 19: >_pathItemCache [](http://example.com/codeflow?start=57&length=14)

The _pathItemCache is per instance of NavigationPath, which should only ever have one path, so we're not really saving anything by having a concurrent dictionary, are we?

src/Microsoft.AspNet.OData.Shared/DefaultEdmPatchMethodHandler.cs
mikepizzo: I think I was confused because the name of the file was still "DefaultEdmPatchMethodHandler"

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
mikepizzo: +1

src/Microsoft.AspNet.OData.Shared/PathItem.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Whether the PathItem is a cast segment
+        /// </summary>
+        public bool IsCastType { internal set; get; }
+
+        /// <summary>
+        /// If the item is a cast segment, Name of the Path Item (eg: derived entity name, entity set name)
+        /// </summary>
+        public string CastTypeName { internal set; get; }
```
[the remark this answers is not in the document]
Sreejithpin: It would have different values, one generic and another the cast type name think it could be used by developer for generic case and then specific cast case, eg: Name = NewFriends, CastTypeName = Microsoft.Test.E2E.AspNet.OData.BulkInsert.MyNewFriend

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
Sreejithpin: The pathitems is already parsed in the deserializer by calling pathhandler.parse (for odata bind its already written by elizabeth) so wanteed to reuse it and not call it again , thats why used both in constructor

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
Sreejithpin: The pathitems is already parsed in the deserializer by calling pathhandler.parse (for odata bind its already written by elizabeth) so wanteed to reuse it and not call it again , thats why used both in constructor

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
Sreejithpin: This was introduced to handle multiple set of path items in the same path like customers(1)/Orders(1)/shipment(1) and customers(1)/Orders(1)/Address(1). so cache customes(1)/orders(1) not to call again . But since we are making seperate navigation paths for each object  it wouldnt be ideal to keep a cache. Might be in later versions we can consider a centralizzed caching if we see such scenarios, even then we should think of heavy cache allocations . So removing it for now

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
Sreejithpin: This was introduced to handle multiple set of path items in the same path like customers(1)/Orders(1)/shipment(1) and customers(1)/Orders(1)/Address(1). so cache customes(1)/orders(1) not to call again . But since we are making seperate navigation paths for each object  it wouldnt be ideal to keep a cache. Might be in later versions we can consider a centralizzed caching if we see such scenarios, even then we should think of heavy cache allocations . So removing it for now

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
Sreejithpin: This was introduced to handle multiple set of path items in the same path like customers(1)/Orders(1)/shipment(1) and customers(1)/Orders(1)/Address(1). so cache customes(1)/orders(1) not to call again . But since we are making seperate navigation paths for each object  it wouldnt be ideal to keep a cache. Might be in later versions we can consider a centralizzed caching if we see such scenarios, even then we should think of heavy cache allocations . So removing it for now

src/Microsoft.AspNet.OData.Shared/NavigationPath.cs
Sreejithpin: Added check in constructor and in calling method

samples/AspNetODataSample.Web/packages.config
Sreejithpin: There was a bug which is fixed in 7.9.1 with out which async would throw error in some scenarios

src/Microsoft.AspNet.OData.Shared/DataModificationExceptionType.cs
Sreejithpin: It was taken crom vocabs and its like that so used same.

PR #2575, opened 2021-09-23
Fixing Patching Complex type when using one derived type over another

### LINKED ISSUE #2572: PATCH doesn't work for nested derived complex types
When patching from one derived type to another derived type for the same base type, it shows errors `Cannot use Delta of type 'derivedType1' on an entity of type 'derivedType2'.`

### Assemblies affected

OData WebApi lib 7.3.0

### Reproduce steps

- Odata schema
```
<EntityType Name="anEntity">
     <Property Name="property1" Type="complexType1">
</EntityType>
<ComplexType Name="complexType1">
    <Property Name="property2" Type="complexType2">
</ComplexType>
<ComplexType Name="complexType2"
    <Property Name="property3" Type="complexType3">
</ComplexType>
<ComplexType Name="complexType3">
     <Property Name="propertyBase" Type="someBaseType">
</ComplexType>
<ComplexType Name="someBaseType" Abstract="true" />
<ComplexType Name="derivedType1" BaseType="someBaseType" />
<ComplexType Name="derivedType2"  BaseType="someBaseType" >
    <Property Name="propertyDerived" Type="Collection(Edm.String)>
</ComplexType>
```

- Existing anEntity

```
{
    "property1": {
        "property2": {
             "property3": {
                  "propertyBase": {
                       "@odata.type": "derivedType1"
                  }
             }
         }
     }
}
```

- Patch call to change `propertyBase`

```
{
    "property1": {
        "property2": {
             "property3": {
                  "propertyBase": {
                       "@odata.type": "derivedType2",
                       "propertyDerived": [ "some string" ]
                  }
             }
         }
     }
}
```

- Error 
`Cannot use Delta of type 'derivedType1' on an entity of type 'derivedType2'`

### Expected result

The patch should work and the `propertyBase` should be updated to `derivedType2`

### Actual result

`Cannot use Delta of type 'derivedType1' on an entity of type 'derivedType2'`

### Additional detail

Some fixes that I have already tried -
- Update Odata library to 7.4.1 => Same error
- Changed `delta.CopyChangedValues` to `delta.Patch` => same error
- Change Path request body from 
```
{
    "property1": {
        "property2": {
             "property3": {
                  "propertyBase": {
                       "@odata.type": "derivedType2",
                       "propertyDerived": [ "some string" ]
                  }
             }
         }
     }
}
```
to
```
{
       "propertyBase": {
             "@odata.type": "derivedType2",
             "propertyDerived": [ "some string" ]
      }
}
```
The delta of the patch is `null`

Sreejithpin:
### Issues

*This pull request fixes #2572 .*

### Description

This PR helps in unblocking of using one derived type to Patch over another derived complex type in the case where a common ancestor is used in declaration

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -45,11 +45,6 @@ namespace Microsoft.AspNet.OData
         private IDictionary<string, object> _dynamicDictionaryCache;
 
         /// <summary>
-        /// To determine if the StructuralType is a Complex type
-        /// </summary>
-        public bool IsComplexType { get; private set; }
-
-        /// <summary>
         /// Initializes a new instance of <see cref="Delta{TStructuralType}"/>.
         /// </summary>
         public Delta()
@@ -135,6 +130,11 @@ namespace Microsoft.AspNet.OData
         /// considered to be changed.</remarks>
         public IList<string> UpdatableProperties
             => _updatableProperties;
+
+        /// <summary>
+        /// To determine if the StructuralType is a Complex type
+        /// </summary>
+        public bool IsComplexType { get; private set; }
 
         /// <inheritdoc/>
         public override void Clear()
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -44,7 +44,10 @@ namespace Microsoft.AspNet.OData
         private HashSet<string> _changedDynamicProperties;
         private IDictionary<string, object> _dynamicDictionaryCache;
 
-        internal bool IsComplexType { get; set; }
+        /// <summary>
+        /// To determine if the StructuralType is a Complex type
+        /// </summary>
+        public bool IsComplexType { get; private set; }
 
         /// <summary>
         /// Initializes a new instance of <see cref="Delta{TStructuralType}"/>.
@@ -355,16 +358,15 @@ namespace Microsoft.AspNet.OData
                 }
                 else
                 {
-                    // Recursively patch the subtree.
-                    bool isDeltaType = TypedDelta.IsDeltaOfT(deltaNestedResource.GetType());
-                    Contract.Assert(isDeltaType, nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
+                    // Recursively patch the subtree.                    
+                    Contract.Assert(TypedDelta.IsDeltaOfT(((object)deltaNestedResource).GetType()), nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
 
-                    Type nestedType = deltaNestedResource.StructuredType;
+                    Type newType = deltaNestedResource.StructuredType;
                     Type originalType = originalNestedResource.GetType();
 
-                    if (deltaNestedResource.IsComplexType && !nestedType.IsAssignableFrom(originalType))
+                    if (deltaNestedResource.IsComplexType && !newType.IsAssignableFrom(originalType))
                     {
-                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, nestedType, originalType);
+                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, newType, originalType, deltaNestedResource.ExpectedClrType);
                     }
 
                     deltaNestedResource.CopyChangedValues(originalNestedResource);
@@ -372,57 +374,53 @@ namespace Microsoft.AspNet.OData
             }
         }
 
-        private dynamic ReAssignComplexDerivedType(TStructuralType original, string nestedResourceName, dynamic originalNestedResource, Type nestedType, Type originalType)
+        private dynamic ReAssignComplexDerivedType(TStructuralType parent, string nestedPropertyName, dynamic originalValue, Type newType, Type originalType, Type declaredType)
         {
             //As per OASIS discussion, changing a complex type from 1 derived type to another is allowed if both derived type have a common ancestor and the property
             //is declared in terms of a common ancestor. The logic below checks for a common ancestor. Create a new object of the derived type in delta request.
             //And copy the common properties.
 
-            Type complexBaseType = nestedType.BaseType;
-            HashSet<Type> lstComplexBaseTypes = new HashSet<Type>();
+            Type newBaseType = newType.BaseType;
+            HashSet<Type> newBaseTypes = new HashSet<Type>();
 
-            while (complexBaseType != null)
+            //Iterate till you find the declaring base type and add all that to hashset
+            while (newBaseType != null && newBaseType != declaredType)
             {
-                lstComplexBaseTypes.Add(complexBaseType);
-                complexBaseType = complexBaseType.BaseType;
+                newBaseTypes.Add(newBaseType);
+                newBaseType = newBaseType.BaseType;
             }
+
+            newBaseTypes.Add(declaredType);
 
             //Here original type is the type for original (T) resource.
             //We will keep  going to base types and finally will get the Common Basetype for the derived complex types in to the originalType variable.
-                        
-            bool foundCommonbase = false;
-
+            
             //The new Original type, means the new complex type (T) which will replace the current complex type.
             dynamic newOriginalNestedResource = null;
 
             while (originalType != null)
             {
-                if (lstComplexBaseTypes.Contains(originalType))
+                if (newBaseTypes.Contains(originalType))
                 {
                     //Now originalType = common base type of the derived complex types.
                     //OriginalNested Resource = T(of current Complex type). We are creating newOriginalNestedResource (T - new complex type).
-                    newOriginalNestedResource = Activator.CreateInstance(nestedType);
+                    newOriginalNestedResource = Activator.CreateInstance(newType);
 
                     //Here we get all the properties of common base type and get value from original complex type(T) and 
                     //copy it to the new complex type newOriginalNestedResource(came as a part of Delta) 
 
                     foreach (PropertyInfo property in originalType.GetProperties())
                     {
-                        object value = property.GetValue(originalNestedResource);
+                        object value = property.GetValue(originalValue);
                         property.SetValue(newOriginalNestedResource, value);
                     }
 
-                    foundCommonbase = true;
+                    _structuredType.GetProperty(nestedPropertyName).SetValue(parent, (object)newOriginalNestedResource);
 
                     break;
                 }
 
                 originalType = originalType.BaseType;
-            }
-
-            if (foundCommonbase)
-            {
-                _structuredType.GetProperty(nestedResourceName).SetValue(original, (object)newOriginalNestedResource);
             }
 
             return newOriginalNestedResource;
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -407,6 +407,8 @@ namespace Microsoft.AspNet.OData
 
                     break;
                 }
+
+                originalBaseType = originalBaseType.BaseType;
             }
 
             if (foundCommonbase)
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -44,6 +44,8 @@ namespace Microsoft.AspNet.OData
         private HashSet<string> _changedDynamicProperties;
         private IDictionary<string, object> _dynamicDictionaryCache;
 
+        internal bool IsComplexType { get; set; }
+
         /// <summary>
         /// Initializes a new instance of <see cref="Delta{TStructuralType}"/>.
         /// </summary>
@@ -88,11 +90,32 @@ namespace Microsoft.AspNet.OData
         /// properties. <c>null</c> means this entity type is not open.</param>
         public Delta(Type structuralType, IEnumerable<string> updatableProperties,
             PropertyInfo dynamicDictionaryPropertyInfo)
+            : this(structuralType, updatableProperties: updatableProperties, dynamicDictionaryPropertyInfo, false)
+        {         
+        }
+
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="Delta{TStructuralType}"/>.
+        /// </summary>
+        /// <param name="structuralType">The derived entity type or complex type for which the changes would be tracked.
+        /// <paramref name="structuralType"/> should be assignable to instances of <typeparamref name="TStructuralType"/>.
+        /// </param>
+        /// <param name="updatableProperties">The set of properties that can be updated or reset. Unknown property
+        /// names, including those of dynamic properties, are ignored.</param>
+        /// <param name="dynamicDictionaryPropertyInfo">The property info that is used as dictionary of dynamic
+        /// properties. <c>null</c> means this entity type is not open.</param>
+        /// <param name="isComplexType">Boolean value to determine if its a complex type</param>
+        public Delta(Type structuralType, IEnumerable<string> updatableProperties,
+            PropertyInfo dynamicDictionaryPropertyInfo, bool isComplexType)
         {
             _dynamicDictionaryPropertyinfo = dynamicDictionaryPropertyInfo;
             Reset(structuralType);
             InitializeProperties(updatableProperties);
+            IsComplexType = isComplexType;
         }
+
+
 
         /// <inheritdoc/>
         public override Type StructuredType
@@ -336,9 +359,62 @@ namespace Microsoft.AspNet.OData
                     bool isDeltaType = TypedDelta.IsDeltaOfT(deltaNestedResource.GetType());
                     Contract.Assert(isDeltaType, nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
 
+                    Type nestedType = deltaNestedResource.StructuredType;
+                    Type originalType = originalNestedResource.GetType();
+
+                    if (deltaNestedResource.IsComplexType && !nestedType.IsAssignableFrom(originalType))
+                    {
+                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, nestedType, originalType);
+                    }
+
                     deltaNestedResource.CopyChangedValues(originalNestedResource);
                 }
             }
+        }
+
+        private dynamic ReAssignComplexDerivedType(TStructuralType original, string nestedResourceName, dynamic originalNestedResource, Type nestedType, Type originalType)
+        {
+            //As per OASIS discussion, changing a complex type from 1 derived type to another is allowed if both derived type have a common ancestor and the property
+            //is declared in terms of a common ancestor. The logic below checks for a common ancestor. Create a new object of the derived type in delta request.
+            //And copy the common properties.
+
+            Type complexBaseType = nestedType.BaseType;
+            HashSet<Type> lst = new HashSet<Type>();
+
+            while (complexBaseType != null)
+            {
+                lst.Add(complexBaseType);
+                complexBaseType = complexBaseType.BaseType;
+            }
+
+            Type originalBaseType = originalType.BaseType;
+            bool foundCommonbase = false;
+
+            while (originalBaseType != null)
+            {
+                if (lst.Contains(originalBaseType))
+                {
+                    dynamic newOriginal = originalNestedResource;
+                    originalNestedResource = Activator.CreateInstance(nestedType);
+
+                    foreach (PropertyInfo property in originalBaseType.GetProperties())
+                    {
+                        object value = property.GetValue(newOriginal);
+                        property.SetValue(originalNestedResource, value);
+                    }
+
+                    foundCommonbase = true;
+
+                    break;
+                }
+            }
+
+            if (foundCommonbase)
+            {
+                _structuredType.GetProperty(nestedResourceName).SetValue(original, (object)originalNestedResource);
+            }
+
+            return originalNestedResource;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -396,7 +396,7 @@ namespace Microsoft.AspNet.OData
             //We will keep  going to base types and finally will get the Common Basetype for the derived complex types in to the originalType variable.
             
             //The new Original type, means the new complex type (T) which will replace the current complex type.
-            dynamic newOriginalNestedResource = null;
+            dynamic newOriginalNestedResource = originalValue;
 
             while (originalType != null)
             {
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -379,28 +379,37 @@ namespace Microsoft.AspNet.OData
             //And copy the common properties.
 
             Type complexBaseType = nestedType.BaseType;
-            HashSet<Type> lst = new HashSet<Type>();
+            HashSet<Type> lstComplexBaseTypes = new HashSet<Type>();
 
             while (complexBaseType != null)
             {
-                lst.Add(complexBaseType);
+                lstComplexBaseTypes.Add(complexBaseType);
                 complexBaseType = complexBaseType.BaseType;
             }
 
-            Type originalBaseType = originalType.BaseType;
+            //Here original type is the type for original (T) resource.
+            //We will keep  going to base types and finally will get the Common Basetype for the derived complex types in to the originalType variable.
+                        
             bool foundCommonbase = false;
 
-            while (originalBaseType != null)
-            {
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -88,12 +90,33 @@ public Delta(Type structuralType, IEnumerable<string> updatableProperties)
...
+        /// <param name="structuralType">The derived entity type or complex type for which the changes would be tracked.
+        /// <paramref name="structuralType"/> should be assignable to instances of <typeparamref name="TStructuralType"/>.
+        /// </param>
+        /// <param name="updatableProperties">The set of properties that can be updated or reset. Unknown property
+        /// names, including those of dynamic properties, are ignored.</param>
+        /// <param name="dynamicDictionaryPropertyInfo">The property info that is used as dictionary of dynamic
+        /// properties. <c>null</c> means this entity type is not open.</param>
+        /// <param name="isComplexType">Boolean value to determine if its a complex type</param>
+        public Delta(Type structuralType, IEnumerable<string> updatableProperties,
+            PropertyInfo dynamicDictionaryPropertyInfo, bool isComplexType)
```
mikepizzo, line 110: Do we need to have a public constructor that sets isComplexType?  Do we expect the user to create and give us a Delta<T>? If so, should we expose a public property for the user to get whether it is complex or not?

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -235,11 +235,11 @@ public virtual object CreateResourceInstance(IEdmStructuredTypeReference structu
                             structuredType.StructuredDefinition(), model);
 
                         return Activator.CreateInstance(readContext.ResourceType, clrType, structuralProperties,
-                            dynamicDictionaryPropertyInfo);
+                            dynamicDictionaryPropertyInfo, structuredType.IsComplex());
```
mikepizzo, line 238: >, structuredType.IsComplex()) [](http://example.com/codeflow?start=57&length=29)

are there other places where we should be setting this?

mikepizzo:
Today the Delta<TStructuralType> has no mapping to the edm type. Adding the isComplexType is providing a coupling to the edm model, although it's a loose coupling (just whether it maps to a complex or entity type, not the specific type.  I think I'm okay with that in principle, and probably prefer separating complex from entity type rather than mapping to a specific edm type, but would like to get @xuzhg's thoughts.

If we are going to differentiate between complex and entity types I would rather do it through derivation, but that would have to wait for a new major version release.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,66 @@ public void CopyChangedValues(TStructuralType original)
                     bool isDeltaType = TypedDelta.IsDeltaOfT(deltaNestedResource.GetType());
                     Contract.Assert(isDeltaType, nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
 
+                    Type nestedType = deltaNestedResource.StructuredType;
+                    Type originalType = originalNestedResource.GetType();
+
+                    if (deltaNestedResource.IsComplexType && !nestedType.IsAssignableFrom(originalType))
```
mikepizzo, line 365: >IsComplexType [](http://example.com/codeflow?start=44&length=13)

If this hasn't been set, should we default to assuming it's complex (and we can reassign) or that it's entity and we can't?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,66 @@ public void CopyChangedValues(TStructuralType original)
...
         }
 
+        private dynamic ReAssignComplexDerivedType(TStructuralType original, string nestedResourceName, dynamic originalNestedResource, Type nestedType, Type originalType)
+        {
+            //As per OASIS discussion, changing a complex type from 1 derived type to another is allowed if both derived type have a common ancestor and the property
+            //is declared in terms of a common ancestor. The logic below checks for a common ancestor. Create a new object of the derived type in delta request.
+            //And copy the common properties.
+
+            Type complexBaseType = nestedType.BaseType;
+            HashSet<Type> lst = new HashSet<Type>();
```
mikepizzo, line 382: >lst [](http://example.com/codeflow?start=26&length=3)

can you make this a more descriptive name?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,66 @@ public void CopyChangedValues(TStructuralType original)
...
             }
         }
 
+        private dynamic ReAssignComplexDerivedType(TStructuralType original, string nestedResourceName, dynamic originalNestedResource, Type nestedType, Type originalType)
+        {
+            //As per OASIS discussion, changing a complex type from 1 derived type to another is allowed if both derived type have a common ancestor and the property
+            //is declared in terms of a common ancestor. The logic below checks for a common ancestor. Create a new object of the derived type in delta request.
+            //And copy the common properties.
+
+            Type complexBaseType = nestedType.BaseType;
```
mikepizzo, line 381: >nestedType.BaseType [](http://example.com/codeflow?start=35&length=19)

Please confirm your start states here and on line 390. It looks like you only check for subtypes, not if one of the actual types is a subtype of the other.  What if nestedType derives from originalType or vice-versa?  To validate, please add a test case for changing engine from V4 to V41 and from V41 to V4.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,66 @@ public void CopyChangedValues(TStructuralType original)
...
+
+            Type originalBaseType = originalType.BaseType;
+            bool foundCommonbase = false;
+
+            while (originalBaseType != null)
+            {
+                if (lst.Contains(originalBaseType))
+                {
+                    dynamic newOriginal = originalNestedResource;
+                    originalNestedResource = Activator.CreateInstance(nestedType);
```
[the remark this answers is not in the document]
mikepizzo: And, if this is a Delta<nestedType>, how do you know whether to set the isComplex property? You would need to have set the edm type when creating the delta<TStructuredType> in order to know whether a property of the TSructuredType was complex or not.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,66 @@ public void CopyChangedValues(TStructuralType original)
                     bool isDeltaType = TypedDelta.IsDeltaOfT(deltaNestedResource.GetType());
```
mikepizzo, line 359: >bool isDeltaType = TypedDelta.IsDeltaOfT(deltaNestedResource.GetType()); [](http://example.com/codeflow?start=20&length=72)

I realize this isn't your change, but we should put this in the Contract.Assert so that we're not making this somewhat expensive call unnecessarily in the non-debug case.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,66 @@ public void CopyChangedValues(TStructuralType original)
...
                     Contract.Assert(isDeltaType, nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
 
+                    Type nestedType = deltaNestedResource.StructuredType;
+                    Type originalType = originalNestedResource.GetType();
```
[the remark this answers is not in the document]
mikepizzo: originalType is the CLR type, not the delta<tstructuraltype>.  please add comments to make it easier to track where we are using the CLR type and where we are using the Delta<TStructuralType>

src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializer.cs
Sreejithpin: Not really. Ideally all resources are created here.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
Sreejithpin: As discussed lets call it user has to call this overlod to set initially?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
Sreejithpin: Updated. Parent to child already taken care. Added tests too

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
Sreejithpin: As discussed entity?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
Sreejithpin: Since deltaNestedResource is a dynamic type it doenst allow to use inside contract.assert. throws a build err


Error CS1974 The dynamically dispatched call to method 'Assert' may fail at runtime because one or more applicable overloads are conditional methods.

mikepizzo: That's a bummer.  

At least we should be able to put the TypedDelta.IsDeltaOfT() call within the Contract.Assert:

Type deltaType = deltaNestedResource.GetType();
Contract.Assert(TypedDelta.IsDeltaOfT(deltaType),...);

mikepizzo: Alternatively, we could cast to object:
 Contract.Assert(TypedDelta.IsDeltaOfT(((object)deltaNestedResource)?.GetType()), nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
mikepizzo: can we just name this newBaseTypes?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,75 @@ public void CopyChangedValues(TStructuralType original)
...
+
+                    foundCommonbase = true;
+
+                    break;
+                }
+
+                originalType = originalType.BaseType;
+            }
+
+            if (foundCommonbase)
```
mikepizzo, line 423: why do you need the foundCommonbase flag at all? why not just put 425 at line 415, before the break;?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,75 @@ public void CopyChangedValues(TStructuralType original)
...
+                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, nestedType, originalType);
+                    }
+
                     deltaNestedResource.CopyChangedValues(originalNestedResource);
                 }
             }
         }
 
+        private dynamic ReAssignComplexDerivedType(TStructuralType original, string nestedResourceName, dynamic originalNestedResource, Type nestedType, Type originalType)
+        {
```
mikepizzo, line 376: This method (and its members) should be named according to what it does, not how how it is used.  i.e., calling parameters "nestedType" and "original" bring in context from the caller that make it difficult to understand locally. Instead, if we just think of this method as reassigning a property, we could have a signature such as:

private dynamic ReAssignComplexProperty(TStructuralType parent, string propertyName, dynamic newValue, Type newType, Type originalType)

This makes it easier to understand this method without understanding how the method is used.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,75 @@ public void CopyChangedValues(TStructuralType original)
...
+
+            //Here original type is the type for original (T) resource.
+            //We will keep  going to base types and finally will get the Common Basetype for the derived complex types in to the originalType variable.
+                        
+            bool foundCommonbase = false;
+
+            //The new Original type, means the new complex type (T) which will replace the current complex type.
+            dynamic newOriginalNestedResource = null;
+
+            while (originalType != null)
```
mikepizzo, line 398: what if we don't find a common ancestor? or do we assert that there should be a common ancestor, since they should both derive from the type of the property?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -336,11 +359,75 @@ public void CopyChangedValues(TStructuralType original)
...
             }
         }
 
+        private dynamic ReAssignComplexDerivedType(TStructuralType original, string nestedResourceName, dynamic originalNestedResource, Type nestedType, Type originalType)
+        {
+            //As per OASIS discussion, changing a complex type from 1 derived type to another is allowed if both derived type have a common ancestor and the property
+            //is declared in terms of a common ancestor. The logic below checks for a common ancestor. Create a new object of the derived type in delta request.
+            //And copy the common properties.
+
+            Type complexBaseType = nestedType.BaseType;
```
mikepizzo, line 381: >complexBaseType [](http://example.com/codeflow?start=17&length=15)

complexBaseType doesn't help me understand if this is the base type of the new type or the original type.  Maybe call it newBaseType?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
mikepizzo: yes; we should set this in the constructor, not the property.  My question was, if the user is setting this in the constructor, shouldn't they have a way to get it on the created instance?  i.e., expose a public getter?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -44,6 +44,8 @@ public class Delta<TStructuralType> : TypedDelta, IDelta where TStructuralType :
         private HashSet<string> _changedDynamicProperties;
         private IDictionary<string, object> _dynamicDictionaryCache;
 
+        internal bool IsComplexType { get; set; }
```
mikepizzo, line 47: >set [](http://example.com/codeflow?start=43&length=3)

should this be private set? it should only be set in the constructor, right?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Builder/TestModels/InheritanceModels.cs
```diff
@@ -54,6 +54,42 @@ public override int WheelCount
...
+
+        public Engine MyEngine { get; set; }
+
+        public V4 MyV4Engine { get; set; }
+    }
+
+    public class Engine
+    {
+        public int Hp { get; set; }
+    }
```
mikepizzo, line 66: can we add another complex property to the base "engine" type, that has derived types, so that we can verify that the recursion works?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
mikepizzo: Not convinced your tests cover the scenario I was looking for.

Both the original and new tests have the type of the original value the same as the expected type of the property (I would still like to understand how changing the type works in this case, though, if it doesn't go through your new code).  The new test I was looking for was a property of declared type "V2", whose current value was type V4 and new value V41, and vice-versa.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
Sreejithpin: If there is not in cae of deserialization it throws error, in case of user creation this will be skipped and it will throw err at old place where it used to

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -44,6 +44,11 @@ public class Delta<TStructuralType> : TypedDelta, IDelta where TStructuralType :
         private HashSet<string> _changedDynamicProperties;
         private IDictionary<string, object> _dynamicDictionaryCache;
 
+        /// <summary>
+        /// To determine if the StructuralType is a Complex type
+        /// </summary>
+        public bool IsComplexType { get; private set; }
```
mikepizzo, line 50: should be moved down with other public properties

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -332,15 +358,74 @@ public void CopyChangedValues(TStructuralType original)
...
-                    // Recursively patch the subtree.
-                    bool isDeltaType = TypedDelta.IsDeltaOfT(deltaNestedResource.GetType());
-                    Contract.Assert(isDeltaType, nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
+                    // Recursively patch the subtree.                    
+                    Contract.Assert(TypedDelta.IsDeltaOfT(((object)deltaNestedResource).GetType()), nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
+
+                    Type newType = deltaNestedResource.StructuredType;
+                    Type originalType = originalNestedResource.GetType();
+
+                    if (deltaNestedResource.IsComplexType && !newType.IsAssignableFrom(originalType))
```
mikepizzo, line 367: >!newType.IsAssignableFrom(originalType)) [](http://example.com/codeflow?start=61&length=40)

does this mean that, if the new object is a base type of the original object, we won't update the type?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/InheritanceTests.cs
```diff
@@ -192,6 +192,225 @@ public async Task Can_Patch_Entity_In_Inheritance()
             Assert.False((bool)result.CanDoAWheelie);
         }
 
+        [Fact]
+        public async Task Can_Patch_Entity_In_Inheritance_DerivedEngine()
```
mikepizzo, line 196: Please validate that the type, as well as properties, are updated in each of these tests

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -332,15 +358,74 @@ public void CopyChangedValues(TStructuralType original)
...
-                    Contract.Assert(isDeltaType, nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
+                    // Recursively patch the subtree.                    
+                    Contract.Assert(TypedDelta.IsDeltaOfT(((object)deltaNestedResource).GetType()), nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
+
+                    Type newType = deltaNestedResource.StructuredType;
+                    Type originalType = originalNestedResource.GetType();
+
+                    if (deltaNestedResource.IsComplexType && newType != originalType)
+                    {
+                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, newType, originalType, deltaNestedResource.ExpectedClrType);
```
mikepizzo, line 369: >originalNestedResource [](http://example.com/codeflow?start=24&length=22)

Does returning null here (in the case that we don't find a common base type) change the current behavior of just keeping the original originalNestedResource?

mikepizzo: In iteration 3 you returned originalNestedResource if no common base type was found. In iteration 4 the behavior changed to return null if no common base type was found. Was this intentional?

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -332,15 +358,74 @@ public void CopyChangedValues(TStructuralType original)
...
+                newBaseType = newBaseType.BaseType;
+            }
+
+            newBaseTypes.Add(declaredType);
+
+            //Here original type is the type for original (T) resource.
+            //We will keep  going to base types and finally will get the Common Basetype for the derived complex types in to the originalType variable.
+            
+            //The new Original type, means the new complex type (T) which will replace the current complex type.
+            dynamic newOriginalNestedResource = null;
```
mikepizzo, line 399: >dynamic newOriginalNestedResource = null; [](http://example.com/codeflow?start=12&length=41)

In iteration 3 you returned originalNestedResource if no common base type was found. In iteration 4 the behavior changed to return null if no common base type was found. Was this intentional, or should newOriginalNestedResource default to originalValue if no common base type is found in order to preserve existing behavior?

mikepizzo (approved):
:shipit:

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Formatter/DeltaTests.cs
```diff
@@ -412,6 +412,28 @@ public async Task PatchShouldSupportNonSettableCollectionProperties()
...
             Assert.Equal(3, query.Orders.Count);
         }
+
+        [Fact]
+        public async Task PatchShouldSupportComplexDerivedTypeTransform()
+        {
+            HttpRequestMessage patch = new HttpRequestMessage(new HttpMethod("MERGE"), BaseAddress + "/odata/DeltaCustomers(6)");
+            dynamic data = new ExpandoObject();
+            data.Addresses = Enumerable.Range(10, 3).Select(i => new DeltaAddress { ZipCode = i });
+
```
xuzhg, line 422: The above test cases update the (6) customers, and you do the similar thing for same customers.

In the controller, it's a static List. Any concurrency problem?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Formatter/DeltaTests.cs
```diff
@@ -412,6 +412,28 @@ public async Task PatchShouldSupportNonSettableCollectionProperties()
             Assert.Equal(3, query.Addresses.Count);
             Assert.Equal(3, query.Orders.Count);
         }
+
+        [Fact]
+        public async Task PatchShouldSupportComplexDerivedTypeTransform()
+        {
+            HttpRequestMessage patch = new HttpRequestMessage(new HttpMethod("MERGE"), BaseAddress + "/odata/DeltaCustomers(6)");
```
xuzhg, line 419: >MERGE [](http://example.com/codeflow?start=78&length=5)

use Patch, maybe replace it for other test cases.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -88,12 +88,33 @@ public Delta(Type structuralType, IEnumerable<string> updatableProperties)
...
+        /// <param name="structuralType">The derived entity type or complex type for which the changes would be tracked.
+        /// <paramref name="structuralType"/> should be assignable to instances of <typeparamref name="TStructuralType"/>.
+        /// </param>
+        /// <param name="updatableProperties">The set of properties that can be updated or reset. Unknown property
+        /// names, including those of dynamic properties, are ignored.</param>
+        /// <param name="dynamicDictionaryPropertyInfo">The property info that is used as dictionary of dynamic
+        /// properties. <c>null</c> means this entity type is not open.</param>
+        /// <param name="isComplexType">Boolean value to determine if its a complex type</param>
+        public Delta(Type structuralType, IEnumerable<string> updatableProperties,
+            PropertyInfo dynamicDictionaryPropertyInfo, bool isComplexType)
```
xuzhg, line 108: >isComplexType [](http://example.com/codeflow?start=61&length=13)

Again, I don't think we should introduce this parameter.
For the nested resource, it's a Delta<T> also.

if it's derived type, the property "StructuredType" is the real type, (Derived) type,  "ExpectedClrType" is the expected type, (Base) type. 

We just replace the value using the derived value. 
 
from developer persepecitve, there's no "difference" between  entity type or complex type.

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -332,15 +358,74 @@ public void CopyChangedValues(TStructuralType original)
...
+                    // Recursively patch the subtree.                    
+                    Contract.Assert(TypedDelta.IsDeltaOfT(((object)deltaNestedResource).GetType()), nestedResourceName + "'s corresponding value should be Delta<T> type but is not.");
+
+                    Type newType = deltaNestedResource.StructuredType;
+                    Type originalType = originalNestedResource.GetType();
+
+                    if (deltaNestedResource.IsComplexType && newType != originalType)
+                    {
+                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, newType, originalType, deltaNestedResource.ExpectedClrType);
+                    }
```
xuzhg, line 370: I think the logic is:

1) if it's the same type, retrieve the original nested resource and call "CopyChangedValues".

2) if it's not the same type, create an empty nested resource and call "CopyChangeValues".

So, ReAssignComplexDerivedType(...) ==> CreateNewNestedResourceInstance(...)

Issue #2584, opened 2021-10-09
Regression in AutoExpand logic


ificator:
Changes in https://github.com/OData/WebApi/pull/2582 have caused `ArgumentNullException` failures when these did not used to occur.

<img width="605" alt="image" src="https://user-images.githubusercontent.com/8387253/136673920-090257bd-806e-4875-b0dc-edfce3e12150.png">

The old method (green arrow) allowed `baseEntityType` to be null - it just returned an empty `List<IEdmStructuralProperty>`. However the new method (red arrow) asserts that `baseEntityType` cannot be null - it throws an `ArgumentNullException`.

I have not drilled in to figure out what `baseEntityType` is null in this case, but it seems the simplest fix is to keep the semantics of the old code and not consider this a failure case.

### Assemblies affected

Microsoft.AspNet.OData 7.5.12

### Reproduce steps

Unclear - will take much more work to try to isolate

### Expected result

`baseEntityType == null` does not result in failure

### Actual result

`baseEntityType == null` results in failure

PR #2585, opened 2021-10-11
Fix Auto-Expand Regression

ificator:
### Issues

Fixes https://github.com/OData/WebApi/issues/2584

### Description

1. Make "GetAutoSelectPaths" and "GetAutoExpandPaths" behave like "GetAutoSelectProperties" when "baseEntityType" is null
2. Avoid some unnecessary allocations

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
@@ -34,22 +34,24 @@ namespace Microsoft.AspNet.OData.Query
                 throw Error.ArgumentNull(nameof(edmModel));
             }
 
-            if (structuredType != null)
+            if (structuredType == null)
             {
-                foreach (IEdmStructuredType edmStructuredType in new SelfAndDerivedEnumerator(structuredType, edmModel))
-                {
-                    // for top type, let's retrieve its properties and the properties from base type of top type if has.
-                    // for derived type, let's retrieve the declared properties.
-                    IEnumerable<IEdmStructuralProperty> properties = edmStructuredType == structuredType
-                            ? edmStructuredType.StructuralProperties()
-                            : edmStructuredType.DeclaredStructuralProperties();
+                throw Error.ArgumentNull(nameof(structuredType));
+            }
 
-                    foreach (IEdmStructuralProperty subProperty in properties)
+            foreach (IEdmStructuredType edmStructuredType in new SelfAndDerivedEnumerator(structuredType, edmModel))
+            {
+                // for top type, let's retrieve its properties and the properties from base type of top type if has.
+                // for derived type, let's retrieve the declared properties.
+                IEnumerable<IEdmStructuralProperty> properties = edmStructuredType == structuredType
+                        ? edmStructuredType.StructuralProperties()
+                        : edmStructuredType.DeclaredStructuralProperties();
+
+                foreach (IEdmStructuralProperty subProperty in properties)
+                {
+                    if (IsAutoSelect(subProperty, property, edmStructuredType, edmModel))
                     {
-                        if (IsAutoSelect(subProperty, property, edmStructuredType, edmModel))
-                        {
-                            return true;
-                        }
+                        return true;
                     }
                 }
             }
@@ -81,7 +83,7 @@ namespace Microsoft.AspNet.OData.Query
 
         private static bool HasAutoExpandProperty(this IEdmModel edmModel, IEdmStructuredType structuredType, IEdmProperty pathProperty, ISet<IEdmStructuredType> visited)
         {
-            if (structuredType == null || visited.Contains(structuredType))
+            if (visited.Contains(structuredType))
             {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
```diff
@@ -347,5 +334,95 @@ private static bool CanExpand(this IEdmModel edmModel, IEdmStructuredType struct
...
+
+                        default:
+                            throw new InvalidOperationException("Enumeration is an invalid state");
+                    }
+                }
+            }
+
+            object IEnumerator.Current => this.Current;
+
+            public void Dispose()
```
corranrogue9, line 392: I actually predict that this will perform better than `Prepend` at least because we are not disposing `this.derivedEnumerator`. Please add that logic in here.

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
```diff
@@ -347,5 +334,95 @@ private static bool CanExpand(this IEdmModel edmModel, IEdmStructuredType struct
...
+            public SelfAndDerivedEnumerator GetEnumerator()
+            {
+                return this;
+            }
+
+            public bool MoveNext()
+            {
+                switch (this.stage)
+                {
+                    case 0:
```
corranrogue9, line 405: Can't this just be:
```
case 0:
  this.stage = 1;
  return true;

default:
  return this.derivedEnumerator.MoveNext();
```

The compiler might be optimizing that out, but I think it will be faster and it's easier to read

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
```diff
@@ -33,28 +34,22 @@ public static bool HasAutoSelectProperty(this IEdmModel edmModel, IEdmStructured
...
-                        ? edmStructuredType.StructuralProperties()
-                        : edmStructuredType.DeclaredStructuralProperties();
-
-                foreach (IEdmStructuralProperty subProperty in properties)
+                foreach (IEdmStructuredType edmStructuredType in new SelfAndDerivedEnumerator(structuredType, edmModel))
                 {
-                    if (IsAutoSelect(subProperty, property, edmStructuredType, edmModel))
+                    // for top type, let's retrieve its properties and the properties from base type of top type if has.
+                    // for derived type, let's retrieve the declared properties.
+                    IEnumerable<IEdmStructuralProperty> properties = edmStructuredType == structuredType
```
corranrogue9, line 43: > edmStructuredType == structuredType 

Wait, we are behaving different for the first element anyway? Why does it need to be part of the same sequence as the derived types? We can have an overload of `IsAutoSelect` that takes the properties sequence, and then this code becomes:
```
if (IsAutoSelect(structuredType.StructuralProperties))
{
  return true;
}

foreach (var edmStructuredType in edmModel.FindAllDerivedTypes(structuredType))
{
  if (IsAutoSelect(structuredType.StructuralProperties))
  {
    return true;
  }
}

return false;
```
Also, we seem to just be re-implementing `Any` here.

corranrogue9: This feedback seems to apply to other cases that are being updated in this PR as well

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
```diff
@@ -347,5 +334,95 @@ private static bool CanExpand(this IEdmModel edmModel, IEdmStructuredType struct
...
+        /// This is a helper that allows us to avoid inefficiencies in the previous pattern:
+        ///     var structuredTypes = new List&lt;IEdmStructuredType&gt;();
+        ///     structuredTypes.Add(structuredType);
+        ///     structuredTypes.AddRange(edmModel.FindAllDerivedTypes(structuredType));
+        ///
+        /// Specifically, the allocation of the list and the resizing driven by the "AddRange" call are
+        /// avoided by leveraging a struct with a simple state machine for enumerating over a type
+        /// and its derived types.
+        /// </summary>
+        private struct SelfAndDerivedEnumerator : IEnumerator<IEdmStructuredType>
```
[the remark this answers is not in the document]
ificator: Here's the result of a simple benchmark:
|  Method |      Mean |    Error |   StdDev |  Gen 0 | Allocated |
|-------- |----------:|---------:|---------:|-------:|----------:|
| Current | 129.77 ns | 2.448 ns | 2.289 ns | 0.0296 |     124 B |
| Prepend |  88.25 ns | 1.603 ns | 1.846 ns | 0.0143 |      60 B |
|  Manual |  29.31 ns | 0.602 ns | 0.618 ns |      - |         - |
|  Custom |  55.67 ns | 0.924 ns | 0.865 ns |      - |         - |

Where:
```csharp
[SimpleJob(RuntimeMoniker.Net472)]
[MemoryDiagnoser]
public class Benchmark
{
    private int first = 1;
    private List<int> rest = new List<int> { 2, 3, 4, 5 };
    
    [Benchmark]
    public int List()
    {
        var l = new List<int>();
        l.Add(this.first);
        l.AddRange(this.rest);
        
        int sum = 0;
        foreach (int i in l)
        {
            sum += i;
        }
        
        return sum;
    }
    
    [Benchmark]
    public int Prepend()
    {
        int sum = 0;
        foreach (int i in this.rest.Prepend(this.first))
        {
            sum += i;
        }
        
        return sum;
    }
    
    [Benchmark]
    public int Manual()
    {
        int sum = this.first;
        foreach (int i in this.rest)
        {
            sum += i;
        }
        
        return sum;
    }

    [Benchmark]
    public int Custom()
    {
        int sum = 0;
        foreach (int i in new SelfAndDerivedEnumerator(this.first, this.rest))
        {
            sum += i;
        }
        
        return sum;
    }
}
```
So there's a good win in terms of memory as expected over `Prepend`, but execution time is more trivial.

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
ificator: Good callout - I didn't think too much about what the code was attempting to do, but on removing the list allocation. There appears to be further optimizations that we could do, but given the "PR complexity" analysis that is done I'd rather keep this one focused on the list change. We can follow up with more involved refactoring in another PR.

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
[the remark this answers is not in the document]
corranrogue9: Gotcha, that makes sense because of the use of `struct` I supposed. Thanks for the benchmarks

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
corranrogue9: I'll create an issue or something to track, I actually got lost in the code and didn't realize the behavior isn't actually changing here so I agree it's outside the scope of this change

corranrogue9 (approved):

KenitoInc (approved):

KenitoInc:
@ificator Kindly fix conflicts

ificator:
@KenitoInc it looks like https://github.com/OData/WebApi/pull/2588 fixed the original issue (null `baseEntityType` asserting when it previously didn't), so I guess the question is do we want my refactoring to avoid allocations? If so, I'll merge those changes in and convert this PR into a PR specifically for that.

KenitoInc:
> @KenitoInc it looks like #2588 fixed the original issue (null `baseEntityType` asserting when it previously didn't), so I guess the question is do we want my refactoring to avoid allocations? If so, I'll merge those changes in and convert this PR into a PR specifically for that.

If we are avoiding allocations for performance improvement, then that's fine.

marabooy:
@ificator could you kindly rebase your changes to master so we can proceed with resolving this issue.  Many thanks for your contribution.

marabooy (approved):

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
```diff
@@ -175,6 +164,11 @@ private static bool HasAutoExpandProperty(this IEdmModel edmModel, IEdmStructure
                 {
                     if (IsAutoSelect(property, pathProperty, edmStructuredType, edmModel, querySettings))
                     {
+                        if (autoSelectProperties == null)
+                        {
+                            autoSelectProperties = new List<SelectModelPath>(1);
```
corranrogue9, line 169: I don't see the value in having to perform this `null` check every iteration. It probably makes more sense to just put line 160 back

src/Microsoft.AspNet.OData.Shared/Query/AutoSelectExpandHelper.cs
```diff
@@ -347,5 +337,103 @@ private static bool CanExpand(this IEdmModel edmModel, IEdmStructuredType struct
...
+            }
+
+            private readonly IEnumerator<IEdmStructuredType> derivedEnumerator;
+            private readonly IEdmStructuredType structuredType;
+
+            private Stage stage;
+
+            public SelfAndDerivedEnumerator(IEdmStructuredType structuredType, IEdmModel edmModel)
+            {
+                if (structuredType == null)
```
corranrogue9, line 367: Given that this is all internal, these `null` checks probably make more sense as `Debug.Assert`s

Sreejithpin:
Hi @ificator , will be great if you can look at comments and rebase

mikepizzo (approved):
:shipit:

KenitoInc:
@ificator Thanks for this PR

Issue #2586, opened 2021-10-12
Cannot use a different Derived complex Type for Patching 


Sreejithpin:
As per Protocol, one Complex derived type can be replaced by another, as long as there is a common ancestor base which is used for declaring the property

We had an issue for nested complex type https://github.com/OData/WebApi/issues/2572 which is fixed
But the same is applicable for root type, ie: Patch on a complex type

### Assemblies affected

OData WebApi lib 7.5.12

### Reproduce steps

Odata schema
<EntityType Name="anEntity">
     <Property Name="property1" Type="complexType1">
</EntityType>
<ComplexType Name="complexType1">
    <Property Name="property2" Type="complexType2">
</ComplexType>
<ComplexType Name="complexType2"
    <Property Name="property3" Type="complexType3">
</ComplexType>
<ComplexType Name="complexType3">
     <Property Name="propertyBase" Type="someBaseType">
</ComplexType>
<ComplexType Name="someBaseType" Abstract="true" />
<ComplexType Name="derivedType1" BaseType="someBaseType" />
<ComplexType Name="derivedType2"  BaseType="someBaseType" >
    <Property Name="propertyDerived" Type="Collection(Edm.String)>
</ComplexType>
Existing anEntity
{
    "property1": {
        "property2": {
             "property3": {
                  "propertyBase": {
                       "@odata.type": "derivedType1"
                  }
             }
         }
     }
}
Patch call to change propertyBase
{
    "property1": {
        "property2": {
             "property3": {
                  "propertyBase": {
                       "@odata.type": "derivedType2",
                       "propertyDerived": [ "some string" ]
                  }
             }
         }
     }
}

Error
Cannot use Delta of type 'derivedType1' on an entity of type 'derivedType2'

### Expected result

The patch should work and the propertyBase should be updated to derivedType2

### Actual result

Cannot use Delta of type 'derivedType1' on an entity of type 'derivedType2'

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

PR #2601, opened 2021-11-15
Update Patch to Return TStructuralType and Handle Patch for Derived Complex type

### LINKED ISSUE #2590: Delta<T> Patch is void, Need capability to return Delta from Patch
1. Currently Delta<T> Patch is void. So we cannot return any updates to Delta to the user including nested Deltas and Collections.
We need to have  a capability to return Delta<T> from Patch so that a user can return Delta from the controller action.

2. Currently Delta<T> cannot be returned from controller action as Delta<T> as root it is not allowed to serialize. We need to unblock the same as well

### Assemblies affected

Odata.WebApi 7.5.x

### Reproduce steps

### Expected result

User should be able to return Delta from Patch Controller action

### Actual result

Currently Delta<T> cannot be returned from controller action 

### Additional detail

Sreejithpin:
### Issues

*This pull request fixes #2590  and #2586 .*

### Description

Updated Patch to return TStructuralType
Updated Patch to handle a different derived complex type from original type (as per oasis standards)

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -343,7 +343,7 @@ namespace Microsoft.AspNet.OData
         /// to the <paramref name="original"/> entity recursively.
         /// </summary>
         /// <param name="original">The entity to be updated.</param>
-        public void CopyChangedValues(TStructuralType original)
+        public TStructuralType CopyChangedValues(TStructuralType original)
         {
             if (original == null)
             {
@@ -403,11 +403,14 @@ namespace Microsoft.AspNet.OData
                     if (deltaNestedResource.IsComplexType && newType != originalType)
                     {
                         originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, newType, originalType, deltaNestedResource.ExpectedClrType);
+                        _structuredType.GetProperty(nestedResourceName).SetValue(original, (object)originalNestedResource);
                     }
 
                     deltaNestedResource.CopyChangedValues(originalNestedResource);
                 }
             }
+
+            return original;
         }
 
         private dynamic ReAssignComplexDerivedType(TStructuralType parent, string nestedPropertyName, dynamic originalValue, Type newType, Type originalType, Type declaredType)
@@ -415,6 +418,11 @@ namespace Microsoft.AspNet.OData
             //As per OASIS discussion, changing a complex type from 1 derived type to another is allowed if both derived type have a common ancestor and the property
             //is declared in terms of a common ancestor. The logic below checks for a common ancestor. Create a new object of the derived type in delta request.
             //And copy the common properties.
+
+            if(newType == originalType)
+            {
+                return originalValue;
+            }
 
             Type newBaseType = newType;
             HashSet<Type> newBaseTypes = new HashSet<Type>();
@@ -451,8 +459,6 @@ namespace Microsoft.AspNet.OData
                         property.SetValue(newOriginalNestedResource, value);
                     }
 
-                    _structuredType.GetProperty(nestedPropertyName).SetValue(parent, (object)newOriginalNestedResource);
-
                     break;
                 }
 
@@ -488,14 +494,20 @@ namespace Microsoft.AspNet.OData
             CopyUnchangedDynamicValues(original);
         }
 
+
         /// <summary>
         /// Overwrites the <paramref name="original"/> entity with the changes tracked by this Delta.
         /// <remarks>The semantics of this operation are equivalent to a HTTP PATCH operation, hence the name.</remarks>
         /// </summary>
         /// <param name="original">The entity to be updated.</param>
-        public void Patch(TStructuralType original)
+        public TStructuralType Patch(TStructuralType original)
         {
-            CopyChangedValues(original);
+            if (IsComplexType)
+            {
+                original = ReAssignComplexDerivedType(null, "", original, _structuredType, original.GetType(), ExpectedClrType) as TStructuralType;                
+            }
+
+            return CopyChangedValues(original);           
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
@@ -402,7 +402,7 @@ namespace Microsoft.AspNet.OData
 
                     if (deltaNestedResource.IsComplexType && newType != originalType)
                     {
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -369,20 +369,28 @@ public void CopyChangedValues(TStructuralType original)
...
-                        originalNestedResource = ReAssignComplexDerivedType(original, nestedResourceName, originalNestedResource, newType, originalType, deltaNestedResource.ExpectedClrType);
+                        originalNestedResource = ReAssignComplexDerivedType(originalNestedResource, newType, originalType, deltaNestedResource.ExpectedClrType);
+                        _structuredType.GetProperty(nestedResourceName).SetValue(original, (object)originalNestedResource);
                     }
 
                     deltaNestedResource.CopyChangedValues(originalNestedResource);
                 }
             }
+
+            return original;
```
corranrogue9, line 380: I'm sure I'm missing something obvious here, but I don't see the value in returning the parameter, especially because the generic type restriction for `TStructuralType` is that it is a `class`, so any modifications to it will be reflected on the underlying object anyway.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceControllers.cs
```diff
@@ -117,6 +117,25 @@ public ITestActionResult Patch(int key, [FromBody]Delta<Window> delta)
             return Ok(window);
         }
 
+        [ODataRoute("Windows({key})/CurrentShape")]
+        public ITestActionResult PatchShape(int key, [FromBody] Delta<Shape> delta)
+        {
+            Window window = _windows.FirstOrDefault(e => e.Id == key);
```
corranrogue9, line 123: Given that there is no `null` check, `First` makes more sense than `FirstOrDefault`

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -342,23 +342,56 @@ public async Task PatchContainingEntity_Matched_DerivedType(string modelMode)
...
+                {
+                    '@odata.type':'#Microsoft.Test.E2E.AspNet.OData.ComplexTypeInheritance.Circle',
+                    'Radius':2,
+                    'Center':{'X':1,'Y':2},
+                    'HasBorder':true
+                },
+                'OptionalShapes': [ ]
+            }";
+
+            StringContent stringContent = new StringContent(content: content, encoding: Encoding.UTF8, mediaType: "application/json");
```
corranrogue9, line 356: Please wrap `IDisposable`s in `using` blocks (for example, `StringContent` and `HttpResponseMessage`)

corranrogue9 (approved):

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
Sreejithpin: Actually not really, and thats why we had to make this change, for complextype we are creating a new type like by doing original = new ...()

src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -455,14 +461,21 @@ public void CopyUnchangedValues(TStructuralType original)
             CopyUnchangedDynamicValues(original);
         }
 
+
         /// <summary>
```
mikepizzo, line 465: nit- extra blank line

mikepizzo (approved):
:shipit:

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -342,23 +342,56 @@ public async Task PatchContainingEntity_Matched_DerivedType(string modelMode)
...
+            request.Content = stringContent;
+            HttpResponseMessage response = await Client.SendAsync(request);
+            string contentOfString = await response.Content.ReadAsStringAsync();
+            Assert.True(HttpStatusCode.OK == response.StatusCode);
+        }
+
+        [Theory]
+        [InlineData("convention")]
+        [InlineData("explicit")]
+        // Patch ~/Widnows(3)
```
gathogojr, line 366:
```suggestion
        // Patch ~/Windows(3)
```

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -342,23 +342,56 @@ public async Task PatchContainingEntity_Matched_DerivedType(string modelMode)
...
+        {
+            string serviceRootUri = string.Format("{0}/{1}", BaseAddress, modelMode).ToLower();
+            string requestUri = serviceRootUri + "/Windows(3)/CurrentShape";
+
+            HttpRequestMessage request = new HttpRequestMessage(new HttpMethod("PATCH"), requestUri);
+
+            // Attempt to PATCH nested resource with delta object of the different CLR type
+            // will result an error.
+            
+
```
gathogojr, line 377:
```suggestion
```

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ComplexTypeInheritance/ComplexTypeInheritanceTests.cs
```diff
@@ -342,23 +342,56 @@ public async Task PatchContainingEntity_Matched_DerivedType(string modelMode)
...
             StringContent stringContent = new StringContent(content: content, encoding: Encoding.UTF8, mediaType: "application/json");
             request.Content = stringContent;
             HttpResponseMessage response = await Client.SendAsync(request);
-            string contentOfString = await response.Content.ReadAsStringAsync();
+            JObject contentOfJObject = await response.Content.ReadAsObject<JObject>();
+            Assert.Equal(2, (int)contentOfJObject["Radius"]);
             Assert.True(HttpStatusCode.OK == response.StatusCode);
         }
 
+
```
gathogojr, line 394:
```suggestion
```

gathogojr (approved):
A few trivial comments otherwise looks good

PR #2622, opened 2022-02-11
Create type annotation only with type name

xuzhg:
### Issues

Port from [Create type annotation only has the type name](https://github.com/OData/AspNetCoreOData/commit/4721f7aa9b84ba5c2ff914fa4ca860d146b87534)

### Description

* If the type name is null, we should not create the type annotation for primitive property and let ODL to decision the type metadata.

- [x] *Test cases added*

### Additional work necessary


mikepizzo (approved):
:shipit:

Sreejithpin (approved):

PR #2631, opened 2022-03-07
updated uses of random number generation in tests so that we log the …

corranrogue9:
…used for later investigation

### Description

*Some tests use a random number generated when creating data. This can cause tests to fail intermittently. This PR logs the seed used for random number generation so that those intermittent failures can be reproduced for investigation.*

- [ ] *Test cases added*

### Additional work necessary

*No additional work necessary.*

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Common/Instancing/RandomSeedGenerator.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Common/Instancing/RandomSeedGenerator.cs
@@ -6,9 +6,11 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Security;
+using System.Threading;
 
 namespace Microsoft.Test.E2E.AspNet.OData.Common.Instancing
 {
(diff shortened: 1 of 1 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Formatter/CollectionPropertyTests.cs
```diff
@@ -69,7 +70,10 @@ public async Task SupportPostCollectionPropertyByEntityPayload()
             // clear respository
             await this.ClearRepositoryAsync("CollectionProperty_Entity");
 
-            var rand = new Random(RandomSeedGenerator.GetRandomSeed());
+            var seed = RandomSeedGenerator.GetRandomSeed();
+            Trace.WriteLine($"Generated seed for random number generator: {seed}");
```
xuzhg, line 74: where do you retrieve the trace information? If it's running on Azure pipeline, is it easy to trace?

Different test cases output the same trace, is it easy to identify?

corranrogue9: I guess I am only expecting someone to need the seed in the event of an intermittent test failure that needs to be investigated, so in the pipeline logs, when it logs the failed test, the trace will be right above the name of the failed test

corranrogue9: Are you thinking I should add the test name to the message or something along those lines?

corranrogue9: @xuzhg, I've updated the change so that we only log 1 time. It is relatively straightforward to find the seed in the build logs.

xuzhg (approved):

PR #2648, opened 2022-05-18
Allow `EnableQueryAttribute` derived classes to specify `IgnoreQueryOptions`

ificator:
### Issues

Allow classes deriving from `EnableQueryAttribute` to pass `IgnoreQueryOptions` flags to `ODataQueryOptions`.

### Description

Deriving from `EnableQueryAttribute` is a supported scenario so that `ApplyQuery` can have its implementation tweaked. However, the `ODataQuerySettings` instance is NOT exposed to derived classes and so it's not possible to pass in `ignoreQueryOptions` to `ODataQueryOptions.ApplyTo`. This change allows the value to be directly set on `ODataQueryOptions` instead of being set via `ApplyTo`.

- [x] *Test cases added*


src/Microsoft.AspNet.OData.Shared/EnableQueryAttribute.cs
```diff
@@ -371,6 +371,11 @@ public int MaxOrderByNodeCount
             }
         }
 
+        /// <summary>
+        /// Gets the query settings.
+        /// </summary>
+        protected ODataQuerySettings QuerySettings => _querySettings;
```
corranrogue9, line 377: Yeah, this whole thing is a mess. I don't feel comfortable signing off as-is because `_querySettings` sort of exists in memory for the lifetime of the process, but it's also very mutable, so it's really easy for a caller to make mistakes. I see two options to address that:

1. Change this code to use the `CopyFrom` method on `ODataQuerySettings`, that way the "original" settings are always preserved
2. Add a new type for `ODataQuerySettings` that is immutable, along with an overload on `ODataQueryOptions` which uses the new type.

ificator: Thank you for calling this out - I forgot that this is a `FilterAttribute` and as a result it's cached. Given that, you're right that `_querySettings` spans requests and should therefore be protected. I'll look into option 1 for now as I don't think 2 is worth it given how little data `CopyFrom` will copy.

habbes: Out of curiosity, does this concern also apply to the existing properties on this class which appear to mutate the `_querySettings` when modified (e.g. the `PageSize`)?

ificator: I believe so - the "normal" use-case is that a developer sticks the attribute on a controller/action and, using attribute syntax, populates the required property values. There's no way to modify those at runtime.

Where thread-safety comes into the picture is if you inherit from `EnableQueryAttribute`, at which point you can theoretically modify the values per-request which, given the semi-static nature of `EnableQueryAttribute` instances, can lead to cross-thread state issues.

So the "bigger picture" questions are how can we allow developers to override some aspects of `EnableQueryAttribute` without introducing potential race conditions.

corranrogue9: @ificator are you waiting on something here? I think I'm still happy with either of the two options I put at the beginning. I don't see thread-safety issues with either of them (happy to be wrong though).

ificator: Sorry for the delay! I've changed my approach to avoid the need to expose `ODataQuerySettings` for now (although it probably should be done at some point) by focusing more on my specific scenario. I think this latest change should be good.

src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
```diff
@@ -530,8 +535,8 @@ private static void ExtractGroupingProperties(List<string> result, IEnumerable<G
         /// query options.</remarks>
         public virtual object ApplyTo(object entity, ODataQuerySettings querySettings, AllowedQueryOptions ignoreQueryOptions)
         {
-            _ignoreQueryOptions = ignoreQueryOptions;
-            return ApplyTo(entity, new ODataQuerySettings());
```
ificator, line 534: I'm assuming this was a bug...

WanjohiSammy:
@ificator 
Please link the Issue ID associated with this change, if available. Also, ensure to add tests that validate the cases introduced by this change.

WanjohiSammy:
@ificator as there has been no activity or response on this PR for a while, I am going to close it for now. If you would like to continue working on it, please feel free to reopen it.

PR #2650, opened 2022-05-27
argument null exception in containerbuilder on .net 6

xuzhg:
### Issues

*This pull request fixes https://github.com/OData/AspNetCoreOData/issues/601.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
@@ -83,15 +83,18 @@ namespace Microsoft.AspNet.OData
         /// <returns>The container built by this builder.</returns>
         public virtual IServiceProvider BuildContainer()
         {
-            /* "services.BuildServiceProvider()" returns IServiceProvider in Microsoft.Extensions.DependencyInjection 1.0 and ServiceProvider in Microsoft.Extensions.DependencyInjection 2.0
-            * (This is a breaking change)[https://github.com/aspnet/DependencyInjection/issues/550].
-            * To support both versions with the same code base in OData/WebAPI we decided to call that extension method using reflection.
-            * More info at https://github.com/OData/WebApi/pull/1082
-            */
+            try
+            {
+                return services.BuildServiceProvider();
+            }
+            catch
+            {
+                MethodInfo buildServiceProviderMethod =
+                    typeof(ServiceCollectionContainerBuilderExtensions)
+                    .GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) });
 
-            MethodInfo buildServiceProviderMethod = typeof(ServiceCollectionContainerBuilderExtensions).GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) });
-
-            return (IServiceProvider)buildServiceProviderMethod.Invoke(null, new object[] { services });
+                return (IServiceProvider)buildServiceProviderMethod.Invoke(null, new object[] { services });
+            }
         }
 
         private static Microsoft.Extensions.DependencyInjection.ServiceLifetime TranslateServiceLifetime(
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.OData.Shared/DeltaOfTStructuralType.cs
```diff
@@ -414,7 +414,7 @@ public TStructuralType CopyChangedValues(TStructuralType original)
             return original;
         }
 
-        private dynamic ReAssignComplexDerivedType(dynamic originalValue, Type newType, Type originalType, Type declaredType)
+        private static dynamic ReAssignComplexDerivedType(dynamic originalValue, Type newType, Type originalType, Type declaredType)
```
xuzhg, line 417: This is a build warning fix.

tools/WebStack.versions.settings.targets
```diff
@@ -4,8 +4,8 @@
   <!-- Set the version number: major, minor, build and release (i.e. alpha, beta or blank for RTM)-->
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">7</VersionMajor>
-    <VersionMinor Condition="'$(VersionMinor)' == ''">5</VersionMinor>
-    <VersionBuild Condition="'$(VersionBuild)' == ''">14</VersionBuild>
+    <VersionMinor Condition="'$(VersionMinor)' == ''">7</VersionMinor>
```
xuzhg, line 7: I will revert the version change to another release PR to change the version.

src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
```diff
@@ -83,15 +83,18 @@ public class DefaultContainerBuilder : IContainerBuilder
...
         public virtual IServiceProvider BuildContainer()
         {
-            /* "services.BuildServiceProvider()" returns IServiceProvider in Microsoft.Extensions.DependencyInjection 1.0 and ServiceProvider in Microsoft.Extensions.DependencyInjection 2.0
-            * (This is a breaking change)[https://github.com/aspnet/DependencyInjection/issues/550].
-            * To support both versions with the same code base in OData/WebAPI we decided to call that extension method using reflection.
-            * More info at https://github.com/OData/WebApi/pull/1082
-            */
-
-            MethodInfo buildServiceProviderMethod = typeof(ServiceCollectionContainerBuilderExtensions).GetMethod(nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider), new[] { typeof(IServiceCollection) });
+            try
```
corranrogue9, line 86: To avoid using exceptions as control flow logic and to avoid masking unexpected exceptions from `BuildServiceProvider`, we could do something like:
```
var buildServiceProviderMethod = typeof(ServiceCollectionContainerBuilderExtensions).GetMethod(
  nameof(ServiceCollectionContainerBuilderExtensions.BuildServiceProvider),
  new[] { typeof(IServiceCollection) });
if (buildServiceProviderMethod.ReturnType == typeof(ServiceProvider))
{
  return services.BuildServiceProvider();
}
else
{
  return (IServiceProvider)buildServiceProviderMethod.Invoke(null, new object[] { services });
}
```
To be even more helpful, we could do:
```
else if (buildServiceProviderMethod.ReturnType == typeof(IServiceProvider))
{
  return (IServiceProvider)buildServiceProviderMethod.Invoke(null, new object[] { services });
}
else
{
  throw new InvalidOperationException("No service provider method was found with the correct return type");
}
```

tools/WebStack.versions.settings.targets
corranrogue9: I don't understand this change. Also, why wouldn't it be `7.6.0` or `7.5.15`?

KenitoInc: You could have left as it is and only change in the release PR

KenitoInc (approved):

tools/WebStack.versions.settings.targets
xuzhg: As i mentioned, i will revert this change in this PR and will have another pr to update the version.

src/Microsoft.AspNet.OData.Shared/DefaultContainerBuilder.cs
mikepizzo: alternatively, can we check for the conditions that would cause services.BuildServiceProvider() to fail, rather than waiting for it to fail and catching an exception?  what is the failure?

If we do have to catch, is there a specific exception that we can catch, rather than catching any exception?

corranrogue9 (approved):

PR #2652, opened 2022-05-31
Fix/issue 551

### LINKED ISSUE #1572: OData deserializer swallows errors if ILogger service is registered
Model binding errors for malformed data when invoking Post() methods (and probably others) are not propagated to the controller method if a service with the `Microsoft.Extensions.Logging.ILogger` interface is registered.

### Assemblies tested
Microsoft.AspNetCore.OData 7.0.1 (from Nuget)

### Reproduction steps

I haven't had time to put together a complete sample project for the issue (my test project includes a number of additional bits specific to our environment), but I did pull out enough into a gist that should be able to demonstrate the issue. If I get some additional time this week, I'll try to put together a more complete sample solution. The gist should be reachable here: https://gist.github.com/nkilzer/63d676bd8ad5817d6708b1fca94ccc44

The issue occurs when sending malformed (e.g. not conforming to datatype, or valdation rules) data to the `[FromBody]` parameter of a `Post()` endpoint - but only when a service for `Microsoft.Extensions.Logging.ILogger` is registered in the application service provider.

### Expected results

The `ModelState.IsValid` property would return false, and `ModelState` would contain a collection of errors about what properties were invalid, and why (as in all normal model-binding scenarios).

### Actual Results

When attempting to send invalid (malformed, etc.) data to the `Post()` endpoint to determine how best to handle errors, I was consistently getting a result where the incoming model parameter would be null, but the `ModelState.IsValid` value was returning true, and there were no errors in the `ModelState` collection.

If the service is not registered, the model binder appears to propagate an error state correctly to the controller method.

### Additional details

I was able to isolate the root cause to this block by creating a decorator for `ILogger` and checking the call-stack at the point that the logging method was called:
https://github.com/OData/WebApi/blob/f9d10191efcb13fee7f995fa4ec2188860d8c6fd/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs#L130-L139

Given that the commit comment indicates that this was part of the addition of OData batching support, I'm not sure I have a useful contribution to what an appropriate fix is - just going back to throwing the exception in all cases seems like it would break the batching behavior.

### Workarounds

My current workaround is to ensure that at any site where `ModelState.IsValid` is checked, that there is also a check for `IsValid && input == null`, as I don't have any use cases that would involve a potentially empty object being posted to the endpoint.

Nthemba:
### Issues

*This pull request fixes #1572 and https://github.com/OData/AspNetCoreOData/issues/551*

### Description

*Model state errors are ignored when sending a post request when ILogger service is registered.*
*A flag has been added to avoid introducing a breaking change. To continue with the existing behavior ( not throwing an error after logging model state error), do not add the flag.*
*However if the expected behavior is to throw the model state error after logging it , add the compatibility flag to your startup as shown below.*

routebuilder.SetCompatibilityOptions(CompatibilityOptions.ThrowExceptionAfterLoggingModelStateError);

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
@@ -31,8 +31,8 @@ namespace Microsoft.AspNet.OData
         DisableCaseInsensitiveRequestPropertyBinding = 0x2,
 
         /// <summary>
-        /// Throw exeception and add ModelState error if ILogger is defined.
+        /// Throw exception after logging ModelState error.
         /// </summary>
-        ThrowModelStateExceptionIfILoggerDefined = 0x4
+        ThrowExceptionAfterLoggingModelStateError = 0x4
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
@@ -28,6 +28,11 @@ namespace Microsoft.AspNet.OData
         /// <summary>
         /// Disable case-insensitive request property binding.
         /// </summary>
-        DisableCaseInsensitiveRequestPropertyBinding = 0x2
+        DisableCaseInsensitiveRequestPropertyBinding = 0x2,
+
+        /// <summary>
+        /// Throw exeception and add ModelState error if ILogger is defined.
+        /// </summary>
+        ThrowModelStateExceptionIfILoggerDefined = 0x4
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
@@ -142,6 +142,7 @@ namespace Microsoft.AspNet.OData.Formatter
                     }
 
                     logger.LogError(ex, String.Empty);
+                    throw ex;
                 };
 
                 List<IDisposable> toDispose = new List<IDisposable>();
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
@@ -140,9 +140,11 @@ namespace Microsoft.AspNet.OData.Formatter
                     {
                         throw ex;
                     }
-
                     logger.LogError(ex, String.Empty);
-                    throw ex;
+                    if (context.HttpContext.Request.GetCompatibilityOptions().HasFlag(CompatibilityOptions.ThrowModelStateExceptionIfILoggerDefined))
+                    {
+                        throw ex;
+                    }
                 };
 
                 List<IDisposable> toDispose = new List<IDisposable>();
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
@@ -140,8 +140,9 @@ namespace Microsoft.AspNet.OData.Formatter
                     {
                         throw ex;
                     }
+
                     logger.LogError(ex, String.Empty);
-                    if (context.HttpContext.Request.GetCompatibilityOptions().HasFlag(CompatibilityOptions.ThrowModelStateExceptionIfILoggerDefined))
+                    if (context.HttpContext.Request.GetCompatibilityOptions().HasFlag(CompatibilityOptions.ThrowExceptionAfterLoggingModelStateError))
                     {
                         throw ex;
                     }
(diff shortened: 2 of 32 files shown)


src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
```diff
@@ -142,6 +142,7 @@ public override async Task<InputFormatterResult> ReadRequestBodyAsync(InputForma
                     }
 
                     logger.LogError(ex, String.Empty);
+                    throw ex;
```
corranrogue9, line 145: Throwing these exceptions is probably going to break some amount of callers. Can we add a behavior flag for this so that we default to the current behavior, and it can be toggled to get the exceptions?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Formatter/ODataInputFormatterTest.cs
```diff
@@ -0,0 +1,111 @@
...
+                request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json"));
+                request.Content = new StringContent(requestbody);
+                
+                // Act
+                using (HttpResponseMessage response = await client.SendAsync(request))
+                {
+                    // Assert
+                    Assert.NotNull(response);
+                    Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
+                }
```
corranrogue9, line 76:
```suggestion
                using (var content = new StringContent(requestbody))
                {
                    request.Content = content;
                
                    // Act
                    using (HttpResponseMessage response = await client.SendAsync(request))
                    {
                        // Assert
                        Assert.NotNull(response);
                        Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
                    }
                }
```

Nthemba:
> For that reason I'd be interested to know what happens when the exception is thrown inside of a batch request.

Changes done on the code base do not alter results of batch requests when ILogger is defined or not or a flag is added or not.

corranrogue9 (dismissed):

gathogojr (dismissed):
:shipit:

src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
```diff
@@ -28,6 +28,11 @@ public enum CompatibilityOptions
         /// <summary>
         /// Disable case-insensitive request property binding.
         /// </summary>
-        DisableCaseInsensitiveRequestPropertyBinding = 0x2
+        DisableCaseInsensitiveRequestPropertyBinding = 0x2,
+
+        /// <summary>
+        /// Throw exeception and add ModelState error if ILogger is defined.
```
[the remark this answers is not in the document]
gathogojr: The reasoning here is that you can only log the `ModelState` error if the `ILogger` service is registered in the DI container. In any case the current behaviour is throw the exception when the `ILogger` service is not registered in the DI container

src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
```diff
@@ -28,6 +28,11 @@ public enum CompatibilityOptions
...
+        /// </summary>
+        ThrowModelStateExceptionIfILoggerDefined = 0x4
```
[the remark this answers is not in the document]
gathogojr: Purely a suggestion if you find that it makes sense

src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs
```diff
@@ -140,8 +140,11 @@ public override async Task<InputFormatterResult> ReadRequestBodyAsync(InputForma
                     {
                         throw ex;
                     }
-
```
gathogojr, line 143: Restore this line. Across the repo, the practice is to leave an empty line after the closing brace, unless it's followed by another closing brace

corranrogue9 (approved):

ElizabethOkerio (dismissed):

habbes:
@Nthemba  Seems like updated the  baseline tests for AspNetCore but not AspNet classic. Do your changes also apply for AspNet classic?

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ServerFactory.cs
```diff
@@ -47,8 +48,9 @@ public static TestServer Create(Type[] controllers, Action<IRouteBuilder> config
                 services.AddMvc();
 #else
                 services.AddMvc(options => options.EnableEndpointRouting = false)
-                    .AddNewtonsoftJson();
+                    .AddNewtonsoftJson();
```
KenitoInc, line 51: Remove the whitespace

test/UnitTest/Microsoft.AspNetCore.OData.Test/Abstraction/ServerFactory.cs
```diff
@@ -156,5 +158,7 @@ protected override bool IsController(TypeInfo typeInfo)
                 return true;
             }
         }
+
```
KenitoInc, line 161: Remove one line space

KenitoInc (changes requested):
Fix public api tests that are failing

gathogojr (approved):

PR #2655, opened 2022-06-20
Update ODL dependency version to 7.11.1

xuzhg:
### Description

We unlist 7.11.0, so let's update ODL dependency version to 7.11.1.

- [x] *Test cases added*

### Additional work necessary


corranrogue9:
We just released `7.12.0`, do we want to take that instead?

corranrogue9 (dismissed):

xuzhg:
> We just released `7.12.0`, do we want to take that instead?

@corranrogue9 @KenitoInc I noticed 7.12.0 contains only one fix for OData.Client, right? 
Web API OData nuget package doesn't reply on OData.Client. Do you think it's better to update to that version?

corranrogue9:
> > We just released `7.12.0`, do we want to take that instead?
> 
> @corranrogue9 @KenitoInc I noticed 7.12.0 contains only one fix for OData.Client, right? Web API OData nuget package doesn't reply on OData.Client. Do you think it's better to update to that version?

Nope, that's fine

PR #2673, opened 2022-07-07
fix delta response serialization error in v4

ElizabethOkerio:
### Issues

### Description
A response for a deep update of a nested delta resource set is truncated if the OData-Version is not specified as v4.01 in the request. This PR removes the check for the OData-Version. This means that the OData-Version need not be passed in the request and the response should be serialized properly.

The fix is to use version v4.01 to serialize delta responses. 

- [ ] *Test cases added*

### Additional work necessary


KenitoInc (commented):
Minor comments.
Overrall I'm happy with the change.

PR #2674, opened 2022-07-07
Serialize navigation properties in Deep insert responses

KenitoInc:
### Issues
In the existing implementation, we serialize navigation properties in 2 scenarios:
- Responses when we use `[AutoExpand]` attribute in controller methods.
- Responses when we use `$expand` query option in GET requests.
In Deep insert, we make a POST request that has nested resources (navigation properties). We expect the response to contain the nested resources. However the `ODataResourceSerializer`doesn't know how to handle Deep insert responses.

### Description
When serializing the response, we check if the request was a POST request. If true, we write the nested resources.
We have introduced new methods `WriteNestedNavigationProperties` and `WriteNestedNavigationPropertiesAsync` which only work when the request is a `POST` request.

Weirdly, `WriteNavigationLink` methods does the following for navigation properties that haven't been expanded:
```csharp
writer.WriteStart(navigationLink);
writer.WriteEnd()
```
We have updated the code to ensure that for `POST` requests, the above logic in `WriteNavigationLink` is not executed. We will write the `navigationLink` in the `WriteNestedNavigationProperties` method.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/EdmChangedObjectCollection.cs
+++ b/src/Microsoft.AspNet.OData.Shared/EdmChangedObjectCollection.cs
@@ -71,7 +71,6 @@ namespace Microsoft.AspNet.OData
             _entityType = entityType;
             _edmType = new EdmDeltaCollectionType(new EdmEntityTypeReference(_entityType, isNullable: true));
             _edmTypeReference = new EdmCollectionTypeReference(_edmType);
-
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.OData.Shared/EdmChangedObjectCollection.cs
+++ b/src/Microsoft.AspNet.OData.Shared/EdmChangedObjectCollection.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNet.OData
         private IEdmEntityType _entityType;
         private EdmDeltaCollectionType _edmType;
         private IEdmCollectionTypeReference _edmTypeReference;
-    
+
         /// <summary>
         /// Initializes a new instance of the <see cref="EdmChangedObjectCollection"/> class.
         /// </summary>
@@ -71,7 +71,7 @@ namespace Microsoft.AspNet.OData
             _entityType = entityType;
             _edmType = new EdmDeltaCollectionType(new EdmEntityTypeReference(_entityType, isNullable: true));
             _edmTypeReference = new EdmCollectionTypeReference(_edmType);
-            
+
         }
 
         /// <summary>
@@ -102,7 +102,7 @@ namespace Microsoft.AspNet.OData
         internal EdmChangedObjectCollection CopyChangedValues(EdmODataAPIHandler apiHandler, ODataEdmAPIHandlerFactory apiHandlerFactory = null)
         {
             EdmChangedObjectCollection changedObjectCollection = new EdmChangedObjectCollection(_entityType);
-            string[] keys = _entityType.Key().Select(x=>x.Name).ToArray();
+            string[] keys = _entityType.Key().Select(x => x.Name).ToArray();
 
             foreach (IEdmChangedObject changedObj in Items)
             {
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -1411,7 +1411,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
 
                 foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
                 {
-                    if (changedProperties == null || changedProperties.Contains(navigationProperty.Name))
+                    if (changedProperties != null && changedProperties.Contains(navigationProperty.Name))
                     {
                         yield return new KeyValuePair<IEdmNavigationProperty, Type>(navigationProperty, typeof(IEdmChangedObject));
                     }
@@ -1426,9 +1426,12 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                 {
                     object obj = null;
 
-                    if (changedProperties == null || changedProperties.Contains(navigationProperty.Name) && delta.DeltaNestedResources.TryGetValue(navigationProperty.Name, out obj))
+                    if (changedProperties != null && changedProperties.Contains(navigationProperty.Name) && delta.DeltaNestedResources.TryGetValue(navigationProperty.Name, out obj))
                     {
-                        yield return new KeyValuePair<IEdmNavigationProperty, Type>(navigationProperty, obj.GetType());
+                        if (obj != null)
+                        {
+                            yield return new KeyValuePair<IEdmNavigationProperty, Type>(navigationProperty, obj.GetType());
+                        }
                     }
                 }
             }
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -1378,7 +1378,8 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             if (complexProperties != null)
             {
                 IEnumerable<string> changedProperties = null;
-                if (null != resourceContext.EdmObject && resourceContext.EdmObject.IsDeltaResource())
+
+                if (resourceContext.EdmObject != null && resourceContext.EdmObject.IsDeltaResource())
                 {
                     IDelta deltaObject = resourceContext.EdmObject as IDelta;
                     changedProperties = deltaObject.GetChangedPropertyNames();
@@ -1405,7 +1406,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
 
             IEnumerable<string> changedProperties = null;
 
-            if (null != resourceContext.EdmObject && resourceContext.EdmObject is IDelta changedObject)
+            if (resourceContext.EdmObject != null && resourceContext.EdmObject is IDelta changedObject)
             {
                 changedProperties = changedObject.GetChangedPropertyNames();
 
@@ -1417,7 +1418,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                     }
                 }
             }
-            else if (null != resourceContext.ResourceInstance && resourceContext.ResourceInstance is IDelta deltaObject)
+            else if (resourceContext.ResourceInstance != null && resourceContext.ResourceInstance is IDelta deltaObject)
             {
                 changedProperties = deltaObject.GetChangedPropertyNames();
                 dynamic delta = deltaObject;
@@ -1825,7 +1826,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             {
                 IEnumerable<IEdmStructuralProperty> structuralProperties = selectExpandNode.SelectedStructuralProperties;
 
-                if (null != resourceContext.EdmObject && resourceContext.EdmObject.IsDeltaResource())
+                if (resourceContext.EdmObject != null && resourceContext.EdmObject.IsDeltaResource())
                 {
                     IDelta deltaObject = resourceContext.EdmObject as IDelta;
                     IEnumerable<string> changedProperties = deltaObject.GetChangedPropertyNames();
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
 using System.Linq;
@@ -1651,8 +1652,8 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
 
         private void WriteNestedNavigationProperties(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
         {
-            Contract.Assert(resourceContext != null);
-            Contract.Assert(writer != null);
+            Debug.Assert(resourceContext != null);
+            Debug.Assert(writer != null);
 
             if (!resourceContext.IsPostRequest)
             {
@@ -1677,8 +1678,8 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
 
         private async Task WriteNestedNavigationPropertiesAsync(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
         {
-            Contract.Assert(resourceContext != null);
-            Contract.Assert(writer != null);
+            Debug.Assert(resourceContext != null);
+            Debug.Assert(writer != null);
 
             if (!resourceContext.IsPostRequest)
             {
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -1655,8 +1655,8 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
 
         private void WriteNestedNavigationProperties(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
         {
-            Debug.Assert(resourceContext != null);
-            Debug.Assert(writer != null);
+            Debug.Assert(resourceContext != null, "resourceContext != null");
+            Debug.Assert(writer != null, "writer != null");
 
             if (!resourceContext.IsPostRequest)
             {
@@ -1681,8 +1681,8 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
 
         private async Task WriteNestedNavigationPropertiesAsync(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
         {
-            Debug.Assert(resourceContext != null);
-            Debug.Assert(writer != null);
+            Debug.Assert(resourceContext != null, "resourceContext != null");
+            Debug.Assert(writer != null, "writer != null");
 
             if (!resourceContext.IsPostRequest)
             {
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -647,6 +647,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                         WriteDynamicComplexProperties(resourceContext, writer);
                         WriteNavigationLinks(selectExpandNode, resourceContext, writer);
                         WriteExpandedNavigationProperties(selectExpandNode, resourceContext, writer);
+                        WriteNestedNavigationProperties(selectExpandNode, resourceContext, writer);
                         WriteReferencedNavigationProperties(selectExpandNode, resourceContext, writer);
                         writer.WriteEnd();
                     }
@@ -689,6 +690,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                         await WriteDynamicComplexPropertiesAsync(resourceContext, writer);
                         await WriteNavigationLinksAsync(selectExpandNode, resourceContext, writer);
                         await WriteExpandedNavigationPropertiesAsync(selectExpandNode, resourceContext, writer);
+                        await WriteNestedNavigationPropertiesAsync(selectExpandNode, resourceContext, writer);
                         await WriteReferencedNavigationPropertiesAsync(selectExpandNode, resourceContext, writer);
                         await writer.WriteEndAsync();
                     }
@@ -1118,7 +1120,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             Contract.Assert(selectExpandNode != null);
             Contract.Assert(resourceContext != null);
 
-            if (selectExpandNode.SelectedNavigationProperties == null)
+            if (selectExpandNode.SelectedNavigationProperties == null || resourceContext.IsPostRequest)
             {
                 return;
             }
@@ -1139,7 +1141,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             Contract.Assert(selectExpandNode != null);
             Contract.Assert(resourceContext != null);
 
-            if (selectExpandNode.SelectedNavigationProperties == null)
+            if (selectExpandNode.SelectedNavigationProperties == null || resourceContext.IsPostRequest)
             {
                 return;
             }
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -647,6 +647,7 @@ internal async Task WriteDeltaNavigationPropertiesAsync(SelectExpandNode selectE
                         WriteDynamicComplexProperties(resourceContext, writer);
                         WriteNavigationLinks(selectExpandNode, resourceContext, writer);
                         WriteExpandedNavigationProperties(selectExpandNode, resourceContext, writer);
+                        WriteNavigationPropertiesForDeepInsert(selectExpandNode, resourceContext, writer);
```
ElizabethOkerio, line 650: is this necessary here? will we support deltas in a deep insert?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1433,6 +1435,39 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+                propertyNamesAndValues.Add(name, value);
+            }
+
+            if (navigationProperties != null)
+            {
+                foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
+                {
+                    if (propertyNamesAndValues.ContainsKey(navigationProperty.Name))
+                    {
+                        /*object obj = propertyNamesAndValues[navigationProperty.Name];
```
ElizabethOkerio, line 1458: remove this if not needed

src/Microsoft.AspNetCore.OData/ResourceContext.cs
```diff
@@ -36,5 +36,13 @@ public HttpRequest Request
                 SerializerContext.Request = value;
             }
         }
+
+        internal bool IsPostRequest
```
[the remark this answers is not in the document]
KenitoInc: Let's keep generic since we are checking if it's a `POST` request. We may re-use it somewhere else.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -689,6 +690,7 @@ internal async Task WriteDeltaNavigationPropertiesAsync(SelectExpandNode selectE
                         await WriteDynamicComplexPropertiesAsync(resourceContext, writer);
                         await WriteNavigationLinksAsync(selectExpandNode, resourceContext, writer);
                         await WriteExpandedNavigationPropertiesAsync(selectExpandNode, resourceContext, writer);
+                        await WriteNavigationPropertiesForDeepInsertAsync(selectExpandNode, resourceContext, writer);
```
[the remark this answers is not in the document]
KenitoInc: Currently we don't deserialize deltas in deep insert. If that changes, we will adjust accordingly.
We are calling this method inside `WriteResourceAsync` since nested resources in deep insert are deserialized as `T` not `Delta<T>`

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1139,7 +1141,7 @@ private async Task WriteNavigationLinksAsync(SelectExpandNode selectExpandNode,
             Contract.Assert(selectExpandNode != null);
             Contract.Assert(resourceContext != null);
 
-            if (selectExpandNode.SelectedNavigationProperties == null)
+            if (selectExpandNode.SelectedNavigationProperties == null || resourceContext.IsPostRequest)
```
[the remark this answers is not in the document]
KenitoInc: I noticed something weird in this logic.
For some background, all navigation properties that are not "expanded" are in the `SelectedNavigationProperties`.
So for each, we call:
```csharp
    await writer.WriteStartAsync(nestedResourceInfo);
    await writer.WriteEndAsync();
```
Weirdly, the above logic doesn't write anything in the output.
This logic has issues when serializing deep insert resources. In this case, we haven't expanded anything so we will call the above methods in the `WriteNavigationLinksAsync` logic then call the methods below in `WriteNavigationPropertiesForDeepInsertAsync` method.
```csharp
    await writer.WriteStartAsync(nestedResourceInfo);
    await WriteComplexAndExpandedNavigationPropertyAsync(navigationProperty.Key, null, resourceContext, writer);
    await writer.WriteEndAsync();
```
Calling `WriteStartAsync(nestedResourceInfo);` twice causes the `DuplicatePropertyNameChecker` to throw an exception thinking we are writing 2 nested properties with the same name.

This is why for deep insert we avoid calling `WriteStartAsync(nestedResourceInfo);` in the the `WriteNavigationLinksAsync` method since we will call it again in the `WriteNavigationPropertiesForDeepInsertAsync` method.

ElizabethOkerio:
add tests

KenitoInc:
> add tests

Noted.
Waiting for the handlers PR first to get in

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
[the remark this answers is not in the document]
gathogojr: @KenitoInc Based on your explanation above, that logic feels a bit brittle

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1433,6 +1435,37 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+                string name = propertyInfo.Name;
+                object value = propertyInfo.GetValue(instance, null);
+                propertyNamesAndValues.Add(name, value);
+            }
+
+            if (navigationProperties != null)
+            {
+                foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
+                {
+                    if (propertyNamesAndValues.ContainsKey(navigationProperty.Name))
```
gathogojr, line 1456: Use `TryGetValue` instead of `ContainsKey`

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1433,6 +1435,37 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+        {
+            ISet<IEdmNavigationProperty> navigationProperties = selectExpandNode.SelectedNavigationProperties;
+            object instance = resourceContext.ResourceInstance;
+            PropertyInfo[] properties = instance.GetType().GetProperties();
+            Dictionary<string, object> propertyNamesAndValues = new Dictionary<string, object>();
+
+            foreach (PropertyInfo propertyInfo in properties)
+            {
+                string name = propertyInfo.Name;
+                object value = propertyInfo.GetValue(instance, null);
```
gathogojr, line 1448: The reason we use this overload of `GetValue` as opposed to the one that takes one argument is?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1586,7 +1619,7 @@ private async Task WriteComplexAndExpandedNavigationPropertyAsync(IEdmProperty e
 
             object propertyValue = resourceContext.GetPropertyValue(edmProperty.Name);
 
-            if (propertyValue == null || propertyValue is NullEdmComplexObject)
+            if (propertyValue == null || propertyValue is NullEdmComplexObject || propertyValue is ODataIdContainer)
```
gathogojr, line 1622: Curious if `is` would throw an exception the l-value is null... If no, the null check is superfluous

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1433,6 +1435,37 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
             }
         }
 
+        private IEnumerable<KeyValuePair<IEdmNavigationProperty, Type>> GetNavigationPropertiesForDeepInsert(SelectExpandNode selectExpandNode, ResourceContext resourceContext)
```
mikepizzo, line 1438: >GetNavigationPropertiesForDeepInsert [](http://example.com/codeflow?start=72&length=36)

Rather than introduce a new GetNavigationPropertiesForDeepInsert (and WriteNavigationPropertiesForDeepInsert[Async]), can we just have the existing GetNavigationPropertiesToWrite understand the deep insert scenario and return the appropriate properties to write there?

src/Microsoft.AspNetCore.OData/ResourceContext.cs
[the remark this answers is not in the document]
KenitoInc: Good catch!

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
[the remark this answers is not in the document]
KenitoInc: Resolved

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1433,6 +1435,37 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+            Dictionary<string, object> propertyNamesAndValues = new Dictionary<string, object>();
+
+            foreach (PropertyInfo propertyInfo in properties)
+            {
+                string name = propertyInfo.Name;
+                object value = propertyInfo.GetValue(instance, null);
+                propertyNamesAndValues.Add(name, value);
+            }
+
+            if (navigationProperties != null)
```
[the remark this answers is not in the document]
KenitoInc: The return type of this method is an `IEnumerable<T>` so we use `yield break` instead.
Fixed

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: Actually, no specific reason. `GetValue(instance);` should work.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: Fixed

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: `Logical OR` is evaluated from left to right. If the `propertyValue == null` is `true`, the rest are not evaluated.
See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-or-operator-

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: Refactored code to use the `GetNavigationPropertiesToWrite` method.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
gathogojr: I don't think you understood what I meant...
Let me rephrase
Would the following throw an exception if `propertyValue` is null?
```csharp
if (propertyValue is NullEdmComplexObject)
{
    // ...
}
```

ElizabethOkerio: I think from his explanation, he means if `propertyValue==null` then

```csharp
if (propertyValue is NullEdmComplexObject)
{
    // ...
}
```
is not evaluated

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationTest.cs
```diff
@@ -978,13 +978,15 @@ public async Task PostEmployee_WithCreateFriends()
             StringContent stringContent = new StringContent(content: content, encoding: Encoding.UTF8, mediaType: "application/json");
             requestForPatch.Content = stringContent;
 
+            var expected = "Friends\":[{\"Id\":1001,\"Name\":\"Friend 1001\",\"Age\":31},{\"Id\":1002,\"Name\":\"Friend 1002\",\"Age\":32},{\"Id\":1003,\"Name\":\"Friend 1003\",\"Age\":33}]";
```
ElizabethOkerio, line 981: is there a nested property here?

ElizabethOkerio: Maybe you can add a test with one object and a nested navigation property within the object.

src/Microsoft.AspNet.OData.Shared/EdmChangedObjectCollection.cs
```diff
@@ -71,7 +71,7 @@ private void Initialize(IEdmEntityType entityType)
             _entityType = entityType;
             _edmType = new EdmDeltaCollectionType(new EdmEntityTypeReference(_entityType, isNullable: true));
             _edmTypeReference = new EdmCollectionTypeReference(_edmType);
-            
+
```
ElizabethOkerio, line 74: nit: remove whitespace.

src/Microsoft.AspNet.OData.Shared/ResourceContext.cs
```diff
@@ -95,6 +95,18 @@ public IEdmModel EdmModel
...
         }
 
+        internal bool IsPostRequest
+        {
+            get
+            {
+#if NETCORE
+                return Request == null ? false : String.Equals(Request.Method, "post", StringComparison.OrdinalIgnoreCase);
+#else
+                return Request == null ? false : String.Equals(Request.Method.ToString(), "post", StringComparison.OrdinalIgnoreCase);
```
habbes, line 105: What is the type of `Request.Method` here?

habbes:
My initial comments and questions from my first-pass review:

What if we have POST request that's not a deep insert, should the same behaviour still apply? Does it write all navigation properties in the object being serialized, or just the ones that were referenced in the deep-insert request?

Does this also affect `POST` requests to an action or to a `/$query` endpoint (which is more of a GET request)?

Does the `WriteNavigationLink` also emit no output when using full metadata mode?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationTest.cs
KenitoInc: In the request payload, `Friends` is a nested (navigation) property.
This test checked if the Friends nested property is in the response.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1399,65 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+            if (navigationProperties == null)
             {
-                IEnumerable<string> changedProperties = null;
-                if (null != resourceContext.EdmObject && resourceContext.EdmObject is IDelta changedObject)
+                yield break;
+            }
+
+            IEnumerable<string> changedProperties = null;
+
+            if (resourceContext.EdmObject != null && resourceContext.EdmObject is IDelta changedObject)
```
gathogojr, line 1409:
```suggestion
            if (resourceContext.EdmObject is IDelta changedObject)
```
Since `resourceContext.EdmObject is IDelta` would evaluate to false if `resourceContext.EdmObject` is `null` - I BELIEVE!

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1399,65 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+
+                foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
+                {
+                    if (changedProperties == null || changedProperties.Contains(navigationProperty.Name))
+                    {
+                        yield return new KeyValuePair<IEdmNavigationProperty, Type>(navigationProperty, typeof(IEdmChangedObject));
+                    }
+                }
+            }
+            else if (resourceContext.ResourceInstance != null && resourceContext.ResourceInstance is IDelta deltaObject)
```
gathogojr, line 1421:
```suggestion
            else if (resourceContext.ResourceInstance is IDelta deltaObject)
```
Since `resourceContext.ResourceInstance is IDelta` would evaluate to `false` if `resourceContext.ResourceInstance` is `null` - I BELIEVE!

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1399,65 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+                changedProperties = deltaObject.GetChangedPropertyNames();
+                dynamic delta = deltaObject;
+
+                foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
                 {
-                    changedProperties = changedObject.GetChangedPropertyNames();
+                    Object obj = null;
 
-                    foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
+                    if (changedProperties == null || changedProperties.Contains(navigationProperty.Name) && delta.DeltaNestedResources.TryGetValue(navigationProperty.Name, out obj))
```
gathogojr, line 1430: Add some brackets in the conditional expressions to make it clear whether `changedProperties == null || changedProperties.Contains(navigationProperty.Name)` or `changedProperties.Contains(navigationProperty.Name) && delta.DeltaNestedResources.TryGetValue(navigationProperty.Name, out obj)` is evaluated first

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1399,65 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+            }
+            else if (resourceContext.ResourceInstance != null && resourceContext.ResourceInstance is IDelta deltaObject)
+            {
+                changedProperties = deltaObject.GetChangedPropertyNames();
+                dynamic delta = deltaObject;
+
+                foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
                 {
-                    changedProperties = changedObject.GetChangedPropertyNames();
+                    Object obj = null;
```
gathogojr, line 1428:
```suggestion
                    object obj = null;
```

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1618,6 +1651,58 @@ private async Task WriteComplexAndExpandedNavigationPropertyAsync(IEdmProperty e
             }
         }
 
+        private void WriteNestedNavigationProperties(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
+        {
+            Contract.Assert(resourceContext != null);
```
gathogojr, line 1656: Why `Contract.Assert` and not `Debug.Assert` in this private method?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1618,6 +1651,58 @@ private async Task WriteComplexAndExpandedNavigationPropertyAsync(IEdmProperty e
...
+
+                writer.WriteStart(nestedResourceInfo);
+                WriteComplexAndExpandedNavigationProperty(navigationProperty.Key, null, resourceContext, writer);
+                writer.WriteEnd();
+            }
+        }
+
+        private async Task WriteNestedNavigationPropertiesAsync(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
+        {
+            Contract.Assert(resourceContext != null);
```
gathogojr, line 1682: Why `Contract.Assert` and not `Debug.Assert` in this private method?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1399,65 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+
+            IEnumerable<string> changedProperties = null;
+
+            if (resourceContext.EdmObject != null && resourceContext.EdmObject is IDelta changedObject)
+            {
+                changedProperties = changedObject.GetChangedPropertyNames();
+
+                foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
+                {
+                    if (changedProperties == null || changedProperties.Contains(navigationProperty.Name))
```
gathogojr, line 1415: If `changedProperties` collection is null, why do we still return that property as a navigation property to write? What about when the collection is empty? Should we treat it the same way we're treating null?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1399,65 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
 
-            if (navigationProperties != null)
+            if (navigationProperties == null)
             {
-                IEnumerable<string> changedProperties = null;
-                if (null != resourceContext.EdmObject && resourceContext.EdmObject is IDelta changedObject)
+                yield break;
+            }
+
+            IEnumerable<string> changedProperties = null;
```
gathogojr, line 1407: Why does `changedProperties` need to be declared here instead of within the respective enclosing blocks?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: I was just consistent with what is in other private methods of the `ODataResourceSerializer`.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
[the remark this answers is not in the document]
habbes: > Weirdly, the above logic doesn't write anything in the output.

There is some metadata which is only emitted when serializing in full-metadata mode. I think that could be one of them.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: Updated to `Debug.Assert`

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: Updated to `Debug.Assert`

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: I think Sreejith left that code that way intentionally. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#operator-precedence

habbes:
Just wanted to shed some light on what `WriteNavigationLink` achieves with:
```c#
WriteStart(nestedResourceInfo);
WriteEnd();
```

Besides validating against duplicate properties, this also emits association and navigation links of navigation properties when the metadata level is full , i.e. when the header is `Accept: application/json; odata.metadata=full`.

image

These links are also emitted when the navigation properties are selected in the response, even if they are not expanded. For example, the links will be written in both `People?$select=Friends` and `People?$expand=Friends`. In the first case, the links will be written but the nested resources won't be expanded.

The links are not emitted in other odata metadata levels.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1396,35 +1400,63 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
...
+                    object obj = null;
 
-                    foreach (IEdmNavigationProperty navigationProperty in navigationProperties)
+                    if (changedProperties == null || changedProperties.Contains(navigationProperty.Name) && delta.DeltaNestedResources.TryGetValue(navigationProperty.Name, out obj))
                     {
-                        if (changedProperties == null || changedProperties.Contains(navigationProperty.Name))
-                        {
-                            yield return new KeyValuePair<IEdmNavigationProperty, Type>(navigationProperty, typeof(IEdmChangedObject));
-                        }
+                        yield return new KeyValuePair<IEdmNavigationProperty, Type>(navigationProperty, obj.GetType());
```
habbes, line 1431: Could you clarify the change from `typeof(IEdmChangeObject)` to `obj.GetType()`

KenitoInc: The type didn't change. I refactored the method.
In the `if` code block, we have `typeof(IEdmChangeObject)`
In the `else if` code block, we have `obj.GetType()`

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
gathogojr: The answer to my question here is?

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
gathogojr: @KenitoInc RE: https://github.com/OData/WebApi/pull/2674/files#r979664329
Operator precedence tables are one thing and providing clarity to the naked eye is another. It's for the same reason why I wouldn't feel comfortable with this: `18 / 2 * 3`. The compiler will have no problem interpreting that but a fair number of developers would have to refer to the operator precedence tables to interpret it accurately.

KenitoInc: Refactored code. So this now is not an issue

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
KenitoInc: Refactored the code to `changedProperties != null`

gathogojr (approved):
LGTM

PR #2707, opened 2022-10-05
Support deserializing top-level collections of entity types.

mikepizzo:
Enables PUT to an entity collection.

### Issues

Currently we have code that prevents us from deserializing a top level collection of entity types in a request payload.

This prevents us, for example, from doing a PUT to a collection-valued navigation property.

### Description

This PR relaxes the constraint that prevents deserializing a top-level collection of entity types in a request payload.

Separately, ODL requires that, for a request payload, if you provide the entity type you must also specify the entity set.  A simple change in ODL would relax that constraint, in the meantime this PR creates a dummy entity set to pass the reader in order to use the entity type when deserializing the collection. This avoids having to mark up the request payload with the type of each of the elements.

- [ ] *Test cases added*

### Additional work necessary

 - [ ] Create issue for ODL
 - [ ] Add new test cases

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceSetDeserializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceSetDeserializer.cs
@@ -54,7 +54,19 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                 throw Error.Argument("edmType", SRResources.ArgumentMustBeOfType, EdmTypeKind.Complex + " or " + EdmTypeKind.Entity);
             }
 
-            ODataReader resourceSetReader = readContext.IsChangedObjectCollection ? messageReader.CreateODataDeltaResourceSetReader() : messageReader.CreateODataResourceSetReader();
+            IEdmEntityType entityType = edmType.AsCollection().ElementType().Definition as IEdmEntityType;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceSetDeserializer.cs
```diff
@@ -54,7 +54,19 @@ public override object Read(ODataMessageReader messageReader, Type type, ODataDe
...
-            ODataReader resourceSetReader = readContext.IsChangedObjectCollection ? messageReader.CreateODataDeltaResourceSetReader() : messageReader.CreateODataResourceSetReader();
+            IEdmEntityType entityType = edmType.AsCollection().ElementType().Definition as IEdmEntityType;
+            IEdmEntitySet entitySet = null;
+            if(entityType != null)
+            {
+                // CreateODataDeltaResourceSetReader requires an entity set if you pass a type for reading a request.
+                // We can relax that check in ODataJsonlightInputContext, line 670.
+                // In the meantime, creating a dummy entity set allows the reader to know what type its reading so the payload
+                // doesn't have to be marked up with odata.type annotations.
+                IEdmEntityContainer container = new EdmEntityContainer(entityType.Namespace, "");
```
xuzhg, line 65: I remember that we can create the reader using entity type only, right?

PR #2734, opened 2022-12-02
Allow using property-paths with respect to declaring type from model.

mirecg:
Solves property path expressions like Stamp/CreatedByUser/Name, where CreatedByUser is interface (e.g. IMyInterface) which do not directly have a Name property, but the Name is in some base interface up in the hierarchy.

Otherwise it internally throws System.ArgumentException: 'Instance property 'Name' is not defined for type 'IMyInterface'.

It is possible to overcome the issue using a cast expression, but this forces to use hardcoded casts on client calling odata endpoints and have deeper knowledge about model behind it.

### Description

*Using IEdmProperty and IEdmModel in ExpressionBinderBase.GetPropertyExpression() we can create property access expression with regards to declaring type of the property*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
@@ -1176,17 +1176,29 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             else
             {
                 return GetFlattenedPropertyExpression(propertyPath)
-                    ?? ConvertNonStandardPrimitives(ExpressionBinderBase.GetPropertyExpression(source, (this.HasInstancePropertyContainer && !propertyPath.Contains("\\") ? "Instance\\" : String.Empty) + propertyName));
+                    ?? ConvertNonStandardPrimitives(ExpressionBinderBase.GetPropertyExpression(source, (this.HasInstancePropertyContainer && !propertyPath.Contains("\\") ? "Instance\\" : String.Empty) + propertyName, property, Model));
             }
         }
 
-        internal static Expression GetPropertyExpression(Expression source, string propertyPath)
+        internal static Expression GetPropertyExpression(Expression source, string propertyPath, IEdmProperty property = null, IEdmModel model = null)
         {
             string[] propertyNameParts = propertyPath.Split('\\');
             Expression propertyValue = source;
-            foreach (var propertyName in propertyNameParts)
+            for (int i = 0; i < propertyNameParts.Length; i++)
             {
-                propertyValue = Expression.Property(propertyValue, propertyName);
+                // property access over declaring type from model without extra casts
+                // - solves property path expressions like Stamp/CreatedByUser/Name, where CreatedByUser is interface which do not directly have a Name property, but the Name is in some base interface up in the hierarchy
+                // - 'property' argument should correspond to the last part of the given 'propertyPath'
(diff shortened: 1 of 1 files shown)


KenitoInc:
Add tests

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1176,17 +1176,29 @@ internal Expression CreatePropertyAccessExpression(Expression source, IEdmProper
...
+                if (property != null && model != null && i == propertyNameParts.Length - 1)
+                {
+                    var propertyDeclaringClrType = EdmLibHelpers.GetClrType(property.DeclaringType, model);
+                    if (propertyDeclaringClrType != null)
+                        propertyValue = Expression.Property(propertyValue, propertyDeclaringClrType, propertyNameParts[i]);
+                    else
+                        propertyValue = Expression.Property(propertyValue, propertyNameParts[i]);
+                }
+                else
+                    propertyValue = Expression.Property(propertyValue, propertyNameParts[i]);
```
habbes, line 1201: Suggestions to meet library's coding conventions
```suggestion
                    Type propertyDeclaringClrType = EdmLibHelpers.GetClrType(property.DeclaringType, model);
                    if (propertyDeclaringClrType != null)
                    {
                        propertyValue = Expression.Property(propertyValue, propertyDeclaringClrType, propertyNameParts[i]);
                    }
                    else
                    {
                        propertyValue = Expression.Property(propertyValue, propertyNameParts[i]);
                    }
                }
                else
                {
                    propertyValue = Expression.Property(propertyValue, propertyNameParts[i]);
                }
```

src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1176,17 +1176,29 @@ internal Expression CreatePropertyAccessExpression(Expression source, IEdmProper
...
+        internal static Expression GetPropertyExpression(Expression source, string propertyPath, IEdmProperty property = null, IEdmModel model = null)
         {
             string[] propertyNameParts = propertyPath.Split('\\');
             Expression propertyValue = source;
-            foreach (var propertyName in propertyNameParts)
+            for (int i = 0; i < propertyNameParts.Length; i++)
             {
-                propertyValue = Expression.Property(propertyValue, propertyName);
+                // property access over declaring type from model without extra casts
+                // - solves property path expressions like Stamp/CreatedByUser/Name, where CreatedByUser is interface which do not directly have a Name property, but the Name is in some base interface up in the hierarchy
```
habbes, line 1190: When you talk about a "base interface up in the hierarchy", do you mean a type that inerhits from `CreatedByUser` or a type that `CreatedByUser` inherits from?

Why not use that type that has the property in the path expressions instead of using a type that is not guaranteed to have that property?

Could you share more information about concrete scenarios where this issue occurs and that this PR aims to fix?

ElizabethOkerio:
@mirecg  Add tests and also respond to some of the questions already asked.

WanjohiSammy:
@mirecg Also link the Issue ID associated with this PR if available.

WanjohiSammy:
@mirecg as there has been no activity or response on this PR for a while, I am going to close it for now. If you would like to continue working on it, please feel free to reopen it.

PR #2751, opened 2023-02-16
AutoExpand does not expand beyond non-default MaxExpansionDepth

### LINKED ISSUE #2748: AutoExpand only goes 2 levels deep irrespective of MaxExpansionDepth
My understanding of AutoExpand is that I don't need to explicitly use `$expand` in the query. However, AutoExpand only returns 2 levels deep even if `MaxExpansionDepth` is set to 0 or any other number. It seems like the `Validate` method in `SelectExpandQueryOption` never gets a chance to execute and set `LevelsMaxLiteralExpansionDepth` to a meaningful number (more details below).

### Assemblies affected

Microsoft.AspNet.Odata 7.6.3

### Reproduce steps

My models look like below:

```
public class Menu
{
    public int Id { get; set; }
    [AutoExpand]
    public List<Tab> Tabs { get; set; }
}

public class Tab
{
    public int Id { get; set; }
    [AutoExpand]
    public List<Item> Items { get; set; }
}

public class Item
{
    public int Id { get; set; }
    [AutoExpand]
    public List<Note> Notes { get; set; }
}

public class Note
{
    public int Id { get; set; }
}
```
    
My controller action is as follows:

```
[HttpGet]
[EnableQuery(MaxExpansionDepth = 0)]
[ODataRoute("Menu")]
public Menu Get()
{
    Menu menu = _service.Get();
    return menu;
}
```

My metadata is as follows:

```
<EntitySet Name="Menu" EntityType="EntityModel.Menu">
    <NavigationPropertyBinding Path="Tabs" Target="Tab" />
</EntitySet>

<EntitySet Name="Tab" EntityType="EntityModel.Tab">
    <NavigationPropertyBinding Path="Items" Target="Item" />
</EntitySet>

<EntitySet Name="Item" EntityType="EntityModel.Item">
    <NavigationPropertyBinding Path="Notes" Target="Note" />
</EntitySet>

<EntitySet Name="Note" EntityType="EntityModel.Note" />
```
  

My API request is http://localhost/api/Menu/

### Expected result

My response should be

```
{
  "@odata.context": "http://localhost/api/$metadata#Menu(Tabs(Items(Notes)))/$entity",
  "Id": 1,
  "Tabs": {
    "Id": 1,
    "Items": [
      {
        "Id": 1,
        "Notes": [
          {
            "Id": 1
          }
        ]
      }
    ]
  }
}
```

### Actual result

My response does not include Notes (third level)

```
{
  "@odata.context": "http://localhost/api/$metadata#Menu(Tabs(Items))/$entity",
  "Id": 1,
  "Tabs": {
    "Id": 1,
    "Items": [
      {
        "Id": 1
      }
    ]
  }
}
```

### Additional detail

These are my observations after stepping through the code:

- When you have AutoExpand attributes in models, it is added to the query parameters in AddAutoSelectExpandProperties in ODataQueryOptions
- AddAutoSelectExpandProperties is called in the ApplyTo method, which is called in ExecuteQuery in EnableQueryAttribute
- Validate method of ODataQueryValidator is called before ApplyTo gets called, and SelectExpand is null at this time
- Since SelectExpand is null, the Validate method of SelectExpandQueryOption is not called
- the code that sets LevelsMaxLiteralExpansionDepth to MaxExpansionDepth is in the Validate method of SelectExpandQueryOption, which is never called, so it uses the initial value of -1, which is then set to ODataValidationSettings.DefaultMaxExpansionDepth (which is 2) in ProcessLevels

This seems to explain why it stops at 2 levels, however it is possible that I am missing something to make this work. Any suggestions are appreciated.

gathogojr:
### Issues

*This pull request fixes #2748.*

### Description

We set the default MaxExpansionDepth when the Validate method of the ODataQueryOptions object is called. The problem currently is that when the MaxExpansionDepth is overridden, for instance by assigning the value from the EnableQuery attribute ([EnableQuery(MaxExpansionDepth = 4)]), we call the Validate method before the AutoExpand navigation properties have been added to the SelectExpandQueryOption object. For that reason, the specified MaxExpansionDepth ends up not being respected for the auto-expanded navigation properties.

This pull request fixes that by calling ODataQueryOptions.AddAutoSelectExpandProperties prior to validating the SelectExpand query option.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/ODataQueryContext.cs
+++ b/src/Microsoft.AspNet.OData.Shared/ODataQueryContext.cs
@@ -156,6 +156,11 @@ namespace Microsoft.AspNet.OData
 
         internal string TargetName { get; private set; }
 
+        /// <summary>
+        /// Gets or sets the query validation settings.
+        /// </summary>
+        internal ODataValidationSettings ValidationSettings { get; set; }
+
         private static IEdmNavigationSource GetNavigationSource(IEdmModel model, IEdmType elementType, ODataPath odataPath)
         {
             Contract.Assert(model != null);
--- a/src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/ODataQueryOptions.cs
@@ -585,6 +585,8 @@ namespace Microsoft.AspNet.OData.Query
                 throw Error.ArgumentNull("validationSettings");
             }
 
+            Context.ValidationSettings = validationSettings;
+
             if (Validator != null)
             {
                 Validator.Validate(this, validationSettings);
@@ -792,6 +794,10 @@ namespace Microsoft.AspNet.OData.Query
                 if (originalSelectExpand != null && originalSelectExpand.LevelsMaxLiteralExpansionDepth > 0)
                 {
                     SelectExpand.LevelsMaxLiteralExpansionDepth = originalSelectExpand.LevelsMaxLiteralExpansionDepth;
+                }
+                else if (Context.ValidationSettings != null && Context.ValidationSettings.MaxExpansionDepth > 0)
+                {
+                    SelectExpand.LevelsMaxLiteralExpansionDepth = Context.ValidationSettings.MaxExpansionDepth;
                 }
             }
         }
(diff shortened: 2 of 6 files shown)


habbes (approved):

src/Microsoft.AspNet.OData.Shared/ODataQueryContext.cs
```diff
@@ -156,6 +156,11 @@ public DefaultQuerySettings DefaultQuerySettings
 
         internal string TargetName { get; private set; }
 
+        /// <summary>
+        /// Gets the query validation settings.
```
habbes, line 160:
```suggestion
        /// Gets or sets the query validation settings.
```

habbes (approved):

xuzhg (approved):

PR #2755, opened 2023-03-01
Apply Query options to Created results

### LINKED ISSUE #2754: QueryOptions not applied to Created(object) response
In POST method in an OData controller method

If we have the code below
```csharp
[EnableQuery]
public async Task<IActionResult> Post([FromBody] Customer customer)
{
    return Ok(customer);
}
```

If we call
```json
POST /Customers?$expand=Orders
{
    "Id": 1000,
    "Name": "Customer 1",
    "Age": 31,
    "Orders": [
        {
            "Id": 2002,
            "Price": 35,
            "Quantity": 2
        },
        {
            "Id": 2001,
            "Price": 10,
            "Quantity": 10
        }
    ]
}
```
The response will be 
```
{
    "@odata.context": "http://localhost:6285/odata/$metadata#Customers(Orders())/$entity",
    "Id": 1000,
    "Name": "Customer 1",
    "Age": 31,
    "Orders": [
        {
            "Id": 2002,
            "Price": 35,
            "Quantity": 2
        },
        {
            "Id": 2001,
            "Price": 10,
            "Quantity": 10,
            "Customers": []
        }
    ]
}
```

However if the controller method is as follows:

```csharp
[EnableQuery]
public async Task<IActionResult> Post([FromBody] Customer customer)
{
    return Created(customer);
}
```

The response will not have the Orders navigation property.

This code is the culprit https://github.com/OData/WebApi/blob/9a06ab9ee8683a2c7365a76607b818e66ac2a357/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs#L235-L237

`Ok()` returns an `OkObjectResult` which is a derived type of `ObjectResult`.

`Created()` returns `CreatedODataResult<TEntity>` which implements `IActionResult` interface. In line 235 `responseContent` will be `null` when we return `Created()` hence query options will not be applied.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData WebApi lib 7.6.4

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

KenitoInc:
### Issues

*This pull request fixes #2754 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
@@ -237,10 +237,10 @@ namespace Microsoft.AspNet.OData
 
                 if (responseContent == null)
                 {
-                    Type createdODataResultType = actionExecutedContext.Result.GetType().GetGenericArguments().Count() > 0 ?
-                        typeof(CreatedODataResult<>).MakeGenericType(actionExecutedContext.Result.GetType().GetGenericArguments()[0]) : null;
-
                     Type actionResultType = actionExecutedContext.Result.GetType();
+                    Type[] actionResultGenericArgs = actionResultType.GetGenericArguments();
+                    Type createdODataResultType = actionResultGenericArgs.Length > 0 ?
+                        typeof(CreatedODataResult<>).MakeGenericType(actionResultGenericArgs[0]) : null;
 
                     // Get the entity object from CreatedODataResult<T> via reflection.
                     // Use the entity object to create an instance of ObjectResult.
--- a/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
@@ -17,6 +17,7 @@ using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter;
 using Microsoft.AspNet.OData.Query;
+using Microsoft.AspNet.OData.Results;
 using Microsoft.AspNet.OData.Routing;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -233,6 +234,26 @@ namespace Microsoft.AspNet.OData
                 // actionExecutedContext.Result might also indicate a status code that has not yet
                 // been applied to the result; make sure it's also successful.
                 ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
+
+                if (responseContent == null)
+                {
+                    Type createdODataResultType = actionExecutedContext.Result.GetType().GetGenericArguments().Count() > 0 ?
+                        typeof(CreatedODataResult<>).MakeGenericType(actionExecutedContext.Result.GetType().GetGenericArguments()[0]) : null;
+
+                    Type actionResultType = actionExecutedContext.Result.GetType();
+
+                    // Get the entity object from CreatedODataResult<T> via reflection.
+                    // Use the entity object to create an instance of ObjectResult.
+                    if (createdODataResultType != null &&
+                        (actionResultType == createdODataResultType || createdODataResultType.IsAssignableFrom(actionResultType)))
+                    {
+                        object entity = ((PropertyInfo)createdODataResultType.GetProperty("Entity")).GetValue(actionExecutedContext.Result);
+                        responseContent = new ObjectResult(entity)
+                        {
+                            StatusCode = response.StatusCode
+                        };
+                    }
+                }
 
                 if (responseContent != null && (responseContent.StatusCode == null || IsSuccessStatusCode(responseContent.StatusCode.Value)))
                 {
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
@@ -249,15 +249,16 @@ namespace Microsoft.Test.E2E.AspNet.OData.BulkOperation
 
         [ODataRoute("Employees")]
         [HttpPost]
+        [EnableQuery]
         public ITestActionResult Post([FromBody] Employee employee)
         {
-            InitEmployees();
+            //InitEmployees();
 
-            var handler = new EmployeeAPIHandler();
+            //var handler = new EmployeeAPIHandler();
 
-            handler.UpdateLinkedObjects(employee, Request.GetModel());
+            //handler.UpdateLinkedObjects(employee, Request.GetModel());
 
-            return Ok(employee);
+            return Created(employee);
         }
 
         [ODataRoute("Employees({key})/Friends")]
--- a/src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNet.OData/EnableQueryAttribute.cs
@@ -21,7 +21,6 @@ using Microsoft.AspNet.OData.Common;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.AspNet.OData.Formatter;
 using Microsoft.AspNet.OData.Query;
-using Microsoft.AspNet.OData.Results;
 using Microsoft.OData.Edm;
 
 namespace Microsoft.AspNet.OData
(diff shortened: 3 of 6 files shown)


src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -234,6 +235,26 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
                 // been applied to the result; make sure it's also successful.
                 ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
 
+                if (responseContent == null)
+                {
+                    Type createdODataResultType = actionExecutedContext.Result.GetType().GetGenericArguments().Count() > 0 ?
+                        typeof(CreatedODataResult<>).MakeGenericType(actionExecutedContext.Result.GetType().GetGenericArguments()[0]) : null;
+
+                    Type actionResultType = actionExecutedContext.Result.GetType();
+
```
habbes, line 244: You seem to be computing `actionExecutedContext.Result.GetType()` and `GetGenericArguments()` multiple times here. Suggestion to compute them only once:
```suggestion
                    Type actionResultType = actionExecutedContext.Result.GetType();
                    Type[] actionResultGenericArgs = actionResultType.GetGenericArguments();
                    Type createdODataResultType = actionResultGenericArgs.Length > 0 ?
                        typeof(CreatedODataResult<>).MakeGenericType(actionResultGenericArgs[0]) : null;

```

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -234,6 +235,26 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
                 // been applied to the result; make sure it's also successful.
                 ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
 
+                if (responseContent == null)
+                {
+                    Type createdODataResultType = actionExecutedContext.Result.GetType().GetGenericArguments().Count() > 0 ?
+                        typeof(CreatedODataResult<>).MakeGenericType(actionExecutedContext.Result.GetType().GetGenericArguments()[0]) : null;
```
habbes, line 241: What happens if the result is neither an `Ok(result)` or `Created(result)`?

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -234,6 +235,26 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
...
+
+                    Type actionResultType = actionExecutedContext.Result.GetType();
+
+                    // Get the entity object from CreatedODataResult<T> via reflection.
+                    // Use the entity object to create an instance of ObjectResult.
+                    if (createdODataResultType != null &&
+                        (actionResultType == createdODataResultType || createdODataResultType.IsAssignableFrom(actionResultType)))
+                    {
+                        object entity = ((PropertyInfo)createdODataResultType.GetProperty("Entity")).GetValue(actionExecutedContext.Result);
+                        responseContent = new ObjectResult(entity)
```
habbes, line 251: Down below there's a statement like:
```c#
responseContent.Value = queryResult
```

What's the impact of modifying `responseContent.Value` in cases where the `responseContent` is actually created here as opposed to coming from the user?

habbes: Also, is it necessary to create a new instance of `ObjectResult`, looking at the code below, most of it only uses the value of `responseContent.Value`, which I assume is the same `entity` in this case. Maybe we can just extract the value and use it directly instead of creating an instance of `ObjectResult` where one was not provided.

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
```diff
@@ -249,15 +249,16 @@ public ITestActionResult PatchEmployees([FromBody] DeltaSet<Employee> coll)
...
         public ITestActionResult Post([FromBody] Employee employee)
         {
-            InitEmployees();
+            //InitEmployees();
 
-            var handler = new EmployeeAPIHandler();
+            //var handler = new EmployeeAPIHandler();
 
-            handler.UpdateLinkedObjects(employee, Request.GetModel());
+            //handler.UpdateLinkedObjects(employee, Request.GetModel());
```
habbes, line 259: Why are these commented out?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
```diff
@@ -249,15 +249,16 @@ public ITestActionResult PatchEmployees([FromBody] DeltaSet<Employee> coll)
...
 
-            return Ok(employee);
+            return Created(employee);
```
habbes, line 261: Maybe you should have created a new test to validate this new fix instead of modifying an existing case since the existing behaviour should continue to be supported.

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
xuzhg: Why do we create the CreatedODataResult<T> if we have generic arguments? Is it certain?

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
xuzhg: +1

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
KenitoInc: The original tests are not being used since they are part of Deep Insert work which is in progress. The final tests should be as modified

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
KenitoInc: They are currently not being used. They will be activated in the Deep insert PR

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
KenitoInc: Fixed

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
KenitoInc: @habbes Most of the results that we return  are a subclass of `ObjectResult` so they will be handled correctly

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
KenitoInc: 1) There is no negative impact since technically w update the `actionExecutedContext.Result` with the created object.
2) I am creating the `responseContent` object since it's `null` in line `238`. 

I think casting `actionExecutedContext.Result` to `ObjectResult` was a way to restrict the type of results we can apply query options on. This PR ensure we also apply query options to `CreatedODataResult` which is not a subclass of `ObjectResult`.

With no restrictions we can use the `actionExecutedContext.Result` directly. So I will keep it that way for now.

src/Microsoft.AspNetCore.OData/EnableQueryAttribute.cs
```diff
@@ -234,6 +235,26 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
                 // been applied to the result; make sure it's also successful.
                 ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
 
+                if (responseContent == null)
+                {
+                    Type actionResultType = actionExecutedContext.Result.GetType();
+                    Type[] actionResultGenericArgs = actionResultType.GetGenericArguments();
+                    Type createdODataResultType = actionResultGenericArgs.Length > 0 ?
+                        typeof(CreatedODataResult<>).MakeGenericType(actionResultGenericArgs[0]) : null;
```
ElizabethOkerio, line 243: can we have more than one `actionResultGenericArgs` here? Why do we choose to get the one at position `zero`?

KenitoInc: [CreatedODataResult](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnet.odata.results.createdodataresult-1) has only one argument

KenitoInc:
Resolved by #2765

PR #2764, opened 2023-03-28
support writing datamodififcation exception message

ElizabethOkerio:
### Issues
The `@Core.DataModificationException` annotation is not written properly in a bulk update response. Some information doesn't get written and only the type is written. This fix ensures that the entire annotation is written with all it's properties,, like the `MessageType`, the `FailedOperation` etc. 
### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceValueSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceValueSerializer.cs
@@ -137,26 +137,30 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             }
             else
             {
-                HashSet<string> structuralPropertyNames = new HashSet<string>();
-
-                foreach(IEdmStructuralProperty structuralProperty in expectedType.DeclaredStructuralProperties())
+                // If a type is an org.OData.Core.V1.DataModificationExceptionType type, 
+                // then we include all it's properties and not just the structural properties.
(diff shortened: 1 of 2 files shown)


KenitoInc (approved):
I'm okay with the change

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceValueSerializer.cs
```diff
@@ -137,26 +137,30 @@ private ODataResourceValue CreateODataResourceValue(object graph, IEdmStructured
             }
             else
             {
-                HashSet<string> structuralPropertyNames = new HashSet<string>();
-
-                foreach(IEdmStructuralProperty structuralProperty in expectedType.DeclaredStructuralProperties())
+                // If a type is an org.OData.Core.V1.DataModificationExceptionType type, 
+                // then we include all it's properties and not just the structural properties.
+                if (expectedType.Definition.FullTypeName().Equals("Org.OData.Core.V1.DataModificationExceptionType"))
```
habbes, line 142: This doesn't look like a clean way to check the exception type. Is there a better way?

habbes (approved):

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceValueSerializer.cs
ElizabethOkerio: We are checking the name of the `IEdmStructuredTypeReference`.

PR #2765, opened 2023-03-31
Make CreatedODataResult inherit from ObjectResult

### LINKED ISSUE #2754: QueryOptions not applied to Created(object) response

KenitoInc:
### Issues

*This pull request fixes #2754 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
@@ -31,15 +31,15 @@ namespace Microsoft.AspNet.OData.Results
         /// Initializes a new instance of the <see cref="CreatedODataResult{T}"/> class.
         /// </summary>
         /// <param name="entity">The created entity.</param>
-        public CreatedODataResult(object entity)
-            :base(entity)
+        public CreatedODataResult(T entity)
+            : base(entity)
         {
             if (entity == null)
             {
                 throw Error.ArgumentNull("entity");
             }
 
-            this._innerResult = (T)entity;
+            this._innerResult = entity;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNet.OData.Results
     /// <remarks>This action result handles content negotiation and the HTTP prefer header. It generates a location
     /// header containing the edit link of the created entity and, if response has status code: NoContent, also
     /// generates an OData-EntityId header.</remarks>
-    public class CreatedODataResult<T> : IActionResult
+    public class CreatedODataResult<T> : ObjectResult
     {
         private readonly T _innerResult;
 
@@ -31,14 +31,15 @@ namespace Microsoft.AspNet.OData.Results
         /// Initializes a new instance of the <see cref="CreatedODataResult{T}"/> class.
         /// </summary>
         /// <param name="entity">The created entity.</param>
-        public CreatedODataResult(T entity)
+        public CreatedODataResult(object entity)
+            :base(entity)
         {
             if (entity == null)
             {
                 throw Error.ArgumentNull("entity");
             }
 
-            this._innerResult = entity;
+            this._innerResult = (T)entity;
         }
 
         /// <summary>
@@ -53,7 +54,7 @@ namespace Microsoft.AspNet.OData.Results
         }
 
         /// <inheritdoc/>
-        public async virtual Task ExecuteResultAsync(ActionContext context)
+        public override async Task ExecuteResultAsync(ActionContext context)
         {
             HttpRequest request = context.HttpContext.Request;
             HttpResponse response = context.HttpContext.Response;
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
@@ -249,15 +249,17 @@ namespace Microsoft.Test.E2E.AspNet.OData.BulkOperation
 
         [ODataRoute("Employees")]
         [HttpPost]
+        [EnableQuery]
         public ITestActionResult Post([FromBody] Employee employee)
         {
-            InitEmployees();
+            //InitEmployees();
 
-            var handler = new EmployeeAPIHandler();
+            //var handler = new EmployeeAPIHandler();
 
-            handler.UpdateLinkedObjects(employee, Request.GetModel());
+            //handler.UpdateLinkedObjects(employee, Request.GetModel());
 
-            return Ok(employee);
+            //return Ok(employee);
+            return Created(employee);
         }
 
         [ODataRoute("Employees({key})/Friends")]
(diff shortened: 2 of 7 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
```diff
@@ -249,15 +249,17 @@ public ITestActionResult PatchEmployees([FromBody] DeltaSet<Employee> coll)
 
         [ODataRoute("Employees")]
         [HttpPost]
+        [EnableQuery]
         public ITestActionResult Post([FromBody] Employee employee)
         {
-            InitEmployees();
+            //InitEmployees();
```
ElizabethOkerio, line 255: is this meant to be added in the next PR?

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
```diff
@@ -23,22 +23,23 @@ namespace Microsoft.AspNet.OData.Results
...
+        public CreatedODataResult(object entity)
+            :base(entity)
         {
             if (entity == null)
             {
                 throw Error.ArgumentNull("entity");
             }
 
-            this._innerResult = entity;
+            this._innerResult = (T)entity;
```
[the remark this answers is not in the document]
xuzhg: https://github.com/OData/WebApi/blob/cc0786b6e445d917569e5cec0b294616b9d7f669/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs#L81-L84 also uses the ObjectResult.

We should use that to process.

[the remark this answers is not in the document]
gathogojr: Maybe use `this._innerResult = entity as T;` and then handle the scenario where the outcome is `null`. If it's acceptable for `entity` to be null, `entity as T` would yield null but that wouldn't be because an error was experienced when attempting to cast `entity` into a `T`.

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
```diff
@@ -23,22 +23,23 @@ namespace Microsoft.AspNet.OData.Results
...
     {
         private readonly T _innerResult;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CreatedODataResult{T}"/> class.
         /// </summary>
         /// <param name="entity">The created entity.</param>
-        public CreatedODataResult(T entity)
+        public CreatedODataResult(object entity)
+            :base(entity)
```
gathogojr, line 35: nit:
```suggestion
            : base(entity)
```

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
gathogojr: If the code is not relevant to this PR, we should delete it instead of commenting it out

gathogojr:
Other than for the few concerns highlighted in the reviews, I like this approach better than the previous one. It saves us the trouble of making changes to the more intricate `EnableQueryAttribute` logic.

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
```diff
@@ -23,22 +23,23 @@ namespace Microsoft.AspNet.OData.Results
...
+    public class CreatedODataResult<T> : ObjectResult
     {
         private readonly T _innerResult;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CreatedODataResult{T}"/> class.
         /// </summary>
         /// <param name="entity">The created entity.</param>
-        public CreatedODataResult(T entity)
+        public CreatedODataResult(object entity)
```
habbes, line 34: Why change the parameter type from `T` to `object`?

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
habbes: I would prefer creating a new test instead of modifying the existing one, since return `Ok()` from a `Post()` request is still supported.

test/UnitTest/Microsoft.AspNetCore.OData.Test/PublicApi/Microsoft.AspNetCore.OData.PublicApi.bsl
```diff
@@ -3158,8 +3158,8 @@ public class Microsoft.AspNet.OData.Results.BadRequestODataResult : Microsoft.As
 	public virtual System.Threading.Tasks.Task ExecuteResultAsync (Microsoft.AspNetCore.Mvc.ActionContext context)
 }
 
-public class Microsoft.AspNet.OData.Results.CreatedODataResult`1 : IActionResult {
-	public CreatedODataResult`1 (T entity)
+public class Microsoft.AspNet.OData.Results.CreatedODataResult`1 : Microsoft.AspNetCore.Mvc.ObjectResult, IActionResult {
+	public CreatedODataResult`1 (object entity)
```
habbes, line 3162: Is this a breaking change?

KenitoInc: I reverted this change

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
KenitoInc: I reverted this change

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/BulkOperation/BulkOperationController.cs
KenitoInc: Deleted the code.
@habbes This controller method is for bulk ops and we don't need to have a test that returns `Ok()`

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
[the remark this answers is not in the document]
KenitoInc: I reverted this change

gathogojr:
@KenitoInc Check out the failing build

ElizabethOkerio (approved):

Issue #2583, opened 2021-10-08
Allow to modify ODataQueryOptions in  controller action


chrisspre:
There are scenarios where a service needs more control over the supplied query options before they are handed over to the QueryProvider (before they are "applied" to IQueryable : `queryOptions.ApplyTo(data)`).

One recent scenario is the ability to drop a certain property for less privileged clients (based on the Authorization token.)

For that I imagine some code like 
```csharp

  var data = /* ... query data store */

  if (! privileged) 
  {
    foreach(var property in privilegedProperties) 
    {
      queryOptions.SelectExpand.SelectExpandClause.SelectedItems.Remove ( property );
    }
  }
  return queryOptions.ApplyTo(data); 
}

Angelinsky7:
i would be also very interested to be able to dynamically add/remove expand property in the controller. 
Sometime i would like that when i'm getting a lot of result, by default nothing is expanded but when i got one result some properties are extended. `AutoExpandAttribute` does not seems to allow that.

PR #2766, opened 2023-04-02
Avoid duplicating instantiation of ODataQueryOptions

ificator:
### Issues

Avoid duplicating instantiation of `ODataQueryOptions`.

### Description

Currently if an action implementation defines an `ODataQueryOptions` parameter that instance is ignored when it comes time for `EnableQueryAttribute` to apply the options to the response. This causes two classes of issues:
1. It's a waste of CPU to reparse potentially complex query strings.
2. Action implementations cannot modify the `ODataQueryOptions` to control how the response is processed.

Related issues: https://github.com/OData/WebApi/issues/2583 

- [x] *Test cases added*


ificator:
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.AspNet.OData/ODataOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// Provides programmatic configuration for the OData service.
+    /// </summary>
+    public class ODataOptions
+    {
+        /// <summary>
+        /// Gets or sets a value indicating if <see cref="ODataQueryOptions"/> should be reused between
+        /// <see cref="ODataQueryParameterBindingAttribute"/> and <see cref="EnableQueryAttribute"/>.
+        /// </summary>
+        public bool ParseODataQueryOptionsOnce { get; set; }
```
mikepizzo, line 21: >ParseODataQueryOptionsOnce [](http://example.com/codeflow?start=20&length=26)

In what scenarios would I *not* want this to be true? 

Is this there to protect backward compatibility, so that *if* someone happens to have code that modifies the query options (i.e., during Apply) without copying them first, that it doesn't affect other uses of QueryOptions?

In a future breaking change release, would we want to remove this option and always re-use the ODataQueryOptions?

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -239,6 +240,27 @@ public static Uri GetNextPageLink(this HttpRequestMessage request, int pageSize,
...
         }
 
+        /// <summary>
+        /// Gets the <see cref="ODataOptions"/>.
+        /// </summary>
+        /// <param name="request">The request that provides access to the <see cref="IServiceProvider"/>.</param>
+        /// <returns>The <see cref="ODataOptions"/> instance.</returns>
+        public static ODataOptions GetODataOptions(this HttpRequestMessage request)
+        {
+            return request.GetRequestContainer().GetRequiredService<ODataOptions>();
```
mikepizzo, line 250: >GetRequiredService [](http://example.com/codeflow?start=49&length=18)

Should we make this GetRequiredService, or be prepared to handle nulls?  We add this in ConfigureDefaultServices, which is called from 
MapODataServiceRoute and EnableDependencyInjection, but are there any scenarios where ConfigureDefaultServices might not be called? What about when applying QueryOptions to a non-OData route?

src/Microsoft.AspNet.OData/ODataOptions.cs
ificator: Personally I think it would ideally always be true, but for safety I made it opt in. I'd be fine with a break change release inverting this logic (i.e. default to `true` and but allow override to `false`).

src/Microsoft.AspNet.OData/Extensions/HttpRequestMessageExtensions.cs
ificator: Interesting call out re: `QueryOptions` for non-OData routes... I hadn't really considered it. I guess we could make it non-required and return a default instance here if one isn't registered. I'd much rather consumers of this method not have to worry about null checks.

KenitoInc (changes requested):

src/Microsoft.AspNet.OData/ODataOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+//------------------------------------------------------------------------------
+
+using Microsoft.AspNet.OData.Query;
+
+namespace Microsoft.AspNet.OData
+{
+    /// <summary>
+    /// Provides programmatic configuration for the OData service.
+    /// </summary>
+    public class ODataOptions
```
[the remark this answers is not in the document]
ificator: I can do that if it makes sense - I only used `ODataOptions` because I saw it defined in the AspNetCore implementation and so thought it was the correct pattern. I'll see what it looks like to switch to this existing thing.

src/Microsoft.AspNet.OData/ODataOptions.cs
ificator: Since I've switched to `CompatibilityOptions` the change will now be active by default.

src/Microsoft.AspNet.OData/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -36,7 +36,7 @@ public HttpRequestMessage Request
 
                 // We add this setting via CompatibilityOptions
                 CompatibilityOptions options = webApiRequestMessage != null ? webApiRequestMessage.Configuration.GetCompatibilityOptions() : CompatibilityOptions.None;
-                DisableCaseInsensitiveRequestPropertyBinding = options.HasFlag(CompatibilityOptions.DisableCaseInsensitiveRequestPropertyBinding) ? true : false;
```
ificator, line 39: `HasFlag` will box and therefore allocate - instead of requiring bit checks for options I created the `HasOption` extension method instead.

mikepizzo (dismissed):
bonus points for a unit test that verifies the queryoptions is not re-used if the compatibility flag is set.

src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs
```diff
@@ -33,6 +34,29 @@ public enum CompatibilityOptions
...
+    public static class CompatibilityOptionsExtensions
+    {
+        /// <summary>
+        /// Determines whether the provided option is set.
+        /// </summary>
+        /// <param name="options">The set options.</param>
+        /// <param name="option">The option to check.</param>
+        /// <returns>True if the option is set, false otherwise.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool HasOption(this CompatibilityOptions options, CompatibilityOptions option)
```
KenitoInc, line 57: This a very good fix!

KenitoInc (commented):
Fix the failing build pipeline

gathogojr (approved):

gathogojr:
Thanks @ificator for your contribution.

PR #2772, opened 2023-04-13
Apply correct handler for Handler related object

KenitoInc:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/DeltaSetOfT.cs
+++ b/src/Microsoft.AspNet.OData.Shared/DeltaSetOfT.cs
@@ -216,7 +216,7 @@ namespace Microsoft.AspNet.OData
                         {
                             operation = DataModificationOperationKind.Update;
 
-                            ODataAPIResponseStatus linkResponseStatus = handler.TryAddRelatedObject(original, out errorMessage);
+                            ODataAPIResponseStatus linkResponseStatus = apiHandlerOfT.TryAddRelatedObject(original, out errorMessage);
 
                             if (linkResponseStatus == ODataAPIResponseStatus.Failure)
                             {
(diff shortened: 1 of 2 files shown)


gathogojr (approved):

src/Microsoft.AspNet.OData.Shared/DeltaSetOfT.cs
```diff
@@ -216,7 +216,7 @@ internal DeltaSet<TStructuralType> CopyChangedValues(IODataAPIHandler apiHandler
                         {
                             operation = DataModificationOperationKind.Update;
 
-                            ODataAPIResponseStatus linkResponseStatus = handler.TryAddRelatedObject(original, out errorMessage);
+                            ODataAPIResponseStatus linkResponseStatus = apiHandlerOfT.TryAddRelatedObject(original, out errorMessage);
```
ElizabethOkerio, line 219: Line 110 we assign `apiHandlerOfT` to handler. Why are the different at this point?

KenitoInc: I explained that in the PR description.

In line 149, we assign the `handler` to `odataPathApiHandler`. This will cause issues in line 219

ElizabethOkerio: Ah I see.

ElizabethOkerio: I think we should do a check..if `odataApiHandler` is null then we should use the `handler`, otherwise then we use `apiHandlerOfT` but I guess even doing it this way is fine.

PR #2778, opened 2023-04-28
Check SelectAllDynamicProperties AND SelectedDynamicProperties

ificator:
### Issues

Sending a request with `$select=*,dynamicProperty1` only returns `dynamicProperty1` instead of all dynamic properties.

### Description

The issue is that `AppendDynamicProperties` makes an incorrect assumption about when `SelectAllDynamicProperties` and `SelectedDynamicProperties` have values. The code is written to assume that if `SelectAllDynamicProperties` is `true` then `SelectedDynamicProperties` must be `null`, but this isn't the case because `SelectExpandNode.BuildSelect` will populate `SelectedDynamicProperties` in `BuildSelectItem` and later set `SelectAllDynamicProperties` when the wildcard is encountered.

The fix is straight forward - instead of using nullity of `SelectedDynamicProperties` to determine if all are selected use `SelectAllDynamicProperties` instead.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -298,12 +298,18 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
             HashSet<string> declaredPropertyNameSet = new HashSet<string>(resource.Properties.Select(p => p.Name));
             List<ODataProperty> dynamicProperties = new List<ODataProperty>();
 
-            // To test SelectedDynamicProperties == null is enough to filter the dynamic properties.
-            // Because if SelectAllDynamicProperties == true, SelectedDynamicProperties should be null always.
-            // So `selectExpandNode.SelectedDynamicProperties == null` covers `SelectAllDynamicProperties == true` scenario.
-            // If `selectExpandNode.SelectedDynamicProperties != null`, then we should test whether the property is selected or not using "Contains(...)".
+            // Due to the early return condition at the start of this method we know that either:
+            //  1. SelectAllDynamicProperties == true
+            //  2. SelectedDynamicProperties != null
+            //  3. SelectAllDynamicProperties == true AND SelectedDynamicProperties != null
+            //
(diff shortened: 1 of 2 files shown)


ificator:
@KenitoInc @gathogojr can you take a quick look at this one as well? thanks!

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -298,12 +298,18 @@ public virtual ODataDeletedResource CreateDeletedResource(SelectExpandNode selec
...
+            //  2. SelectedDynamicProperties != null
+            //  3. SelectAllDynamicProperties == true AND SelectedDynamicProperties != null
+            //
+            // Previous code simply checked the nullity of SelectedDynamicProperties here, but that ignored case 3 since which is possible
+            // since SelectExpandNode.BuildSelectExpand will set SelectAllDynamicProperties to true even if SelectedDynamicProperties is
+            // not-null. Instead we need to check both properties in order to determine if a dynamic property should be selected.
             IEnumerable<KeyValuePair<string, object>> dynamicPropertiesToSelect =
-                dynamicPropertyDictionary.Where(x => selectExpandNode.SelectedDynamicProperties == null || selectExpandNode.SelectedDynamicProperties.Contains(x.Key));
+                dynamicPropertyDictionary.Where(
+                    x => selectExpandNode.SelectAllDynamicProperties ||
```
mikepizzo, line 311: >selectExpandNode.SelectAllDynamicProperties  [](http://example.com/codeflow?start=25&length=44)

Perhaps optimize so we don't have to loop in this case: 

if (selectExpandNode.SelectAllDynamicProperties) { dynamicPropertiesToSelect = dynamicPropertyDictionary }
else {
  ...
} 

Better yet might be to simply move the checks into the following loop and avoid creating the separate list and extra loop through the dynamic properties.

ificator: Good call

ificator:
@mikepizzo any other comments or can we merge this?

mikepizzo (approved):
:shipit:

Issue #2413, opened 2021-02-11
Nested $select on optional complex property throws exception when complex property is null


UmairB:
When an entity has a nullable complex property (which is implemented as an owned type in entity framework), a nested $select on the complex property will throw an error whenever the complex property is null.

### Assemblies affected

Tried on Microsoft.AspNetCore.OData version 7.5.1 and the 8 beta, being used with Microsoft.EntityFrameworkCore.SqlServer v5.0.2

### Reproduce steps

Setup the following classes
```
public class EntityWithOwnedType
{
    public int Id { get; set; }

    public SingleOwnedType SingleOwnedType { get; set; }
}

public class SingleOwnedType
{
    public int SomeId { get; set; }

    public string Name { get; set; }
}
```
with the following entity framework core configuration
```
builder.HasKey(e => e.Id);
builder.Property(e => e.Id).UseIdentityColumn();

builder.OwnsOne(e => e.SingleOwnedType, b =>
{
    b.WithOwner();
});
```
and the following odata configuration
```
var entityType = builder.EntitySet<EntityWithOwnedType>("EntityWithOwnedTypes").EntityType;
entityType.ComplexProperty(e => e.SingleOwnedType);
```

### Expected result

The following endpoint `/entityWithOwnedTypes?$select=Id,SingleOwnedType($select=SomeId)` would return
```
{
    "@odata.context": "https://localhost:5001/odata/$metadata#EntityWithOwnedTypes(Id,SingleOwnedType/SomeId)",
    "value": [
        {
            "Id": 1,
            "SingleOwnedType": null
        }
    ]
}
```
if `SingleOwnedType` is null on `EntityWithOwnedType`

### Actual result

Invalid, partial json is returned with the error "Microsoft.OData.ODataException: The property 'SomeId[Nullable=False]' of type 'Edm.Int32' has a null value, which is not allowed."

### Additional detail

Not quite sure what the expected behaviour is here. The error occurs for `SingleOwnedType($select=SomeId)` but for `SingleOwnedType($select=Name)` it returns the following valid json:

```
{
    "@odata.context": "https://localhost:5001/odata/$metadata#EntityWithOwnedTypes(Id,SingleOwnedType/Name)",
    "value": [
        {
            "Id": 1,
            "SingleOwnedType": {
                "Name": null
            }
        }
    ]
}
```
[...]

PR #2780, opened 2023-05-13
Check HandleNullPropagation in ProjectElement

ificator:
### Issues

If an object is null, but a `$select` is provided that specifically requests a property on that object the results are unexpected. At the minimum it'll return an empty instance of the object, but if there's a required property defined for the object's type it'll result in a failed request.

Consider the following request where `foo` exists with `id` "blah", but `bah` is null:

`.../foo?$select=id,bah($select=baz)`

Expected:
```
{
  "id": "blah",
  "bah": null
}
```

Actual:
```
{
  "id": "blah",
  "bah": { }
}
```

The following issue goes into a lot of detail for the scenario that results in request failure: https://github.com/OData/WebApi/issues/2413

### Description

The problem appears to be that `HandleNullPropagation` is not applied in `ProjectElement` - after doing so both of the problematic scenarios mentioned above are resolved.

I'd say it's pretty cut and dry if it weren't for the fact that there's a unit test asserting the old behavior... does that mean it's deliberate? I'm hoping the test was just written to code and not to an actual expectation of behavior.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -350,7 +350,17 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
             Type wrapperGenericType = GetWrapperGenericType(isInstancePropertySet, isTypeNamePropertySet, isContainerPropertySet);
             wrapperType = wrapperGenericType.MakeGenericType(elementType);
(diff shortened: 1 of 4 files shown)


ificator:
Looks like an E2E test is also asserting the old behavior. I won't fix that until we have consensus that this change should move forward.

ificator:
ping @KenitoInc @habbes @xuzhg @corranrogue9 @ElizabethOkerio @gathogojr @mikepizzo

mikepizzo (approved):
Agree with expected result.

ificator:
:shipit:?

PR #2785, opened 2023-05-26
Server-side paging should not apply to collections of non-entity types

### LINKED ISSUE #2738: OData WebAPI lib applies service driven pagination to collections of complex types.
Normally OData WebAPI lib doesn't apply any service driven pagination for complex type collections. However when $select option is queried service driven pagination is applied both to expanded entity type collections and complex type collections while building projection. Thus link generator fails to generate NextLink because navigation property is obviously a null.

### Assemblies affected

OData WebApi lib 7.x

### Reproduce steps

Run sample application attached to this issue. Execute url: http://localhost:42184/odata/Entities?$select=ComplexTypes

### Expected result

JSON output with two entities and all nested ComplexTypes objects.

### Actual result

Application crushes. 
System.ArgumentNullException
  HResult=0x80004003
  Message=Value cannot be null. (Parameter 'navigationProperty')
  Source=Microsoft.AspNetCore.OData
  StackTrace:
   at Microsoft.AspNet.OData.Builder.NavigationSourceLinkBuilderAnnotation.BuildNavigationLink(ResourceContext instanceContext, IEdmNavigationProperty navigationProperty) in D:\a\1\s\src\Microsoft.AspNet.OData.Shared\Builder\NavigationSourceLinkBuilderAnnotation.cs:line 269

[WebApplication1.zip](https://github.com/OData/WebApi/files/10220488/WebApplication1.zip)

gathogojr:
### Issues

*This pull request fixes #2738.*

### Description

Server-side paging should not apply to collections of non-entity types.

Currently, if a primitive, enum or complex collection property is included in the `$select` property list and server-side paging is configured, the `SelectExpandBinder` applies paging to the collection if the items exceed the configured `PageSize`. In the case of complex collection property, the attempt fails with an exception since the next-link can't be successfully generated.

When the `$select` query option is not present on the URL, no paging is applied for non-entity type collections. This should be the expected behaviour even when a `$select` is present.

This pull request addresses the described issue.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -539,15 +539,15 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             // for $select=NS.SubType/Property, we don't care about the leading type segment, because with or without the type segment
             // the "Property" property value expression should be built into the property container.
 
-            PropertySegment firstSturucturalPropertySegment = firstNonTypeSegment as PropertySegment;
-            if (firstSturucturalPropertySegment != null)
+            PropertySegment firstStructuralPropertySegment = firstNonTypeSegment as PropertySegment;
+            if (firstStructuralPropertySegment != null)
             {
                 // $select=abc/..../xyz
                 SelectExpandIncludedProperty newPropertySelectItem;
-                if (!currentLevelPropertiesInclude.TryGetValue(firstSturucturalPropertySegment.Property, out newPropertySelectItem))
+                if (!currentLevelPropertiesInclude.TryGetValue(firstStructuralPropertySegment.Property, out newPropertySelectItem))
                 {
-                    newPropertySelectItem = new SelectExpandIncludedProperty(firstSturucturalPropertySegment, navigationSource);
-                    currentLevelPropertiesInclude[firstSturucturalPropertySegment.Property] = newPropertySelectItem;
+                    newPropertySelectItem = new SelectExpandIncludedProperty(firstStructuralPropertySegment, navigationSource);
+                    currentLevelPropertiesInclude[firstStructuralPropertySegment.Property] = newPropertySelectItem;
                 }
 
                 newPropertySelectItem.AddSubSelectItem(remainingSegments, pathSelectItem);
@@ -818,7 +818,7 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             }
 
             int? modelBoundPageSize = querySettings == null ? null : querySettings.PageSize;
-            propertyValue = ProjectAsWrapper(propertyValue, subSelectExpandClause, structuralProperty.Type.ToStructuredType(), pathSelectItem.NavigationSource,
+            propertyValue = ProjectAsWrapper(propertyValue, subSelectExpandClause, propertyStructuredType, pathSelectItem.NavigationSource,
                 pathSelectItem.OrderByOption, // $orderby=...
                 pathSelectItem.TopOption, // $top=...
                 pathSelectItem.SkipOption, // $skip=...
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1057,23 +1045,26 @@ private Expression AddOrderByQueryForSource(Expression source, OrderByClause ord
                     _settings.EnableConstantParameterization);
             }
 
-            if (_settings.PageSize.HasValue || modelBoundPageSize.HasValue || hasTopValue || hasSkipvalue)
+            if (structuredType is IEdmEntityType)
             {
-                // don't page nested collections if EnableCorrelatedSubqueryBuffering is enabled
-                if (!_settings.EnableCorrelatedSubqueryBuffering)
+                if (_settings.PageSize.HasValue || modelBoundPageSize.HasValue)
```
habbes, line 1050: Why don't we care about to and skip anymore?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -818,7 +818,7 @@ private Expression CreateTotalCountExpression(Expression source, bool? countOpti
             }
 
             int? modelBoundPageSize = querySettings == null ? null : querySettings.PageSize;
-            propertyValue = ProjectAsWrapper(propertyValue, subSelectExpandClause, structuralProperty.Type.ToStructuredType(), pathSelectItem.NavigationSource,
+            propertyValue = ProjectAsWrapper(propertyValue, subSelectExpandClause, propertyStructuredType, pathSelectItem.NavigationSource,
```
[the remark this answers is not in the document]
gathogojr: @habbes A few lines above this one, we have this variable declaration:
```csharp
IEdmStructuredType propertyStructuredType = structuralProperty.Type.ToStructuredType();
```
https://github.com/OData/WebApi/blob/4abd149629fd3f20a19b039a6dcb7bd37abc46b8/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs#L813

All I'm doing here is to use that variable in place of the longer expression.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
gathogojr: @habbes `PageSize` and `ModelBoundPageSize` apply to server-side paging. `$top` and `$top` in the manner they're used here apply to client-side paging. We want to still allow the client to do something like this:
```
EntitySet?$select=PrimitiveOrEnumOrComplexCollectionProperty($top=2)
EntitySet?$select=PrimitiveOrEnumOrComplexCollectionProperty($skip=3;$top=2)
```

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1057,23 +1045,26 @@ private Expression AddOrderByQueryForSource(Expression source, OrderByClause ord
                     _settings.EnableConstantParameterization);
             }
 
-            if (_settings.PageSize.HasValue || modelBoundPageSize.HasValue || hasTopValue || hasSkipvalue)
+            if (structuredType is IEdmEntityType)
```
xuzhg, line 1048: don't give a variable name for 'is' ?

xuzhg (approved):

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
gathogojr: @xuzhg In this case we just need to know we're dealing with an `IEdmEntityType` and not an `IEdmComplexType`. We have no use for the variable.

habbes (approved):

PR #2790, opened 2023-06-15
Fix $filter on nested collections

### LINKED ISSUE #1696: $filter on nested collection items does not work
I want to make some filter over items in a nested collection:
`ClassTypes?$filter=customAttributes/any(d: d/attributeDefinitionId eq 48)`

I'm getting error, that property `attributeDefinitionId` is not filterable, although it is. I can filter on property when accessing entity directly.

### Assemblies affected

OData WebApi lib 6.1.0
OData WebApi lib 7.1.0

Probably also other between those (those two I checked).

### Reproduce steps

Suppose I have an entity, with collection as prop:

```
public class ClassType {
    [Key]
    public long Id { get; set; }
    
    // some more properties here

    public ICollection<ClassTypeCustomAttribute> CustomAttributes { get; set; }
}

public class ClassTypeCustomAttribute {
    [Key, Column(Order = 1)]
    public long AttributeDefinitionId { get; set; }

    // some more properties here
}
```

In model builder I marked `ClassType.CustomAttributes` property to be both expandable and filterable, and marked `ClassTypeCustomAttribute.AttributeDefinitionId` to be filterable.

Expanding works well, filtering by `ClassTypeCustomAttribute.AttributeDefinitionId` works fine, when querying `ClassTypeCustomAttribute` entity directly.

However when I want to query `ClassType` and filter out by `CustomAttributes` property, it fails. Heres exemplary query:
`ClassTypes?$filter=customAttributes/any(d: d/attributeDefinitionId eq 48)`

### Expected result

It should definitely be able to be filtered.

### Actual result

It gives me error:
```
{
    "error": {
        "code": "",
        "message": "The query specified in the URI is not valid. The property 'attributeDefinitionId' cannot be used in the $filter query option.",
        "innererror": {
            "message": "The property 'attributeDefinitionId' cannot be used in the $filter query option.",
            "type": "Microsoft.OData.ODataException",
            "stacktrace": "   at Microsoft.AspNet.OData.Query.Validators.FilterQueryValidator.ValidateSingleValuePropertyAccessNode(SingleValuePropertyAccessNode propertyAccessNode, ODataValidationSettings settings)\r\n   at Microsoft.AspNet.OData.Query.Validators.FilterQueryValidator.ValidateLogicalOperator(BinaryOperatorNode binaryNode, ODataValidationSettings settings)\r\n   at Microsoft.AspNet.OData.Query.Validators.FilterQueryValidator.ValidateLogicalOperator(BinaryOperatorNode binaryNode, ODataValidationSettings settings)\r\n   at Microsoft.AspNet.OData.Query.Validators.FilterQueryValidator.ValidateAnyNode(AnyNode anyNode, ODataValidationSettings settings)\r\n   at Microsoft.AspNet.OData.Query.FilterQueryOption.Validate(ODataValidationSettings validationSettings)\r\n   at Microsoft.AspNet.OData.Query.Validators.ODataQueryValidator.Validate(ODataQueryOptions options, ODataValidationSettings validationSettings)\r\n   at Microsoft.AspNet.OData.Query.ODataQueryOptions.Validate(ODataValidationSettings validationSettings)\r\n   at Microsoft.AspNet.OData.EnableQueryAttribute.ValidateQuery(HttpRequestMessage request, ODataQueryOptions queryOptions)\r\n   at PerfectGym.Api.Core.ODataQueryableAttribute.ValidateQuery(HttpRequestMessage request, ODataQueryOptions queryOptions) in D:\\PerfectGym\\PerfectGym.Api.V2\\Core\\ODataQueryableAttribute.cs:line 81\r\n   at Microsoft.AspNet.OData.EnableQueryAttribute.<>c__DisplayClass0_0.<OnActionExecuted>b__3(ODataQueryContext queryContext)\r\n   at Microsoft.AspNet.OData.EnableQueryAttribute.ExecuteQuery(Object responseValue, IQueryable singleResultCollection, IWebApiActionDescriptor actionDescriptor, Func`2 modelFunction, IWebApiRequestMessage request, Func`2 createQueryOptionFunction)\r\n   at Microsoft.AspNet.OData.EnableQueryAttribute.OnActionExecuted(Object responseValue, IQueryable singleResultCollection, IWebApiActionDescriptor actionDescriptor, IWebApiRequestMessage request, Func`2 modelFunction, Func`2 createQueryOptionFunction, Action`1 createResponseAction, Action`3 createErrorAction)"
        }
    }
}
```

### Additional detail

I did some debugging of the AspNet.OData library (in version both 6.1 and 7.1), and found some very cumbersome logic for calculating which property is filterable and which is not.
Check this exact method `EdmLibHelpers.IsNotFilterable`:
https://github.com/OData/WebApi/blob/12d998c0e60ed9bcf1a45dbdc650a51a3578b763/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs#L285

It is called when applying query validator. When the validation is applied to abovementioned property `attributeDefinitionId`:
- parameter `edmProperty` points to this property (I mean `ClassTypeCustomAttribute.AttributeDefinitionId`)
- parameter `pathEdmProperty` is null (it is taken from `_property` field in validator which is set to null on initializing the validator)
- parameter `pathEdmStructuredType` is pointing to EDM type for `ClassTypeCustomAttribute` as expected
- parameter `edmModel` is the whole EDM model
- parameter `enableFilter` is some default from query settings (`false` in my case)

So far so good. But check on call to method `GetModelBoundQuerySettings` 
https://github.com/OData/WebApi/blob/12d998c0e60ed9bcf1a45dbdc650a51a3578b763/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs#L301

Here you see, `pathEdmProperty` is passed, which is `null`.

Inside of this `GetModelBoundQuerySettings` method lays logic for calculating dictionary, that maps properties to whether they are filterable or not.

https://github.com/OData/WebApi/blob/12d998c0e60ed9bcf1a45dbdc650a51a3578b763/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs#L567

And this is the part that is very shady: this dictionary is not calculated for the type we are checking, but for the parent one: `ClassType`.
These settings are merged with some other dictionary, calculated from property, but since property is null, nothing is merged in fact.

And here lays the dragon: we check if some property from nested type is filterable, by checking if property with the same name, but for parent type is filterable!
I mean, we check if `ClassTypeCustomAttribute.AttributeDefinitionId` is filterable by searching dictionary that contains properties of `ClassType`.
For example, it by accident, there would property with the same name: `ClassType.AttributeDefinitionId` and it would be filterable, it would work!

### LINKED ISSUE #2261: Cannot filter on child model property without enabling property in parent [Filter]
I'm trying to filter on a child collection of my OData model by adding `[Filter]` attributes. I have added the `[Filter]` attribute to my top-level model, and that works. However I am getting an error message saying that the property on my child model cannot be used as a filter unless I add it to the attribute on my main model.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*
- System.Web.OData v6.0.0.0 _(I think)_

### Reproduce steps

Filter string: `children/any(x: x/name eq 'bar')`

Controller and models:

```csharp
[ODataRoutePrefix("repro")]
public class ReproController : ODataController
{
    public static List<ModelA> Data { get; set; } = new List<ModelA>
    {
        new ModelA(),
    };

    [HttpGet]
    [ODataRoute]
    [EnableQuery(AllowedQueryOptions = AllowedQueryOptions.Filter)]
    public Task Repro(ODataQueryOptions<ModelA> options)
    {
        options.Validate(new ODataValidationSettings
        {
            AllowedQueryOptions = AllowedQueryOptions.Filter,
        });

        return Task.CompletedTask;
    }
}

[Filter("children")]
public class ModelA
{
    [Key]
    public string Id { get; set; } = Guid.NewGuid().ToString();

    public IEnumerable<ModelB> Children { get; set; } = new List<ModelB>
    {
        new ModelB(),
    };
}

[Filter("name")]
public class ModelB
{
    public string Name { get; set; } = "foo";
}
```

### Expected result

- Filter validation succeeds.

### Actual result

- Validation fails: `The property 'name' cannot be used in the $filter query option.`

### Additional detail

It does work when I add `"name"` to the `Filter` attribute on `ModelA`, but that doesn't seem like it should be necessary.

gathogojr:
### Issues

*This pull request fixes #1696 and fixes #2261.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBoundQuerySettings/FilterAttributeTest/FilterAttributeTest.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBoundQuerySettings/FilterAttributeTest/FilterAttributeTest.cs
@@ -11,6 +11,8 @@ using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using Microsoft.AspNet.OData.Extensions;
 using Microsoft.Test.E2E.AspNet.OData.Common.Execution;
+using Microsoft.Test.E2E.AspNet.OData.Common.Extensions;
+using Newtonsoft.Json.Linq;
 using Xunit;
 
 namespace Microsoft.Test.E2E.AspNet.OData.ModelBoundQuerySettings.FilterAttributeTest
@@ -20,9 +22,11 @@ namespace Microsoft.Test.E2E.AspNet.OData.ModelBoundQuerySettings.FilterAttribut
         private const string CustomerBaseUrl = "{0}/enablequery/Customers";
         private const string OrderBaseUrl = "{0}/enablequery/Orders";
         private const string CarBaseUrl = "{0}/enablequery/Cars";
+        private const string AuthorBaseUrl = "{0}/enablequery/Authors";
         private const string ModelBoundCustomerBaseUrl = "{0}/modelboundapi/Customers";
         private const string ModelBoundOrderBaseUrl = "{0}/modelboundapi/Orders";
         private const string ModelBoundCarBaseUrl = "{0}/modelboundapi/Cars";
+        private const string ModelBoundAuthorBaseUrl = "{0}/modelboundapi/Authors";
 
         public FilterAttributeTest(WebHostTestFixture fixture)
             :base(fixture)
@@ -31,8 +35,11 @@ namespace Microsoft.Test.E2E.AspNet.OData.ModelBoundQuerySettings.FilterAttribut
 
         protected override void UpdateConfiguration(WebRouteConfiguration configuration)
         {
-            configuration.AddControllers(typeof(CustomersController), typeof(OrdersController),
-                    typeof(CarsController));
+            configuration.AddControllers(
+                typeof(CustomersController),
+                typeof(OrdersController),
+                typeof(CarsController),
+                typeof(AuthorsController));
             configuration.JsonReferenceLoopHandling =
                 Newtonsoft.Json.ReferenceLoopHandling.Ignore;
             configuration.Expand();
@@ -185,6 +192,28 @@ namespace Microsoft.Test.E2E.AspNet.OData.ModelBoundQuerySettings.FilterAttribut
             {
                 Assert.Contains("cannot be used in the $filter query option.", result);
             }
+        }
+
+        [Theory]
+        [InlineData(AuthorBaseUrl + "?$filter=Books/any(d: d/BookId eq 7)")]
+        [InlineData(ModelBoundAuthorBaseUrl + "?$filter=Books/any(d: d/BookId eq 7)")]
+        public async Task FilterOnNestedCollection(string entitySetUrl)
+        {
+            var requestUrl = string.Format(entitySetUrl, BaseAddress);
+            var request = new HttpRequestMessage(HttpMethod.Get, requestUrl);
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json;odata.metadata=minimal"));
+            var client = new HttpClient();
+
+            var response = await client.SendAsync(request);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            var content = await response.Content.ReadAsObject<JObject>();
+
+            var authors = content.GetValue("value") as JArray;
+            Assert.NotNull(authors);
+
+            var author = Assert.Single(authors) as JObject;
+            Assert.Equal(3, author.GetValue("AuthorId"));
         }
     }
 }
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -1002,7 +1002,8 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             bool hasTopValue = topOption != null && topOption.HasValue;
             bool hasSkipvalue = skipOption != null && skipOption.HasValue;
 
-            if (structuredType is IEdmEntityType entityType)
+            IEdmEntityType entityType = structuredType as IEdmEntityType;
+            if (entityType != null)
             {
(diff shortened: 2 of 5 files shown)


test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBoundQuerySettings/FilterAttributeTest/FilterAttributeTest.cs
```diff
@@ -186,5 +193,27 @@ public async Task FilterSingleNavigationOrComplexProperty(string entitySetUrl, i
                 Assert.Contains("cannot be used in the $filter query option.", result);
             }
         }
+
+        [Theory]
+        [InlineData(AuthorBaseUrl + "?$filter=Books/any(d: d/BookId eq 7)")]
+        [InlineData(ModelBoundAuthorBaseUrl + "?$filter=Books/any(d: d/BookId eq 7)")]
```
xuzhg, line 199: You changed codes in SelectExpandBinder, but the test doesn't refresh the changes?

xuzhg (approved):

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/ModelBoundQuerySettings/FilterAttributeTest/FilterAttributeTest.cs
gathogojr: The change in `SelectExpandBinder` was to address a warning by the analyzer.
I had `if (structuredType is IEdmEntityType entityType)` and `if (structuredType is IEdmEntityType)` in the same method (introduced in my last PR) and the analyzer for the AspNet Web API project wasn't too happy about that.

habbes (approved):

KenitoInc (approved):

PR #2796, opened 2023-07-19
Fix broken serialization for $select used in non-Edm scenarios

### LINKED ISSUE #1929: .Net core OData EnableQueryAttribute not working with $select
My controller action works fine with OData $select operation (without EnableQueryAttribute). However as I need a default pageSize to be specified I need to use the `EnableQuery(PageSize=100)` on my controller action. Once I do this OData throws and ODataException with message

> The query specified in the URI is not valid. Could not find a property named 'PropertyA' on type 'Microsoft.AspNet.OData.Query.Expressions.SelectSomeAndInheritance_1OfPropertiesViewModel'.

This only happens when I use the EnableQueryAttribute.

(NB. I am aware that I can specify a pageSize with the ODataQueryOptions.ApplyTo (iQueryable, new ODataQuerySettings{PageSize=100} but this is not suitable as it causes the query to execute. ))

### Assemblies affected

 I am using version 7.2.2 of Microsoft.AspNetCore.OData

### LINKED ISSUE #2038: OData adding unwanted properties in the response
*OData adding unwanted properties in the response when using the default System.Text.Json serializer in .net core 3.1. Changing the json serializer to Newtonsoft fixes the issue*

### Assemblies affected

*AspnetCore OData WebApi lib 7.3.0*

### Reproduce Steps

I created a .net core 3.1 web api and enabled the query options on the controller actions.
Startup.cs:
--------------
Have below in the ConfigureServices() method:
services.AddOData();
services.AddControllers(options => {
                options.EnableEndpointRouting = false;
}

Have below in the Configure() method :
app.UseMvc(routeBuilder => {
                routeBuilder.EnableDependencyInjection();
                routeBuilder.Expand().Select().OrderBy().Filter();
            });

OData url: GET /api/students?$expand=students($filter=id eq '123' or id eq '124';$select=id,nickname,name)

### Expected result

*{
	"students": [
            {
                "id": "123",
                "nickname": "Bill",
                "name": "William Smith",                
            },
            {
                "id": "124",
                "nickname": "John",
                "name": "John Doe",
            }
        ]
}*

### Actual result

*{
    "instance": {
        "students": [
            {
                "id": "123",
                "nickname": "Bill",
                "name": "William Smith",                
            },
            {
                "id": "124",
                "nickname": "John",
                "name": "John Doe",
            }
        ]
    },
    "container": {},
    "modelID": "fd9f6c72-dab4-4935-b492-d03c75348a90",
    "untypedInstance": {
        "students": [
            {
                "id": "123",
                "nickname": "Bill",
                "name": "William Smith",                
            },
            {
                "id": "124",
                "nickname": "John",
                "name": "John Doe",
            }
        ]
    },
    "instanceType": null,
    "useInstanceForProperties": true
}*

### LINKED ISSUE #2174: System.Text.Json support
The latest version of `Microsoft.AspNetCore.OData` (v7.4.1) still depends on `NewtonSoft.Json` for correct JSON serialization, e.g. when using `$select`.

Using OData with AspNetCore's current default serializer `System.Text.Json` will not serialize projection queries correctly. 

Please see #1929 or #2038 for further details.

I'm happy that endpoint routing is finally supported. When can we expect `System.Text.Json` support, so that we can combine AspNetCore.OData with the latest AspNetCore default stack - without the "fallback to NewtonSoft workaround" using `Microsoft.AspNetCore.Mvc.NewtonsoftJson`?

gathogojr:
### Issues

*This pull request fixes #1929, fixes #2174, and fixes #2038.*

### Description

Following the release of .NET Core 3.1, serialization broke when `$select` query option is applied to the URL in non-Edm scenarios. This happened because in ASP.NET Core 3.1, the default serialization library was changed from Newtonsoft.Json to System.Text.Json. There existed custom converters based on Newtonsoft.Json that were used in non-Edm scenarios but they could no longer applied after the default serialization library changed.

This pull request implements custom converters based on System.Text.Json that will be applied in non-Edm scenarios for .NET Core 3.1 or higher.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
@@ -1,0 +1,311 @@
+//-----------------------------------------------------------------------------
+// <copyright file="SelectExpandWrapperJsonConverter.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved. 
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+#if NETCOREAPP3_1_OR_GREATER
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.AspNet.OData.Common;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNet.OData.Query.Expressions
+{
+    /// <summary>
+    /// Represents a custom <see cref="JsonConverter"/> to serialize <see cref="SelectExpandWrapper{TElement}"/> instances to JSON.
+    /// </summary>
+    internal class SelectExpandWrapperJsonConverter : JsonConverterFactory
+    {
+        private static readonly Func<IEdmModel, IEdmStructuredType, IPropertyMapper> _mapperProvider =
+            (IEdmModel model, IEdmStructuredType type) => new JsonPropertyNameMapper(model, type);
+
+        /// <inheritdoc/>
+        public override bool CanConvert(Type typeToConvert)
+        {
+            if (typeToConvert == null)
+            {
+                throw Error.ArgumentNull(nameof(typeToConvert));
+            }
+
+            if (!typeToConvert.IsGenericType)
+            {
+                return false;
+            }
+
+            return typeof(ISelectExpandWrapper).IsAssignableFrom(typeToConvert);
+        }
+
+        /// <inheritdoc/>
+        public override JsonConverter CreateConverter(
+            Type typeToConvert,
+            JsonSerializerOptions options)
+        {
+            if (typeToConvert == null || !typeToConvert.IsGenericType)
+            {
+                return null;
+            }
+
+            Type genericType = typeToConvert.GetGenericTypeDefinition();
+            Type type = typeToConvert.GetGenericArguments()[0];
+
+            if (genericType == typeof(SelectExpandBinder.SelectSome<>))
+            {
+                return (JsonConverter)Activator.CreateInstance(
+                    typeof(SelectSomeOfTJsonConverter<>).MakeGenericType(new Type[] { type }));
+            }
+
+            if (genericType == typeof(SelectExpandBinder.SelectSomeAndInheritance<>))
+            {
+                return (JsonConverter)Activator.CreateInstance(
+                    typeof(SelectSomeAndInheritanceOfTJsonConverter<>).MakeGenericType(new Type[] { type }));
+            }
+
+            if (genericType == typeof(SelectExpandBinder.SelectAll<>))
+            {
+                return (JsonConverter)Activator.CreateInstance(
+                    typeof(SelectAllOfTJsonConverter<>).MakeGenericType(new Type[] { type }));
+            }
+
+            if (genericType == typeof(SelectExpandBinder.SelectAllAndExpand<>))
+            {
+                return (JsonConverter)Activator.CreateInstance(
+                    typeof(SelectAllAndExpandOfTJsonConverter<>).MakeGenericType(new Type[] { type }));
+            }
+
+            if (genericType == typeof(SelectExpandWrapper<>))
+            {
+                return (JsonConverter)Activator.CreateInstance(
+                    typeof(SelectExpandWrapperOfTJsonConverter<>).MakeGenericType(new Type[] { type }));
+            }
+
+            return null;
+        }
+
+        private class JsonPropertyNameMapper : IPropertyMapper
+        {
+            private IEdmModel _model;
+            private IEdmStructuredType _type;
+
+            public JsonPropertyNameMapper(IEdmModel model, IEdmStructuredType type)
+            {
+                _model = model;
+                _type = type;
+            }
+
+            public string MapProperty(string propertyName)
+            {
+                IEdmProperty property = _type.Properties().First(s => s.Name == propertyName);
+                PropertyInfo propertyInfo = GetPropertyInfo(property);
+
+                Newtonsoft.Json.JsonPropertyAttribute jsonPropertyAttribute;
+                JsonPropertyNameAttribute jsonPropertyNameAttribute;
+
+                // NOTE: Preceding .NET Core 3.1 Newtonsoft was the default serialization library
+                // We check the Newtonsoft JsonPropertyAttribute first to prevent breaking changes
+                if ((jsonPropertyAttribute = GetNewtonsoftJsonPropertyAttribute(propertyInfo)) != null && !string.IsNullOrWhiteSpace(jsonPropertyAttribute.PropertyName))
+                {
+                    return jsonPropertyAttribute.PropertyName;
+                }
+
+                if ((jsonPropertyNameAttribute = GetSystemTextJsonPropertyNameAttribute(propertyInfo)) != null && !string.IsNullOrWhiteSpace(jsonPropertyNameAttribute.Name))
+                {
+                    return jsonPropertyNameAttribute.Name;
+                }
+
+                return property.Name;
+            }
+
+            private PropertyInfo GetPropertyInfo(IEdmProperty property)
+            {
+                ClrPropertyInfoAnnotation clrPropertyAnnotation = _model.GetAnnotationValue<ClrPropertyInfoAnnotation>(property);
+                if (clrPropertyAnnotation != null)
+                {
+                    return clrPropertyAnnotation.ClrPropertyInfo;
+                }
+
+                ClrTypeAnnotation clrTypeAnnotation = _model.GetAnnotationValue<ClrTypeAnnotation>(property.DeclaringType);
+                Contract.Assert(clrTypeAnnotation != null);
+
+                PropertyInfo propertyInfo = clrTypeAnnotation.ClrType.GetProperty(property.Name);
+                Contract.Assert(propertyInfo != null);
+
+                return propertyInfo;
+            }
+
+            private static Newtonsoft.Json.JsonPropertyAttribute GetNewtonsoftJsonPropertyAttribute(PropertyInfo property)
+            {
+                return property.GetCustomAttributes(typeof(Newtonsoft.Json.JsonPropertyAttribute), inherit: false)
+                    .OfType<Newtonsoft.Json.JsonPropertyAttribute>().SingleOrDefault();
+            }
+
+            private static JsonPropertyNameAttribute GetSystemTextJsonPropertyNameAttribute(PropertyInfo property)
+            {
+                return property.GetCustomAttributes(typeof(JsonPropertyNameAttribute), inherit: false)
+                    .OfType<JsonPropertyNameAttribute>().SingleOrDefault();
+            }
+        }
+
+        /// <summary>
+        /// Represents a custom <see cref="JsonConverter"/> to serialize <see cref="SelectExpandBinder.SelectSome{TElement}"/> instances to JSON.
+        /// </summary>
+        private class SelectSomeOfTJsonConverter<TEntity>
+            : JsonConverter<SelectExpandBinder.SelectSome<TEntity>>
+        {
+            /// <inheritdoc/>
+            public override SelectExpandBinder.SelectSome<TEntity> Read(
+                ref Utf8JsonReader reader,
+                Type typeToConvert,
+                JsonSerializerOptions options)
+            {
+                throw new NotImplementedException(
+                    Error.Format(
+                        SRResources.JsonConverterDoesNotSupportRead,
+                        typeof(SelectExpandBinder.SelectSome<TEntity>).Name));
+            }
+
+            /// <inheritdoc/>
+            public override void Write(
+                Utf8JsonWriter writer,
+                SelectExpandBinder.SelectSome<TEntity> value,
+                JsonSerializerOptions options)
+            {
+                JsonSerializer.Serialize(
+                    writer,
+                    value.ToDictionary(_mapperProvider), typeof(IDictionary<string, object>),
+                    options);
+            }
+        }
+
+        /// <summary>
+        /// Represents a custom <see cref="JsonConverter"/> to serialize <see cref="SelectExpandBinder.SelectSomeAndInheritance{TElement}"/> instances to JSON.
+        /// </summary>
+        private class SelectSomeAndInheritanceOfTJsonConverter<TEntity>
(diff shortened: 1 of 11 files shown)


KenitoInc (approved):

test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Build.AspNetCore3x/NonEdm/NonEdmTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            : base(fixture)
+        {
+        }
+
+        protected override void UpdateConfiguration(WebRouteConfiguration configuration)
+        {
+            configuration.AddControllers(typeof(CustomersController));
+            configuration.Select().Filter().OrderBy().Expand().Count().MaxTop(null);
+
+            // Force use of System.Text.Json output formatter
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc Because in the test fixture, `AddNewtonsoftJson` method is called and that causes the Newtonsoft.Json formatters to have highest precedence: https://github.com/OData/WebApi/blob/2c211035d8e25e193d1c5a3b58999581cfa05a4d/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/Common/Execution/WebHostTestFixture.cs#L239

By deliberately inserting the System.Text.Json output formatter at index 0 of the formatters collection, we ensure that the formatters I added are used for the tests in this test file particularly.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+
+            public JsonPropertyNameMapper(IEdmModel model, IEdmStructuredType type)
+            {
+                _model = model;
+                _type = type;
+            }
+
+            public string MapProperty(string propertyName)
+            {
+                IEdmProperty property = _type.Properties().Single(s => s.Name == propertyName);
```
habbes, line 100: Do we need to validate that the properties names are unique at this point? Shouldn't this be handled by EdmLib or the writer? Why not use `.First()` or `.FirstOrDefault()`?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+                PropertyInfo propertyInfo = GetPropertyInfo(property);
+
+                Newtonsoft.Json.JsonPropertyAttribute jsonPropertyAttribute;
+                JsonPropertyNameAttribute jsonPropertyNameAttribute;
+
+                // NOTE: Preceding .NET Core 3.1 Newtonsoft was the default serialization library
+                // We check the Newtonsoft JsonPropertyAttribute first to prevent breaking changes
```
habbes, line 107: 👍
I was wondering why we're referencing Newtonsoft.Json in System.Text.Json converter. Thanks for clearing that up.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+                return info;
+            }
+
+            private static Newtonsoft.Json.JsonPropertyAttribute GetJsonPropertyAttribute(PropertyInfo property)
+            {
+                return property.GetCustomAttributes(typeof(Newtonsoft.Json.JsonPropertyAttribute), inherit: false)
+                    .OfType<Newtonsoft.Json.JsonPropertyAttribute>().SingleOrDefault();
+            }
+
+            private static JsonPropertyNameAttribute GetJsonPropertyNameAttribute(PropertyInfo property)
```
habbes, line 145: Feel free to ignore this, but I think it would be less confusing for the names if something about the library is mentioned in the name, e.g. `GetNewtonsoftJsonPropertyAttribute`, Get `GetSystemJsonPropertyNameAttribute`

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNet.OData.Query.Expressions
+{
+    /// <summary>
+    /// Represents a custom <see cref="JsonConverter"/> to serialize <see cref="SelectExpandWrapper{TElement}"/> instances to JSON.
+    /// </summary>
+    internal class SelectExpandWrapperJsonConverter : JsonConverterFactory
+    {
+        public static readonly Func<IEdmModel, IEdmStructuredType, IPropertyMapper> _mapperProvider =
```
habbes, line 26: Where is the `IPropertyMapper` defined? And does it only map the property name or also the value?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+                    return jsonPropertyAttribute.PropertyName;
+                }
+                else if ((jsonPropertyNameAttribute = GetJsonPropertyNameAttribute(propertyInfo)) != null && !string.IsNullOrWhiteSpace(jsonPropertyNameAttribute.Name))
+                {
+                    return jsonPropertyNameAttribute.Name;
+                }
+                else
+                {
+                    return property.Name;
+                }
```
habbes, line 119: Nit: This statement doesn't need to be in an `else` block.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+                Newtonsoft.Json.JsonPropertyAttribute jsonPropertyAttribute;
+                JsonPropertyNameAttribute jsonPropertyNameAttribute;
+
+                // NOTE: Preceding .NET Core 3.1 Newtonsoft was the default serialization library
+                // We check the Newtonsoft JsonPropertyAttribute first to prevent breaking changes
+                if ((jsonPropertyAttribute = GetJsonPropertyAttribute(propertyInfo)) != null && !string.IsNullOrWhiteSpace(jsonPropertyAttribute.PropertyName))
+                {
+                    return jsonPropertyAttribute.PropertyName;
+                }
+                else if ((jsonPropertyNameAttribute = GetJsonPropertyNameAttribute(propertyInfo)) != null && !string.IsNullOrWhiteSpace(jsonPropertyNameAttribute.Name))
```
habbes, line 112: nit: The `else` is unnecessary.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+
+            /// <inheritdoc/>
+            public override void Write(
+                Utf8JsonWriter writer,
+                SelectExpandBinder.SelectSomeAndInheritance<TEntity> value,
+                JsonSerializerOptions options)
+            {
+                JsonSerializer.Serialize(
+                    writer,
+                    value.ToDictionary(_mapperProvider),
```
habbes, line 209: Does the Newtonsoft converter version also use `ToDictionary`?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandWrapperJsonConverterTests.cs
```diff
@@ -0,0 +1,227 @@
...
+            public Dictionary<string, object> Properties { get; private set; }
+
+            public override void ToDictionaryCore(
+                Dictionary<string, object> dictionary,
+                IPropertyMapper propertyMapper,
+                bool includeAutoSelected)
+            {
+                foreach (var kvp in Properties)
+                {
+                    dictionary.Add(kvp.Key, kvp.Value);
```
habbes, line 216: I thought the `propertyMapper` would be used to map the property names.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
gathogojr: @habbes Yes it does https://github.com/OData/WebApi/blob/b810e7185d45104f16fcd8ae1fc18cfeef923658/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperConverter.cs#L47

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandWrapperJsonConverterTests.cs
gathogojr: @gathogojr This is a quick and dirty implementation of the `PropertyContainer`. The logic for the `PropertyContainer` and its derivations didn't change in this pull request.

Be that as it may, I should explain that for the tests in this file, it really doesn't matter.

If you look at the helper method named `SerializerPayloadUsingSelectExpandWriterJsonConverter`, we set the `UseInstanceForProperty` property of the `SelectExpandWrapper` to `true`
```csharp
selectExpandWrapper.UseInstanceForProperties = true;
```
By doing that, we're indicating that we want all the structural properties on the instance.
In addition, we also initialize the `Instance` property of the `SelectExpandWrapper<T>`:
```
selectExpandWrapperOfT.Instance = new SampleEntity
{
    Property = "foobar"
};
```
By doing that, the following logic in the `SelectExpandWrapper` will be executed. By design, it falls back on the instance:
- https://github.com/OData/WebApi/blob/b810e7185d45104f16fcd8ae1fc18cfeef923658/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs#L135
- https://github.com/OData/WebApi/blob/b810e7185d45104f16fcd8ae1fc18cfeef923658/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapper.cs#L87

gathogojr: @habbes I have gone further and modified the logic such that the `MockPropertyContainer` makes use of the `IPropertyMapper` to map property names. In addition, I added tests for advanced scenarios including those where `UseInstanceForProperties` is set to `false` and others where properties are decorated with `JsonProperty` attribute (Newtonsoft) and `JsonPropertyName` attribute (System.Text.Json) and both.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
gathogojr: `IPropertyMapper` is an interface defined in `Microsoft.AspNet.OData.Query` namespace. It maps property names only. It's relevant where the property name to be written on the wire is different from the property name on the type. For example:
```csharp
public class TestEntity
{
    [JsonProperty("Prop")]
    public string Property { get; set; }
}
```

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
ElizabethOkerio: I don't understand this answer..

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,311 @@
...
+            {
+                _model = model;
+                _type = type;
+            }
+
+            public string MapProperty(string propertyName)
+            {
+                IEdmProperty property = _type.Properties().First(s => s.Name == propertyName);
+                PropertyInfo propertyInfo = GetPropertyInfo(property);
+
```
ElizabethOkerio, line 106: nit: Can the `Newtonsoft.Json` be added as an import

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
```diff
@@ -0,0 +1,311 @@
...
+                Newtonsoft.Json.JsonPropertyAttribute jsonPropertyAttribute;
```
ElizabethOkerio, line 107: nit: add Newtonsoft.Json as import

gathogojr: @ElizabethOkerio Because there are types named the same between Newtonsoft.Json and System.Text.Json libraries and we only use the Newtonsoft types a few places, I thought it'd be cleaner to not import Newtonsoft.Json. It'd require me to alias the import anyway to prevent types from clashing.

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Query/Expressions/SelectExpandWrapperJsonConverterTests.cs
```diff
@@ -0,0 +1,305 @@
...
+            var converter = converterFactory.CreateConverter(typeof(T), jsonSerializerOptions) as JsonConverter<T>;
+
+            var utf8JsonReader = new Utf8JsonReader();
+
+            try
+            {
+                // Act
+                converter.Read(ref utf8JsonReader, typeof(object), jsonSerializerOptions);
+            }
+            catch (NotImplementedException ex)
```
KenitoInc, line 198: We can't use `Assert.Throws` here?

KenitoInc (approved):

ElizabethOkerio (approved):

habbes (approved):

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandWrapperJsonConverter.cs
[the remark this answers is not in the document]
gathogojr: @habbes In Newtonsoft.Json, the attribute is `JsonPropertyAttribute` while in System.Text.Json the attribute is `JsonPropertyNameAttribute`. The naming of the two variables is based on that differentiation. And yes, I could have gone further and prepended with `newtonsoft` to differentiate further.

PR #2803, opened 2023-08-31
Fix for auto-generated swagger missing $filter for EntitySet

### LINKED ISSUE #2802: Swagger generator is missing $filter from EntitySet "get" definitions
For swagger generation, the ODataSwaggerConverter currently includes many query parameters ($expand, $select, $count, etc) for EntitySet "get" path definitions, but is missing the $filter parameter. Can this please be added? I think something like this to the [ODataSwaggerUtilities.cs CreateSwaggerPathForEntitySet() method](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/ODataSwaggerUtilities.cs#L43)
```
.Parameter("$filter", "query", "Filter by some expression", "string")
```

sakana280:
### Issues

This pull request fixes #2802

### Description

Adds the $filter parameter to the auto-generated swagger.

- [x] *Test cases added*

Builds and all tests pass inside VS2019 Community. I couldn't get build.cmd working sorry, it either hung or errored from FxCop.

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/ODataSwaggerUtilities.cs
+++ b/src/Microsoft.AspNet.OData.Shared/ODataSwaggerUtilities.cs
@@ -40,6 +40,7 @@ namespace Microsoft.AspNet.OData
                         .Description("Returns the EntitySet " + entitySet.Name)
                         .Tags(entitySet.Name)
                         .Parameters(new JArray()
+                            .Parameter("$filter", "query", "Filter by some expression", "string")
                             .Parameter("$expand", "query", "Expand navigation property", "string")
                             .Parameter("$select", "query", "select structural property", "string")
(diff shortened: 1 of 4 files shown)


sakana280:
@microsoft-github-policy-service agree

sakana280:
I've just noticed I broke one of the E2E tests and have just pushed an update to fix it. Some of the other E2E tests fail during "run all" tests in Visual Studio for WebApiOData.E2E.AspNetCore.sln, but then pass when the tests are run individually. QueryForLongSelectList is the only E2E test i couldn't get passing, but it is also failing in master, on my PC at least, so I don't think adding $filter to the swagger broke that test.

xuzhg (approved):

gathogojr (approved):
:shipit:

PR #2804, opened 2023-09-07
Fix case-insensitive property matching

### LINKED ISSUE #2635: Case-Insensitive Property Matching NullRefs
After upgrading to `7.5.14` we observed a `NullReferenceException` in one of our test cases. After drilling in we found it was related to https://github.com/OData/WebApi/pull/2311 which swapped to allowing case-insensitive property matching during deserialization. We've worked around this by opting into the old behavior, but the nullref needs to be fixed.

### Assemblies affected

Microsoft.AspNet.OData 7.5.14

### Reproduce steps

1. Send a POST request to update the value of `myProperty` using `myproperty` as the name

### Expected result

Request should succeed

### Actual result

Request fails

### Additional detail

It looks like the property is found (i.e. it determines it's NOT dynamic), but then a different property name lookup files to find the property type:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Object.GetType()
   at Microsoft.AspNet.OData.Formatter.EdmPrimitiveHelpers.ConvertPrimitiveValue(Object value, Type type)
   at Microsoft.AspNet.OData.Formatter.Deserialization.DeserializationHelpers.SetDeclaredProperty(Object resource, EdmTypeKind propertyKind, String propertyName, Object propertyValue, IEdmProperty edmProperty, ODataDeserializerContext readContext)
   at Microsoft.AspNet.OData.Formatter.Deserialization.DeserializationHelpers.ApplyProperty(ODataProperty property, IEdmStructuredTypeReference resourceType, Object resource, ODataDeserializerProvider deserializerProvider, ODataDeserializerContext readContext)
   at Microsoft.AspNet.OData.Formatter.Deserialization.ODataResourceDeserializer.ApplyStructuralProperties(Object resource, ODataResourceWrapper resourceWrapper, IEdmStructuredTypeReference structuredType, ODataDeserializerContext readContext)
   at Microsoft.AspNet.OData.Formatter.Deserialization.ODataResourceDeserializer.ApplyResourceProperties(Object resource, ODataResourceWrapper resourceWrapper, IEdmStructuredTypeReference structuredType, ODataDeserializerContext readContext)
   at Microsoft.AspNet.OData.Formatter.Deserialization.ODataResourceDeserializer.ReadResource(ODataResourceWrapper resourceWrapper, IEdmStructuredTypeReference structuredType, ODataDeserializerContext readContext)
   at Microsoft.AspNet.OData.Formatter.Deserialization.ODataResourceSetDeserializer.<ReadResourceSet>d__5.MoveNext()
```

KenitoInc:
### Issues

*This pull request fixes #2635 .*

### Description

If we have an entitytype `Book`
```csharp
public class Book
{
    public int Id { get; set; }
    public string? Title { get; set; }
    public IDictionary<string, object> DynamicProperties { get; set; }
}
```
And we send a POST request to update the value of `Title` on an open-type using `title` as the name and have a null value
```json
{
  "title": null
}
```
An exception is thrown.

By default, we set the WebAPI requests to be `case-insensitive`, however we usually don't set `EnablePropertyNameCaseInsensitive=true` in the `ODataReaderSettings` that we use to intialize the `ODataMessageReader`. This causes the `title` property to be deserialized as an `ODataUntypedValue` with RawValue as string "null". That's why we treat the value as a primitive value and this fails in the `ConvertPrimitiveValue` method.

If we set `EnablePropertyNameCaseInsensitive` as `true` in the `ODataReaderSettings`, the `title` property is deserialized as an `ODataNullValue`.

This PR ensures that set the correct `EnablePropertyNameCaseInsensitive` in the `ODataReaderSettings`

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs
@@ -82,6 +82,15 @@ namespace Microsoft.AspNet.OData.Formatter
                 oDataReaderSettings.Version = version;
                 oDataReaderSettings.MaxProtocolVersion = version;
 
+                ODataPath path = internalRequest.Context.Path;
+                ODataDeserializerContext readContext = getODataDeserializerContext();
+                readContext.Path = path;
+                readContext.Model = model;
+                readContext.ResourceType = type;
+                readContext.ResourceEdmType = expectedPayloadType;
+
+                oDataReaderSettings.EnablePropertyNameCaseInsensitive = !readContext.DisableCaseInsensitiveRequestPropertyBinding;
+
                 IODataRequestMessage oDataRequestMessage = getODataRequestMessage();
 
                 string preferHeader = RequestPreferenceHelpers.GetRequestPreferHeader(internalRequest.Headers);
@@ -99,13 +108,6 @@ namespace Microsoft.AspNet.OData.Formatter
 
                 ODataMessageReader oDataMessageReader = new ODataMessageReader(oDataRequestMessage, oDataReaderSettings, model);
                 registerForDisposeAction(oDataMessageReader);
-
-                ODataPath path = internalRequest.Context.Path;
-                ODataDeserializerContext readContext = getODataDeserializerContext();
-                readContext.Path = path;
-                readContext.Model = model;
-                readContext.ResourceType = type;
-                readContext.ResourceEdmType = expectedPayloadType;
 
 #if NETCORE
                 result = await deserializer.ReadAsync(oDataMessageReader, type, readContext);
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingTest.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingTest.cs
@@ -420,7 +420,7 @@ namespace Microsoft.AspNet.OData.Test.Routing
         [Theory]
         [InlineData("CollectionOfPrimitiveTypeFunction(intValues=@p)?@p=[1,2,4,7,8")] // missing "]"
         [InlineData("CanMoveToAddress(address=@p)?@p={\"@odata.")] // not valid complex payload
-        [InlineData("EntityTypeFunction(product=@p)?@p={\"@odata.type\":\"Microsoft.AspNet.OData.Test.Routing.Product\",\"id\":9,\"Name\":\"Phone\"}")] // should be "ID"
+        [InlineData("EntityTypeFunction(product=@p)?@p={\"@odata.type\":\"Microsoft.AspNet.OData.Test.Routing.Product\",\"id1\":9,\"Name\":\"Phone\"}")] // should be "ID"
         public async Task RoutesIncorrectly_ForBadFunctionParameters(string uri)
         {
             // Arrange & Act
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataModelBinderConverter.cs
@@ -232,6 +232,7 @@ namespace Microsoft.AspNet.OData.Formatter
 
             IWebApiRequestMessage request = readContext.InternalRequest;
             ODataMessageReaderSettings oDataReaderSettings = request.ReaderSettings;
+            oDataReaderSettings.EnablePropertyNameCaseInsensitive = !readContext.DisableCaseInsensitiveRequestPropertyBinding;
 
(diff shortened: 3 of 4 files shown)


test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Routing/ODataRoutingTest.cs
```diff
@@ -420,7 +420,7 @@ public async Task RoutesCorrectly(string httpMethod, string uri, string expected
         [Theory]
         [InlineData("CollectionOfPrimitiveTypeFunction(intValues=@p)?@p=[1,2,4,7,8")] // missing "]"
         [InlineData("CanMoveToAddress(address=@p)?@p={\"@odata.")] // not valid complex payload
-        [InlineData("EntityTypeFunction(product=@p)?@p={\"@odata.type\":\"Microsoft.AspNet.OData.Test.Routing.Product\",\"id\":9,\"Name\":\"Phone\"}")] // should be "ID"
+        [InlineData("EntityTypeFunction(product=@p)?@p={\"@odata.type\":\"Microsoft.AspNet.OData.Test.Routing.Product\",\"id1\":9,\"Name\":\"Phone\"}")] // should be "ID"
         public async Task RoutesIncorrectly_ForBadFunctionParameters(string uri)
```
xuzhg, line 424: why changing from "id" to "id1"? It doesn't look like a case-sensitive? it's a different name

KenitoInc: This is an existing test that was meant to fail when properties are case-insensitive. Since we support case-insensitive property binding by default, I opted to change the property name.

gathogojr: @KenitoInc Wasn't the right action to remove the test case instead? Consider the comment at the end of the test case - `// should be "ID"`. Isn't that misleading now?

KenitoInc: I believe that the wrong property name counts as "Bad function parameter" (see test name)
`// should be "ID"` indicates that the correct property name should have been `ID`

gathogojr: Strictly speaking, relatively to `ID`, `id1` is not a wrong property name - it's a different property name. It doesn't count as bad function parameter.

gathogojr (approved):
:shipit:

habbes:
@KenitoInc I don't fully understand the problem statement is the description. Are the properties `title` and `Title` related? You say that we send a POST request to update the value of `Title` on an open type using `title` as the name. What is the intended outcome of this `POST`? Would the created entity have both a structural property `Title` and a dynamic property `title` with the same value? The rest of the description only mentions `title` and not `Title`. As a result, I'm not sure whether and how the two are related, what the expected outcome or what exactly this PR intends to fix.

src/Microsoft.AspNet.OData.Shared/Formatter/ODataInputFormatterHelper.cs
```diff
@@ -82,6 +82,15 @@ internal static class ODataInputFormatterHelper
...
                 oDataReaderSettings.MaxProtocolVersion = version;
 
+                ODataPath path = internalRequest.Context.Path;
+                ODataDeserializerContext readContext = getODataDeserializerContext();
+                readContext.Path = path;
+                readContext.Model = model;
+                readContext.ResourceType = type;
+                readContext.ResourceEdmType = expectedPayloadType;
+
+                oDataReaderSettings.EnablePropertyNameCaseInsensitive = !readContext.DisableCaseInsensitiveRequestPropertyBinding;
```
habbes, line 92: Do you have test cases for both scenarios (when enabled and disabled)?

KenitoInc: Added the tests

habbes:
> @KenitoInc I don't fully understand the problem statement is the description. Are the properties `title` and `Title` related? You say that we send a POST request to update the value of `Title` on an open type using `title` as the name. What is the intended outcome of this `POST`? Would the created entity have both a structural property `Title` and a dynamic property `title` with the same value? The rest of the description only mentions `title` and not `Title`. As a result, I'm not sure whether and how the two are related, what the expected outcome or what exactly this PR intends to fix.

Adding context based on our offline chat: In the scenario described in the description, the expectation is that the `title` field in the request body will be bound to the `Title` property of the `Book` object and should not be treated as dynamic property. The reason it was being treated as a dynamic property is because the case-insensitive setting (which is enabled by default in WebAPI)  was not properly propagated in ODL, causing the property to be treated as a dynamic property.

habbes (approved):

PR #2807, opened 2023-09-21
enable write dynamic complex for delta entity and complex

### LINKED ISSUE #2806: Cannot serialize EdmDeltaComplexObject in odata 7x and above
ODataResourceSerializer fails to serialize resource of type **EdmDeltaComplexObject** for Delta entities.

### Assemblies affected

OData 7x and above

### Reproduce steps

While trying to serialize the resource of type `EdmDeltaComplexObject`

```
public class Sample
{
    public string Description
    {  get; set;  }

    public SampleItemCollection Samplelist
    {  get; set;  }
}

public class SampleItemCollection : EdmDeltaComplexObject
{
     public IDictionary<string, object> DynamicProperties { get; set; }
}
```

When an instance of the above Sample class is edited and a delta sync request is sent to fetch the changes,
Expected response is:

```
{
  ""samplelist"": {
    ""key-samplelist"": {
      ""title"": ""Title"",
    }
  },
  ""description"": "an edit to sample item"
}
```

Actual response seen in v7x:

```
{
  ""samplelist"": {
  },
  ""description"": "an edit to sample item"
}
```

### Additional detail

Reasoning:
In OData V5x & 6x Complex type property is treated as a struct property, and serialization for a complex property was performed similar to primitive property.
however, in OData v7x, complex properties are treated as a resource and classified as DynamicComplexProperties.

It looks like a feature gap/regression bug and solution would be to call the WriteDynamicComplexProperties for delta entities.

xuzhg:
### Issues

*This pull request fixes #2806.*

### Description

* Enable write complex dynamic for delta

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -2087,6 +2087,7 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
                 {
                     writer.WriteStart(resource);
                     WriteDeltaComplexProperties(selectExpandNode, resourceContext, writer);
+                    WriteDynamicComplexProperties(resourceContext, writer);
                     WriteDeltaNavigationProperties(selectExpandNode, resourceContext, writer);
                     writer.WriteEnd();
                 }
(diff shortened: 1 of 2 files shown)


habbes (approved):

gathogojr (approved):
:shipit:

PR #2811, opened 2023-10-09
Fixes #2810, make action/function link works using lower camel case

### LINKED ISSUE #2810: GeneateActionLink should apply lowercase for bindingParameterType
Currently, while generating an action link for an odata action, the LinkGenerationHelpers.GenerateActionLink method looks for a parameter type which does not match the ones in the IEdmModel.
This is because the method uses the bindingParameter.Type parameter which does not enforce lowercase, however the model schema elements are lowercased.
The solution would be to replace the Type with the fullName parameter.

### Assemblies affected

AspNet.OData lib 7x

### Reproduce steps

Consider an action called simpleAction, which is applied on a type 'Plan' with namespace microsoft.simple.plan,
```
ActionConfiguration simpleAction= builder.EntitySet<Plan>("plan").EntityType
    .Action("simpleAction");
```
Here, the model contains the declared type "microsoft.simple.plan" and the bindingParameter's type shows "microsoft.simple.Plan".
this results in a mismatch and a NullReferenceException.

### Expected result
The result should return a link generated for actions
```
"#microsoft.simple.simpleAction": {
        "title": "simpleAction",
        "target": "http://localhost/V3.0/plans('lsPtJMLRzUyfE12cVfCv-f8ABBP1')/microsoft.simple.simpleAction"
      }
```

### Actual result

System.NullReferenceException: Object reference not set to an instance of an object.

xuzhg:
### Issues

*This pull request fixes #2810.*

### Description

- [x] *Test cases added*

### Additional work necessary


habbes:
@xuzhg in the issue thread you said that the customer scenario is not the intended usage of `OnModelCreating`. Is there an alternative/recommended way the customer can implement the intended use-case without resorting to changing names in `OnModelCreating`? Maybe that could be explored before changing the library?

gathogojr (approved):

xuzhg:
> @xuzhg in the issue thread you said that the customer scenario is not the intended usage of `OnModelCreating`. Is there an alternative/recommended way the customer can implement the intended use-case without resorting to changing names in `OnModelCreating`? Maybe that could be explored before changing the library?

There are a lot of ways, for example, Customer can use 'Name' property to change the type of name before running conventions. 
Or customer can call "HasActionLink" or "HasFunctionLink" to override default, etc. We'd better to retrieve the action/function then use the binding type of it to build link, not use 'bindingParameterTypeName' string from action configuration.

habbes (approved):

PR #2821, opened 2023-11-16
Add a service to config the OData version for delta payload

### LINKED ISSUE #2816: OData-prefix missing in delta responses
OData-prefix missing in delta responses. I was able to figure the issue here, it looks like the [ODataOutputFormatterHelper ](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs) class, line 163 applies an override which always assigns odata version v4.01 to all delta payloads, no matter what we provide for the "OData-MaxVersion" in the request header.

I see that the specification for V4.01 states that odata-prefix will be dropped for optimization purposes. 
I believe fixing this and providing flexibility for customers to choose the version would resolve this issue.

Similarly, the deleted Entity delta response also uses the v4.01 for response, and the responses thus have a different nested delta format (explained below)

 
### Assemblies affected

OData 7x and above

### Reproduce steps

Any delta response for odatav7x can be considered for this.

### Expected result

```
{
    "@odata.context": "https://sample.com/beta/$metadata/entity",
    "@odata.type": "microsoft.graph.plannerPlanConfiguration",
    "@odata.etag": "etag",
    "key": "value"
}
```

Deleted Entity response:
```
{
  ""@odata.context"": ""http://localhost/V3.0/$metadata#tasks/$deletedEntity"",
  ""id"": ""taskId"",
  ""reason"": ""deleted""
}
```

### Actual result

```
{
    "@context": "https://sample.com/beta/$metadata/entity",
    "@type": "microsoft.graph.plannerPlanConfiguration",
    "@etag": "etag",
    "key": "value"
}
```

Deleted Entity response:
```
{
  ""@context"": ""http://localhost/V3.0/$metadata#tasks/$deletedEntity"",
  ""@removed"": {
    ""reason"": ""deleted""
  },
  ""@id"": ""taskId""
}
```

xuzhg:
### Issues

*This pull request fixes #2816.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
@@ -127,7 +127,8 @@ namespace Microsoft.AspNet.OData.Formatter
             Func<IServiceProvider, ODataMessageWrapper> getODataMessageWrapper,
             Func<IEdmTypeReference, ODataSerializer> getEdmTypeSerializer,
             Func<Type, ODataSerializer> getODataPayloadSerializer,
-            Func<ODataSerializerContext> getODataSerializerContext)
+            Func<ODataSerializerContext> getODataSerializerContext,
+            ODataVersion deltaVersion)
         {
             if (model == null)
             {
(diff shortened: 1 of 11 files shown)


xuzhg:
close it since it's fixed in other PR. Thanks.

PR #2829, opened 2024-01-25
fixes #2816: Set the response version to delta response and delta patch response

### LINKED ISSUE #2816: OData-prefix missing in delta responses

xuzhg:
### Issues

*This pull request fixes #2816.*

### Description

Currently, ODataVersion.4.01 is hard coded for all delta response. So, customer can't switch between 4.0 and 4.01.

However, it may be true for delta patch since it's only included in OData version 4.01, but not in OData version 4.0. For other delta responses, it should be valid in both versions.

This PR is to change to 4.01 only for delta patch response but use the version setting for other delta response. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
@@ -158,7 +158,8 @@ namespace Microsoft.AspNet.OData.Formatter
             ODataMessageWriterSettings writerSettings = internalRequest.WriterSettings;
             writerSettings.BaseUri = baseAddress;
 
-            if (serializer.ODataPayloadKind == ODataPayloadKind.Delta)
+            // Set the response version to v4.01 when the request is a delta patch, but not any time we write a delta response payload.
+            if (serializer.ODataPayloadKind == ODataPayloadKind.Delta && internalRequest.Method == ODataRequestMethod.Patch)
             {
                 writerSettings.Version = ODataVersion.V401;
             }
--- a/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DeltaQueryTests/DeltaQueryTests.cs
+++ b/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/DeltaQueryTests/DeltaQueryTests.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Test.E2E.AspNet.OData
         }
 
         [Fact]
-        public async Task DeltaVerifyReslt()
+        public async Task DeltaVerifyResult()
         {
             HttpRequestMessage get = new HttpRequestMessage(HttpMethod.Get, BaseAddress + "/odata/TestCustomers?$deltaToken=abc");
             get.Headers.Add("Accept", "application/json;odata.metadata=minimal");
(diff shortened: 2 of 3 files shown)


mikepizzo (dismissed):
LGTM!

gathogojr (dismissed):

src/Microsoft.AspNet.OData.Shared/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -158,7 +158,8 @@ internal static bool TryGetCharSet(MediaTypeHeaderValue mediaType, IEnumerable<s
             ODataMessageWriterSettings writerSettings = internalRequest.WriterSettings;
             writerSettings.BaseUri = baseAddress;
 
-            if (serializer.ODataPayloadKind == ODataPayloadKind.Delta)
+            // Set the response version to v4.01 when the request is a delta patch, but not any time we write a delta response payload.
+            if (serializer.ODataPayloadKind == ODataPayloadKind.Delta && internalRequest.Method == ODataRequestMethod.Patch)
```
ElizabethOkerio, line 162: What's the internal request method of a normal delta response? Isn't it `Patch` also?

ElizabethOkerio: or this is for other methods like GET, POST etc..

xuzhg: That's 'Get', right?!

mikepizzo: PATCH is the verb used when the client is doing a delta (aka bulk) update.

The delta response payload is also used when the client retrieves changes using the deltaLink using GET.

Bulk updates are only supported for 4.01, so it's appropriate for the response from a delta update to be 4.01, but a request to get changes (which also returns in delta format) has been around since 4.0, and is valid in either 4.0 or 4.01 format.

PR #2836, opened 2024-03-29
fixes #2835: omit the ODataIdContainer from model builder, change the…

### LINKED ISSUE #2835: Perf: OData performance regressions found between 6.0.0 and 7.7.4

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData WebApi lib 7.x

### Reproduce steps

* The Reflection loading is only happening on the initial request for an organization, I can see that it tries to load the Annotation Values and if not found in the EdmModel goes out and calls the EdmLibHelpers.GetMatchingTypes() where the GetClrType is invoked and we see the significant CPU Spike. Subsequent request I see are successfully getting the Annotation from the model.

image

This by itself isn’t a problem, but Dataverse is hosting up to 300 Organizations on an environment and each Organization is trying to load over 1700 assemblies and many more types which is what is generating the spike in CPU load. 

Looking at this block here 
[https://github.com/OData/WebApi/blob/7dbaf97e17acecfb842edd002d7085d423189d87/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs#L1751-L1769](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FOData%2FWebApi%2Fblob%2F7dbaf97e17acecfb842edd002d7085d423189d87%2Fsrc%2FMicrosoft.AspNet.OData.Shared%2FFormatter%2FSerialization%2FODataResourceSerializer.cs%23L1751-L1769&data=05%7C02%7Csaxu%40microsoft.com%7Cff9957af35bf4206c06408dc500d6498%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638473265712869437%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=5PSK8R73uJoS8Byfk3gsW9%2BDGayPZVDifw7qVgks%2F6M%3D&reserved=0)
### Expected result

image

It appears that we are loading all the types available just so that we can ignore the ODataIdContainer type in the GetPropertiesToWrite() method. 

We also found this change last year as well around this code block that seems relevant [fix regression issue. by ElizabethOkerio · Pull Request #2769 · OData/WebApi · GitHub](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FOData%2FWebApi%2Fpull%2F2769&data=05%7C02%7Csaxu%40microsoft.com%7Cff9957af35bf4206c06408dc500d6498%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638473265712883566%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=JYw0SIEx0Erx5cUyEemeeGxvRnvGcBxJMUvgrKlQEz4%3D&reserved=0)

### Actual result

*What is actually happening.*

### Additional detail

xuzhg:
… serializer

### Issues

*This pull request fixes #2835.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
@@ -551,7 +551,11 @@ namespace Microsoft.AspNet.OData.Builder
                 bool isCollection;
                 IEdmTypeConfiguration mappedType;
 
-                PropertyKind propertyKind = GetPropertyType(property, out isCollection, out mappedType);
+                PropertyKind propertyKind = GetPropertyType(property, out isCollection, out mappedType, out bool isValid);
+                if (!isValid)
+                {
+                    continue;
+                }
 
                 if (propertyKind == PropertyKind.Primitive || propertyKind == PropertyKind.Complex || propertyKind == PropertyKind.Enum)
                 {
@@ -694,17 +698,25 @@ namespace Microsoft.AspNet.OData.Builder
 
         // figures out the type of the property (primitive, complex, navigation) and the corresponding edm type if we have seen this type
         // earlier or the user told us about it.
-        private PropertyKind GetPropertyType(PropertyInfo property, out bool isCollection, out IEdmTypeConfiguration mappedType)
+        private PropertyKind GetPropertyType(PropertyInfo property, out bool isCollection, out IEdmTypeConfiguration mappedType, out bool IsValid)
         {
             Contract.Assert(property != null);
+
+            IsValid = true;
+            mappedType = null;
+            isCollection = false;
+
+            if (typeof(ODataIdContainer) == property.PropertyType)
+            {
+                IsValid = false;
+                return (PropertyKind)(int.MaxValue);
+            }
 
             // IDictionary<string, object> is used as a container to save/retrieve dynamic properties for an open type.
             // It is different from other collections (for example, IEnumerable<T> or IDictionary<string, int>)
             // which are used as navigation properties.
             if (typeof(IDictionary<string, object>).IsAssignableFrom(property.PropertyType))
             {
-                mappedType = null;
-                isCollection = false;
                 return PropertyKind.Dynamic;
             }
 
@@ -712,16 +724,12 @@ namespace Microsoft.AspNet.OData.Builder
             // It is different from other collections (for example, IDictionary<string,IDictionary<string, int>>)          
             if (typeof(IODataInstanceAnnotationContainer).IsAssignableFrom(property.PropertyType))
             {
-                mappedType = null;
-                isCollection = false;
-
                 return PropertyKind.InstanceAnnotations;
             }
 
             PropertyKind propertyKind;
             if (TryGetPropertyTypeKind(property.PropertyType, out mappedType, out propertyKind))
             {
-                isCollection = false;
                 return propertyKind;
             }
 
@@ -729,6 +737,13 @@ namespace Microsoft.AspNet.OData.Builder
             if (TypeHelper.IsCollection(property.PropertyType, out elementType))
             {
                 isCollection = true;
+
+                if (typeof(ODataIdContainer) == elementType)
+                {
+                    IsValid = false;
+                    return (PropertyKind)(int.MaxValue);
+                }
+
                 if (TryGetPropertyTypeKind(elementType, out mappedType, out propertyKind))
                 {
                     return propertyKind;
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
@@ -291,17 +291,20 @@ namespace Microsoft.AspNet.OData.Formatter
             }
 
             string typeName = edmSchemaType.FullName();
-            IEnumerable<Type> matchingTypes = GetMatchingTypes(typeName, assembliesResolver);
 
-            if (matchingTypes.Count() > 1)
+            // Calling "ToList()" to avoid multiple enumerates.
+            IList<Type> matchingTypes = GetMatchingTypes(typeName, assembliesResolver).ToList();
+
+            if (matchingTypes.Count > 1)
             {
                 throw Error.Argument("edmTypeReference", SRResources.MultipleMatchingClrTypesForEdmType,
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -694,41 +698,52 @@ private void MapStructuralProperty(StructuralTypeConfiguration type, PropertyInf
...
             {
-                isCollection = false;
                 return propertyKind;
             }
 
             Type elementType;
             if (TypeHelper.IsCollection(property.PropertyType, out elementType))
             {
                 isCollection = true;
+
```
mikepizzo, line 740: yikes -- there should never be a reason for having a collection of ODataIdContainers, but doesn't hurt to check for it.

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1748,26 +1748,22 @@ private async Task WriteStreamPropertiesAsync(SelectExpandNode selectExpandNode,
                     changedProperties = deltaObject.GetChangedPropertyNames();
                 }
 
-                foreach (KeyValuePair<IEdmStructuralProperty, PathSelectItem> complexProperty in complexProperties)
+                if (changedProperties != null)
                 {
-                    if (changedProperties == null || changedProperties.Contains(complexProperty.Key.Name))
+                    IDictionary<IEdmStructuralProperty, PathSelectItem> newDic = new Dictionary<IEdmStructuralProperty, PathSelectItem>();
```
mikepizzo, line 1753: maybe name this more appropriately to make it easier for readers of the code to know what it is for, not just that it is a dictionary.
```suggestion
                    IDictionary<IEdmStructuralProperty, PathSelectItem> propertiesToReturn = new Dictionary<IEdmStructuralProperty, PathSelectItem>();
```

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -694,41 +698,52 @@ private void MapStructuralProperty(StructuralTypeConfiguration type, PropertyInf
 
         // figures out the type of the property (primitive, complex, navigation) and the corresponding edm type if we have seen this type
         // earlier or the user told us about it.
-        private PropertyKind GetPropertyType(PropertyInfo property, out bool isCollection, out IEdmTypeConfiguration mappedType)
+        private PropertyKind GetPropertyType(PropertyInfo property, out bool isCollection, out IEdmTypeConfiguration mappedType, out bool IsValid)
```
mikepizzo, line 701: nit: the property is a valid property, it just shouldn't be mapped.  Maybe rename to something like "Unmapped"?

mikepizzo (changes requested):

mikepizzo (commented):
A few minor comments; otherwise, looks good.

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
```diff
@@ -694,41 +698,52 @@ private void MapStructuralProperty(StructuralTypeConfiguration type, PropertyInf
...
         {
             Contract.Assert(property != null);
 
+            IsValid = true;
+            mappedType = null;
+            isCollection = false;
+
+            if (typeof(ODataIdContainer) == property.PropertyType)
```
mikepizzo, line 709: should we use isassignablefrom in case the user has derived from ODataIdContainer?

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
```diff
@@ -291,17 +291,20 @@ public static Type GetClrType(IEdmType edmType, IEdmModel edmModel, IWebApiAssem
...
+
+            if (matchingTypes.Count > 1)
             {
                 throw Error.Argument("edmTypeReference", SRResources.MultipleMatchingClrTypesForEdmType,
-                    typeName, String.Join(",", matchingTypes.Select(type => type.AssemblyQualifiedName)));
+                    typeName, String.Join(",", matchingTypes.Select(t => t.AssemblyQualifiedName)));
             }
 
-            edmModel.SetAnnotationValue<ClrTypeAnnotation>(edmSchemaType, new ClrTypeAnnotation(matchingTypes.SingleOrDefault()));
+            Type type = matchingTypes.Count == 0 ? null : matchingTypes[0];
```
mikepizzo, line 304: maybe make this name more explicit: "clrType"?

```suggestion
            Type clrType = matchingTypes.Count == 0 ? null : matchingTypes[0];
```

I generally like to avoid naming types "type" because it is confusing with "Type", but in this case it really is a type. Naming it "clrType" helps clarify that we did mean the clrType, not something else we were calling "type" (like the edmType)

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
xuzhg: I changed to "mapped".

src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
xuzhg: #Resolved.

src/Microsoft.AspNet.OData.Shared/Builder/ODataConventionModelBuilder.cs
xuzhg: I'd love to take it.

src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs
xuzhg: #Resolved

mikepizzo (dismissed):

PR #2842, opened 2024-06-13
Make sure 'top' doesn't apply on a null collection

ificator:
### Issues
`$top` is attempted on null collections, resulting in failures like this:
```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. --> 
System.ArgumentNullException: Value cannot be null. 
Parameter name: source
	at System.Linq.OrderedEnumerable`2..ctor(Enumerable`1 source, Func` keySelector, IComparer`1comparer, Boolean decending)
	at lambda_method(Closure_Site) 
	--End of inner exception stack trace--
	at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor) 
	at System.Reflection.RuntimeMethodInfo. UnsafeInvokeInternal(Object obj, Object [] parameters, Object arguments) 
	at System.Delegatre.DynamicInvokeImpl
	at Microsot.AspNet.OData.Query.Expressions.SelectExpandBinder.Bind
	at Microsot.AspNet.OData.Query.ODataQueryOptions.ApplySelectExpand
```

### Description
`ProjectCollection` does have logic for handling null propagation, but it's currently using the modified source (e.g. the one that wraps the original in `OrderBy`, `Take`, `Skip`). As a result, when the `source` is null the check doesn't apply prior to the execution of those operations, which all require a non-null collection.

The fix is to simply perform the null check on the original `source`.

NOTE: The same issue exists in https://github.com/OData/AspNetCoreOData but I'll leave it to maintainers to port the fix.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
@@ -978,8 +978,12 @@ namespace Microsoft.AspNet.OData.Query.Expressions
 
         // new CollectionWrapper<ElementType> { Instance = source.Select((ElementType element) => new Wrapper { }) }
         [SuppressMessage("Microsoft.Maintainability", "CA1502:AvoidExcessiveComplexity", Justification = "These are simple conversion function and cannot be split up.")]
-        private Expression ProjectCollection(Expression source, Type elementType,
-            SelectExpandClause selectExpandClause, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource,
+        private Expression ProjectCollection(
+            Expression source,
+            Type elementType,
+            SelectExpandClause selectExpandClause,
+            IEdmStructuredType structuredType,
+            IEdmNavigationSource navigationSource,
             OrderByClause orderByClause,
             long? topOption,
             long? skipOption,
@@ -1005,9 +1009,10 @@ namespace Microsoft.AspNet.OData.Query.Expressions
             //      (ElementType element) => new Wrapper { }
             LambdaExpression selector = Expression.Lambda(projection, element);
 
+            Expression modifiedSource = source;
             if (orderByClause != null)
             {
-                source = AddOrderByQueryForSource(source, orderByClause, elementType);
+                modifiedSource = AddOrderByQueryForSource(modifiedSource, orderByClause, elementType);
             }
 
             bool hasTopValue = topOption != null && topOption.HasValue;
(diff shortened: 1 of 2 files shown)


xuzhg (approved):

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1005,9 +1009,10 @@ private Expression AddOrderByQueryForSource(Expression source, OrderByClause ord
             //      (ElementType element) => new Wrapper { }
             LambdaExpression selector = Expression.Lambda(projection, element);
 
+            Expression modifiedSource = source;
             if (orderByClause != null)
             {
-                source = AddOrderByQueryForSource(source, orderByClause, elementType);
```
xuzhg, line 1010: Weird, save the source to a new variable fixes the issue?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -978,8 +978,12 @@ private Expression AddOrderByQueryForSource(Expression source, OrderByClause ord
 
         // new CollectionWrapper<ElementType> { Instance = source.Select((ElementType element) => new Wrapper { }) }
         [SuppressMessage("Microsoft.Maintainability", "CA1502:AvoidExcessiveComplexity", Justification = "These are simple conversion function and cannot be split up.")]
-        private Expression ProjectCollection(Expression source, Type elementType,
-            SelectExpandClause selectExpandClause, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource,
```
xuzhg, line 982: How about the "skip" since it's at same level as "top"?

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
ificator: There's still one reference to "source" in the null propagation code. That's the fix - making sure the null check uses the original source instead of the modified.

src/Microsoft.AspNet.OData.Shared/Query/Expressions/SelectExpandBinder.cs
ificator: Skip should be fixed as well since it's using modifiedSource, although I don't have a test for that.

habbes:
Comment to trigger pipeline.

PR #2845, opened 2024-06-24
Fixes #2844: Enable property with instance annotation without propert…

### LINKED ISSUE #2844: WebApi v7 doesn't support property Instance annotation without property value
WebApi v7 doesn't support property instance annotations when the property value is not present.

### Assemblies affected

OData WebApi v7

### Reproduce steps

Considering this payload, where references is an OpenComplexTypeObject. When this payload is passed, we expect the webapi library to identify the key "https://a/task@task.task" as a property instance annotation. 
However, the v7 library is missing this feature and returns nothing for the property Instance Annotations.

```
{
    "details", new JObject
    {
        {
            "references", new JObject
            {
                {
                    "https://a/task@task.task", new JObject
                    {
                        { "@odata.type", "microsoft.taskServices.externalReference" },
                        { "alias", "abcd"},
                        { "type", "Other" },
                    }
                }
            }
        },
    }
}
```

### Expected result
We expect the request to fail with a BadRequest like it did in v6.
(or)
When a prefer header is sent with the request,
```
request.Headers.Add("Prefer", "odata.include-annotations=*");
```
The InstanceAnnotations in the ODataResourceWrapper of the ODataResourceDeserializer, should populate the InstanceAnnotations for the property.

### Actual result

ODataResourceWrapper.ResourceBase.Properties.InstanceAnnotations.Count = 0 for the property "references" 
and the request passes with the response showing 
```"details": null```

### Additional detail
In WebApi lib v6, the ODataResourceDeserializer considered ```references``` as a complex property and threw an error like:
```ODataJsonLightPropertyAndValueDeserializer_ComplexValuePropertyAnnotationWithoutProperty```
However, in v7, ```references``` is considered as a resource. The deserializer is not sure how to map the annotation as the property value is missing.
I was informed that V8 has a fix for this. So, porting that to v7 would be a quick fix.

### PR DESCRIPTION
**Pull request by xuzhg**
…y value

### Issues

*This pull request fixes #2844 .*

### Description

* Enable property with instance annotations without property value

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataReaderExtensions.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataReaderExtensions.cs
@@ -189,6 +189,12 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                     itemsStack.Push(deltaResourceSetWrapper);
                     break;
 
+                case ODataReaderState.NestedProperty:
+                    Contract.Assert(itemsStack.Count > 0, "The nested property info should be a non-null primitive value within resource wrapper.");
+                    ODataResourceWrapper resourceParentWrapper = (ODataResourceWrapper)itemsStack.Peek();
+                    resourceParentWrapper.NestedPropertyInfos.Add((ODataPropertyInfo)reader.Item);
+                    break;
+
                 case ODataReaderState.DeltaLink:
                 case ODataReaderState.DeltaDeletedLink:
 
--- a/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/ODataReaderExtensionsTests.cs
+++ b/test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/ODataReaderExtensionsTests.cs
@@ -1,0 +1,333 @@
+//-----------------------------------------------------------------------------
+// <copyright file="ODataDeserializerTest.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved. 
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNet.OData.Formatter;
+using Microsoft.AspNet.OData.Formatter.Deserialization;
+using Microsoft.AspNet.OData.Test.Common;
+using Microsoft.OData;
+using Microsoft.OData.Edm;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNet.OData.Test.Formatter.Deserialization
+{
+    public partial class ODataReaderExtensionsTests
+    {
+        private static readonly EdmModel Model;
+
+        static ODataReaderExtensionsTests()
+        {
+            Model = new EdmModel();
+
+            // Address
+            EdmComplexType address = new EdmComplexType("NS", "Address");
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataReaderExtensions.cs [hunk 1]
```diff
@@ -189,6 +189,12 @@ private static void ReadCollectionItem(ODataReader reader, Stack<ODataItemBase>
                     itemsStack.Push(deltaResourceSetWrapper);
                     break;
 
+                case ODataReaderState.NestedProperty:
+                    Contract.Assert(itemsStack.Count > 0, "The nested property info should be a non-null primitive value within resource wrapper.");
+                    ODataResourceWrapper resourceParentWrapper = (ODataResourceWrapper)itemsStack.Peek();
```
mikepizzo, line 194: This assumes that the only time we would get a NestedProperty is when reading a property info with no value. Is that guaranteed to be true? Might the PropertyInfo be an ODataProperty (i.e., have a value)? Now or in the future?  In that case would it also be added to ODataResource.Properties?

mikepizzo (approved):

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/ODataReaderExtensionsTests.cs
```diff
@@ -0,0 +1,333 @@
...
+
+            // Assert
+            Assert.NotNull(item);
+            ODataResourceWrapper resource = Assert.IsType<ODataResourceWrapper>(item);
+            Assert.Equal(3, resource.NestedResourceInfos.Count);
+            Assert.Equal(new[] { "Location", "Order", "Orders" }, resource.NestedResourceInfos.Select(n => n.NestedResourceInfo.Name));
+        }
+
+        [Fact]
+        public async Task ReadResourceWithPropertyWithoutValueButWithInstanceAnnotationsWorksAsExpected()
```
ElizabethOkerio, line 122: Do we already support writing these annotations?

ElizabethOkerio (approved):

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/ODataReaderExtensionsTests.cs
```diff
@@ -0,0 +1,333 @@
...
+        {
+            // Arrange
+            // Property 'Name' without value but with instance annotations
+            const string payload =
+            "{" +
+                "\"@odata.context\":\"http://localhost/$metadata#Customers/$entity\"," +
+                "\"CustomerID\": 17," +
+                "\"Name@Custom.PrimitiveAnnotation\":123," +
```
vasrinwork, line 130: Could we add a test where the there is a property without value but with instance annotations in a nestedResourceinfo?
For example, in the below case, will the resource.ResourceBase.Properties show the Instance Annotation count as 1? 
or only the NestedResourceInfo 'names' will have the instance annotation count as 1?
{
  "customerId":17,
  "details":{
    "names":{
      "Name@Custom.PrimitiveAnnotation":123
    }
  }
}

vasrinwork:
@xuzhg could you provide an eta on when this PR is expected to be merged?

test/UnitTest/Microsoft.AspNet.OData.Test.Shared/Formatter/Deserialization/ODataReaderExtensionsTests.cs
xuzhg: Yes

PR #2859, opened 2024-10-31
Fixes #2858: Add Virtual method for creating Delta response

### LINKED ISSUE #2858: No virtual method for creating Delta ComplexNestedResourceInfo
I am writing a custom Odata serializer that inherits from ODataResourceSerializer, overriding the CreateResource method. My goal is to conditionally filter out some properties to hide them from certain apps calling our service. We have success in filtering out some structural properties from our resource. I am also able to filter out complex nested resources through overriding CreateComplexNestedResourceInfo. 

However, I am now trying to filter out a complex property for delta and noticed that the code doesn't run through CreateComplexNestedResourceInfo. It creates the complex resource through private methods in [WriteDeltaResource](https://github.com/OData/WebApi/blob/651c83ddb20703d13083d737ae8006ac0a3ec013/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs#L2089)

We need a way to override creating the delta complex resource, where if the overridden method returns null, that complex property won't show up in the serialized odata output. Something like a virtual "CreateDeltaComplexNestedResourceInfo".

### Assemblies affected

OData WebApi lib 7.7.6.0

xuzhg:
Fixes #2858: Add Virtual method for creating Delta response
1) Add a new virtual method to create the ODataNestedResourceInfo for complex and navigation in the delta
2) Add a test to test the calling

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
@@ -462,6 +462,29 @@ namespace Microsoft.AspNet.OData.Formatter.Serialization
         }
 
         /// <summary>
+        /// Creates the <see cref="ODataNestedResourceInfo"/> to be written while writing this delta nested property (complex or entity).
+        /// </summary>
+        /// <param name="property">The property for which the nested resource info is being created.</param>
+        /// <param name="resourceContext">The context for the property instance being written.</param>
+        /// <returns>The nested resource info to be written. Returns 'null' will omit this property serialization.</returns>
+        /// <remarks>It enables customer to get more control by overriding this method. </remarks>
+        public virtual ODataNestedResourceInfo CreateDeltaNestedResourceInfo(IEdmProperty property, ResourceContext resourceContext)
+        {
+            if (property == null)
+            {
+                throw Error.ArgumentNull(nameof(property));
+            }
+
+            return property.Type != null ?
+                new ODataNestedResourceInfo
+                {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -461,6 +461,29 @@ public virtual string CreateETag(ResourceContext resourceContext)
...
         }
 
+        /// <summary>
+        /// Creates the <see cref="ODataNestedResourceInfo"/> to be written while writing this delta nested property (complex or entity).
+        /// </summary>
+        /// <param name="property">The property for which the nested resource info is being created.</param>
+        /// <param name="resourceContext">The context for the property instance being written.</param>
+        /// <returns>The nested resource info to be written. Returns 'null' will omit this property serialization.</returns>
+        /// <remarks>It enables customer to get more control by overriding this method. </remarks>
+        public virtual ODataNestedResourceInfo CreateDeltaNestedResourceInfo(IEdmProperty property, ResourceContext resourceContext)
```
habbes, line 471: Why do we use `Delta` in the name of the method? Why not just `CreateNestedResourceInfo`?

habbes: I see that we have another `CreateDynamicComplexNestedResourceInfo` method. Are there other scenarios of nested resource info that we would want to allow the customer to override?

habbes (approved):

PR #2863, opened 2024-11-20
Fixes #2861: Pass the IServiceProvider to ODataUriParser

### LINKED ISSUE #2861: Microsoft.AspNet.OData 7.7.5.0 - ODataResourceDeserializer to take in IServiceProvider

Currently the ODataResourceDeserializer::ReadInline calls 
ODataResourceDeserializerHelpers.ApplyIdToPath which calls:
ODataUriParser parser = new ODataUriParser(readContext.Model, new Uri(serviceRoot), resourceWrapper.ResourceBase.Id);
without specifying the 4th argument - the IServiceProvider.

This is an issue as the custom ODataUriResolver can't be used from the IServiceProvider, and default one is being used causing  issues down the line.

We are using Microsoft.AspNet.OData 7.7.5.0 version and updating to the latest version could be risky. We ask this to be fixed in 7.7.5.0 if possible. This is a backward compatible change, to take in IServiceProvider (default to null) in all public and protected methods of ODataResourceDeserializer and pass it down accordingly.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*
Microsoft.AspNet.OData 7.7.5.0 +

### Reproduce steps

Create a custom ODataUriResolver and register it in IServiceProvider as ODataUriResolver.
Parse this URL:
/api/data/v9.0/systemusers(4cf0864c-c224-ef11-8408-0022480ac456)/Microsoft.Dynamics.CRM.RetrievePrincipalAccess(Target=@Target)?@Target=%7B%22%40odata.id%22%3A%22incidents(e21d1a87-1793-ef11-ac2d-6045bd09fcb8)%22%7D

### Expected result

Custom ODataUriResolver needs to be called for both parts of the URL.

### Actual result

Notice that the custom resolver is called for
systemusers(4cf0864c-c224-ef11-8408-0022480ac456)/
part but it is never called for
incidents(e21d1a87-1793-ef11-ac2d-6045bd09fcb8)
part (default one is called).

xuzhg:
Fixes #2861: Pass the IServiceProvider to ODataUriParser

### DIFF
--- a/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializerHelpers.cs
+++ b/src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializerHelpers.cs
@@ -91,7 +91,17 @@ namespace Microsoft.AspNet.OData.Formatter.Deserialization
                     {
                         try
                         {
-                            path = new Routing.ODataPath(new ODataUriParser(readContext.Model, new Uri(navigationSource.Path.Path, UriKind.Relative)).ParsePath());
+                            ODataUriParser parser;
+                            if (readContext != null && readContext.InternalRequest != null && readContext.InternalRequest.RequestContainer != null)
+                            {
+                                parser = new ODataUriParser(readContext.Model, new Uri(navigationSource.Path.Path, UriKind.Relative), readContext.InternalRequest.RequestContainer);
+                            }
+                            else
+                            {
+                                parser = new ODataUriParser(readContext.Model, new Uri(navigationSource.Path.Path, UriKind.Relative));
+                            }
+
+                            path = new Routing.ODataPath(parser.ParsePath());
                         }
                         catch (ODataException)
                         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.OData.Shared/Formatter/Deserialization/ODataResourceDeserializerHelpers.cs
```diff
@@ -91,7 +91,17 @@ internal static ODataDeserializerContext GenerateNestedReadContext(ODataNestedRe
                     {
                         try
                         {
-                            path = new Routing.ODataPath(new ODataUriParser(readContext.Model, new Uri(navigationSource.Path.Path, UriKind.Relative)).ParsePath());
+                            ODataUriParser parser;
+                            if (readContext != null && readContext.InternalRequest != null && readContext.InternalRequest.RequestContainer != null)
```
gathogojr, line 95: nit:
```suggestion
                            if (readContext?.InternalRequest?.RequestContainer != null)
```

gathogojr:
@xuzhg Add a test to verify that if request container is injected, it is used as expected

xuzhg:
> @xuzhg Add a test to verify that if request container is injected, it is used as expected

I tried, but it seems it's hard to construct such test. I am failed.

